PR #2, opened 2019-10-16
tutorial_vqls

andreamari:

**Title:** Tutorial on Variational Quantum Linear Solvers

**Possible Drawbacks:**

I still need to check if sphinx building process works in this repo.


implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+def variational_block(weights):
+    """Variational circuit mapping the ground state |0> to the ansatz state |x>.
+    """
+    # We first prepare an equal superposition of all the states of the computational basis.
+    for idx in range(n_qubits):
+        qml.Hadamard(wires=idx)
+
+    # Model A
+    # For this specific example we use a very simple variational circuit.
+    for idx, element in enumerate(weights[0]):
```
co9olguy: what is the content of `weights[i]` for i>0?

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+##############################################################################
+# Hadamard test
+# --------------
+# We define a PennyLane ``qnode`` object representing a model of the actual quantum computation.
+#
+# The circuit is based on the
+# `Hadamard test <https://en.wikipedia.org/wiki/Hadamard_test_(quantum_computation)>`_
+# and will be used to estimate the coefficients :math:`\mu_{l,l',j}` defined in the introduction.
+#
+# A graphical representation of this circuit is given in the figure placed below the title of this tutorial.
```
co9olguy:
```suggestion
# A graphical representation of this circuit is shown at the top of this tutorial. 
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+# A graphical representation of this circuit is given in the figure placed below the title of this tutorial. 
+
+@qml.qnode(dev_mu)
+def local_hadamard_test(weights, A_idx=None, A_dag_idx=None, Z_idx=None, part=None):
+
+    # Variational circuit generating a guess for the solution vector |x>
+    variational_block(weights)
+
+    # First Hadamard gate applied to the ancillary qubit.
+    qml.Hadamard(wires=ancilla_idx)
```
co9olguy: maybe switch the hadamard steps on the ancilla and the variational ansatz to match the order in the figure?

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+    # For estimating the imaginary part of the coefficient "mu", we must add a "-i" phase gate.
+    if part == "Im" or part == "im":
+        qml.PhaseShift(3 / 2 * np.pi, wires=ancilla_idx)
```
co9olguy: Shown as a rotation `R` in the figure. Can we make consistent?

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+    U_b()
+
+    # Controlled application of the unitary A_idx^dag.
+    CA_dag(A_dag_idx)
+
+    # Second Hadamard gate applied to the ancillary qubit.
+    qml.Hadamard(wires=ancilla_idx)
+
+    # Expectation value of Z for the ancillary qubit.
+    return qml.expval.PauliZ(wires=ancilla_idx)
```
co9olguy:
```suggestion
    return qml.expval(qml.PauliZ(wires=ancilla_idx))
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+
+##############################################################################################
+# To get the real and immaginary parts of :math:`\mu_{l,l',j}`, one needs to run the previous
```
co9olguy:
```suggestion
# To get the real and imaginary parts of :math:`\mu_{l,l',j}`, one needs to run the previous
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+# quantum circuit with and without a phase-shift of the ancillary qubit. This is automatically
+# done by the following function.
+
+
+def mu(weights, l=None, lp=None, j=None):
+    """Generates the matrix element \mu_{l, lp ,j} = <0| V^dag A_lp^dag U Z_j  U^dag A_l V |0>,
+    where Z_j is the Z operator applied to the qubit with index j.
+    If j=-1, then Z_{-1} is replaced with the identity.
+    This quantity can be used to compute a "local" version of the variational cost.
```
co9olguy:
```suggestion
    This quantity can be used to compute a "local" version of the variational cost function.  
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+##############################################################################
+# Variational optimization
+# -----------------------------
+#
+# We first initialize the variational weights with random parameters (with a fixed seed).
+
+
+# Random number generator with seed.
+rng = np.random.RandomState(rng_seed)
```
co9olguy: more common pattern is 
```
np.random.seed(rng_seed)
np.random.randn(q_depth, n_qubits)
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+# Random initial weights.
+w = q_delta * rng.randn(q_depth, n_qubits)
+
+##############################################################################
+# To minimize the cost function we use the Adam optimizer.
+opt = qml.GradientDescentOptimizer(eta)
```
co9olguy: comment does not match code

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+for it in range(steps):
+    w = opt.step(cost_loc, w)
+    cost = cost_loc(w)
+    print("Step {:3d}       Cost_L = {:9.7f}".format(it, cost))
+    cost_history.append(cost)
+
+
+##############################################################################
+# We plot the cost function with respect to the optimization steps.
+# We remark that this quantity is not an abstract optimization tool,
```
co9olguy: "tool" seems a bit off; suggest rewording

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+    cost = cost_loc(w)
+    print("Step {:3d}       Cost_L = {:9.7f}".format(it, cost))
+    cost_history.append(cost)
+
+
+##############################################################################
+# We plot the cost function with respect to the optimization steps.
+# We remark that this quantity is not an abstract optimization tool, 
+# but it actually quantifies the error between the generated state 
+# and the exact solution of the problem.
```
co9olguy: still true for local cost?

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+##############################################################################
+# Classical algorithm
+# ^^^^^^^^^^^^^^^^^^^
+# To solve the problem in a classical way, we use the explicit matrix representation in 
+# terms of numerical NumPy arrays.
+
+Id = np.identity(2)
+Z = np.array([[1, 0], [0, -1]])
+X = np.array([[0, 1], [1, 0]])
+H = np.sqrt(2) * np.array([[0, 1], [1, 0]])
```
co9olguy: - This isn't hadamard gate?
- appears not to be used

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+##############################################################################
+# The solution can be computed via a matrix inversion:
+
+A_inv = np.linalg.inv(A_num)
+x = np.dot(A_inv, b)
+
+##############################################################################
+# To check that the solution is correct, we compute and print the error.
+
+err = np.sum(np.abs(np.dot(A_num, x) - b) ** 2)
```
co9olguy: can use `np.linalg.norm`

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+print("|A x - b|^2 = ", err)
+
+# Finally, in order to compare x with the quantum state |x>, we square and normalize its elements.
+c_probs = x ** 2
+c_probs = c_probs / np.sum(c_probs)
```
co9olguy: can use `np.linalg.norm` here as well
```
c_probs = x ** 2 / np.linalg.norm(x)
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+##############################################################################
+# Preparation of the quantum solution
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+##############################################################################
+# Given the quantum variational weights ``w`` that we have previously optimized,
```
co9olguy:
```suggestion
# Given the variational weights ``w`` that we have previously optimized,
```

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+# we can generate the quantum state :math:`|x\rangle`.
+
+
+@qml.qnode(dev_x)
```
co9olguy: Since this is the only place where `dev_x` is used, I would just define the device here, rather than at the top.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+    # Arbitrary observable.
+    # In this case we are actually interested only on the state probabilities.
+    obs = [qml.expval.Identity(k) for k in range(0, n_qubits)]
+    return obs
+
+
+prepare_x(w)
+
+# In PennyLane one can easily extract the squared modulus of the state amplitudes,
```
co9olguy: only works for simulators

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+    # In this case we are actually interested only on the state probabilities.
+    obs = [qml.expval.Identity(k) for k in range(0, n_qubits)]
+    return obs
+
+
+prepare_x(w)
+
+# In PennyLane one can easily extract the squared modulus of the state amplitudes, 
+# evaluated before the final measurement.
+q_probs = list(dev_x.probability().values())
```
co9olguy: Might be better to estimate these using expectation values of Pauli Z measurements? that way can be executed on hardware?

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
+r"""
+.. _vqls:
```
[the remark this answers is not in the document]
andreamari: Ok thanks!

implementations/tutorial_vqls.py
andreamari: Done. But I think the correct formula should be this one: 
`c_probs = (x / np.linalg.norm(x)) ** 2`

implementations/tutorial_vqls.py
andreamari: ok, done.

implementations/tutorial_vqls.py
andreamari: Replaced with a basis state sampling.

implementations/tutorial_vqls.py
andreamari: Using expectation values would require computing many moments of Z measurements. I would say that, experimentally, it is better to directly take samples of computational basis states (i.e. measuring Z_1 Z_2 ... Z_n).

I have implemented the computational basis measurement with a simple `qml.Hermitian()` but if this a problem for the hardware I could try to rewrite it only in terms of `qml.PauliZ()`.

implementations/tutorial_vqls.py
andreamari: ok, done.

implementations/tutorial_vqls.py
andreamari: changed `3 / 2 * np.pi` with `-np.pi / 2` .    Do you have a suggestion for a better phase-shift symbol? I have taken it from  https://en.wikipedia.org/wiki/Quantum_logic_gate

implementations/tutorial_vqls.py
andreamari: ok changed.

implementations/tutorial_vqls.py
andreamari: fixed

implementations/tutorial_vqls.py
andreamari: changed in this way.
`
# We plot the cost function with respect to the optimization steps.
# We remark that this is not an abstract mathematical quantity
# since it also represents a bound for the error between the generated state
# and the exact solution of the problem.
`

implementations/tutorial_vqls.py
andreamari: changed in this way:
```
# We plot the cost function with respect to the optimization steps.
# We remark that this is not an abstract mathematical quantity
# since it also represents a bound for the error between the generated state
# and the exact solution of the problem.
```

implementations/tutorial_vqls.py
andreamari: Changed as suggested.

andreamari: Actually this step always gave a trivial error of 0.0, and it is also not very useful for the tutorial.
So I removed this part.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+Now we discuss two alternative methods which could be used to experimentally solve the minimization problem.
+
+First method
+>>>>>>>>>>>>>
+
+Let us write :math:`C_G` more explicitly:
+
+.. math::
+
+    C_G = 1- \frac{ \sum_{l, l'}  c_l c_{l'}^* \langle 0|  V^\dagger A_{l'}^\dagger U \color{blue}{|0\rangle \langle 0|} U^\dagger A_l  V |0\rangle}
```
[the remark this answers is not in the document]
andreamari: It is referenced in the text a few lines below (second method).  But if it still creates confusion I can make everything black.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+def U_b():
+    """Unitary matrix rotating the ground state to the problem vector |b> = U_b |0>.
+    """
+    for idx in range(n_qubits):
+        qml.Hadamard(wires=idx)
+
+
+def CA(idx):
+    """Controlled versions of the unitary components A_l of the problem matrix A.
+    """
```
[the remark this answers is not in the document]
andreamari: ok

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+
+.. math::
+
+    P =  \frac{1}{2} + \frac{1}{2n}\sum_{j=0}^{n-1} Z_j,
+
+where :math:`Z_j` is the Pauli :math:`Z` operator locally applied to the :math:`j\rm{th}` qubit. This gives a new cost function:
+
+.. math::
+
+    C_L = 1- \frac{ \sum_{l, l'}  c_l c_{l'}^* \langle 0|  V^\dagger A_{l'}^\dagger U \color{blue}{P} U^\dagger A_l  V |0\rangle}
```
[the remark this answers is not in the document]
andreamari: Because it replaces the previous blue ground-state projector.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+
+A simple example
+^^^^^^^^^^^^^^^^^
+
+In this tutorial we consider the following simple example based on a system of 3 qubits (plus an ancilla)
+which is very similar to the one experimentally tested in Ref. [1]:
+
+.. math::
+        A  &=&  c_0 A_0 + c_1 A_1 + c_2 A_2 = \mathbb{1} + 0.2 X_0 Z_1 + 0.2 X_0, \\
+        |b\rangle &=& U |0 \rangle = H_0 \otimes H_1 \otimes H_2 |0\rangle,
```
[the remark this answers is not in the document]
andreamari: done. I also added more space between the two lines.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+# quantum circuit with and without a phase-shift of the ancillary qubit. This is automatically
+# done by the following function.
+
+
+def mu(weights, l=None, lp=None, j=None):
+    r"""Generates the matrix element \mu_{l, lp ,j} = <0| V^dag A_lp^dag U Z_j  U^dag A_l V |0>,
+    where Z_j is the Z operator applied to the qubit with index j.
+    If j=-1, then Z_{-1} is replaced with the identity.
+    This quantity can be used to compute a "local" version of the variational cost.
+    """
```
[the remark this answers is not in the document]
andreamari: Replaced with a more simple `"""Generates the coefficients to compute the "local" cost function C_L."""`. The full formula for \mu is already given in the ReST in the introduction.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+    elif idx == 1:
+        qml.CNOT(wires=[ancilla_idx, 0])
+        qml.CZ(wires=[ancilla_idx, 1])
+
+    elif idx == 2:
+        qml.CNOT(wires=[ancilla_idx, 0])
+
+    # If necessary, further unitary components can be added below.
+    # elif idx==3:
+    # ...
```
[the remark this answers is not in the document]
andreamari: I just removed it. It could create unnecessary confusion even as a ReST block.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+    """Adjoint of the controlled unitary components of A.
+    """
+    if idx == 0:
+        CA(0)  # In this particular example the operations are self-adjoint.
+    elif idx == 1:
+        CA(1)
+    elif idx == 2:
+        CA(2)
+    elif idx == 3:
+        CA(3)
```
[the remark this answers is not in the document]
andreamari: A removed the CA_dag() and U_dag() functions. Each time I need them, I added comment saying that, for this specific example, A_l and U are self-adjoint.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,558 @@
...
+
+    # Controlled Z operator at position Z_idx. If Z_idx=-1, apply the identity.
+    if Z_idx != -1:
+        qml.CZ(wires=[ancilla_idx, Z_idx])
+
+    # Unitary U_b associated to the problem vector |b>.
+    U_b()
+
+    # Controlled application of the unitary A_idx^dag.
+    CA_dag(A_dag_idx)
```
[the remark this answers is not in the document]
andreamari: A removed the CA_dag() and U_dag() functions.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+many variational layers. 
+
+A simple example
+^^^^^^^^^^^^^^^^^
+
+In this tutorial we consider the following simple example based on a system of 3 qubits (plus an ancilla) 
+which is very similar to the one experimentally tested in Ref. [1]:
+
+.. math::
+        A  &=&  c_0 A_0 + c_1 A_1 + c_2 A_2 = \mathbb{1} + 0.2 X_0 Z_1 + 0.2 X_0, \\
```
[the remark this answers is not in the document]
andreamari: Everywhere in this tutorial Z_j, X_j, H_j  are operators acting on qubit "j" but living in the full Hilbert space.  
E.g.,    Z_1= Identity_0 \otimes Z \otimes Identity_2 .
This simplifies many formulas and it is the same notation used in the paper.

Unfortunately, in my early draft I mixed the two notations, but now all should be consistent.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+# Coefficients of the linear combination A = c_0 A_0 + c_1 A_1 ...
+c = np.array([1.0, 0.2, 0.2])
+
+
+def U_b():
+    """Unitary matrix rotating the ground state to the problem vector |b> = U_b |0>.
+    """
+    for idx in range(n_qubits):
+        qml.Hadamard(wires=idx)
```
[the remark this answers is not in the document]
andreamari: Here I use idx to apply a layer of H gates, one for each qubit. 
The matrix U_b doesn't have an index because it represent a single unitary which generates the problem state |b>.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,556 @@
...
+
+    # If necessary, further unitary components can be added below.
+    # elif idx==3:
+    # ...
+
+
+def U_b_dag():
+    """Adjoint of the unitary matrix U_b, where |b> = U_b |0>.
+    """
+    U_b()  # In this particular example the operation is self-adjoint.
```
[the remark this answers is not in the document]
andreamari: I removed all the functions for the adjoint operations. Now when I need this, I write a comment saying that, for this specific example, U is self-adjoint.

implementations/tutorial_vqls.py
andreamari: Initially `weights[j] ` represented the wights for the layer `j`. However, since in this example we use a single variational layer, I removed the layer axis and so `weights[0]` is now simply `weights`. For the same reason I removed the unnecessary hyper-parameter `q_depth` at the top of the code.

andreamari:
I should have replied to all comments and corrections. From my side there are still two open questions: 
1. Should I turn black the currently blue projectors |0><0| and P.
2. The current phase-shift gate in the circuit figure is R_\phi with \phi=-\pi/2. If this can be confused as an amplitude rotation we could use S^\dagger as in ref. [1], or something else.

implementations/tutorial_vqls.py
```diff
@@ -0,0 +1,518 @@
...
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+##############################################################################
+# Given the variational weights ``w`` that we have previously optimized,
+# we can generate the quantum state :math:`|x\rangle`. By measuring :math:`|x\rangle`
+# in the computational basis we can estimate the probability of each basis state.
+#
+# For this task, we initialize a new PennyLane device and define the associated
+# *qnode* object.
```
co9olguy:
```suggestion
# *qnode* circuit.
```

implementations/tutorial_vqls.py
```diff
@@ -137,42 +137,45 @@
...
-where :math:`Z_j, X_j, H_j` represent the Pauli Z, Pauli X and Hadamard gates applied to the qubit with index :math:`j`.
+where :math:`Z_j, X_j, H_j` represent the Pauli :math:`Z`, Pauli :math:`X` and Hadamard gates applied to the qubit with index :math:`j`.
 
-This problem is computationally quite easy since a single layer of local rotations is enough to generate the 
+This problem is computationally quite easy since a single layer of local rotations is enough to generate the
 solution state, i.e., we can use the following simple ansatz:
 
 .. math::
-        |x\rangle = V(w) |0\rangle = R_y(w_1) \otimes  R_y(w_2) \otimes  R_y(w_3) |0\rangle.
+        |x\rangle = V(w) |0\rangle = \Big [  R_y(w_0) \otimes  R_y(w_1) \otimes  R_y(w_2) \Big ]  H_0  H_1  H_2 |0\rangle.
```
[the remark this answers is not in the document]
andreamari: One could define, similarly to H_0, H_1, H_2, the rotation operators Ry_0 Ry_1 and Ry_2 living in the full Hilbert space but acting non-trivially on qubits 0,1 and 2. But this would require too many sub_indeces and creating confusion. So, in this case, I preferred to explicitly write the tensor product of three single-qubit operations.

implementations/tutorial_vqls.py
```diff
@@ -251,7 +251,7 @@ def variational_block(weights):
     """Variational circuit mapping the ground state |0> to the ansatz state |x>."""
     # We first prepare an equal superposition of all the states of the computational basis.
     for idx in range(n_qubits):
-        qml.Hadamard(wires=idx)
+     qml.Hadamard(wires=idx)
```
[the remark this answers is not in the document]
andreamari: fixed

andreamari:
Changes applied, including a new figure with a new phase-gate symbol.
As explained in my reply to the specific comment, I would keep the `\otimes` in the formula.
But I am also open to an alternative notation.

PR #3, opened 2019-10-16
Add rotoselect tutorial

angusjlowe:
**Title:** Rotoselect (circuit structure learning)

**Summary:** Tutorial for implementing circuit structure learning algorithm that employs a closed form expression for the optimal angle parameters given a fixed slice in parameter space. Depth of the circuit is held constant while the optimal generators of rotation are learned. The performance of the algorithm is compared with vanilla gradient descent and an arbitrary circuit ansatz. 

**Relevant references:** https://arxiv.org/abs/1905.09692 

**Possible Drawbacks:** n/a

**Related GitHub Issues:** Issue #223 from PennyLane repo.


.gitignore
```diff
@@ -2,3 +2,4 @@ _build
 __pycache__
 tutorial
 app
+.idea
```
co9olguy:
```suggestion
.idea

```

angusjlowe:
@co9olguy No, apart from updating paths to figures

PR #8, opened 2019-10-21
Update advanced usage tutorial with tensor observable feature

AroosaIjaz:


beginner/tutorial_advanced_usage.py
```diff
@@ -56,21 +56,34 @@ def circuit1(param):
...
+@qml.qnode(dev)
+def circuit2(param):
+    qml.RX(param, wires=0)
+    qml.CNOT(wires=[0, 1])
+    return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+print(circuit2(np.pi / 2))
+
+##############################################################################
+# This outcome is expected, as both spins are either up or down.
```
co9olguy:
```suggestion
# Notice how this expectation value differs from the local versions above.
```

beginner/tutorial_advanced_usage.py
```diff
@@ -56,21 +56,34 @@ def circuit1(param):
...
 #
-#     It is important to emphasize that the expectation values in ``circuit`` are both **local**,
+#     It is important to emphasize that the expectation values in ``circuit1`` are both **local**,
 #     i.e., this circuit is evaluating :math:`\left\langle \sigma_z\right\rangle_0` and :math:`\left\langle \sigma_z\right\rangle_1`,
-#     not :math:`\left\langle \sigma_z\otimes \sigma_z\right\rangle_{01}` (where the subscript denotes which wires the
-#     observable is located on).
+#     not :math:`\left\langle \sigma_z\otimes \sigma_z\right\rangle_{01}` (where the subscript
+#     denotes which wires the observable is located on). 
+#
+# In order to measure :math:`\langle\sigma_z \otimes \sigma_z \rangle _{01}`, PennyLane's tensor
```
co9olguy:
```suggestion
# In order to measure a tensor product observable like :math:`\langle\sigma_z \otimes \sigma_z \rangle _{01}`, 
```

beginner/tutorial_advanced_usage.py
```diff
@@ -56,21 +56,34 @@ def circuit1(param):
...
+# observable feature can be used.
```
co9olguy:
```suggestion
# the matrix multiplication operator ``@`` can be used:
```

beginner/tutorial_advanced_usage.py
```diff
@@ -56,21 +56,34 @@ def circuit1(param):
...
+
+##############################################################################
+# This outcome is expected, as both spins are either up or down.
 #
 # We may even mix different return types, for example expectation values and variances:
 
 
 @qml.qnode(dev)
-def circuit1(param):
+def circuit3(param):
```
co9olguy: Note that there is already a `circuit3` below, so if we want every circuit to have a different index, we'll need to update below

josh146: I generally don't think this is strictly necessary --- most people reading should be familiar with Python, and should follow that a new definition (using the same name) is being created.

co9olguy: in that case, we could just call every one `circuit`

beginner/tutorial_advanced_usage.py
```diff
@@ -79,7 +79,7 @@ def circuit2(param):
 
 
 @qml.qnode(dev)
-def circuit3(param):
+def circuit(param):
```
co9olguy: @AroosaIjaz note that there is still a `circuit3` below

AroosaIjaz: The one in the other section? okay, I will update that too.

PR #12, opened 2019-10-28
Tutorial coherent vqls

andreamari:

**Title:** New implementaion: *the coherent variational quantum linear solver (CVQLS)*.

**Summary:** . A tutorial on the VQLS already exists. In this new tutorial, the matrix associated the linear problem is applied as a coherent quantum operation. 

**Relevant references:** As far as I know, this is a new approach, never considered in the context of variational solvers. For the (non-coherent) VQLS, see e.g.: Bravo-Prieto et al.. arXiv:1909.05820, (2019).
 
**Possible Drawbacks:**  ---

**Related GitHub Issues:** ---


implementations.rst
```diff
@@ -69,6 +69,12 @@ algorithms using PennyLane and near-term quantum hardware.
     :figure: implementations/doubly_stochastic/single_shot.png
     :description: :doc:`Doubly stochastic gradient descent <app/tutorial_doubly_stochastic>`
 
+.. customgalleryitem::
+    :tooltip: Coherent Variational Quantum Linear Solver
```
josh146: Do you want to put a tooltip that is different from the title? Maybe a bit more explanatory/a short summary?

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+In the code presented below we solve this particular problem, by following the general scheme of the CVQLS previously discussed.
+Eventually we will compare the quantum solution with the classical one.
+
+"""
+
+
+##############################################################################
+# General setup
+# ------------------------
+# This Python code requires *PennyLane* and the plotting library *matplotlib*.
```
josh146: Might as well extend the module docstring down to here. This avoids an unfortunate bug in our server Sphinx, in which the first heading directly following the module docstring is not rendered correctly.

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+
+# Pennylane
+import pennylane as qml
+from pennylane import numpy as np
+
+# Plotting
```
josh146: this comment is probably not needed

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+
+"""
+
+
+##############################################################################
+# General setup
+# ------------------------
+# This Python code requires *PennyLane* and the plotting library *matplotlib*.
+
+# Pennylane
```
josh146: this comment is probably not needed

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+# amplitudes with just three gates:
+# a local :math:`R_y` rotation, a controlled-:math:`R_y` and a controlled-NOT.
+
+
+def U_c():
+    """Unitary matrix rotating the ground state of the ancillary qubits to |sqrt(c)> = U_c |0>."""
+    # Circuit mapping |00> to sqrt_c[0] |00> + sqrt_c[1] |01> + sqrt_c[2] |10>
+    qml.RY(-2 * np.arccos(sqrt_c[0]), wires=ancilla_idx)
+    qml.CRY(-2 * np.arctan(sqrt_c[2] / sqrt_c[1]), wires=[ancilla_idx, ancilla_idx + 1])
+    qml.CNOT(wires=[ancilla_idx + 1, ancilla_idx])
```
josh146: Does it make sense to make this function take `sqrt_c` (or even `c`) as an argument?

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+# function, it is useful to express everything in terms of expectation values through the
+# the Bayes' theorem:
+#
+# .. math::
+#   |\langle b | \Psi \rangle|^2=
+#   P( \mathrm{sys}=\mathrm{ground}\,|\, \mathrm{anc} = \mathrm{ground}) =
+#   P( \mathrm{all}=\mathrm{ground})/P( \mathrm{anc}=\mathrm{ground})
+#
+# To evaluate the two probabilities appearing on the r.h.s. of the previous equation
+# we initialize two PennyLane devices with the ``default.qubit`` backend,
```
josh146:
```suggestion
# we initialize two PennyLane devices with the ``default.qubit`` device,
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+# directly make use the measurement samples. However, since we want to optimize the cost
+# function, it is useful to express everything in terms of expectation values through the
+# the Bayes' theorem:
+#
+# .. math::
+#   |\langle b | \Psi \rangle|^2=
+#   P( \mathrm{sys}=\mathrm{ground}\,|\, \mathrm{anc} = \mathrm{ground}) =
+#   P( \mathrm{all}=\mathrm{ground})/P( \mathrm{anc}=\mathrm{ground})
+#
+# To evaluate the two probabilities appearing on the r.h.s. of the previous equation
```
josh146:
```suggestion
# To evaluate the two probabilities appearing on the right hand side of the previous equation
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+def global_ground(weights):
+    # Circuit gates
+    full_circuit(weights)
+    # Projector on the global ground state.
+    P = np.zeros((2 ** tot_qubits, 2 ** tot_qubits))
+    P[0, 0] = 1.0
+    return qml.expval(qml.Hermitian(P, wires=range(tot_qubits)))
+
+
+dev_partial = qml.device("default.qubit", wires=tot_qubits)
```
josh146: It seems like `dev_partial` and `dev_full` are identical - why not just use define device?

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+dev_full = qml.device("default.qubit", wires=tot_qubits)
+
+@qml.qnode(dev_full)
+def global_ground(weights):
+    # Circuit gates
+    full_circuit(weights)
+    # Projector on the global ground state.
+    P = np.zeros((2 ** tot_qubits, 2 ** tot_qubits))
+    P[0, 0] = 1.0
+    return qml.expval(qml.Hermitian(P, wires=range(tot_qubits)))
```
josh146: We should add a `StateProjector` observable 🙂

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+
+##############################################################################
+# Given the variational weights ``w`` that we have previously optimized,
+# we can generate the quantum state :math:`|x\rangle`. By measuring :math:`|x\rangle`
+# in the computational basis we can estimate the probability of each basis state.
+#
+# For this task, we initialize a new PennyLane device and define the associated
+# *qnode* circuit.
```
josh146:
```suggestion
# QNode.
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,550 @@
...
+
+    # Variational circuit generating a guess for the solution vector |x>
+    variational_block(weights)
+
+    # We assume that the system is measured in the computational basis.
+    # If we label each basis state with a decimal integer j = 0, 1, ... 2 ** n_qubits - 1,
+    # this is equivalent to a measurement of the following diagonal observable.
+    basis_obs = qml.Hermitian(np.diag(range(2 ** n_qubits)), wires=range(n_qubits))
+
+    return qml.sample(basis_obs)
```
josh146: Perhaps we should make `qml.sample()` a shortcut to sample the computational basis state...

implementations/tutorial_coherent_vqls.py
andreamari: It could be done but we should consider that the notation would become a bit heavy, e.g.  `U_c(c) `and `U_c_dagger(c)`, and would create a strange asymmetry with respect to U_b().

andreamari:
Applied all suggestions but, for the moment, I am still not using `c` as an argument for `U_c()` (see my reply to Josh's comment).

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+
+
+Coherently applying :math:`A`
+>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+
+The problem of coherently applying a liner combination of unitary operations has been already studied in Ref. [2]
+and here we follow a very similar approach.
+
+Without loss of generality we can assume that the coefficients :math:`c=(c_1, c_2, \dots c_L)` appearing
+in the definition of :math:`A` represent a positive and normalized probability distribution. i.e.
```
co9olguy:
```suggestion
in the definition of :math:`A` represent a positive and normalized probability distribution. i.e.,
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+8. Apply :math:`U_b^\dagger` to the system.
+9. Measure the system in the computational basis. The probability of finding it
+   in the ground state (given the ancillary qubits measured in their ground state),
+   is :math:`|\langle 0 | U_b^\dagger |\Psi \rangle|^2 = |\langle b | \Psi \rangle|^2`.
+
+So, with sufficiently many shots of the previous experiment, one can directly estimate
+the cost function of the problem.
+
+Importantly, the operations of steps 7 and 8 commute. Therefore all the measurements can be
+delayed at the end of the quantum circuit as represented in the figure at the top of this tutorial,
```
co9olguy:
```suggestion
delayed until the end of the quantum circuit (as shown in the figure at the top of this tutorial),
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+making the structure of the experiment more straightforward.  
+
+A simple example
+^^^^^^^^^^^^^^^^
+
+In this tutorial we apply the previous theory to the following simple example 
+based on a system of 3 qubits, which was already considered in Ref. [1] and also reproduced in PannyLane (VQLS):
```
co9olguy:
```suggestion
based on a system of 3 qubits, which was already considered in Ref. [1] and also reproduced in PennyLane (VQLS):
```

co9olguy: we can add a link now to the other tutorial

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+import pennylane as qml
+from pennylane import numpy as np
+import matplotlib.pyplot as plt
+
+##############################################################################
+# Setting of the main hyper-parameters of the model
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+n_qubits = 3                # Number of system qubits.
+m = 2                       # Number of ancillary qubits
```
co9olguy:
```suggestion
m = 2                       # Number of ancillary qubits.
```
If you're gonna go with periods, you missed one here :)

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+
+##############################################################################
+# We need to embed the square root of the probability distribution ``c`` into the amplitudes
+# of the ancillary state. It is easy to check that one can always embed 3 positive
+# amplitudes with just three gates:
+# a local :math:`R_y` rotation, a controlled-:math:`R_y` and a controlled-NOT.
+
+
+def U_c():
+    """Unitary matrix rotating the ground state of the ancillary qubits to |sqrt(c)> = U_c |0>."""
```
co9olguy:
```suggestion
    """Unitary matrix rotating the ground state of the ancillary qubits 
    to |sqrt(c)> = U_c |0>."""
```
(trying to avoid the horizontal scroll bar in the built html page

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+    # Controlled-A2
+    qml.CNOT(wires=[ancilla_idx + 1, 0])
+
+
+##############################################################################
+# The circuit for preparing the problem vector :math:`|b\rangle` is very simple:
+
+
+def U_b():
+    """Unitary matrix rotating the system ground state to the problem vector |b> = U_b |0>."""
```
co9olguy:
```suggestion
    """Unitary matrix rotating the system ground state to the 
    problem vector |b> = U_b |0>."""
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+##############################################################################
+# Full quantum circuit
+# --------------------
+#
+# Now, we can define the full circuit associated to the CVQLS protocol presented in the introduction and
+# corresponding to the figure at the top of this tutorial.
+
+
+def full_circuit(weights):
+    """Full quantum circuit necessary for the CVQLS protocol, without the final measurement."""
```
co9olguy:
```suggestion
    """Full quantum circuit necessary for the CVQLS protocol, 
    without the final measurement."""
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+    # For this particular problem adjoint(U_b)=U_b
+    U_b()
+
+    # Adjoint of U_c, applied to the ancillary qubits.
+    U_c_dagger()
+
+
+##############################################################################
+# To estimate the overlap of the ground state with the post-selected state, one could
+# directly make use the measurement samples. However, since we want to optimize the cost
```
co9olguy:
```suggestion
# directly make use of the measurement samples. However, since we want to optimize the cost
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+# function, it is useful to express everything in terms of expectation values through the
```
co9olguy:
```suggestion
# function, it is useful to express everything in terms of expectation values through 
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+# the Bayes' theorem:
```
co9olguy:
```suggestion
# Bayes' theorem:
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+    P_anc = np.zeros((2 ** m, 2 ** m))
+    P_anc[0, 0] = 1.0
+    return qml.expval(qml.Hermitian(P_anc, wires=range(n_qubits, tot_qubits)))
+
+
+##############################################################################
+# Variational optimization
+# -----------------------------
+#
+# In order to variationally solve our lineaer problem, we first define the cost function
```
co9olguy:
```suggestion
# In order to variationally solve our linear problem, we first define the cost function
```

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,541 @@
...
+# :math:`C = 1- |\langle b | \Psi \rangle|^2` that we are going to minimize.
+# As explained above, we express it in terms of expectation values thorugh the Bayes' theorem.
```
co9olguy:
```suggestion
# As explained above, we express it in terms of expectation values through Bayes' theorem.
```

implementations/tutorial_coherent_vqls.py
andreamari: Removed all of them :-)

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,544 @@
+r"""
+.. _coherent_vqls:
+
+Coherent Variational Quantum Linear Solver
+==========================================
+*Author: Andrea Mari*
+
+In this tutorial we propose and implement an algorithm that we call
+*coherent variational quantum linear solver* (CVQLS). 
+This is inspired by the VQLS proposed in Ref. [1], with an important difference:
```
josh146:
```suggestion
This is inspired by the :doc:`VQLS <tutorial_vqls>` proposed in Ref. [1], with an important difference: 
```

josh146: Although, you might even want to explicitly say something along the lines of "see the previous demo on VQLS for more details" etc.

implementations/tutorial_coherent_vqls.py
andreamari: done

andreamari:
> There seems to be an unrelated image _static/thumbs/bloch.gif connected to this PR. Do you know why it is showing up in this PR as new/changed?

That image is created each time I run `sphinx-build` (I don't know why) and I probably forgot to discard that change in one of my past commits. Anyway now I should have removed that change form this PR.

implementations/tutorial_coherent_vqls.py
```diff
@@ -0,0 +1,544 @@
...
+
+
+Coherently applying :math:`A`
+>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+
+The problem of coherently applying a liner combination of unitary operations has been already studied in Ref. [2]
+and here we follow a very similar approach.
+
+Without loss of generality we can assume that the coefficients :math:`c=(c_1, c_2, \dots c_L)` appearing
+in the definition of :math:`A` represent a positive and normalized probability distribution. i.e.,
```
co9olguy:
```suggestion
in the definition of :math:`A` represent a positive and normalized probability distribution, i.e.,
```

co9olguy:
Looks great, ready to go :)

PR #16, opened 2019-11-01
Ported state prep tutorial to pytorch

josh146:
Updates the State Preparation demo to use PyTorch

### DIFF
--- a/implementations/tutorial_state_preparation.py
+++ b/implementations/tutorial_state_preparation.py
@@ -1,8 +1,8 @@
 r"""
 .. _state_preparation:
 
-State preparation on Rigetti Forest
-===================================
+State preparation with Rigetti Forest + PyTorch
+===============================================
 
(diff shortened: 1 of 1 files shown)


implementations/tutorial_state_preparation.py
```diff
@@ -31,9 +31,15 @@
...
+#
+# To start, we import PennyLane, NumPy, and PyTorch for the optimization:
 
 import pennylane as qml
-from pennylane import numpy as np
+import numpy as np
+import torch
+from torch.autograd import Variable
+
+np.random.seed(42)
```
ixfoduap: Hehehe 42

co9olguy:
> Looks good! (Though I'm not fluent in PyTorch)
> What is the motivation to use PyTorch here?

We want to move away from every single demo using "default.qubit" and numpy interface.

PR #24, opened 2020-01-03
Data uploading classifier with fidelity calculated directly

quantshah:
The tutorial used the internal `dev._state` to compute fidelities which lead to errors:

Now, the fidelities are calculated by directly finding the expectation value of the Hermitian observable representing the expected output state. This solves the issue and we do not need to compute the fidelity from the internal state anymore.

**Relevant references:**
https://discuss.pennylane.ai/t/problem-using-cirq-plugin-as-backend/274


josh146:
Thanks for this @quantshah!

PR #29, opened 2020-01-18
[REDESIGN] Allows demos to be filtered by frameworks

josh146:
**Summary:**

Allows demos to now be filtered by tags. Each tutorial thumbnail now accepts tags, for example:

```rest
.. customgalleryitem::
    :tooltip: Do arbitrary state preparation on a real quantum computer.
    :figure: implementations/state_preparation/NOON.png
    :description: :doc:`app/tutorial_state_preparation`
    :tags: forest pytorch
```

These correspond to filters that can be chosen by the user on the right column:
image

**Possible Drawbacks:**

The filter buttons currently act as radio buttons - only one can be selected at a time. We could change these to checkboxes (more than one selected at a time), but this means filters have to be clicked a second time to be de-selected.

I think for now that radio buttons are sufficient. As the demos grow, we can think about how to modify the UI.


mariaschuld:
Looks great. I built it and it works like a charm!

In the longer term we could add other tags, like the optimizer, which measurement or template is used etc.

josh146:
> Would they be better at the bottom of the page for the time being?

The only issue with putting them at the bottom is that when you click on one, the filters would jump upwards as the page 're-paginates', leading to bad UI.

Could put them in the sidebar (where the local TOC goes), but the sidebar is 'global' and embedded in all pages, so it's a bit more work.

josh146:
<h3>Website build</h3>
<strong>Commit:</strong> 2f8ab3f3aa8f250a25943640a83f091c6643b5e6

<strong>Circle build number:</strong> 476

<strong>Website zip:</strong> <a href=https://476-214003948-gh.circle-artifacts.com/0/tmp/qml_html.zip>https://476-214003948-gh.circle-artifacts.com/0/tmp/qml_html.zip</a>

<strong>Website url:</strong> <a href=https://476-214003948-gh.circle-artifacts.com/0/_build/html/index.html>https://476-214003948-gh.circle-artifacts.com/0/_build/html/index.html</a>

mariaschuld:
I think it's really cool with the filters! If they are easy to hide initially, that may be even better. But I think they invite to play around without being too overwhelmed by the tutorials.

josh146:
@co9olguy, I've modified the tags so that they display now in the sidebar, rather than on top of the thumbnails; see [the bot comment](https://github.com/XanaduAI/qml/pull/29#issuecomment-586825812) for a live demo :slightly_smiling_face:

co9olguy:
Looks great @josh146 
Now we can easily see what frameworks we need to have broader tutorial coverage for as well

PR #37, opened 2020-02-10
Tutorials: VQE with parallel QPUs on Rigetti-Forest

trbromley:
This tutorial shows:
- A comparison of parallel and sequential evaluation on two simulators
- Example of using PennyLane-Forest package
- How to create potential energy surface for a molecule
- Example of using the qchem package
- EDIT: Corrected links in previous QChem tutorials

Not included:
- Details of circuit depths for running on device topology
- Mention of new parametric circuits feature, which does come in to play but might be more naturally mentioned elsewhere.

Credit to @agran2018 for data files!
Credit to @jswinarton for the help with CircleCI + docker!


.circleci/config.yml
```diff
@@ -7,6 +7,12 @@ jobs:
   build:
     docker:
       - image: circleci/python:3.7.2
+      - image: rigetti/quilc
+        command:
+        - -S
+      - image: rigetti/qvm
+        command:
+        - -S
```
josh146: Thanks @jswinarton for helping with this!

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+#
+# Each atomic separation results in a different qubit Hamiltonian. To find the corresponding
+# Hamiltonian, we use the :func:`~.generate_hamiltonian` function of the :mod:`~.qchem` package.
+# Further details on the mapping from the electronic Hamiltonian of a molecule to a qubit
+# Hamiltonian can be found in the :ref:`Quantum Chemistry with PennyLane <qchem-beginner>` and
+# :ref:`A brief overview of VQE <qchem-implementations>` tutorials.
+#
+# We begin by creating a dictionary containing molecular separations and corresponding data files
+# saved in ```.xyz``` format. The data files can be downloaded by clicking :download:`here
+# <../implementations/vqe_parallel/h2.zip>`.
```
josh146: Awesome!

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+hamiltonians = [
+    qchem.generate_hamiltonian(
+        mol_name=str(separation),
+        mol_geo_file=file,
+        mol_charge=0,
+        multiplicity=1,
+        basis_set="sto-3g",
+        n_active_electrons=2,
+        n_active_orbitals=2,
+    )[0]
```
josh146: Might be worth mentioning why we only take the 0th element (i.e., the actual hamiltonian lol)

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+# ----------------------------
+#
+# The fifteen Pauli terms comprising each Hamiltonian can conventionally be evaluated in a
+# sequential manner: we evaluate one expectation value at a time before moving on to the next.
+# However, this task is highly suited to parallelization. With access to multiple QPUs,
+# we can split up evaluating the terms between the QPUs and gain an increase in processing speed.
+#
+# Note that some of the Pauli terms commute, and so they can be evaluated in practice with fewer
+# than fifteen quantum circuit runs. Nevertheless, these quantum circuit runs can still be
+# parallelized to multiple QPUs.
```
josh146: Worth having as a note box?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+devs = devs_4 + devs_7
+
+##############################################################################
+# We must also define a circuit to prepare the ground state. The simple circuit below is able to
+# accurately capture the ground state encoded into qubits. It has a single free parameter setting
+# a Y-rotation on the third qubit.
+
+
+def circuit(param, wires):
+    qml.BasisState(np.array([1, 1, 0, 0]), wires=[0, 1, 2, 3])
```
josh146: Is it worth explaining why we use this basis state (that it is the HF state?)

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+##############################################################################
+# Finally, the energies as functions of rotation angle can be given using :class:`~.VQECost`.
+
+energies = [qml.VQECost(circuit, h, devs) for h in hamiltonians]
+
+##############################################################################
+# Calculating the potential energy surface
+# ----------------------------------------
+#
+# :class:`~.VQECost` returns a :class:`QNodeCollection` which can be evaluated using the input
```
josh146: Just checking that the second link resolves?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+# parameters to the ansatz circuit. As of ``v0.8.0`` of PennyLane, the :class:`QNodeCollection`
```
josh146:
```suggestion
# parameters to the ansatz circuit. The :class:`QNodeCollection`

```

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+# can be evaluated asynchronously by specifying the ``parallel = True`` keyword argument.
```
josh146:
```suggestion
# can be evaluated asynchronously by passing the keyword argument ``parallel=True``.

```

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,192 @@
...
+# With the default ``parallel = False``, the QNodes are evaluated sequentially.
```
josh146:
```suggestion
# When ``parallel=False`` (the default behaviour), the QNodes are instead evaluated sequentially.
```

implementations/tutorial_vqe_parallel.py
trbromley: This is fixed, although the `qchem`-based functions are not linking. This is also the same for the VQE tutorial (https://pennylane.ai/qml/app/tutorial_vqe.html, see "PennyLane contains the VQECost...").

I think this is something to do with Intersphinx not being able to see the qchem package.

josh146:
Oh, I think this is because of entry points. In this case, you have to specify the package directly, i.e.,

```rest
:func:`~pennylane_qchem.qchem.generate_haniltonian`
```

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,194 @@
...
+
+energies = [qml.VQECost(circuit, h, devs) for h in hamiltonians]
+
+##############################################################################
+# Calculating the potential energy surface
+# ----------------------------------------
+#
+# :class:`~.VQECost` returns a :class:`QNodeCollection` which can be evaluated using the input
+# parameters to the ansatz circuit. The :class:`QNodeCollection`
+
```
soranjh: remove the empty line here?

.circleci/config.yml
josh146: One question though - should quilc instead be using the -R command, since http mode is deprecated? Also, if the ports aren't specified, does it fallback to the default value?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,194 @@
...
+# speed up in processing:
+
+print("Speedup: {0:.2f}".format(t_seq / t_par))
+
+##############################################################################
+# Can you think of other ways to combine multiple QPUs to improve the
+# performance of quantum algorithms? To conclude the tutorial, let's plot the calculated
+# potential energy surface:
+
+plt.plot(surface, linewidth=2.2, marker="o")
```
soranjh: Might be interesting to plot both surfaces and show that they are exactly the same.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,198 @@
+# coding=utf-8
+r"""
+VQE with parallel QPUs on Rigetti-Forest
+========================================
+
+This tutorial showcases how using asynchronously evaluated parallel QPUs can speed up calculation
+of the potential energy surface of molecular hydrogen.
```
agran2018: molecular hydrogen (:math:`H_2`)

trbromley: Thanks!

.circleci/config.yml
trbromley: :thinking: I'm not sure, I was just using the man page here: https://github.com/rigetti/quilc/blob/master/doc/man/quilc.1 although maybe it is outdated. There it looks like `-R` and `-S` are equivalent. Do you have some more up to date info?

Yes, if no port is specified then `quilc` is available on port `5555` and `qvm` is available on `5000`. These defaults seem to be what the qc in `pyquil` looks for, although PL-Forest suggest that the default port of the compiler is `6000` (https://pennylane-forest.readthedocs.io/en/latest/usage.html).

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+# .. note::
+#    Some of the Pauli terms commute, and so they can be evaluated in practice with fewer than
+#    fifteen quantum circuit runs. Nevertheless, these quantum circuit runs can still be
+#    parallelized to multiple QPUs.
+#
+# Let's suppose we have access to two QPUs: ``Aspen-4-4Q-E`` and ``Aspen-7-4Q-C`` from
+# Rigetti. We can evaluate the expectation value of each Hamiltonian with eight terms run on
+# ``Aspen-4-4Q-E`` and seven terms run on ``Aspen-7-4Q-C``, as summarized by the diagram below:
+#
+# .. figure:: /implementations/vqe_parallel/diagram.png
```
ixfoduap: Did @trbromley make this image?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+
+# Use forest.qpu if hardware access is available
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
+devs_7 = [qml.device("forest.qvm", device="Aspen-7-4Q-C") for _ in range(7)]
+devs = devs_4 + devs_7
+
+##############################################################################
+# We must also define a circuit to prepare the ground state. The simple circuit below is able to
+# prepare states of the form :math:`\alpha |1100\rangle + \beta |0011\rangle` which encode the
+# ground state wave function of the hydrogen molecule in qubits. The circuit has a single free
```
ixfoduap:
```suggestion
# ground state wave function of the hydrogen molecule. The circuit has a single free

```

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+# parameter setting a Y-rotation on the third qubit.
```
ixfoduap:
```suggestion
# parameter, which controls a Y-rotation on the third qubit.

```

ixfoduap: Also, why do we use a Y rotation?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+    t1 = time.time()
+
+    print("Evaluation time: {0:.2f} s".format(t1 - t0))
+    return s, t1 - t0
+
+
+print("Evaluating the potential energy surface sequentially")
+surface, t_seq = calculate_surface(parallel=False)
+
+print("\nEvaluating the potential energy surface in parallel")
```
ixfoduap: For my learning: what does the \n do?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+surface, t_par = calculate_surface(parallel=True)
+
+##############################################################################
+# We have seen how a :class:`~.pennylane.QNodeCollection` can be evaluated in parallel and the
```
ixfoduap: This sentence reads a bit awkwardly, but not sure how to improve it...

.circleci/config.yml
trbromley: I've updated to -R after an in-person discussion.

implementations/tutorial_vqe_parallel.py
trbromley: Added!

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,208 @@
...
+# Use forest.qpu if hardware access is available
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
+devs_7 = [qml.device("forest.qvm", device="Aspen-7-4Q-C") for _ in range(7)]
+devs = devs_4 + devs_7
+
+##############################################################################
+# We must also define a circuit to prepare the ground state. The simple circuit below is able to
+# prepare states of the form :math:`\alpha |1100\rangle + \beta |0011\rangle` which encode the
+# ground state wave function of the hydrogen molecule. The circuit has a single free
+
```
josh146: extra line!

josh146:
```suggestion
```

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,208 @@
...
+# Use forest.qpu if hardware access is available
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
+devs_7 = [qml.device("forest.qvm", device="Aspen-7-4Q-C") for _ in range(7)]
+devs = devs_4 + devs_7
+
+##############################################################################
+# We must also define a circuit to prepare the ground state. The simple circuit below is able to
+# prepare states of the form :math:`\alpha |1100\rangle + \beta |0011\rangle` which encode the
+# ground state wave function of the hydrogen molecule. The circuit has a single free
```
josh146: I find this bit a bit confusing, in that the HF state comes out of thin air, and I have the following questions:

1. How is this relevant wrt the variational circuit below? _Answer by Tom: the circuit implements the parametrized state `a(r)|1100> + b(r) |0011>` where `a^2 + b^2 = 1`_

2. What is the connection between  `a(r)|1100> + b(r) |0011>` and H2?

3. If I have another molecule, how do I calculate the HF state?

Maybe this doesn't need to be explained here, but in another tutorial and a link placed here.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,208 @@
+# coding=utf-8
+r"""
+.. _vqe-parallel:
+
+VQE with parallel QPUs on Rigetti-Forest
```
josh146: Not sure it has a hyphen!
```suggestion
VQE with parallel QPUs on Rigetti Forest
```

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,208 @@
...
+def circuit(param, wires):
+    qml.BasisState(np.array([1, 1, 0, 0]), wires=[0, 1, 2, 3])
+    qml.RY(param, wires=2)
+    qml.CNOT(wires=[2, 3])
+    qml.CNOT(wires=[2, 0])
+    qml.CNOT(wires=[3, 1])
+
+
+##############################################################################
+# Each molecular separation has a different Y-rotation parameter. The values of these
```
agran2018: It might be convenient to rephrase the first sentence as follows:

"The ground state for each inter-atomic distance characterizes by a different Y-rotation angle, this is because the contribution of the Hartree-Fock (:math:`|1100\rangle`) and doubly-excited (:math:`|0011\rangle`) configurations to ground-state wave function varies along the potential energy surface."

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,208 @@
...
+#
+# The objective of this tutorial is to evaluate the potential energy surface of molecular
+# hydrogen. This is achieved by finding the ground state energy of :math:`H_{2}` for a range of
+# atomic separations.
+#
+# Each atomic separation results in a different qubit Hamiltonian. To find the corresponding
+# Hamiltonian, we use the :func:`~.pennylane_qchem.qchem.generate_hamiltonian` function of the
+# :mod:`~.pennylane_qchem.qchem` package. Further details on the mapping from the electronic
+# Hamiltonian of a molecule to a qubit Hamiltonian can be found in the :ref:`Quantum Chemistry
+# with PennyLane <qchem-beginner>` and :ref:`A brief overview of VQE <qchem-implementations>`
```
josh146: I'm not sure if this will work, but if linking to a whole page rather than a section, `:doc:` links are preferable to `:ref:` links where possible.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,208 @@
...
+# We can use this feature to compare the sequential and parallel times required to calculate the
+# potential energy surface. The following function calculates the surface:
+
+
+def calculate_surface(parallel=True):
+    s = []
+    t0 = time.time()
+
+    for i, e in enumerate(energies):
+        print("Running for molecular separation {} Å".format(list(data.keys())[i]))
```
agran2018:
```suggestion
        print("Running for inter-atomic distance {} Å".format(list(data.keys())[i]))

```

implementations/tutorial_vqe_parallel.py
trbromley: @agran2018 @ixfoduap @soranjh any thoughts?

soranjh: If the focus of this tutorial is on q-parallel computing, then probably we can assume that the user knows the basics of calculating the HF state from previous tutorials. I agree with @josh146 that we can just cite those tutorials for people who want more details.

implementations/tutorial_vqe_parallel.py
trbromley: Done!

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+#
+#    pip install pennylane-qchem
+#    pip install pennylane-forest
+#    pip install dask[delayed]
+#
+# Finding the qubit Hamiltonians of :math:`H_{2}`
+# -----------------------------------------------
+#
+# The objective of this tutorial is to evaluate the potential energy surface of molecular
+# hydrogen. This is achieved by finding the ground state energy of :math:`H_{2}` for a range of
```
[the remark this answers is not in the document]
trbromley: Thanks! Am putting these in manually

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+    1.1: "vqe_parallel/h2_1.10.xyz",
+    1.3: "vqe_parallel/h2_1.30.xyz",
+    1.5: "vqe_parallel/h2_1.50.xyz",
+    1.7: "vqe_parallel/h2_1.70.xyz",
+    1.9: "vqe_parallel/h2_1.90.xyz",
+    2.1: "vqe_parallel/h2_2.10.xyz",
+}
+
+##############################################################################
+# The next step is to create the qubit Hamiltonians for each value of molecular separation.
```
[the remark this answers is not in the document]
trbromley: Thanks!

implementations/tutorial_vqe_parallel.py
trbromley: Nice! Thanks a lot, have added this in.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# To do this, start by instantiating a device for each term:
+
+# Use forest.qpu if hardware access is available
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
+devs_7 = [qml.device("forest.qvm", device="Aspen-7-4Q-C") for _ in range(7)]
+devs = devs_4 + devs_7
+
+##############################################################################
+# We must also define a circuit to prepare the ground state. The simple circuit below is able to
```
agran2018: "Within a minimal basis approximation, the ground state wave function of the hydrogen molecule :math:`\vert \Psi \rangle = \alpha \vert 1100 \rangle + \beta \vert 0011 \rangle` consists of only the Hartree-Fock component and a doubly-excited configuration where the two electrons occupy the highest-energy molecular orbitals."

implementations/tutorial_vqe_parallel.py
agran2018: > Is it worth explaining why we use this basis state (that it is the HF state?)

I think that a short explanation like this would be OK:

"Notice that in close analogy with post-Hartree-Fock electron correlation methods, we initialize the qubit register to encode the HF state. Then, we apply an arbitrary rotation on the third qubit and the subsequent CNOT gates to prepared the correlated wave function of the molecule."

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+surface_seq, t_seq = calculate_surface(parallel=False)
+
+print("\nEvaluating the potential energy surface in parallel")
+surface_par, t_par = calculate_surface(parallel=True)
+
+##############################################################################
+# We have seen how a :class:`~.pennylane.QNodeCollection` can be evaluated in parallel. This results
+# in a speed up in processing:
+
+print("Speedup: {0:.2f}".format(t_seq / t_par))
```
co9olguy: Just above you use "speed up" (two words)

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# .. figure:: /implementations/vqe_parallel/diagram.png
+#    :width: 65%
+#    :align: center
+#
+# To do this, start by instantiating a device for each term:
+
+# Use forest.qpu if hardware access is available
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
+devs_7 = [qml.device("forest.qvm", device="Aspen-7-4Q-C") for _ in range(7)]
```
co9olguy: Getting `Lattice with name 'Aspen-7-4Q-C' not found.` when I try to run locally

implementations/tutorial_vqe_parallel.py
co9olguy: I think for this demo, we should try to keep the technical material a bit lighter. Enough to understand what the basic setting is (VQE on H2 molecule), but no more

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,200 @@
...
+
+##############################################################################
+# This tutorial requires the ``pennylane-qchem``, ``pennylane-forest`` and ``dask``
+# packages, which are installed separately using:
+#
+# .. code-block:: bash
+#
+#    pip install pennylane-qchem
+#    pip install pennylane-forest
+#    pip install dask[delayed]
```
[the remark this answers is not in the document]
josh146: That is really strange --- the `pip` and setuptools documentation both specify non-quotations should work fine (and is what I am much more used to seeing, see https://pip.pypa.io/en/stable/reference/pip_install/#examples). We should update it though, Maria had the same problem at some stage.

@co9olguy, what was the error message?

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# Finding the qubit Hamiltonians of :math:`H_{2}`
+# -----------------------------------------------
+#
+# The objective of this tutorial is to evaluate the potential energy surface of molecular
+# hydrogen. This is achieved by finding the ground state energy of :math:`H_{2}` as we increase
+# the bond length between the hydrogen atoms.
+#
+# Each inter-atomic distance results in a different qubit Hamiltonian. To find the corresponding
+# Hamiltonian, we use the :func:`~.pennylane_qchem.qchem.generate_hamiltonian` function of the
```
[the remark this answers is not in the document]
josh146: :cry:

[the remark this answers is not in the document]
josh146: Oh, you know why? I must have never fixed it in the QML version of the theme, only in the PennyLane version of the theme. We should consider having the theme in a single central location for all repos, so that bug fixes don't have to be made multiple times.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+# ``Aspen-4-4Q-E`` and seven terms run on ``Aspen-7-4Q-C``, as summarized by the diagram below:
+#
+# .. figure:: /implementations/vqe_parallel/diagram.png
+#    :width: 65%
+#    :align: center
+#
+# To do this, start by instantiating a device for each term:
+
+# Use forest.qpu if hardware access is available
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
```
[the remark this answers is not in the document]
josh146: This can be considered a 'bug', and we should make a ticket to fix this. Unfortunately, it is likely due to race conditions that require a broader refactor.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+    1.7: "vqe_parallel/h2_1.70.xyz",
+    1.9: "vqe_parallel/h2_1.90.xyz",
+    2.1: "vqe_parallel/h2_2.10.xyz",
+}
+
+##############################################################################
+# The next step is to create the qubit Hamiltonians for each value of the inter-atomic distance.
+
+hamiltonians = [
+    qchem.generate_hamiltonian(
```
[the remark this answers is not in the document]
trbromley: Yes! I've gotten rid of `n_active_electrons` and `n_active_orbitals`. The rest do not have a default. We could also remove the keywords, although I find that including keywords makes it easier.

implementations/tutorial_vqe_parallel.py
trbromley: Thanks, made it two words here.

implementations/tutorial_vqe_parallel.py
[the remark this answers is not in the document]
trbromley: I agree that ideally we have as many logical devices as there are physical devices, but doing this causes an error. The error is a `quilc` timeout error.

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+        n_active_orbitals=2,
+    )[0]  # We take the zero element (the Hamiltonian). The other element is the qubit number
+    for separation, file in data.items()
+]
+
+##############################################################################
+# Each Hamiltonian can be written as a linear combination of fifteen tensor products of Pauli
+# matrices.
+
+print([len(h.ops) for h in hamiltonians])
```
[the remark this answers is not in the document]
trbromley: Done! It now looks like this:
image

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,209 @@
...
+# Rigetti. We can evaluate the expectation value of each Hamiltonian with eight terms run on
+# ``Aspen-4-4Q-E`` and seven terms run on ``Aspen-7-4Q-C``, as summarized by the diagram below:
+#
+# .. figure:: /implementations/vqe_parallel/diagram.png
+#    :width: 65%
+#    :align: center
+#
+# To do this, start by instantiating a device for each term:
+
+# Use forest.qpu if hardware access is available
```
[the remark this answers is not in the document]
trbromley: Done, now it looks like this:
image

implementations/tutorial_vqe_parallel.py
trbromley: Thanks Josh! I think we should definitely look to answer these somewhere as PL-qchem evolves. Here I have gone for a simple explanation that hopefully gives a hint of an explanation for 1 and 2.

implementations/tutorial_vqe_parallel.py
trbromley: Thanks @agran2018 and @co9olguy, I've gone for a brief explanation:
image

implementations/tutorial_vqe_parallel.py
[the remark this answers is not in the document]
trbromley: Actually I did a bit of a restructure, which hopefully makes it easier (TLDR: got rid of the list comprehension)
image

implementations/tutorial_vqe_parallel.py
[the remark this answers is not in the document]
co9olguy: yes please :laughing:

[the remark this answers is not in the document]
trbromley: Have done a quick fix for this now (thanks @josh146) by editing the xanadu_theme layout - I guess the public repo for the theme can wait for later :grin:

implementations/tutorial_vqe_parallel.py
```diff
@@ -0,0 +1,212 @@
...
+# To do this, start by instantiating a device for each term:
+
+devs_4 = [qml.device("forest.qvm", device="Aspen-4-4Q-E") for _ in range(8)]
+devs_7 = [qml.device("forest.qvm", device="Aspen-7-4Q-D") for _ in range(7)]
+devs = devs_4 + devs_7
+
+##############################################################################
+# .. note::
+#     You can swap out ``forest.qvm`` for ``forest.qpu`` if hardware access is available.
+#
```
trbromley: Added a warning here that the user needs the `qvm` and `quilc` services.

implementations/tutorial_vqe_parallel.py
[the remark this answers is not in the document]
trbromley: Although it still doesn't seem to be fixed :thinking:

implementations/tutorial_vqe_parallel.py
```diff
@@ -46,11 +46,12 @@
 # :doc:`../tutorial/tutorial_quantum_chemistry` and :doc:`../app/tutorial_vqe`
 # tutorials.
 #
-# We begin by creating a dictionary containing molecular separations and corresponding data files
-# saved in ``.xyz`` format. The data files can be downloaded by clicking :download:`here
-# <../implementations/vqe_parallel/h2.zip>`.
+# We begin by creating a dictionary containing a selection of bond lengths and corresponding data
+# files saved in `XYZ <https://en.wikipedia.org/wiki/XYZ_file_format>`__ format. These files
+# follow a standard format for specifying the geometry of a molecule and can be downloaded as a
+# Zip from :download:`here <../implementations/vqe_parallel/vqe_parallel.zip>`.
```
co9olguy:
```suggestion
# zip file from :download:`here <../implementations/vqe_parallel/vqe_parallel.zip>`.

```

implementations/tutorial_vqe_parallel.py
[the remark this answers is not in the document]
co9olguy: Got it. Let's definitely look into this some more

implementations/tutorial_vqe_parallel.py
```diff
@@ -189,15 +202,15 @@ def calculate_surface(parallel=True):
 # We have seen how a :class:`~.pennylane.QNodeCollection` can be evaluated in parallel. This results
 # in a speed up in processing:
 
-print("Speedup: {0:.2f}".format(t_seq / t_par))
+print("Speed up: {0:.2f}".format(t_seq / t_par))
```
co9olguy:
```suggestion
print("Speed up: {0:.2f}x".format(t_seq / t_par))

```

PR #41, opened 2020-02-13
Port state preparation tutorial to use default.qubit

trbromley:
**Summary**

Port the state preparation tutorial from using PL-Forest to using `default.qubit` This fixes https://github.com/XanaduAI/qml/issues/35.

**Benefits**

- This tutorial can be a major bottleneck in building the documentation when using later versions of pyquil - it can take on the order of days. Currently we have `pyquil` pinned to `v2.10` in the requirements, which is now relatively outdated as the latest release is `v2.17`. This PR means that none of the tutorials will use PL-Forest, and we can update to pyquil `v2.17` when required without having issues with build time.

**Possible drawbacks**

- We would no longer have example use of Forest in the tutorials. However, there are incoming tutorials (e.g. https://github.com/XanaduAI/qml/pull/37 and https://github.com/XanaduAI/qml/pull/40) that will add this back in and be compatible with later releases of pyquil.

This closes https://github.com/XanaduAI/qml/pull/39. We tried to port to PL-Qiskit, but the build time was still a bit too slow.

### DIFF
--- a/implementations/tutorial_state_preparation.py
+++ b/implementations/tutorial_state_preparation.py
@@ -1,15 +1,15 @@
 r"""
 .. _state_preparation:
 
-State preparation with Rigetti Forest + PyTorch
-===============================================
+State preparation with PyTorch
+==============================
 
 In this notebook, we build and optimize a circuit to prepare arbitrary
 single-qubit states, including mixed states. Along the way, we also show
 how to:
 
 1. Construct compact expressions for circuits composed of many layers.
-2. Succintly evaluate expectation values of many observables.
+2. Succinctly evaluate expectation values of many observables.
 3. Estimate expectation values from repeated measurements, as in real
    hardware.
 """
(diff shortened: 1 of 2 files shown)


implementations/tutorial_state_preparation.py
```diff
@@ -94,15 +94,9 @@ def layer(params, j):
 
 
 ##############################################################################
-# To set up the device, we select a plugin that is compatible with
-# evaluating expectations through sampling: the ``forest.qvm`` plugin. The
-# syntax is slightly different than for other plugins, we need to also
-# feed a ``device`` keyword specifying the number of qubits in the format
-# ``[number of qubits]q-pyqvm``. The keyword ``shots`` indicates the
-# number of samples used to estimate expectation values.
-#
```
trbromley: @ixfoduap - we are changing the device for this tutorial from `forest.qvm` to `default.qubit` because the build time is currently really long.

Have simplified this paragraph as it was focused on `forest.qvm`. Although I'm unsure on the sentence: "To set up the device, we select a plugin that is compatible with evaluating expectations through sampling". Where in the tutorials do the samples come into play?

implementations/tutorial_state_preparation.py
```diff
@@ -94,15 +94,9 @@ def layer(params, j):
...
+# We use the ``default.qubit`` device.
```
josh146:
```suggestion
# Here, we use the ``default.qubit`` device to perform the optimization, but this can be changed to
# any other supported device.
```

implementations/tutorial_state_preparation.py
ixfoduap: There was a version of this tutorial where the expectation value was calculated using "shots", instead of analytically. In the current version we're not doing that, so this sentence can be removed.

trbromley: Great, thank you!

PR #45, opened 2020-02-16
Change ising_hamiltonian function to only use l

JB1984:
There is a bug in the ising_hamiltonian function. It works in the tutorial as long as you use a range in the training phase of 2 but when you changed this to 3 or more it would raise an exception. I believe this is due to the fact that the training epoch iterator is "i" and for some reason the ising_hamiltonian function also used "i" even though I don't believe that is should (or has to). The q_weights passed to it are size (4,3) so I believe the better way to do this would be to use "RY(weights[l, l-1], wires=w)"

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,7 @@ jobs:
           name: Comment artifacts
           command: |
             if [ -z "$GH_AUTH_TOKEN" ]; then
-              echo "No credentials available, artifacts will not be commented.""
+              echo "No credentials available, artifacts will not be commented."
             else
               echo "Credentials are available. Commenting artifacts to GitHub PR..."
               pip3 install pygithub --user && python3 .circleci/comment.py
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,13 @@ jobs:
 
       - run:
           name: Comment artifacts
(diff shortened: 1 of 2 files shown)


josh146:
Thanks @JB1984! Tagging @mariaschuld and @AroosaIjaz for a review.

mariaschuld:
Well spotted @JB1984. The reason it was not failing is that it interpreted the ``i`` from the training steps as the ``i`` selecting the weight. Although I am a bit confused why the results are stable under this bug...

In terms of a fix, why would you suggest ``l-1``? I would do the following:

``` python
for idx, w in enumerate(wires):
      RY(weights[l, idx + 1], wires=w)
```

The ``idx +1`` runs from 1 to 2, which are the indices of the RY weights.

mariaschuld:
Overall, we coded up the template explicitly, but PennyLane 0.8 has ``QAOAEmbedding`` as a flexible template. So one could delete the entire code section and just do

```
from pennylane.templates import QAOAEmbedding
```

at the top.

Would you like to turn the PR into this rather? The advantage is that the user can easily experiment with other settings...

JB1984:
Yea your fix is much better.

mariaschuld:
@JB1984 feel free to make the changes!

JB1984:
Alright sounds good, will work on it tonight!

PR #48, opened 2020-02-19
Fix redesign

mariaschuld:
Fixing the merge conflict present in redesign branch.


mariaschuld:
I see that the fix is already in ``redesign``. Not sure why the PR built there complains. Closing this.

PR #51, opened 2020-02-20
Better fix for iteration issue

JB1984:
fix iteration issue in ising_hamiltonian function

**Related GitHub Issues:** See #45 for a discussion on this bug.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,10 +67,10 @@ jobs:
           name: Comment artifacts
           command: |
             if [[ -z "${GH_AUTH_TOKEN}" ]]; then
+              echo "GitHub token not available, skipping uploading artifacts..."
+            else
               echo "Commenting artifacts..."
               pip3 install pygithub --user && python3 .circleci/comment.py
-            else
-              echo "GitHub token not available, skipping uploading artifacts..."
             fi
 
   deploy:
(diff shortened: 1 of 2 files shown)


co9olguy:
Thanks @JB1984! We'll wait for @mariaschuld to give a check before merging, since she wrote this demo.

.circleci/config.yml
```diff
@@ -65,7 +65,13 @@ jobs:
...
       - run:
           name: Comment artifacts
-          command: pip3 install pygithub --user && python3 .circleci/comment.py
+          command: |
+            if [[ -z "${GH_AUTH_TOKEN}" ]]; then
+              echo "GitHub token not available, skipping uploading artifacts..."
+            else
+              echo "Commenting artifacts..."
+              pip3 install pygithub --user && python3 .circleci/comment.py
+            fi
```
mariaschuld: @josh146 I had an unintentional change like this happen before - do you know what causes it? Is it just the circleci test that got updated somewhere?

In other words, safe to merge?

josh146: Sorry @mariaschuld for the confusion, this change was made by me in commits 
42e858b and 833ea05! Initially, CircleCI was failing --- I realised this was a bug in my configuration due to not taking into account PRs from forks.

josh146: Safe to merge :slightly_smiling_face:

PR #78, opened 2020-05-05
Fix QGAN tutorial

trbromley:
This PR fixes some issues with the QGAN tutorial:

- There was a typo in the CNOT gate of the discriminator. This meant that the state of the generator did not match the state of the real data circuit.
- To check the state of the generator and real data circuit, a section has been added printing out their respective Bloch vectors.
- A hadamard gate was added to find a case that fitted nicely. The resultant Bloch vectors now agree.

One drawback: The `Prob(fake classified as real)` value is 0.5 with these settings whereas before it was close to zero. This should be ok, as this tutorial is more for demonstration.


demonstrations/tutorial_QGAN.py
```diff
@@ -92,13 +94,14 @@ def generator(w):
...
 def discriminator(w):
+    qml.Hadamard(wires=0)
     qml.RX(w[0], wires=0)
     qml.RX(w[1], wires=2)
     qml.RY(w[2], wires=0)
     qml.RY(w[3], wires=2)
     qml.RZ(w[4], wires=0)
     qml.RZ(w[5], wires=2)
-    qml.CNOT(wires=[1, 2])
+    qml.CNOT(wires=[0, 2])
```
trbromley, line 104: This is a typo and doesn't work with `wires=[1, 2]`

josh146:
<h3>Website build</h3>
<strong>Commit:</strong> b02b99e7f5f0c0eb8780d15e549d6e2bcd4888d4

<strong>Circle build number:</strong> 660

<strong>Website zip:</strong> <a href=https://660-214003948-gh.circle-artifacts.com/0/tmp/qml_html.zip>https://660-214003948-gh.circle-artifacts.com/0/tmp/qml_html.zip</a>

<strong>Website url:</strong> <a href=https://660-214003948-gh.circle-artifacts.com/0/_build/html/index.html>https://660-214003948-gh.circle-artifacts.com/0/_build/html/index.html</a>

demonstrations/tutorial_QGAN.py
```diff
@@ -78,7 +79,8 @@ def real(phi, theta, omega):
 # provided as a workspace for the generator, while the discriminator’s
 # output will be on wire 2.
 
-def generator(w):
+def generator(w, wires=None):
```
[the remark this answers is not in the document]
josh146: Since ``wires`` is not being used, maybe better to have the following? Otherwise it might be a bit confusing.
```suggestion
def generator(w, **kwargs):
```

demonstrations/tutorial_QGAN.py
```diff
@@ -65,8 +65,9 @@
 # rotated (from the starting state :math:`\left|0\right\rangle`) to some
 # arbitrary, but fixed, state.
 
-def real(phi, theta, omega):
-    qml.Rot(phi, theta, omega, wires=0)
+def real(angles, wires=None):
```
josh146, line 68:
```suggestion
def real(angles, **kwargs):
```

demonstrations/tutorial_QGAN.py
```diff
@@ -190,7 +193,7 @@ def gen_cost(gen_weights):
 ##############################################################################
 # We begin by creating the optimizer:
 
-opt = tf.keras.optimizers.SGD(0.1)
+opt = tf.keras.optimizers.SGD(0.4)
```
josh146, line 196: This is just to increase convergence?

demonstrations/tutorial_QGAN.py
```diff
@@ -250,5 +253,19 @@ def gen_cost(gen_weights):
...
+# real data circuit. An easy way to access the state of the first qubit is through its
+# `Bloch sphere <https://en.wikipedia.org/wiki/Bloch_sphere>`__ representation:
+
+obs = [qml.PauliX(0), qml.PauliY(0), qml.PauliZ(0)]
+
+bloch_vector_real = qml.map(real, obs, dev, interface="tf")
+bloch_vector_generator = qml.map(generator, obs, dev, interface="tf")
+
+print("Real Bloch vector: {}".format(bloch_vector_real([phi, theta, omega])))
+print("Generator Bloch vector: {}".format(bloch_vector_generator(gen_weights)))
```
josh146, line 271: Nice! I was wondering what you were using `map` for before when you mentioned it 😆

demonstrations/tutorial_QGAN.py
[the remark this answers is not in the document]
trbromley: Nice, thanks!

demonstrations/tutorial_QGAN.py
trbromley: Yes, it means we can almost certainly do it in 50 steps

co9olguy:
> One drawback: The `Prob(fake classified as real)` value is 0.5 with these settings whereas before it was close to zero. This should be ok, as this tutorial is more for demonstration.

This is what one expects for a trained GAN. The goal is to make the generator learn the correct distribution, which means the discriminator can't tell whether something is real or fake. In the case of a single datapoint, it means it's guess is as good as a coin flip

PR #94, opened 2020-05-24
Local Cost Functions

Tstorwick:
**Title:**

**Summary:**

**Relevant references:** https://arxiv.org/abs/2001.00550

**Possible Drawbacks:**

**Related GitHub Issues:**


josh146:
<h3>Website build</h3>
<strong>Commit:</strong> edd4206ac2d97bccc97b8515c6b5ab44d68330c6

<strong>Circle build number:</strong> 764

<strong>Built <code>demos</code> directory:</strong> <a href=https://764-214003948-gh.circle-artifacts.com/0/tmp/qml_demos.zip>https://764-214003948-gh.circle-artifacts.com/0/tmp/qml_demos.zip</a>

<strong>Backreferences zip:</strong> <a href=https://764-214003948-gh.circle-artifacts.com/0/tmp/qml_backreferences.zip>https://764-214003948-gh.circle-artifacts.com/0/tmp/qml_backreferences.zip</a>

<strong>Website zip:</strong> <a href=https://764-214003948-gh.circle-artifacts.com/0/tmp/qml_html.zip>https://764-214003948-gh.circle-artifacts.com/0/tmp/qml_html.zip</a>

<strong>Website url:</strong> <a href=https://764-214003948-gh.circle-artifacts.com/0/_build/html/index.html>https://764-214003948-gh.circle-artifacts.com/0/_build/html/index.html</a>

co9olguy:
Hi @Tstorwick, thanks for the PR contribution! Looking forward to having a deeper look.

Tstorwick:
> Hi @Tstorwick, thanks for the PR contribution! Looking forward to having a deeper look.

No problem! Thanks for taking a look at it :)

Tstorwick:
@josh146 This new version still has the wrong incoding, but for the life of me i cant seem to fix it. I've tried even manually replacing the characters and it still doesnt work.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+global_circuit = qml.QNode(global_cost_simple, dev)
+
+local_circuit = qml.QNode(local_cost_simple, dev)
+
+
+######################################################################
+# To analyze each of the circuits, we provide some random initial
+# :math:`\theta`\ 's for each rotation
```
josh146, line 115: I like the look of θ’s, but probably best to say 'parameters' 😆 
```suggestion
# parameters for each rotation
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# will worsen as the number of qubits increases.
+#
+
+fig = plt.figure()
+ax = fig.add_subplot(111, projection="3d")
+surf = ax.plot_surface(X, Y, global_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
```
josh146, line 179: Visualization plots are really nice, and very effective!

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# becomes much more trainable.
+#
+
+fig = plt.figure()
+ax = fig.add_subplot(111, projection="3d")
+surf = ax.plot_surface(X, Y, local_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
```
josh146, line 193: I suppose a big question I have is, are we solving the same problem when we switch between a global observable and a local observable? It might be the case that the problem can _only_ be formulated in terms of a global cost function!

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+######################################################################
+# Those are some nice pictures, but how do they refect actual
+# trainability? Lets try training both the local, and global cost
+# functions. Because we have a visualization of the total cost landscape,
+# lets pick a point to exaggerate the problem. The worst point in the
+# landscape is :math:`(\pi,0)`, so lets use that.
```
josh146, line 201: For example, here the two optimization landscapes have identical local minima. I wonder if this is always guaranteed; can we construct a counterexample?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+        print("Cost after step {:5d}: {: .7f}".format(i + 1, cost_global(params_global)))
+    if cost_global(params_global) < 0.1:
+        break
+print(global_circuit.draw())
+
+
+######################################################################
+# After 100 steps, the cost function is still exactly 1. Clearly we are in
+# an untrainable area. Now, lets limit the ourselves to the local cost
+# function and see how it performs.
```
josh146, line 222: This is a very effective way of showing the inability to train the system 😆

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+        print("Cost after step {:5d}: {: .7f}".format(i + 1, cost_local(params_local)))
+    if cost_local(params_local) < 0.05:
+        break
+print(local_circuit.draw())
+
+
+######################################################################
+# It trained! And much faster than the global case. However, we know our
+# local cost function is faithful to the global one, but just how much
+# have we trained it?
```
josh146, line 243: When you say faithful, is this due to the local minima intersecting from the landscape plots, or via some other intuition/result?

demonstrations.rst
```diff
@@ -83,6 +83,12 @@ algorithms using PennyLane and near-term quantum hardware.
     :description: :doc:`demos/quantum_neural_net`
     :tags: autograd strawberryfields photonics
 
+.. customgalleryitem::
+    :tooltip: Local cost functions to avoid barren plateaus
+    :figure: demonstrations/<local_cost_functions>/Cerezo_et_al_local_cost_functions.png
+    :description: :doc:`demos/tutorial_local_cost_functions`
+    :tags: barren plateaus cost functions local global autograd
+
```
mariaschuld, line 91: I wonder why this entry is duplicated?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+==========
+
+Barren Plateaus
+---------------
+
+Barren plateaus are large regions of the cost functions parameter space
+where the variance of the gradient is almost 0; or, put another way, the
+cost function landscape is flat. This means that a variational circuit
+initialized in one of these areas will be untrainable any gradient based
+algorithm.
```
mariaschuld, line 12: That's a great explanation of barren plateaus @Tstorwick !

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+from pennylane import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.ticker import LinearLocator, FormatStrFormatter
+
+
+######################################################################
+# Visualizing the problem
+# =======================
+#
+# To start, lets look at the toy problem of learning the identity gate
```
mariaschuld, line 41:
```suggestion
# To start, lets look at the task of learning the identity gate
```

Better to avoid repetition of "problem"?

demonstrations.rst
```diff
@@ -197,6 +203,12 @@ algorithms using PennyLane and near-term quantum hardware.
     :description: :doc:`demos/tutorial_vqe_qng`
     :tags: chemistry
 
+.. customgalleryitem::
+    :tooltip: Local and global cost functions
```
trbromley, line 207:
```suggestion
    :tooltip: Understand the difference between local and global cost functions
```
The tooltips seem to typically be stated in an imperative style.

demonstrations.rst
```diff
@@ -197,6 +203,12 @@ algorithms using PennyLane and near-term quantum hardware.
...
+    :figure: demonstrations/local_cost_functions/Cerezo_et_al_local_cost_functions.png
```
trbromley, line 208: Is it ok to use this image here, since it's taken directly from the paper?
The other two plots in the tutorial look good:
image

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
+"""
+Background
+==========
+
+Barren Plateaus
+---------------
+
+Barren plateaus are large regions of the cost functions parameter space
```
trbromley, line 8:
```suggestion
Barren plateaus are large regions of the cost function's parameter space
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# across multiple qubits. This will help us visualize the problem and get
+# a sense of what is happening in the cost landscape
+#
+# First we define a number of wires we want to train on. The work by
+# Cerezo et al. shows that circuits are trainible under certain ranges, so
+# how many quibits we train on will effect our results.
+#
+# .. figure:: ../demonstrations/local_cost_functions/Cerezo_et_al_local_cost_functions.png
+#    :align: center
+#    :width: 90%
```
trbromley, line 51: It feels like this image is better suited earlier on, around the second paragraph where we're talking about literature.

It would also be good to:
- make the image a bit smaller, it is quite big on my screen and you can tell some blurriness/pixelation
- Give credit for the image

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+
+######################################################################
+# Next, we want to define our qnodes, containing our ansatz. For this
+# simple example, an ansatz that works well is simply a rotation along X,
+# and a rotation along Y, repeated across all the qubits.
+#
+# We will also define our cost functions here. Since we are trying to
+# learn the identity gate, a natural cost function is
+#
+# .. math:: C = 1-(P_{|0>})
```
trbromley, line 65: Do we need the brackets in these equations?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# We will apply this across all qubits for our global cost function, i.e.:
+#
+# .. math:: C_{G} = 1-(P_{|0>})
+#
+# and only on one qubit for our local cost function:
+#
+# .. math:: C_{L} = 1-(P_{|0>,1}).
```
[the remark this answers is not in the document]
trbromley: It would be good to define P here, I was originally thinking of them as projectors but I think they are probabilities (maybe p could be lower case).

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# and a rotation along Y, repeated across all the qubits.
+#
+# We will also define our cost functions here. Since we are trying to
+# learn the identity gate, a natural cost function is
+#
+# .. math:: C = 1-(P_{|0>})
+#
+# We will apply this across all qubits for our global cost function, i.e.:
+#
+# .. math:: C_{G} = 1-(P_{|0>})
```
trbromley, line 69:
```suggestion
# .. math:: C_{G} = 1-(P_{|00 \ldots 0\rangle})
```
This makes it look more multiqubit

demonstrations/tutorial_local_cost_functions.py
trbromley, line 73:
```suggestion
# .. math:: C_{L} = 1-(P_{|0\rangle}).
```
Given the above change, we can probably drop the 1

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# and only on one qubit for our local cost function:
+#
+# .. math:: C_{L} = 1-(P_{|0>,1}).
+#
+# To implement this, we will define a separate qnode for the local cost
```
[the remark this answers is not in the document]
trbromley: Note this occurs a few times.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# function and the global cost function.
+#
+
+
+def global_cost_simple(rotations):
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+    for i in range(wires):
```
trbromley, line 83:
```suggestion
```
Can probably skip this line.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+        qml.RY(rotations[1][i], wires=i)
+    return qml.probs(wires=range(wires))
+
+
+def local_cost_simple(rotations):
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+    for i in range(wires):
```
trbromley, line 91:
```suggestion
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+ax = fig.add_subplot(111, projection="3d")
+surf = ax.plot_surface(X, Y, global_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
+
+
+######################################################################
+# However, when we change to the local cost function, the cost lanscape
```
trbromley, line 183:
```suggestion
# However, when we change to the local cost function, the cost landscape
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+ax = fig.add_subplot(111, projection="3d")
+surf = ax.plot_surface(X, Y, local_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
+
+
+######################################################################
+# Those are some nice pictures, but how do they refect actual
```
trbromley, line 197:
```suggestion
# Those are some nice pictures, but how do they reflect actual
```

demonstrations/tutorial_local_cost_functions.py
trbromley, line 201:
```suggestion
# landscape is :math:`(\pi,0)`, so let's use that.
```

trbromley: Is it local minima or maxima? Looked like the highest point to me.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# trainability? Lets try training both the local, and global cost
+# functions. Because we have a visualization of the total cost landscape,
+# lets pick a point to exaggerate the problem. The worst point in the
+# landscape is :math:`(\pi,0)`, so lets use that.
+#
+
+rotations = [[3 for i in range(wires)], [0 for i in range(wires)]]
```
trbromley, line 204:
```suggestion
rotations = [[np.pi for i in range(wires)], [0 for i in range(wires)]]
```
If we put exactly on that point do things not train?

demonstrations/tutorial_local_cost_functions.py
trbromley: It might also be interesting to print the gradient.

demonstrations/tutorial_local_cost_functions.py
trbromley: Yeah I also was wondering what we mean by faithful :thinking:

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+
+
+######################################################################
+# Interestingly, the global cost function is still 1. If we trained the
+# local cost function, why hasnt the global cost function changed?
+#
+# The answer is that we have trained the global cost a *little bit*, but
+# not enough to see a change with only 10000 shots. To see the effect,
+# we'll need to increase the number of shots to an unreasonable amount.
+# Instead the making the backend analytic gives us the exact
```
trbromley, line 256:
```suggestion
# Instead making the backend analytic gives us the exact
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+params_local = rotations
+for i in range(steps):
+    # update the circuit parameters
+    params_local = opt.step(cost_local, params_local)
+
+    if (i + 1) % 5 == 0:
+        print("Cost after step {:5d}: {: .7f}".format(i + 1, cost_local(params_local)))
+    if cost_local(params_local) < 0.05:
+        break
+print(local_circuit.draw())
```
trbromley, line 237: I like that you print the circuit as a way to visualize the end result!
However, when I look at the output, we seem to have only trained the first qubit parameters and hence broken out of our 2D space:
image

trbromley: Which I guess makes sense since we do not have any entangling gates and it is a local measurement :thinking: But then it feels like a departure from what we plotted above.

trbromley: Maybe the plots above need to vary the angle only on the first qubit in both cases, and adopt a fixed angle for all the other qubits.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+    + str(
+        cost_global([[3.0 for i in range(wires)], [0 for i in range(wires)]])
+        - cost_global(params_local)
+    )
+)
+
+
+######################################################################
+# Our circuit has definitely been trained, but not a useful amount. If we
+# attempt to use this circuit, we are no better than random chance.
```
trbromley, line 277: What do you mean by "no better than random chance"?

demonstrations/tutorial_local_cost_functions.py
trbromley: Maybe it's ok as is... I'm confused :laughing:

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+from matplotlib.ticker import LinearLocator, FormatStrFormatter
+
+
+######################################################################
+# Visualizing the problem
+# =======================
+#
+# To start, lets look at the toy problem of learning the identity gate
+# across multiple qubits. This will help us visualize the problem and get
+# a sense of what is happening in the cost landscape
```
mariaschuld, line 43:
```suggestion
# a sense of what is happening in the cost landscape.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+#
+# First we define a number of wires we want to train on. The work by
+# Cerezo et al. shows that circuits are trainible under certain ranges, so
+# how many quibits we train on will effect our results.
```
mariaschuld, line 47:
```suggestion
# how many qubits we train on will effect our results.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+    for i in range(wires):
+        qml.RY(rotations[1][i], wires=i)
+    return qml.probs(wires=[0])
+
+
+def cost_local(rotations):
+    # result = circuit(rotations)
+    # return sum(abs(result[i]) for i in range(wires))
```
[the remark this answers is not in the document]
Tstorwick: Thanks!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Changed to : "Our circuit has definitely been trained, but not a useful amount. If we attempt to use this circuit, it would act the same as if we never trained at all."

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+# where the global function has a barren plateau. The significance of this
+# is that we can now train from every starting location in this example.
+#
+# But, how often does this problem occur? If we wanted to train this
+# circuit from a random starting point, how often would we be stuck in a
+# plateau? To investigate this, lets attempt to train the global cost
+# function using random starting positions and count how many times we run
+# into a barren plateau.
+#
+# Let's use a more likely number of qubits: n=10. We will say that after
```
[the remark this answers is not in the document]
Tstorwick: Changed to "Let's use a number of qubits we are more likely to use in a real variational circuit: n=10."

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,432 @@
...
+        print("---Switching Locality---")
+        locality += 1
+        continue
+    elif runCost < 0.1 and locality >= wires:
+        break
+print(tunable_circuit.draw())
+
+
+######################################################################
+# Now the circuit can be trained, even though we started from a place
```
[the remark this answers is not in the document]
Tstorwick: Thanks!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: I think i just need to be more clear in the explanation. I compress the parameter space to 2D for the purposes of plotting, but  for the actual training I was going by the assumption that we would train this "blind", and in the full 2n space.

demonstrations.rst
```diff
@@ -83,6 +83,12 @@ algorithms using PennyLane and near-term quantum hardware.
     :description: :doc:`demos/quantum_neural_net`
     :tags: autograd strawberryfields photonics
 
+.. customgalleryitem::
+    :tooltip: Local cost functions to avoid barren plateaus
+    :figure: demonstrations/<local_cost_functions>/Cerezo_et_al_local_cost_functions.png
+    :description: :doc:`demos/tutorial_local_cost_functions`
+    :tags: barren plateaus cost functions local global autograd
```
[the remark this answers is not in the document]
Tstorwick: Done!

demonstrations/tutorial_local_cost_functions.py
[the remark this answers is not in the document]
Tstorwick: changed!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Thanks Josh :)

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Haha, no, unfortunately things do not train directly on the point. I didn't realize this and tore my hair out over it for a week.

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Thanks!

Tstorwick:
> Hey @Tstorwick, thanks for contributing the tutorial. The idea is great, and your explanations clear, but I think there is a bit more work needed.
> 
> @josh146 and @trbromley have already left such a detailed review. Let me only add some high-level comments.
> 
> Something is still off in the way things are designed. If the RX and RY rotation turns a qubit into state a|0> + b|1>, then the "local circuit" produces |a|^2. The global circuit should be just (|a|^2)*6 (since you use the same rotation angle on every qubit, and the chance of each qubit being in state 0 is |a|^2). I checked, and this is consistent with your code.
> 
> But this is why I don't think you are actually reproducing the results from the paper. Your results may not a property of a local or global measurement, but a property of smaller and larger probabilities, if I am not mistaken. (But I could be wrong). So I think it would be good to go back to the paper and really check that what you are doing contains the complexity of what they prove. Especially since it is not clear if the entire observation could be made on circuits where the qubits are not entangled.
> 
> And here a few more things that sprung to my mind:
> 
> * Why do you initialise all rotation angles for RX in the same value, and also for RY?
> * the original circuit ansatz proposed in the paper is actually implemented as a [PennyLane template](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.layers.SimplifiedTwoDesign.html)...would that help? You are using unentangled qubits in the tutorial. So would you consider using the ansatz instead, which uses entanglers and actually corresponds to the paper?
> * my local build fails due to this line missing for 3D plotting: `from mpl_toolkits.mplot3d import Axes3D` Include in the imports?
> * It would be great to add a sentence explaining P_{|0>} (and what quantum observable it corresponds to, since the paper uses expectation values as outputs).
> * The tutorial takes very long to build. I like the idea to "artificially shorten" the runtime (i.e. number of training steps), and load pre-trained parameters in the tutorial, telling the user that they can reproduce by increasing the number of steps to XYZ. (See embeddings_metric_learning tutorial)

Hi Maria! I couldn't figure out how to reply directly, but the quote option will do :). 

So, as I understood the paper, the local vs global measurement advantage was linked to smaller and large probabilities. A local function means a simpler state, and therefore a more "drastic" gradient when training. I thought it was just trivially easier to see when the qubits arent entangled. That said, a proper implementation should have entanglements, so I added them. I originally had a more complex problem after this toy version, that included entanglements and non-trivial search spaces, but runtimes were getting extreme (~2-3 hours) so I cut it. 

"Why do you initialise all rotation angles for RX in the same value, and also for RY?", Originally, I was showing all the starting points on a plot, but decided to cut it for readability. This is just a relic I missed. Good catch!

"my local build fails due to this line missing for 3D plotting: `from mpl_toolkits.mplot3d import Axes3D` Include in the imports?" I tried on a clean python environment locally, and it seemed to work for me. Any more information on the error?

"The tutorial takes very long to build. I like the idea to "artificially shorten" the runtime (i.e. number of training steps), and load pre-trained parameters in the tutorial, telling the user that they can reproduce by increasing the number of steps to XYZ. (See embeddings_metric_learning tutorial)" Good Idea, I'll give implementing this a shot!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: It should be minima, it is on my machine at least. I think I constructed a counter-example earlier; however remember that this plot is a subspace of our problem space (that I had to do for plotting), and the two functions might not have the same minima in a different subspace. That, in turn, begs some questions: is there always a subspace where they have a minima in the same location? Is it easier to find this space and then solve the local version than solving the global version?

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Updated the discription

demonstrations.rst
Tstorwick: I actually meant to ask this question. Do I need permission to reproduce this? Or do I only need a citation?

demonstrations/tutorial_local_cost_functions.py
Tstorwick: I'm honestly not 100% sure. I cant construct an example of a global only problem in my head (aside from the trivial one of 1 qubit, but one could equally as well call that only local), but there may well be one.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+In a quick explanation, a global cost function is the type that is traditionally used.
+Information from the entire measurement is used to analyze the result of the 
+circuit, and a cost function is calculated from this to quantify the circuits 
+performance. A local cost function only considers information from a few qubits, 
+and attempts to analyze the behavior of the entire circuit from this limited scope.
+
+
+Cerezo et al. also handily prove that these local cost functions are bounded by the 
+global ones, i.e. if a global cost function is formulated in the manner discribed
+by Cerezo, then its corrisponding local cost function will always be
```
josh146, line 43:
```suggestion
by Cerezo, then its corresponding local cost function will always be 
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+Reproduced from [#Cerezo2020].
+   
+In a quick explanation, a global cost function is the type that is traditionally used.
+Information from the entire measurement is used to analyze the result of the 
+circuit, and a cost function is calculated from this to quantify the circuits 
+performance. A local cost function only considers information from a few qubits, 
+and attempts to analyze the behavior of the entire circuit from this limited scope.
+
+
+Cerezo et al. also handily prove that these local cost functions are bounded by the
```
josh146, line 41: Nice word, handily!

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+######################################################################
+# Visualizing the problem
+# --------------
+#
+# To start, lets look at the task of learning the identity gate
+# across multiple qubits. This will help us visualize the problem and get
+# a sense of what is happening in the cost landscape.
+#
+# First we define a number of wires we want to train on. The work by
+# Cerezo et al. shows that circuits are trainible under certain regimes, so
```
josh146, line 70:
```suggestion
# Cerezo et al. shows that circuits are trainable under certain regimes, so
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+# how many qubits we train on will effect our results.
+
+wires = 6
+dev = qml.device("default.qubit", wires=wires, shots=10000, analytic=False)
+
+
+######################################################################
+# Next, we want to define our QNodes, containing our ansatz. For this
```
josh146, line 78:
```suggestion
# Next, we want to define our QNodes and our circuit ansatz. For this
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+#
+# .. math:: C = 1-p_{|0\rangle}
+#
+# We will apply this across all qubits for our global cost function, i.e.:
+#
+# .. math:: C_{G} = 1-p_{|00 \ldots 0\rangle}
+#
+# and only on one qubit for our local cost function:
+#
+# .. math:: C_{L} = 1-p_{|0>}.
```
josh146, line 94:
```suggestion
# .. math:: C_{L} = 1-p_{| 0\rangle}.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+# simple example, an ansatz that works well is simply a rotation along X,
+# and a rotation along Y, repeated across all the qubits.
+#
+# We will also define our cost functions here. Since we are trying to
+# learn the identity gate, a natural cost function is the probablity of measuring the 
+# zero state.
+#
+# .. math:: C = 1-p_{|0\rangle}
+#
+# We will apply this across all qubits for our global cost function, i.e.:
```
josh146, line 88:
```suggestion
# We will apply this across all qubits for our global cost function, i.e.,:
```

demonstrations/tutorial_local_cost_functions.py
josh146: In this particular case, do we want to define _which_ qubit the marginal probability `p_{| 0\rangle}` refers to?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+######################################################################
+# Next, we want to define our QNodes, containing our ansatz. For this
+# simple example, an ansatz that works well is simply a rotation along X,
+# and a rotation along Y, repeated across all the qubits.
+#
+# We will also define our cost functions here. Since we are trying to
+# learn the identity gate, a natural cost function is the probablity of measuring the 
+# zero state.
+#
+# .. math:: C = 1-p_{|0\rangle}
```
josh146, line 86: Might be good to define `p_{}` in the text, so that it can be better intepreted by the reader :)

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+global_circuit = qml.QNode(global_cost_simple, dev)
+
+local_circuit = qml.QNode(local_cost_simple, dev)
+
+
+######################################################################
+# To analyze each of the circuits, we provide some random initial
+# parameters for each rotation
```
josh146, line 134:
```suggestion
# parameters for each rotation.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+#
+
+RX = np.random.uniform(low=-np.pi, high=np.pi)
+RY = np.random.uniform(low=-np.pi, high=np.pi)
+rotations = [[RX for i in range(wires)], [RY for i in range(wires)]]
+
+
+######################################################################
+# And look at the results:
```
josh146, line 143: It sounds a bit better to avoid starting the sentence with 'and', e.g.,
```suggestion
# Examining the results:
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+print("Local Cost: {: .7f}".format(cost_local(rotations)))
+print("--- Global Circuit ---")
+print(global_circuit.draw())
+print("--- Local Circuit")
+print(local_circuit.draw())
+
+
+######################################################################
+# With this simple example, we can visualize the cost function, and see
+# the barren plateau effect graphically. Although there are 2n (where n is the
```
josh146, line 156: (so that it renders using LaTeX :))
```suggestion
# the barren plateau effect graphically. Although there are :math:`2n` (where :math:`n` is the 
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+print("--- Local Circuit")
+print(local_circuit.draw())
+
+
+######################################################################
+# With this simple example, we can visualize the cost function, and see
+# the barren plateau effect graphically. Although there are 2n (where n is the 
+# number of qubits) parameters, in order to make the cost landscape plot-able 
+# we must constrain ourselves. We will assume that all the X rotations are the 
+# same, and all the Y rotations are the same.
```
josh146, line 159: Does it make more sense to say 'We will consider the case where all X rotations have the same value, and...'?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+# local cost function is bounded by the global one, but just how much
+# have we trained it?
+#
+
+cost_global(params_local)
+
+
+######################################################################
+# Interestingly, the global cost function is still 1. If we trained the
+# local cost function, why hasnt the global cost function changed?
```
josh146, line 273:
```suggestion
# local cost function, why hasn't the global cost function changed?
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+        locality += 1
+        continue
+    elif runCost < 0.1 and locality >= wires:
+        break
+print(tunable_circuit.draw())
+
+
+######################################################################
+# A more thorough analysis.
+# --------------
```
josh146, line 352: Good practice in restructured text markup to have the title underline match the length of the title:
```suggestion
# A more thorough analysis
# ------------------------
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+# function and the global cost function.
+#
+
+
+def global_cost_simple(rotations):
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+        qml.RY(rotations[1][i], wires=i)
+    for i in range(wires-1):
+        qml.CNOT(wires=[i,i+1])
```
josh146, line 106: @Tstorwick it might be good to use the 'ring' pattern here; this ensures that the entanglement is non-local (and not classically simulatable by a tensor network simulator!)

E.g.:

```suggestion
    qml.broadcast(qml.CNOT, wires=wires, pattern="ring")
```

(see the broadcast documentation here: https://pennylane.readthedocs.io/en/stable/code/api/pennylane.broadcast.html)

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,471 @@
...
+    
+##############################################################################
+# References
+# ----------
+#
+# .. [#Cerezo2020]
+#    
+#   Cerezo, M., Sone, A., Volkoff, T., Cincio, L., &amp; Coles, P. (2020, February 04). 
+#   Cost-Function-Dependent Barren Plateaus in Shallow Quantum Neural Networks. 
+#   arXiv:2001.00550<https://arxiv.org/abs/2001.00550>
```
josh146, line 470:
```suggestion
#   `arXiv:2001.00550 <https://arxiv.org/abs/2001.00550>`__
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,556 @@
...
+surf = ax.plot_surface(X, Y, global_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
+
+
+######################################################################
+# However, when we change to the local cost function, the cost landscape
+# becomes much more trainable.
```
josh146, line 201: It might be worth specifying the figure number in the paper that these plots reproduce

josh146:
```suggestion
# However, when we change to the local cost function, the cost landscape
# becomes much more trainable as the size of the barren plateau decreases.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,556 @@
...
+######################################################################
+# Those are some nice pictures, but how do they reflect actual
+# trainability? Lets try training both the local, and global cost
+# functions.
+# To use pennylane, we will have to change our cost function
+# a bit to work with the suite of optimizers we have available. We will go from:
+# 
+# .. math:: C_{L} = 1-\sum p_{|0\rangle}.
+#
+# Where we sum the marginal probabilities of each quibt, to:
```
josh146, line 222:
```suggestion
# To simplify this model, let's modify our cost function from
# 
# .. math:: C_{L} = 1-\sum p_{|0\rangle},
#
# where we sum the marginal probabilities of each qubit, to
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,556 @@
...
+# 
+# .. math:: C_{L} = 1-p_{|0\rangle}.
+# 
+# where we only consider the probability of a single qubit to be in the 0 state. 
+# This is also a valid local cost function.
```
josh146, line 227:
```suggestion
# .. math:: C_{L} = 1-p_{|0\rangle},
# 
# where we only consider the probability of a single qubit to be in the 0 state. 
```

mariaschuld:
@Tstorwick Sorry, I missed your reply from a while back (must be because I wasn't mentioned by @). But it makes a lot of sense, thanks for the comments. Is this ready for final review?

Tstorwick:
@mariaschuld It is! Let me know what you think.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+#
+# We will apply this across all qubits for our global cost function, i.e.,:
+#
+# .. math:: C_{G} = \langle | \psi(theta) | \left(I - |00 \ldots 0\rangle \langle 00 \ldots 0|\right) \rangle = 1-p_{|00 \ldots 0\rangle}
+#
+# and instead, we will sum the individual contributions from each qubit 
+# for the local cost function:
+#
+# .. math:: C_L = \langle \psi(\theta) | \left(I - \frac{1}{n} \sum_j |0\rangle \langle 0|_j\right)|\psi(\theta)\rangle = 1 - \sum_j p_{|0\rangle_j}.
+#
```
mariaschuld, line 95: I still feel strange about this, we are doing a single-qubit computation n times here and sum the result :)

mariaschuld: Maybe one could mention that this is essentially what this circuit does, to pre-empt the comment from readers?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+print("--- Global Circuit ---")
+print(global_circuit.draw())
+print("--- Local Circuit")
+print(local_circuit.draw())
+
+
+######################################################################
+# With this simple example, we can visualize the cost function, and see
+# the barren plateau effect graphically. Although there are :math:`2n` (where :math:`n` is the 
+# number of qubits) parameters, in order to make the cost landscape plot-able
```
mariaschuld, line 152: Love plot-able :) but is it a word?

```suggestion
# number of qubits) parameters, in order to plot the cost landscape 
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+print("--- Local Circuit")
+print(local_circuit.draw())
+
+
+######################################################################
+# With this simple example, we can visualize the cost function, and see
+# the barren plateau effect graphically. Although there are :math:`2n` (where :math:`n` is the 
+# number of qubits) parameters, in order to make the cost landscape plot-able 
+# we must constrain ourselves. We will consider the case where all X rotations 
+# have the same value, and all the Y rotations have the same value.
```
mariaschuld, line 154: I love your plots, but one should be careful to note here that of course, this is not conclusive: It only shows the phenomenon for a very restricted case. For example, what if the barren plateau is related to this nice trick of making parameters constant?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+from pennylane import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.ticker import LinearLocator, FormatStrFormatter
+
+
+######################################################################
+# Visualizing the problem
+# --------------
+#
+# To start, lets look at the task of learning the identity gate
```
mariaschuld, line 64:
```suggestion
# To start, let's look at the task of learning the identity gate
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+surf = ax.plot_surface(X, Y, local_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
+
+
+######################################################################
+# Those are some nice pictures, but how do they reflect actual
+# trainability? Lets try training both the local, and global cost
```
mariaschuld, line 211:
```suggestion
# trainability? Let us try training both the local, and global cost
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+# 
+# .. math:: C_{L} = 1-\sum p_{|0\rangle},
+#
+# where we sum the marginal probabilities of each qubit, to
+# 
+# .. math:: C_{L} = 1-p_{|0\rangle},
+# 
+# where we only consider the probability of a single qubit to be in the 0 state. 
+#  
+# While we're at it, lets make our ansatz a little more like one we would encounter while
```
mariaschuld, line 223:
```suggestion
# While we're at it, let us make our ansatz a little more like one we would encounter while 
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+surf = ax.plot_surface(X, Y, local_Z, cmap="viridis", linewidth=0, antialiased=False)
+ax.set_zlim(0, 1)
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
+
+######################################################################
+# It seems our changes didn't significantly alter the overall cost landscape.
+# This probably isn't a general trend, but it is a nice surprise.
+# Now, lets get back to training the local, and global cost functions.
```
mariaschuld, line 299:
```suggestion
# Now, let us get back to training the local, and global cost functions.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+    if (i + 1) % 1 == 0:
+        print("Cost after step {:5d}: {: .7f}".format(i + 1, cost_global(params_global)))
+    if cost_global(params_global) < 0.1:
+        break
+print(global_circuit.draw())
+
+
+######################################################################
+# After 100 steps, the cost function is still exactly 1. Clearly we are in
+# an untrainable area. Now, lets limit the ourselves to the local cost
```
mariaschuld, line 322:
```suggestion
# an "untrainable" area. Now, let us limit ourselves to the local cost
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+        qml.RX(rotations[0][i], wires=i)
+        qml.RY(rotations[1][i], wires=i) 
+    return [qml.probs(wires=i) for i in range(wires)]
+
+global_circuit = qml.QNode(global_cost_simple, dev)
+
+local_circuit = qml.QNode(local_cost_simple, dev)
+
+def cost_local(rotations):
+    return 1 - np.sum(local_circuit(rotations)[:,0])/wires
```
mariaschuld, line 119: You use the non-entangled cost here if I am not mistaken?

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+def cost_global(rotations):
+    return 1 - global_circuit(rotations)[0]
```
mariaschuld, line 123: And here too. I guess that is a bug? I hope the plots will still be nice! :)

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+ax.zaxis.set_major_locator(LinearLocator(10))
+ax.zaxis.set_major_formatter(FormatStrFormatter("%.02f"))
+plt.show()
+
+######################################################################
+# It seems our changes didn't significantly alter the overall cost landscape.
+# This probably isn't a general trend, but it is a nice surprise.
+# Now, lets get back to training the local, and global cost functions.
+# Because we have a visualization of the total cost landscape,
+# lets pick a point to exaggerate the problem. The worst point in the
```
mariaschuld, line 301: Why is that the worst point?

demonstrations/tutorial_local_cost_functions.py
mariaschuld: It's great you added entanglers, but they don't do anything to the first qubit you measure, unfortunately. 

If you want to fix this, measure the last qubit, which should get information from every other one, or use boradcasting pattern "ring" so the last qubit gets entangled with the first before you measure!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: @mariaschuld So, the tutorial has changed a little bit. The large confusion was that Patrick Coles has two papers with *almost* identical cost functions. We original had contacted him for some questions we had on one of them, and he directed us to the other; it has better explanations. This is what caused this whole confusion on smaller and larger probability. Firstly, in the paper we are replicating, he demonstrates with a simple example how the global cost function of the zero state over all qubits is more-barren than a local cost function which is the sum of the marginal probabilities of each qubit. This is what I've implemented first.

After that we change the circuit to include entanglements (to be like a more common ansatz) and change the cost function to be the marginal probability of 1 qubit (to a: work with pennylane, currently a cost function with multiple terms throws an error in qml.jacobian when attempting optimization, and b: so that we can implement a version of a local cost function that we can slowly increase to a global one). This other 1 qubit cost function is the cost function implemented in the second Patrick Coles paper. The transitioning from local to global was not implemented in any paper, is probably novel work, and I think is the most interesting part of this tutorial.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+more robust to noise, and have more well-behaved gradients with no
+plateaus for shallow circuits.
+
+
+.. figure:: ../demonstrations/local_cost_functions/Cerezo_et_al_local_cost_functions.png
+   :align: center
+   :width: 50%
+   :caption: Reproduced from [#Cerezo2020].
+   
+Traditionally, most variational quantum algorithms are constructed to use global cost functions.
```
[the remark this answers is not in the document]
Tstorwick: Perhaps it is a little soon to call anything in quantum variational circuits "traditional" .

demonstrations/tutorial_local_cost_functions.py
Tstorwick: See the other comment, but I'll leave this open so I don't forget to change it based on your response.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+
+Cerezo et al. also handily prove that these local cost functions are bounded by the 
+global ones, i.e. if a global cost function is formulated in the manner discribed
+by Cerezo, then its corresponding local cost function will always be 
+less than or equal to the global.
+
+Imports
+--------------
+
+In this notebook, we investigate the effect of barren plateaus in
```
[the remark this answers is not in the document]
Tstorwick: Done!

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+######################################################################
+# Next, we want to define our QNodes and our circuit ansatz. For this
+# simple example, an ansatz that works well is simply a rotation along X,
+# and a rotation along Y, repeated across all the qubits.
+#
+# We will also define our cost functions here. Since we are trying to
+# learn the identity gate a natural cost function is the probability of measuring the 
+# zero state, denoted here as :math:`p_{|0\rangle}`.
```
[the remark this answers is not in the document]
Tstorwick: Good catch!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Done!

demonstrations/tutorial_local_cost_functions.py
Tstorwick: It's in the center of the plateau. I've updated the text to reflect this.

demonstrations/tutorial_local_cost_functions.py
mariaschuld: Great, thanks for the explanation, that clarifies a lot for me!

It would be great to put a bit of this background info into the tutorial, so people can go and look the example up, and they understand where you add new stuff.

Just a minor point though!

mariaschuld:
Hey @Tstorwick , with your explanation about where to find the example and how it is meant in the bigger context, I finally understand the scope of the tutorial. Really nice!

Since I blocked you for so long I approve the PR, but would recommend these two minor changes:

- As mentioned I'd put the repeated plotting code into a function and make calls to that function, so the reader can focus on the important bits
- Thanks for changing the title - I am now not so sure any more my request was so clever. "Alleviating" is more accurate but much less catchy. If you prefer "Avoid" I'm happy for you to change back, but something like "Getting out of barren plateaus" could also work.

Good job!

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+######################################################################
+# It seems our changes didn't significantly alter the overall cost landscape.
+# This probably isn't a general trend, but it is a nice surprise.
+# Now, let us get back to training the local, and global cost functions.
+# Because we have a visualization of the total cost landscape,
+# lets pick a point to exaggerate the problem. One of the worst points in the
+# landscape is :math:`(\pi,0)` as it is in the middle of the plateau, so let's use that.
+
+
+rotations = [[3 for i in range(wires)], [0 for i in range(wires)]]
```
co9olguy, line 281: Just as a Python tip, you can also do something like
```
[[3] * len(wires), [0] * len(wires)]
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+        qml.RY(rotations[1][i], wires=i)
+    qml.broadcast(qml.CNOT, wires=range(wires), pattern="chain")
+    return qml.probs(range(locality))
+
+
+def cost_tunable(rotations):
+    # result = circuit(rotations)
+    # return sum(abs(result[i]) for i in range(wires))
```
co9olguy, line 371:
```suggestion
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
+"""
+Alleviating barren plateaus with local cost functions
```
co9olguy, line 2: @Tstorwick Please credit yourself as the author of this tutorial! :smile: 
See the [barren plateaus example](https://pennylane.ai/qml/demos/tutorial_barren_plateaus.html) for how to do this (email not needed if you don't want)

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+    if not has_been_trained:
+        plateau = plateau + 1
+    print("Trained: {:5d}".format(trained))
+    print("Plateau'd: {:5d}".format(plateau))
+
+
+######################################################################
+# In the global case, anywhere between 70-80% of starting positions are
+# untrainable, a significant number. It is likely as the complexity of our
+# ansatz, and the number of qubits increases, this factor will increase.
```
co9olguy, line 508:
```suggestion
# untrainable, a significant number. It is likely that, as the complexity of our
# ansatz—and the number of qubits—increases, this factor will increase.
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+#
+# Comparing that to our local cost function, every single area trained,
```
co9olguy, line 510:
```suggestion
# We can compare that to our local cost function, where every single area trained,
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+# complex problems.
+    
+    
+##############################################################################
+# References
+# ----------
+#
+# .. [#Cerezo2020]
+#    
+#   Cerezo, M., Sone, A., Volkoff, T., Cincio, L., &amp; Coles, P. (2020, February 04).
```
co9olguy, line 522:
```suggestion
#   Cerezo, M., Sone, A., Volkoff, T., Cincio, L., and Coles, P. (2020). 
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+Barren plateaus are large regions of the cost function's parameter space
+where the variance of the gradient is almost 0; or, put another way, the
+cost function landscape is flat. This means that a variational circuit
+initialized in one of these areas will be untrainable using any gradient based
+algorithm.
+
+
+
+In `"Cost-Function-Dependent Barren Plateaus in Shallow Quantum Neural
+Networks" <https://arxiv.org/abs/2001.00550>`__ Cerezo et al. demonstrate the
```
co9olguy, line 20:
```suggestion
In `"Cost-Function-Dependent Barren Plateaus in Shallow Quantum Neural
Networks" <https://arxiv.org/abs/2001.00550>`__ [#Cerezo2020]_, Cerezo et al. demonstrate the
```

co9olguy (changes requested):
Thanks for the submission @Tstorwick! Hope you had fun making it :smile: 

We'd like to push this towards release now, ideally this week. 

I've left a few suggestions for final polish, spelling and grammar, etc. before release, but all of them are minor. If you are happy with the suggestions, we can merge it and :shipit:

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+Alleviating barren plateaus with local cost functions
+==========
+.. meta::
+    :property="og:description": Local cost functions are cost formulations for variational quantum computing that are more robust to barren plateaus, and noise.
+    :property="og:image": ../demonstrations/local_cost_functions/Cerezo_et_al_local_cost_functions.png
+    
+Barren Plateaus
+---------------
+
+Barren plateaus are large regions of the cost function's parameter space
```
co9olguy, line 11: Might be nice to cross-link to the [barren plateau tutorial](https://pennylane.ai/qml/demos/tutorial_barren_plateaus.html) here?
```suggestion
`Barren plateaus <https://pennylane.ai/qml/demos/tutorial_barren_plateaus.html>`__ are large regions of the cost function's parameter space
```

.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
       - run:
           name: Build tutorials
-          no_output_timeout: 30m
+          no_output_timeout: 60m
```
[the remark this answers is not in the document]
josh146: It might not be anymore. Previously it was, as the tutorial used to take quite long to build (now it only takes ~10 minutes)

demonstrations/tutorial_local_cost_functions.py
Tstorwick: Thanks! Done :)

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+where the variance of the gradient is almost 0; or, put another way, the
```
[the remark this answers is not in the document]
Tstorwick: It is, (at least, the gradient must be almost 0) but that's not how I've seen barren plateaus typically defined. Papers seem to prefer this definition, so I went with it.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,525 @@
...
+
+rotations = [[3 for i in range(wires)], [0 for i in range(wires)]]
+opt = qml.GradientDescentOptimizer(stepsize=0.2)
+steps = 100
+params_global = rotations
+for i in range(steps):
+    # update the circuit parameters
+    params_global = opt.step(cost_global, params_global)
+
+    if (i + 1) % 1 == 0:
```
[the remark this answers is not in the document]
Tstorwick: Right, haha. I had that for testing and forgot about it ._.

.circleci/config.yml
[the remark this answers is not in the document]
Tstorwick: I think it shouldnt be, but I dont think i have the permissions to change this file?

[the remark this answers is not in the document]
josh146: I changed this file originally 🙂 It's fine to leave it as is for now.

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,520 @@
...
+######################################################################
+# Next, we want to define our QNodes and our circuit ansatz. For this
+# simple example, an ansatz that works well is simply a rotation along X,
+# and a rotation along Y, repeated across all the qubits.
+#
+# We will also define our cost functions here. Since we are trying to
+# learn the identity gate, a natural cost function is 1 minus the probability of measuring the 
+# zero state, denoted here as :math:`1 - p_{|0\rangle}`.
+#
+# .. math:: C = \langle  \psi(\theta) | \left(I - |0\rangle \langle 0|\right) \rangle  | \psi(\theta)  \rangle   =1-p_{|0\rangle}
```
co9olguy, line 88:
```suggestion
# .. math:: C = \langle  \psi(\theta) | \left(I - |0\rangle \langle 0|\right)  | \psi(\theta)  \rangle   =1-p_{|0\rangle}
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,520 @@
...
+#
+
+
+def global_cost_simple(rotations):
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+        qml.RY(rotations[1][i], wires=i)
+    return qml.probs(wires=range(wires))
+
+
```
co9olguy, line 115:
```suggestion
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,520 @@
...
+######################################################################
+# Of course, now that we've changed both our cost function and our circuit,
+# we will need to scan the cost landscape again.
+
+
+global_surface = generate_surface(cost_global)
+plot_surface(global_surface)
+
+
```
co9olguy, line 264:
```suggestion
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,520 @@
...
+#
+
+
+def tunable_cost_simple(rotations):
+    for i in range(wires):
+        qml.RX(rotations[0][i], wires=i)
+        qml.RY(rotations[1][i], wires=i)
+    qml.broadcast(qml.CNOT, wires=range(wires), pattern="chain")
+    return qml.probs(range(locality))
+
```
co9olguy, line 364:
```suggestion
```

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -0,0 +1,520 @@
...
+
+def cost_tunable(rotations):
+    return 1 - tunable_circuit(rotations)[0]
+
```
co9olguy, line 368:
```suggestion
```

co9olguy (approved):
Thanks for this @Tstorwick!

PR #96, opened 2020-05-30
10% val accuracy increase

mertall:
I was recently tinkering with your model and got 
Epoch 50/50
 - 0s - loss: 0.0893 - accuracy: 1.0000 - val_loss: 0.8956 - val_accuracy: 0.8000

**Title:**
10% increase in accuracy on quanvolutional  model

**Summary:**
Added a deep layer with 50 units, raised epochs to 50 and used adam optimizer. 
**Relevant references:**

**Possible Drawbacks:**
More epochs means a bit more time to run.
**Related GitHub Issues:**
No issues, just a suggestion to increase the models accuracy. I found it pretty exciting, I will continue to try to improve the accuracy, probably be trying to create a larger test set and train set.


josh146:
Thanks for this nice contribution @mridulsar!

co9olguy:
Thanks @mridulsar, we're just waiting for @andreamari to provide a review (he wrote the original tutorial). Hopefully next week.

andreamari:
Thanks @mridulsar, sorry for my late review.
We are glad that you worked on this tutorial and that you have found a way to improve the accuracy. However, having a large accuracy is not the main goal of this tutorial. The main goal is to present a an example which is **as simple as possible** and, possibly, **as fast as possible** to be executed.

For this reason I would avoid to increase the number of epochs to 50 and to add the new dense layer with 50 nodes. This would make the example too heavy and also too classical (i.e., the model would be dominated by the classical dense layers, making the contribution of the initial quantum layer less clear).

A change that we could still take into consideration is the switch from the SGD optimizer to the Adam optimizer, but only if it gives better results.

mertall:
Hey! Sorry for such a long gap between responses. I got caught up with school. I found that the adam optimizer gives a 10% increase. I attached a graph of the results with adam optimizer.
quanvolutional_results_adam

andreamari:
@mertall thanks, your plots look good! There are still some minor problems:
- The parameter `n_epochs` is still set to 50, while it should be 30 (as in the plots that you posted).
- If we use `optimizer='adam'`, we should remove the line `eta = 0.05      # Learning rate` since it is not used anymore. 
- Some CI tests are failing. It could be due to the superfluous line `eta = 0.05      # Learning rate`.

mertall:
awesome, thanks for those suggestions. I just committed some changes, I think the checks should pass. I will recheck in a few minutes

andreamari:
@josh146 any idea about the problem with failing tests?

josh146:
This is a recent installation bug with Cirq, I've pushed a commit that should fix it :)

mertall:
I was looking over the bug and noticed that the error called on protobuf... it states protobuf~=3.12.0...
From here I took a look at circleci folder in the github and noticed that there is a command with upgrades all packages. I think maybe the protobuf package was updated to 3.12.2 while it looks like cirq is only compatible with 3.12.0. I would test this but my knowledge in utilizing circle ci to check something like version is over my head. I would be happy to understand how to do so if there are any educational resources that you would recommend.

josh146:
Thanks @mertall for catching that! It looks like the Cirq installation bug is now fixed --- previous versions required pinning `protobuf~=3.8`.

mertall:
thats great to hear! When you say they require pinning, does this mean they are held at ver 3.8? I noticed you changed the version of pip, I am curious to understand what influence this had on the system. First time I have seen something like circleci, seems super powerful I am a fan.

josh146:
@mertall, do you want to reopen this, and I can then merge it in!

mertall:
oops, sorry for some reason I thought it got merged. Thank you @josh146 !

PR #97, opened 2020-06-03
Adds VQT Tutorial (WIP)

Lucaman99:
**Title:**

Adds a tutorial on the [Variational Quantum Thermalizer](https://arxiv.org/abs/1910.02071)

**Summary:**

This is a WIP pull request for the VQT tutorial. The tutorial goes into detailed background on what the VQT is and motivation behind how it is constructed. It then goes on to simulate the VQT for Ising and Heisenberg Hamiltonians.

I haven't done many of the tasks required for acceptance of this PR (made final edits to the Notebook, added a thumbnail, etc.), and it probably needs extensive revision, but I wanted to open a WIP PR now to get feedback on my work thus far.

[Here is an NBViewer link to the tutorial](https://nbviewer.jupyter.org/github/Lucaman99/openqml/blob/master/notebooks/vqt_good_copy.ipynb)

**Relevant references:**

The original VQT paper is linked above.

**Possible Drawbacks:**

None that I can think of.

**Related GitHub Issues:**


josh146:
Thanks @Lucaman99! The rendered version of your tutorial as of commit 
447a912 is available here: https://809-214003948-gh.circle-artifacts.com/0/_build/html/demos/tutorial_vqt.html.

Might be worth going through it quickly, and just making sure that any formatting/rendering issues are fixed up.

Lucaman99:
Hey @josh146 - I made a bunch of updates and had to re-render the Python Notebook (hence the force push to overwrite the remote changes). The explanation part is now much shorter and to the point.

josh146:
Thanks @Lucaman99! I made a couple of commits just to fix up the rendering. Here is the rendered version of your tutorial now: https://820-214003948-gh.circle-artifacts.com/0/_build/html/demos/tutorial_vqt.html

Any changes from hereon should be made to the converted `tutorial_vqt.py` file to preserve the rendering fixes 🙂

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+    # Calculates the expectation value of the Hamiltonian, with respect to the preparred states
+
+    return qml.expval(qml.Hermitian(ham_matrix, wires=range(qubit)))
+
+
+qnode = qml.QNode(quantum_circuit, dev)
+
+# Tests and draws the QNode
+
+results = qnode([1 for i in range(0, 12 * depth)], qubits, [1, 0, 1, 0], 3)
```
josh146, line 462: Minor suggestion:
```suggestion
results = qnode(np.ones([12 * depth]), qubits, [1, 0, 1, 0], 3)
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+
+
+######################################################################
+# There is one more thing we must do before implementing the cost
+# function: writing a method that allows us to calculate the entropy of a
+# state. Actually implementing a function that calculates the Von Neumann
+# entropy is not too involved. We will take a probability distribution as
+# our argument, each entry of which corresponds to the digonal elements of
+# the :math:`1`-qubit subsystems. The entropy of a collection of
+# subsystems is the same as the sum of the entropies of the indiivdual
```
josh146, line 473:
```suggestion
# subsystems is the same as the sum of the entropies of the individual
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+
+    return total_entropy
+
+
+######################################################################
+# Finally, we define the cost function. More specifically, this is an
+# **exact** version of the VQT cost function. Instead of sampling from our
+# classical probability distribution, we simply calculate the probability
+# corresponding to every basis state, and thus calculate the energy
+# expectation exactly for each iteration. this is not how the VQT would
```
josh146, line 496:
```suggestion
# expectation exactly for each iteration. This is not how the VQT would
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+    dist_params = params[0:qubit]
+    params = params[qubit:]
+
+    # Creates the probability distribution
+
+    distribution = prob_dist(dist_params)
+
+    # Generates a list of all computational basis states, of our qubit system
+
+    s = [[int(i) for i in list(bin(k)[2:].zfill(qubit))] for k in range(0, 2 ** qubit)]
```
josh146, line 519: We could make use of the `itertools` module for this, which may be more efficient/easier to read:
```pycon
>>> import itertools
>>> qubit = 3
>>> s = itertools.product([0, 1], repeat=qubit)
>>> for i in s:
...     print(i)
(0, 0, 0)
(0, 0, 1)
(0, 1, 0)
(0, 1, 1)
(1, 0, 0)
(1, 0, 1)
(1, 1, 0)
(1, 1, 1)
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+######################################################################
+# Finally, we optimize the cost function:
+#
+
+# Creates the optimizer
+
+iterations = 0
+params = [random.randint(-100, 100) / 100 for i in range(0, (12 * depth) + qubit)]
+
+out = minimize(exact_cost, x0=params, method="COBYLA", options={"maxiter": 1000})
```
josh146, line 553: It has been on our 'todo' list to have a demonstration that uses PennyLane with a SciPy gradient-free optimizer, so very nice to see it working so well here 🙂

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+    final_density_matrix_2 = np.zeros((2 ** qubit, 2 ** qubit))
+
+    # Prepares the optimal parameters, creates the distribution and the bitstrings
+
+    dist_params = params[0:qubit]
+    unitary_params = params[qubit:]
+
+    distribution = prob_dist(dist_params)
+
+    s = [[int(i) for i in list(bin(k)[2:].zfill(qubit))] for k in range(0, 2 ** qubit)]
```
josh146, line 582: Same comment here regarding using `itertools` (although this is a nifty list comprehension workaround!)

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+
+    # Runs the circuit in the case of the optimal parameters, for each bitstring, and adds the result to the final density matrix
+
+    for i in s:
+        qnode(unitary_params, qubits, i, 3)
+        state = device.state
+        for j in range(0, len(i)):
+            state = np.sqrt(distribution[j][i[j]]) * state
+        final_density_matrix_2 = np.add(final_density_matrix_2, np.outer(state, np.conj(state)))
```
josh146, line 591: Minor suggestion, perhaps an in-place addition here?
```suggestion
        final_density_matrix_2 += np.outer(state, np.conj(state))
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+    # Prepares the optimal parameters, creates the distribution and the bitstrings
+
+    dist_params = params[0:qubit]
+    unitary_params = params[qubit:]
+
+    distribution = prob_dist(dist_params)
+
+    s = [[int(i) for i in list(bin(k)[2:].zfill(qubit))] for k in range(0, 2 ** qubit)]
+
+    # Runs the circuit in the case of the optimal parameters, for each bitstring, and adds the result to the final density matrix
```
josh146, line 584: Adding in a line break here to avoid horizontal scrolling on the website 
```suggestion
    # Runs the circuit in the case of the optimal parameters, for each bitstring,
    # and adds the result to the final density matrix.
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+######################################################################
+# Then, we can compare it to the target:
+#
+
+seaborn.heatmap(abs(final_density_matrix))
+
+
+######################################################################
+# As you can see, the two images are not completely the same, but there is
+# definitely some resemblance between them!
```
josh146, line 641: Awesome!

Lucaman99:
Thanks for all of these suggestions @josh146, I'll get working on making changes right away! 🙂

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+process. This initial density matrix will be parametrized by a
+collection of parameters :math:`\boldsymbol\theta`, which will describe
+the probabilities corresponding to different pure states. In this
+implementation of the algorithm, we will use the idea of a **factorized
+latent space** where the initial density matrix describing our quantum
+system in completely un-correlated. It is simply a tensor product of
+multiple :math:`2 \times 2`, diagonal (in the computational basis)
+density matrices, each corresponding to one qubit. This works well for
+scalability of the algorithm, because instead of requiring
+:math:`|\boldsymbol\theta| = 2^n`, for an :math:`n` qubit, diagonal
```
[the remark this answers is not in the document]
Lucaman99: Yep, `\boldsymbol\theta` is the set of parameters, so `|\boldsymbol\theta|` is just the size of the set

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+# function is called a sigmoid, and is a common choice as an activation
+# function in machine learning methods. We can implement the sigmoid as:
+#
+
+# Creates the probability distribution according to the theta parameters
+
+
+def sigmoid(x):
+
+    return math.exp(x) / (math.exp(x) + 1)
```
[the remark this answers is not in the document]
Lucaman99: Not sure, wasn't really thinking about it, I'll change to np (better to be consistent)

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,825 @@
...
+    \mathcal{L}(\theta, \ \phi) \ = \ \beta \langle \hat{H} \rangle \ - \ S_\theta \ = \
+    \beta \ \text{Tr} (\hat{H} \ \rho_{\theta \phi}) \ - \ S_\theta \ = \ \beta \ \text{Tr}( \hat{H} \ \hat{U}(\phi)
+    \rho_{\theta} \hat{U}(\phi)^{\dagger} ) \ - \ S_\theta
+
+Where :math:`\rho_\theta` is the initial density matrix, :math:`U(\phi)`
+is the paramterized ansatz, and :math:`S_\theta` is the von Neumann
+entropy of :math:`\rho_{\theta \phi}`. It is important to note that the
+von Neumann entropy of :math:`\rho_{\theta \phi}` is the same as the von
+Neumann entropy of :math:`\phi_{\theta}`, since entropy is invariant
+under unitary transformations:
```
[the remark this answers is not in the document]
Lucaman99: Sorry, \phi_\theta should have been \rho_\theta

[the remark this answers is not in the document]
ixfoduap: Also, `phi` hasn't been defined yet

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+
+.. math::
+    S(\rho') = \ - \text{Tr} (\rho' \log \rho') &= \ - \text{Tr} ( U \rho U^{\dagger} \log (U \rho U^{\dagger}))  \\
+             &= - \text{Tr} ( U \rho U^{\dagger} \log \rho)  \\
+             &= \ - \text{Tr} ( U \rho \log \rho U^{\dagger}) \\
+             &= - \text{Tr} ( \rho \log \rho U^{\dagger} U) \\
+             &= \ - \text{Tr} ( \rho \log \rho) \\
+             &= \ S(\rho)
```
ixfoduap, line 110: Remove this, no need for this derivation in a PL tutorial

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+
+We repeat the algorithm with new parameters until we minimize free
```
ixfoduap, line 112: What parameters are we optimizing over exactly? I suppose it's both the `thetas` in the input state and the `phis` in the circuit, right? Needs to be more clearly explained

ixfoduap: How are the new parameters chosen?

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+energy. Once we have done this, we have arrived at the thermal state.
+This comes from the fact that our free energy cost function is equivalent
```
ixfoduap, line 114:
```suggestion
This comes from the fact that the free energy cost function is equivalent
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+Since relative entropy must be positive, and is clearly :math:`0` when
+:math:`\rho_{\theta \phi} \ = \ \rho_{\text{Thermal}}`, it follows that
+relative entropy, and hence :math:`\mathcal{L}(\theta, \ \phi)` (since
+it only differs from rleative entropy by an overall additive constant),
+are minimized when
+:math:`\rho_{\theta \phi} \ = \ \rho_{\text{Thermal}}`. So, we know that
+we have to minimize :math:`\mathcal{L}` to find the thermal state. More
+specifically, when
+:math:`\mathcal{L}(\theta, \ \phi) \ = \ - \log Z_{\beta}`, then we have
+minimized the cost function and have found the thermal state.
```
ixfoduap: All this can be removed. It's enough to say that relative entropy is minimized when both states are equal, so we are training the circuit to produce a state that is equal to the target state.

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+
+We repeat the algorithm with new parameters until we minimize free
+energy. Once we have done this, we have arrived at the thermal state.
+This comes from the fact that our free energy cost function is equivalent
+to the relative entropy between :math:`\rho_{\theta \phi}` and our
+target thermal state. Relative entropy is defined as:
+
+.. math:: D(\rho_1 || \rho_2) \ = \ \text{Tr} (\rho_1 \log \rho_1) \ - \ \text{Tr}(\rho_1 \log \rho_2)
+
+If we let :math:`\rho_1` be :math:`\rho_{\theta \phi}` and
```
ixfoduap: Here you can just say that this is minimized when both states are equal

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+:math:`\mathcal{L}(\theta, \ \phi) \ = \ - \log Z_{\beta}`, then we have
+minimized the cost function and have found the thermal state.
+
+For a diagramatic representation of how this works, check out Figure 3
+from the `original VQT paper <https://arxiv.org/abs/1910.02071>`__.
+
+The 3-Qubit Ising Model on a Line
+---------------------------------
+
+We will begin by consdering the Ising model on a linear graph, for 3
```
ixfoduap, line 146:
```suggestion
We begin by consdering the Ising model on a linear graph, for three
```
It's better style to use words (e.g. two, seven) to denote numbers smaller than ten

ixfoduap: Also, why just three qubits? I guess it's too late, but I would assume that we could push qubit number a bit further, no?

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+qubits. This is a fairly simple model, and will act as a good test to
+see if the VQT is working as it is supposed to.
```
ixfoduap, line 148: Nice!

demonstrations/tutorial_vqt.py
ixfoduap: And then quote the equivalence between relative entropy and the cost function

demonstrations/tutorial_vqt.py
Lucaman99: I agree, 3 qubits isn't a lot, however, I didn't want the simulations to require a ton of steps, and thus take a long time to converge (this tutorial is already pushing the 10 minute cap on runtime).

Lucaman99: It eventually does a simulation for the case of 4 quits as well

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,826 @@
...
+In this Notebook, we will be discussing how to go about implementing and
+experimenting a recently proposed quantum algorithm called the
+**Variational Quantum Thermalizer**. Essentially, this algorithm is able
+to use a variational approach to reconstruct the thermal state of a
+given Hamiltonian at a given temperature. This is a task that is
+performed much more efficiently on a quantum device than a classical
+simulations performing the same calculations (for large enough systems).
+In fact, the original paper demonstrates that the VQT is actually a
+generalization of VQE, and as the effective "temperature" of our
+simulation approaches zero, our algorithm similarly approaches the VQE.
```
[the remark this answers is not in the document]
Lucaman99: @ixfoduap Is it enough to link the VQE tutorial in the gallery, which I do in the initial explanation?

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+#
+
+# Creates the target density matrix
+
+
+def create_target(qubit, beta, ham, graph):
+
+    # Calculates the matrix form of the density matrix, by taking the exponential of the Hamiltonian
+
+    h = ham(num_qubits, graph)
```
ixfoduap, line 263: :pig:

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+    y = -1 * float(beta) * h
```
ixfoduap, line 264: Is this `float` necessary?

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+    new_matrix = scipy.linalg.expm(np.array(y))
+    norm = np.trace(new_matrix)
+    final_target = (1 / norm) * new_matrix
+
+    # Calculates the entropy, the expectation value, and the final cost
+
+    entropy = -1 * np.trace(np.matmul(final_target, scipy.linalg.logm(final_target)))
```
ixfoduap, line 271: You can use `@` instead of `matmul`, so `X @ Y` to perform matrix multiplication of `X` and `Y`

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+
+    print("Expectation Value: " + str(ev))
+    print("Entropy: " + str(entropy))
+    print("Final Cost: " + str(real_cost))
+
+    return final_target
+
+
+######################################################################
+# Finally, we can calculate the thermal state corresponding to our
```
ixfoduap, line 285:
```suggestion
# Finally, we can calculate the thermal state corresponding to the
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+# Hamiltonian and inverse temperature, and visualize it using the
+# ``seaborn`` data visualization library:
+#
+
+# Plots the final density matrix
+
+final_density_matrix = create_target(num_qubits, beta, create_hamiltonian_matrix, interaction_graph)
+seaborn.heatmap(abs(final_density_matrix))
```
ixfoduap, line 293: Technically, you're visualizing the absolute value of the density matrix.

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+
+
+######################################################################
```
ixfoduap, line 296: I think it's fine to be exhaustive and to explain things in detail, but at this point I feel like I've gone through a lot of tutorial and still haven't gotten to VQT. So let's consider shortening the beginning

Lucaman99: Sounds good, did you have anything in particular in mind that we could cut-out/shorten?

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+
+    matrix = np.zeros((2 ** n, 2 ** n))
+
+    for i in graph.edges:
+        m = 1
+        for j in range(0, n):
+            if j == i[0] or j == i[1]:
+                m = np.kron(m, qml.PauliX.matrix)
+            else:
+                m = np.kron(m, np.identity(2))
```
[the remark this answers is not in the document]
Lucaman99: The full section of the code is:

```
m = 1
  for j in range(0, n):
    if j == i:
      m = np.kron(m, qml.PauliZ.matrix)
    else:
      m = np.kron(m, np.identity(2))
```

So `m` is only `1` for the first cycle of the loop.

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,815 @@
...
+
+######################################################################
+# The model that we are investigating lives on a linear graph, which we
+# will construct using ``networkx`` for the purposes of eventually
+# constructing our Hamiltonian:
+#
+
+# Creates the graph of interactions for the Heisenberg grid, then draws it
+
+interaction_graph = nx.Graph()
```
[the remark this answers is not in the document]
Lucaman99: I might be wrong, but from what I recall, a line graph is something different https://en.wikipedia.org/wiki/Line_graph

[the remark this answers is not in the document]
Lucaman99: However, I'll take a look at the NetworkX documentation and see if there's a feature that implements this 🙂

demonstrations/tutorial_vqt.py
```diff
@@ -233,44 +201,41 @@ def create_hamiltonian_matrix(n, graph):
...
     new_matrix = scipy.linalg.expm(np.array(y))
     norm = np.trace(new_matrix)
     final_target = (1 / norm) * new_matrix
 
     # Calculates the entropy, the expectation value, and the final cost
 
-    entropy = -1 * np.trace(np.matmul(final_target, scipy.linalg.logm(final_target)))
-    ev = np.trace(np.matmul(final_target, h))
-    real_cost = beta * np.trace(np.matmul(final_target, h)) - entropy
+    entropy = -1 * np.trace(final_target @ scipy.linalg.logm(final_target))
```
ixfoduap, line 236: you can use `from scipy.linalg import logm` in the preamble and then just use `logm` here

demonstrations/tutorial_vqt.py
```diff
@@ -282,12 +247,12 @@ def create_target(qubit, beta, ham, graph):
...
-# Finally, we can calculate the thermal state corresponding to our
-# Hamiltonian and inverse temperature, and visualize it using the
-# ``seaborn`` data visualization library:
+# Finally, the thermal state corresponding to the pre-defined
+# Ising Hamiltonian and inverse temperature can be calculated. 
+# We also visualize it using the ``seaborn`` library:
 #
 
-# Plots the final density matrix
+# Plots the absolute value of the final density matrix
```
ixfoduap, line 255:
```suggestion
# Plots the entry-wise absolute value of the final density matrix
```

demonstrations/tutorial_vqt.py
```diff
@@ -300,41 +265,36 @@ def create_target(qubit, beta, ham, graph):
...
-# matrix. As we discussed earlier in this demonstration, we will be using the
-# factorized latent space model. We just have to decide how we will define
-# each probability in the factorized space. We will let the probability
-# associated with the :math:`j`-th one-qubit system be:
+# Now that we know exactly what the thermal state should look like, we
+# attempt to construct it with the VQT. We begin by calculating the
+# classical probability distribution, which gives the probabilities
+# corresponding to each basis state in the initial, factorized density  
+# matrix. We let the probability associated with the :math:`i`-th one-qubit 
+# system be:
```
ixfoduap, line 273:
```suggestion
# system be given by the sigmoid function:
```

demonstrations/tutorial_vqt.py
```diff
@@ -300,41 +265,36 @@ def create_target(qubit, beta, ham, graph):
...
 #
-# The motivation behind this choice is the fact that this function has a
-# range of :math:`0` to :math:`1`, which is natural for defining
-# probability without constraining our parameters. In addition, this
-# function is called a sigmoid, and is a common choice as an activation
-# function in machine learning methods. We can implement the sigmoid as:
+# The motivation behind this choice comes from the fact that this function has a
+# range of :math:`0` to :math:`1`. This means that we don't need to constrain 
+# the values of the parameters to some subset of :math:`\mathbb{R}`. We can implement 
+# this function (which is called a sigmoid) as:
```
ixfoduap, line 280:
```suggestion
# The motivation behind this choice is that the sigmoid has a
# range of :math:`0` to :math:`1`, meaning we don't need to constrain 
# the parameter values to some subset of :math:`\mathbb{R}`. We can implement 
# this function as:
```

demonstrations/tutorial_vqt.py
```diff
@@ -300,41 +265,36 @@ def create_target(qubit, beta, ham, graph):
...
+# probabilities that correspond to each one-qubit system
 # in the factorized latent space:
 #
 
 # Creates the probability distributions for each of the one-qubit systems
 
-
 def prob_dist(params):
 
     dist = []
```
ixfoduap, line 300: I think you can just use 
```
def prob_dist(params):
    return [[sigmoid(i), 1 - sigmoid(i)] for i in params]
```
Also maybe good to use `p` instead of `i`, i.e. `[[sigmoid(p), 1 - sigmoid(p)] for  in params]`

demonstrations/tutorial_vqt.py
```diff
@@ -345,15 +305,14 @@ def prob_dist(params):
 
 
 ######################################################################
-# Now, with this done, we have to define the quantum parts of our circuit.
-# Befor any qubit register is passed through the variational circuit, we
-# must prepare it in a given basis state. Thus, we can write a function
-# that takes a list of bits, and returns a quantum circuit that prepares
-# the corresponding basis state (in the computational basis):
+# Now, we can define the quantum components of the VQT.
+# The first step of the VQT circuit is to prepare a given
```
ixfoduap, line 309:
```suggestion
# The first step of the circuit is to prepare a given
```

demonstrations/tutorial_vqt.py
```diff
@@ -0,0 +1,783 @@
...
+    for i in range(0, len(prep_state)):
+        if prep_state[i].val == 1:
+            qml.PauliX(wires=i)
+
+
+######################################################################
+# Now, we build the parametrized circuit, through which we pass the initial
+# states. We use a multi-layered ansatz, where each layer is composed
+# of :math:`RX`, :math:`RZ`, and :math:`RY` gates on each qubit, followed
+# by exponentiated :math:`CNOT` gates placed between qubits that share an
```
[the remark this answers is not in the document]
Lucaman99: I mean CRX gates, but I get what you're saying, the language is sloppy

PR #103, opened 2020-06-14
Creates New QGRNN Tutorial

Lucaman99:
**Title:**

Creates a new and improved version of the QGRNN tutorial in https://github.com/XanaduAI/qml/pull/98

This tutorial is still a work-in-progress, hence the draft PR.

**Summary:**

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

https://github.com/XanaduAI/qml/pull/98


Lucaman99:
@ixfoduap The first part of this tutorial deals with generating data to be used in the neural network. I know that part of your feedback for the VQT was that the tutorial should actually dive into implementing the VQT faster, and this tutorial kind of has the same problem. I've done a lot of thinking, and I can't really think of any good way to expedite the process of generating the data. If you have any suggestions, let me know! 🙂

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+
+######################################################################
+# We define the QGRNN to be a variational ansatz of the form:
+#
+# .. math:: U_{H}(\boldsymbol\gamma, \ \boldsymbol\theta) \ = \ \displaystyle\prod_{i \ = \ 1}^{P} \Bigg[ \displaystyle\prod_{j \ = \ 1}^{Q} e^{-i \gamma_{j} H_{j}(\boldsymbol\theta) }\Bigg],
+#
+
+
+######################################################################
+# where we have:
```
ixfoduap, line 53:
```suggestion
# where
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+#
+
+
+######################################################################
+# .. math:: \hat{H}_{j}(\boldsymbol\theta) \ = \ \displaystyle\sum_{(a,b) \in E} \displaystyle\sum_{c} V_{jabc} \hat{A}_{a}^{jc} \otimes \hat{B}_{b}^{jc} \ + \ \displaystyle\sum_{v \in V} \displaystyle\sum_{d} W_{jvd} \hat{C}_{v}^{jd}.
```
ixfoduap, line 58: This is very confusing. What are V, A, B, W and C?

ixfoduap: In fact, this whole section is confusing. Let's chat more offline

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+
+
+######################################################################
+# Learning an Ising Model with the QGRNN
+# ---------------------------------------
+#
+
+
+######################################################################
+# In this demonstration, we attempt to learn the parameters corresponding
```
ixfoduap, line 131:
```suggestion
# We now attempt to learn the parameters corresponding
```
Since we use "in this demo" above

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+######################################################################
+# Getting Started
+# ^^^^^^^^^^^^^^^^
+#
+
+
+######################################################################
+# We begin by importing the necessary dependencies:
+#
+
```
ixfoduap, line 146: What's the goal here? What are we trying to do? This is not clear to me and therefore probably merits a small paragraph to help lead the reader

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+# the simulation.
+#
+
+
+qubit_number = 4
+qubits = range(qubit_number)
+
+
+######################################################################
+# In order to use the QGRNN, we need access to quantum data. In the real
```
ixfoduap, line 166: Why do we need access to quantum data? And having access to quantum data in "the real world" is a problem! Or do you conveniently have some lying around?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+# world, this wouldn’t be a problem, but in this simulation, we have to
+# generate the quantum data ourselves. To do this, we must have *a priori*
```
ixfoduap, line 168: Why do we need this a priori knowledge?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+# knowledge of the target Hamiltonian and the interaction graph. We choose
+# the Hamiltonian to be a transerve field Ising model of the form:
+#
+# .. math:: \hat{H}(\boldsymbol\theta) \ = \ \displaystyle\sum_{(i, j) \in E} \theta_{ij}^{(1)} Z_{i} Z_{j} \ + \ \displaystyle\sum_{i} \theta_{i}^{(2)} Z_{i} \ + \ \displaystyle\sum_{i} X_{i},
+#
+# where :math:`E` is the set of edges in the interaction graph.
```
ixfoduap, line 174: The notion of an interaction graph is not clear

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+ising_graph = nx.cycle_graph(qubit_number)
+
+nx.draw(ising_graph)
+print("Edges: {}" .format(ising_graph.edges))
+
+
+######################################################################
+# We then can initialize the “unknown” target parameters that describe the
+# Hamiltonian, :math:`\alpha \ = \ \{\alpha^{(1)}, \ \alpha^{(2)}\}`,
+# randomly
```
ixfoduap, line 190:
```suggestion
# We then can randomly initialize the “unknown” target parameters that 
# describe the Hamiltonian, :math:`\alpha \ = \ \{\alpha^{(1)}, \ \alpha^{(2)}\}`,
# 
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,633 @@
...
+    return [interaction, bias]
+
+# Creates and prints the parameters for our simulation
+matrix_params = create_params(ising_graph)
+print("Target parameters: {}".format(matrix_params))
+
+
+######################################################################
+# Finally, we use this information to generate the matrix form of the
+# Ising model Hamiltonian (in the computational basis):
```
ixfoduap, line 211:
```suggestion
# Ising model Hamiltonian in the computational basis:
```

demonstrations/tutorial_qgrnn.py
Lucaman99: I meant that in a real-world situation where someone would be using this algorithm, they would have access to some source of quantum data that has evolved under an unknown Hamiltonian (if they didn't, they would have no use for the algorithm in the first place).

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# neural network ansatz (`QGNN <https://arxiv.org/abs/1909.12264>`__).
+
+######################################################################
+# The Idea
+# --------
+#
+
+
+###################################################################### 
+# Recall that a graph is defined as a set of **nodes**, along with a set of
```
ixfoduap, line 26:
```suggestion
# A graph is defined as a set of **nodes**, along with a set of 
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# to nodes and edges, which we call **weights**.
+# It is usually convenient to think of a graph visually:
+#
+# .. image:: ../demonstrations/qgrnn/graph.png
+#     :width: 70%
+#     :align: center
+# 
+# In recent years, the idea of a 
+# `graph neural network <https://arxiv.org/abs/1812.08434>`__ has been
+# receving a lot of attention from the machine learning research community.
```
ixfoduap, line 38:
```suggestion
# receving a lot of attention from the machine learning community. 
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# It is usually convenient to think of a graph visually:
+#
+# .. image:: ../demonstrations/qgrnn/graph.png
+#     :width: 70%
+#     :align: center
+# 
+# In recent years, the idea of a 
+# `graph neural network <https://arxiv.org/abs/1812.08434>`__ has been
+# receving a lot of attention from the machine learning research community. 
+# A **graph neural network** seeks
```
ixfoduap, line 39: So many parentheses in this paragraph! No bueno

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# to learn a representation (a mapping of data into a
+# lower-dimensional vector space) of a given graph with features assigned
+# to nodes and edges such that the each of the vectors in the learned
+# representation preserves not only the features, but also the overall 
+# topology of the graph (which nodes are connected by edges). The 
+# quantum graph neural network attempts to do something similar, but for
+# features that are quantum-mechanical (for instance, a
+# collection of quantum states).
```
ixfoduap, line 47:
```suggestion
# to nodes and edges.  Each of the vectors in the learned
# representation preserves not only the features, but also the overall 
# topology of the graph, i.e., which nodes are connected by edges. The 
# quantum graph neural network attempts to do something similar, but for
# features that are quantum-mechanical; for instance, a
# collection of quantum states.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+#
+
+
+######################################################################
+# Consider the class of Hamiltonians that are **quadratic**, meaning that
```
ixfoduap, line 52:
```suggestion
# Consider the class of qubit Hamiltonians that are **quadratic**, meaning that 
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# quantum graph neural network attempts to do something similar, but for
+# features that are quantum-mechanical (for instance, a
+# collection of quantum states).
+#
+
+
+######################################################################
+# Consider the class of Hamiltonians that are **quadratic**, meaning that 
+# the terms of the Hamiltonian represent either interactions between two 
+# qubits, or bias on one qubits.
```
ixfoduap, line 54:
```suggestion
# qubits, or the energy of individual qubits. 
```
Bias is a term used in analogy with weights and biases of a neural network or Boltzmann machine

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# features that are quantum-mechanical (for instance, a
+# collection of quantum states).
+#
+
+
+######################################################################
+# Consider the class of Hamiltonians that are **quadratic**, meaning that 
+# the terms of the Hamiltonian represent either interactions between two 
+# qubits, or bias on one qubits. 
+# This class of Hamiltonians is naturally described by graphs, with
```
ixfoduap, line 55: Excellent!

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# field Ising model, which is defined as
+#
+# .. math:: \hat{H}_{\text{Ising}}(\boldsymbol\theta) \ = \ \displaystyle\sum_{(i, j) \in E} \theta_{ij}^{(1)} Z_{i} Z_{j} \ + \ \displaystyle\sum_{i} \theta_{i}^{(2)} Z_{i} \ + \ \displaystyle\sum_{i} X_{i},
+#
+# where :math:`\boldsymbol\theta \ = \ \{\theta^{(1)}, \ \theta^{(2)}\}`.
+# In this Hamiltonian, the set :math:`E` that determines which pairs of qubits 
+# have :math:`ZZ` interactions is exactly the set of edges for some graph.
+# The :math:`\theta^{(1)}` parameters correspond to the edge weights and
+# the :math:`\theta^{(2)}` 
+# parameters correspond to weights on the nodes.
```
ixfoduap, line 69: Maybe a good place to introduce the terminology of _interaction graph_

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+#
+
+
+######################################################################
+# This result implies that we can think about **quantum circuits** with
```
ixfoduap, line 74: It's up to you, but I find that `**bold**` is too strong and it's best to use `*italics*` instead

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# parameters correspond to weights on the nodes.
+#
+
+
+######################################################################
+# This result implies that we can think about **quantum circuits** with 
+# graph-theoretic properties. Recall that the time-evolution operator 
+# with respect to some Hamiltonian :math:`H` is defined as:
+#
+# .. math:: U \ = \ e^{-it H}
```
ixfoduap, line 78:
```suggestion
# .. math:: U \ = \ e^{-it H}.
```
Don't forget to add punctuation in equations, a lesson from the previous tutorial!

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+#
+# Thus, we have a clean way of taking our graph-theoretic Hamiltonians and turning 
+# them into unitaries (quantum circuits) that possess the same correspondance to a graph.
+# In the case of the Ising Hamiltonian, we have:
```
ixfoduap, line 82:
```suggestion
# Thus, we have a clean way of taking graph-theoretic Hamiltonians and turning 
# them into unitaries (quantum circuits) that preserve the same correspondance to a graph.
# In the case of the Ising Hamiltonian, we have:
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# for some parametrized Hamiltonian, :math:`H(\boldsymbol\mu)`, of 
+# quadratic order.
+
+######################################################################
+# Using the QGRNN
+# ^^^^^^^^^^^^^^^^
+#
+
+######################################################################
+# Since the QGRNN ansatz is equivalent to the
```
ixfoduap, line 111: Great!

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+
+######################################################################
+# Since the QGRNN ansatz is equivalent to the 
+# approximate time evolution of some quadratic Hamiltonian, we can use it
+# to learn the dynamics of a quantum system.
+#
+# Continuing on with the Ising model example, let's imagine we have some system
+# governed by :math:`\hat{H}_{\text{Ising}}(\boldsymbol\alpha)` for an unknown set of
+# target parameters, 
+# :math:`\boldsymbol\alpha`. Let's also imagine we have access to a bunch of copies of some
```
ixfoduap, line 118:
```suggestion
# :math:`\boldsymbol\alpha`. Let's also suppose we have access to a bunch of copies of some 
```
repeated use of imagine

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# governed by :math:`\hat{H}_{\text{Ising}}(\boldsymbol\alpha)` for an unknown set of
+# target parameters, 
+# :math:`\boldsymbol\alpha`. Let's also imagine we have access to a bunch of copies of some 
+# low-energy state with respect to the target Hamiltonian, :math:`|\psi_0\rangle`. In addition, 
+# we have access to a collection of time evolved states, 
+# :math:`\{ |\psi(t_1)\rangle, \ |\psi(t_2)\rangle, \ ..., \ |\psi(t_N)\rangle \}`, defined by:
+#
+# .. math:: |\psi(t_k)\rangle \ = \ e^{-i t_k \hat{H}_{\text{Ising}}(\boldsymbol\alpha)} |\psi_0\rangle
+#
+# From here, we randomly pick a given number of time-evolved states
```
ixfoduap, line 125: Good to clarify that we know the times `t_i` as well, perhaps as a classical label accompanying the state. It may be good to come up with an example, but let's think about that later

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# from our collection. For some state that we choose, which is
+# evolved to some arbitrary time :math:`t_k`, we compare it
+# to
+#
+# .. math:: U_{\hat{H}_{\text{Ising}}}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle \ \approx \ e^{-i t_k \hat{H}_{\text{Ising}}(\boldsymbol\mu)} |\psi_0\rangle
+#
+# by feeding one of the copies of :math:`|\psi_0\rangle` into a quantum circuit 
+# with the QGRNN ansatz, with some "guessed" set of parameters :math:`\boldsymbol\mu`.
+# We then use a classical optimizer to maximize the average 
+# "similarity" between the time-evolved states and the states outputted
```
ixfoduap, line 135: outputted? Sounds awkward

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# from the QGRNN.
+#
+# Clearly, as the QGRNN states becomes more "similar" to
```
ixfoduap, line 138:
```suggestion
# As the QGRNN states becomes more "similar" to 
```
Pro tip: avoid using "clearly", "it is trivial", "is easy to see" and other similar expressions. They don't really help, can be removed without consequence, and when present can be annoying for readers that do not agree with the claimed simplicity

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,663 @@
...
+# .. math:: U_{\hat{H}_{\text{Ising}}}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle \ \approx \ e^{-i t_k \hat{H}_{\text{Ising}}(\boldsymbol\mu)} |\psi_0\rangle
+#
+# by feeding one of the copies of :math:`|\psi_0\rangle` into a quantum circuit 
+# with the QGRNN ansatz, with some "guessed" set of parameters :math:`\boldsymbol\mu`.
+# We then use a classical optimizer to maximize the average 
+# "similarity" between the time-evolved states and the states outputted
+# from the QGRNN.
+#
+# Clearly, as the QGRNN states becomes more "similar" to 
+# each time-evolved state (for each sampled time), it follows that
```
ixfoduap, line 139:
```suggestion
# each time-evolved state for each sampled time, it follows that 
```

demonstrations/tutorial_qgrnn.py
Lucaman99: I'm actually starting to think that defining calling the graph an _internaction graph_. might not be necessary. I'm trying out some new wording right now, so I'll let you know if I add it.

Lucaman99: Nevermind, it's necessary 😆

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# state by numerically finding the lowest eigenvalue of the Hamiltonian
+# matrix and comparing it to the energy expectation of the low-energy state:
+#
+
+# Finds the energy expectation
+
+def expectation_value(vector, matrix):
+    return np.inner(np.conj(vector), (matrix @ vector))
+
+energy_exp = expectation_value(low_energy_state, ham_matrix)
```
ixfoduap, line 314:
```suggestion
energy_exp = np.inner(np.conj(low_energy_state), (ham_matrix @ low_energy_state))
```
and remove the function definition above. Unless it's used later? I haven't read the whole tutorial yet

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+def ground_state_energy(matrix):
+
+    # Finds the eigenstates of the matrix
+    val = np.linalg.eig(matrix)[0]
+
+    # Returns the minimum eigenvalue
+    return min(val)
+
+ground_state = ground_state_energy(ham_matrix)
```
ixfoduap, line 325: I think here too you don't need a function definition. Should be able to just use
`ground_state = min(np.linalg.eig(ham_matrix)[0])`

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+print("Ground State Energy: {}".format(ground_state))
+
+
+######################################################################
+# This shows us that we have in fact found a low-energy, non-ground state, 
+# as the energy expectation is slightly greater than the energy of the true ground
+# state. This, however, is only half of the information we need. We also require 
+# a collection of time-evolved states. 
+# Luckily, evolving a state forward in time is fairly straightforward, all we
```
ixfoduap, line 334:
```suggestion
# Evolving a state forward in time is fairly straightforward: all we 
```
Again, is it a matter of luck or a matter of fact?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# a collection of time-evolved states. 
+# Luckily, evolving a state forward in time is fairly straightforward, all we 
+# have to do is multiply the initial state by a time-evolution unitary. We define 
+# this operation as a custom gate in PennyLane:
+#
+
+
+def state_evolve(hamiltonian, qubits, time):
+
+    U = scipy.linalg.expm(complex(0, -1) * hamiltonian * time)
```
ixfoduap, line 342:
```suggestion
    U = scipy.linalg.expm(-1j* hamiltonian * time)
```
`1j` is square-root of -1 in python

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+    qml.QubitUnitary(U, wires=qubits)
+
+
+######################################################################
+# We don't actually generate this time-evolved quantum data quite yet,
```
ixfoduap, line 347:
```suggestion
# We don't actually generate time-evolved quantum data quite yet, 
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# Learning the Hamiltonian
+# ^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+
+
+######################################################################
+# With the quantum data defined, we are able to construct the QGRNN and
+# learn the target Hamiltonian. As was discussed above, we wish to use the
+# QGRNN to approximate time-evolution of the target Hamiltonian, with
+# guessed parameters. It follows that we let each of the three
```
ixfoduap, line 362:
```suggestion
# learn the target Hamiltonian. We wish to use the
# QGRNN to approximate time-evolution of the target Hamiltonian.
#. We let each of the three terms
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# collections of :math:`ZZ`, :math:`Z`, and :math:`X` terms from the Ising
+# Hamiltonian. This gives us:
+#
+
+
+def qgrnn_layer(param1, param2, qubits, graph, trotter):
+
+    # Applies a layer of coupling gates (based on a graph)
+    for count, i in enumerate(graph.edges):
+        qml.MultiRZ(2 * param1[count] * trotter, wires=[i[0], i[1]])
```
ixfoduap, line 373: Very nice!

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+    for count, i in enumerate(qubits):
+        qml.RZ(2 * param2[count] * trotter, wires=i)
+
+    # Applies a layer of RX gates
+    for i in qubits:
+        qml.RX(2 * trotter, wires=i)
+
+
+######################################################################
+# As was mentioned in the first section, the QGRNN has two qubit
```
ixfoduap, line 385:
```suggestion
# As was mentioned in the first section, the QGRNN has two
```
They involve more than one qubit, right?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# registers. In one register, some piece of quantum data
+# :math:`|\psi(t)\rangle` is prepared and in the other we have
+# :math:`U_{H}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle`. We need a
+# way to measure the similarity between these states. 
+# One way that we can do this is by using the fidelity, which is
```
ixfoduap, line 390:
```suggestion
# We can do this by using the fidelity, which is
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+
+def swap_test(control, register1, register2):
+
+    qml.Hadamard(wires=control)
+    for i in range(0, len(register1)):
+        qml.CSWAP(wires=[int(control), register1[i], register2[i]])
+    qml.Hadamard(wires=control)
+
+
```
ixfoduap, line 405: Maybe nice to add a short sentence explaining that the fidelity of the states is related to the probability of measuring |0> in the control qubit after the swap test, which in turn is determined by the expectation value of a computational basis measurement in the control qubit

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+######################################################################
+# Before creating the full QGRNN and the cost function, we
+# define a few more fixed values. Among these fixed values is a "guessed"
```
ixfoduap, line 408:
```suggestion
# define a few more fixed values. Among these is a "guessed"
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+reg1 = list(range(qubit_number))  # First qubit register
+reg2 = list(range(qubit_number, 2 * qubit_number))  # Second qubit register
+
+control = 2 * qubit_number  # Index of control qubit
+trotter_step = 0.01  # Trotter step size
+
+# Defines the interaction graph for the new qubit system
+
+new_ising_graph = nx.Graph()
```
ixfoduap, line 424: I think you can just use `new_ising_graph = nx.complete_graph(nr_nodes)`

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+new_ising_graph.add_edges_from([(4, 5), (5, 6), (6, 7), (4, 6), (7, 4), (5, 7)])
+
+print("Edges: " + str(new_ising_graph.edges))
+nx.draw(new_ising_graph)
+
+
+######################################################################
+# Part of the idea behind the QGRNN is that
+# we don’t know the interaction graph, and it has to be learned. In this case, the graph  
+# is learned **automatically** as the target parameters are learned. The
```
ixfoduap, line 435:
```suggestion
# is learned **automatically** as the target parameters are optimized. The
```
To avoid repeated use of learned

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# :math:`\boldsymbol\mu` parameters that correspond to edges that don't exist in 
+# the target graph will simply approach :math:`0`.
+#
+# With this done, we implement the QGRNN circuit for some given time value:
+#
+
+
+def qgrnn(params1, params2, time=None):
+
+    # Prepares the low energy state in the two qubit registers
```
ixfoduap, line 445:
```suggestion
    # Prepares the low energy state in the two registers
```
Across the tutorial: it may be better to just use register as opposed to qubit register

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+    # Applies the SWAP test between the registers
+    swap_test(control, reg1, reg2)
+
+    # Returns the results of the SWAP test
+    return qml.expval(qml.PauliZ(control))
+
+
+######################################################################
+# Notice how the value returned from the circuit is
```
ixfoduap, line 464: A version of this paragraph maybe better after the definition of the SWAP test circuit.

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+# function to minimize. We know that
+# :math:`| \langle \psi(t) | U_{H}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle |^2`
+# approaches :math:`1` as the states become more similar, thus we choose
+# to minimize the quantity
+# :math:`1 \ - \ | \langle \psi(t) | U_{H}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle |^2`.
+# Since we are interested in calculating this value for many different
+# pieces of quantum data, the final cost function is the **average
+# infidelity** between registers:
+#
+# .. math:: \mathcal{L}(\boldsymbol\mu, \ \Delta) \ = \ 1 \ - \ \frac{1}{N} \displaystyle\sum_{t \ = \ 1}^{N} | \langle \psi(t) | \ U_{H}(\boldsymbol\mu, \ \Delta) \ |\psi_0\rangle |^2,
```
ixfoduap, line 480: Why use `1-fidelity` and not just `-fidelity`? Constant terms are irrelevant in optimization. We don't get to use `infidelity`, but maybe that's a good thing

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+#
+# where we use :math:`N` pieces of quantum data.
+#
+# Before creating the cost function, we must define a few more fixed
```
ixfoduap, line 484: "Must" we do this? I don't see them appearing in the definition of the cost function.

Best to define the cost function and then define the device and qnodes

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+######################################################################
+# The last step is to define and execute the optimizer. We use Adam,
+# with a step-size of :math:`0.3`:
+#
+
+
+iterations = 0
+
+optimizer = qml.AdamOptimizer(stepsize=0.3)
```
ixfoduap, line 541: Have you tried without specifying the stepsize? Why choose this specific value?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+for i in range(0, steps):
+    qgrnn_params = optimizer.step(cost_function, qgrnn_params)
+
+print("Final Parameters: {}".format(qgrnn_params))
+
+
+######################################################################
+# With the learned parameters, we construct a visual representation 
+# of the Hamiltonian to which they correspond:
```
ixfoduap, line 555:
```suggestion
# of the Hamiltonian:
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+#
+
+new_ham_matrix = create_hamiltonian_matrix(
+    qubit_number, nx.complete_graph(qubit_number), [qgrnn_params[0:6], qgrnn_params[6:10]]
+)
+plt.matshow(new_ham_matrix)
+plt.show()
+
+
+######################################################################
```
ixfoduap, line 565: This is nice but I don't find it super helpful

Lucaman99: Haha, totally agreed, the only reason I kept it in is because it is nice to look at 😆

Lucaman99: I'll remove it

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,609 @@
...
+
+    # Creates the interaction component of the Hamiltonian
+    for count, i in enumerate(graph.edges):
+        m = 1
+        for j in range(0, n):
+            if i[0] == j or i[1] == j:
+                m = np.kron(m, qml.PauliZ.matrix)
+            else:
+                m = np.kron(m, np.identity(2))
+        matrix = np.add(matrix, params[0][count] * m)
```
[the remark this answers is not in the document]
Lucaman99: Unfortunately, this doesn't work. Josh commented the same thing on the VQT tutorial 🙂 . The "+" operation concatenates arrays together while np.add does matrix addition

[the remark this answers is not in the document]
Lucaman99: Oh wait, this does work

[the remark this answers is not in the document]
Lucaman99: Oops, my bad, when Josh commented this, I think it was for the case of a list, not a Numpy array

demonstrations/tutorial_qgrnn.py
Lucaman99: Yeah, this was more of a stylistic thing (so I can call it the infidelity cost function), but I see what you're saying

demonstrations/tutorial_qgrnn.py
Lucaman99: Actually, in this instance, it is important that the nodes are the indices of the qubits, with this function, node numbering will start at 0 rather than 4

Lucaman99:
Updated HTML version of the tutorial: https://1077-214003948-gh.circle-artifacts.com/0/_build/html/demos/tutorial_qgrnn.html

Note that the only reason that the number of executions is set to one is to ensure that the website builds in a reasonable amount of time when being reviewed (so please disregard the terrible training results!) 🙂

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,618 @@
...
+######################################################################
+# We conclude this demonstration by verifying that the QGRNN found the
+# target parameters by looking at the
+# exact values of the target and learned parameters:
+
+# Inserts 0s where there is no edge present in target parameters
+
+target_params = matrix_params[0] + matrix_params[1]
+target_params.insert(1, 0)
+target_params.insert(4, 0)
```
josh146, line 599: I'm not 100% sure what is intended/happening here in these lines?

josh146, line 599: I'm not 100% sure what is intended/happening here in these lines?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,618 @@
...
+# Since the QGRNN ansatz is equivalent to the
+# approximate time evolution of some quadratic Hamiltonian, we can use it
+# to learn the dynamics of a quantum system.
+#
+# Continuing with the Ising model example, let's imagine we have some system
+# governed by :math:`\hat{H}_{\text{Ising}}(\boldsymbol\alpha)` for an unknown set of
+# target parameters,
+# :math:`\boldsymbol\alpha` and an unknown interaction graph :math:`G`. Let's also
+# suppose we have access to copies of some
+# low-energy state with respect to the target Hamiltonian, :math:`|\psi_0\rangle`. In addition,
```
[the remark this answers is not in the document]
Lucaman99: No, I meant a state with an energy expectation value that is close to the ground state energy. I'll make the explanation more explicit

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,618 @@
...
+    for i in times_sampled:
+        result = qnode(weight_params, bias_params, time=i)
+        total_cost += -1 * result
+
+    # Prints the value of the cost function
+    if iterations % 5 == 0:
+        print("Fidelity at Step " + str(iterations) + ": " + str((-1 * total_cost / N)._value))
+        print("Parameters at Step " + str(iterations) + ": " + str(params._value))
+        print("---------------------------------------------")
+
```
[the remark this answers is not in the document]
Lucaman99: The only issue with this approach is that, in order to display the value of the cost function, the number of quantum circuit executions would have to double, which would double the runtime (which is already very high). Is there any way that the value of the cost function can be determined from optimizer.step?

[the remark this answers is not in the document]
josh146: Oh, that's a very good point! And something we should maybe add to the optimizers (I don't think there is currently a way of extracting the cost evaluation). Maybe worth adding a comment to https://github.com/XanaduAI/pennylane/issues/683 regarding this.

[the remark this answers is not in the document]
Lucaman99: Sure, that would be great!

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,618 @@
...
+# Before creating the cost function, we must define a few more fixed
+# variables:
+#
+
+N = 15  # The number of pieces of quantum data that are used for each step
+max_time = 0.1  # The maximum value of time that can be used for quantum data
+
+
+######################################################################
+# We then define the infidelity cost function:
```
AroosaIjaz, line 519:
```suggestion
# We then define the fidelity cost function:
```

demonstrations/tutorial_qgrnn.py
AroosaIjaz: I agree with Josh

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,618 @@
...
+#     :align: center
+#
+# In recent years, the concept of a
+# `graph neural network <https://arxiv.org/abs/1812.08434>`__ (GNN) has been
+# receving a lot of attention from the machine learning community.
+# A GNN seeks
+# to learn a representation (a mapping of data into a
+# lower-dimensional vector space) of a given graph with features assigned
+# to nodes and edges. Each of the vectors in the learned
+# representation preserves not only the features, but also the overall
```
AroosaIjaz, line 44:
```suggestion
# representation preserves not only the node features, but also the overall
```

demonstrations.rst
```diff
@@ -203,6 +203,12 @@ algorithms using PennyLane and near-term quantum hardware.
     :description: :doc:`demos/tutorial_quantum_metrology`
     :tags: cirq metrology autograd
 
+.. customgalleryitem::
+    :tooltip: Learning the dynamics of quantum systems on graphs
```
ixfoduap, line 207: I think this sentence used below is clearer: "Using a quantum graph recurrent neural network to learn quantum dynamics"

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+
+"""
+
+######################################################################
+# In this demonstration, we investigate quantum graph
+# recurrent neural networks (QGRNN), which are the quantum analogue of a
+# classical graph recurrent neural network, and a subclass of the more
+# general quantum graph
+# neural network ansatz. Both the QGNN and QGRNN were introduced by
+# `Verdon et al. (2019) <https://arxiv.org/abs/1909.12264>`__.
```
ixfoduap, line 19:
```suggestion
# neural network ansatz. Both the QGNN and QGRNN were introduced in
# `this paper (2019) <https://arxiv.org/abs/1909.12264>`__.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+
+######################################################################
+# The Idea
+# --------
+#
+
+
+######################################################################
+# A graph is defined as a set of *nodes*, along with a set of
```
ixfoduap, line 28:
```suggestion
# A graph is defined as a set of *nodes* along with a set of
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# **edges**, which represent interactions or relationships between nodes.
```
ixfoduap, line 29:
```suggestion
# **edges**, which represent connections between nodes.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+#
+# .. math::
+#
+#     \hat{H}_{\text{Ising}}(\boldsymbol\theta) \ = \ \displaystyle\sum_{(i, j) \in E}
+#     \theta_{ij}^{(1)} Z_{i} Z_{j} \ + \ \displaystyle\sum_{i} \theta_{i}^{(2)} Z_{i} \ + \
+#     \displaystyle\sum_{i} X_{i},
+#
+# where :math:`\boldsymbol\theta \ = \ \{\theta^{(1)}, \ \theta^{(2)}\}`.
+# In this Hamiltonian, the set :math:`E` that determines which pairs of qubits
+# have :math:`ZZ` interactions is exactly the set of edges for some graph. With
```
ixfoduap, line 72:
```suggestion
# have :math:`ZZ` interactions can be represented by the set of edges for some graph. With
```
"is exactly" is too strong

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+
+
+######################################################################
+# This result implies that we can think about *quantum circuits* with
+# graph-theoretic properties. Recall that the time-evolution operator
+# with respect to some Hamiltonian :math:`H` is defined as:
+#
+# .. math:: U \ = \ e^{-it H}.
+#
+# Thus, we have a clean way of taking graph-theoretic Hamiltonians and turning
```
ixfoduap, line 87:
```suggestion
# Thus, we have a clean way of taking quadratic Hamiltonians and turning
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+#
+#     \exp \Big[ -it \Big( \displaystyle\sum_{(i, j) \in E} \theta_{ij}^{(1)} Z_{i} Z_{j} \ + \
+#     \displaystyle\sum_{i} \theta_{i}^{(2)} Z_{i} \ + \ \displaystyle\sum_{i} X_{i} \Big) \Big]
+#     \ \approx \ \displaystyle\prod_{k \ = \ 1}^{t / \Delta} \Bigg[ \displaystyle\prod_{j \ = \
+#     1}^{Q} e^{-i \Delta \hat{H}_{\text{Ising}}^{j}(\boldsymbol\theta)} \Bigg]
+#
+# where :math:`\hat{H}_{\text{Ising}}^{j}(\boldsymbol\theta)` is the :math:`j`-th term of the
+# Ising Hamiltonian and :math:`\Delta` is some small number.
+#
+# The circuit at which we have arrived is a specific instance of the **Quantum Graph
```
ixfoduap, line 111:
```suggestion
# This circuit is a specific instance of the **Quantum Graph
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# Recurrent Neural Network**, which in general is defined as a variational ansatz of
+# the form
+#
+# .. math::
+#
+#     U_{H}(\boldsymbol\mu, \ \boldsymbol\gamma) \ = \ \displaystyle\prod_{i \ = \ 1}^{P} \Bigg[
+#     \displaystyle\prod_{j \ = \ 1}^{Q} e^{-i \gamma_j H^{j}(\boldsymbol\mu)} \Bigg],
+#
+# for some parametrized Hamiltonian, :math:`H(\boldsymbol\mu)`, of
+# quadratic order.
```
ixfoduap, line 121:
```suggestion
# for some parametrized quadratic Hamiltonian, :math:`H(\boldsymbol\mu)`.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# Continuing with the Ising model example, let's imagine we have some system
+# governed by :math:`\hat{H}_{\text{Ising}}(\boldsymbol\alpha)` for an unknown set of
+# target parameters,
+# :math:`\boldsymbol\alpha` and an unknown interaction graph :math:`G`. Let's also
+# suppose we have access to copies of some
+# low-energy state of the target Hamiltonian, :math:`|\psi_0\rangle`. In addition,
+# we have access to a collection of time-evolved states,
+# :math:`\{ |\psi(t_1)\rangle, \ |\psi(t_2)\rangle, \ ..., \ |\psi(t_N)\rangle \}`, defined by:
+#
+# .. math:: |\psi(t_k)\rangle \ = \ e^{-i t_k \hat{H}_{\text{Ising}}(\boldsymbol\alpha)} |\psi_0\rangle
```
ixfoduap, line 142:
```suggestion
# .. math:: |\psi(t_k)\rangle \ = \ e^{-i t_k \hat{H}_{\text{Ising}}(\boldsymbol\alpha)} |\psi_0\rangle.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+#
+# We call the low-energy states and the collection of time-evolved states *quantum data*.
+# From here, we randomly pick a given number of time-evolved states
```
ixfoduap, line 145:
```suggestion
# From here, we randomly pick a number of time-evolved states
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# from our collection. For some state that we choose, which is
+# evolved to some time :math:`t_k`, we compare it
+# to
+#
+# .. math::
+#
+#     U_{\hat{H}_{\text{Ising}}}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle \ \approx \ e^{-i t_k
+#     \hat{H}_{\text{Ising}}(\boldsymbol\mu)} |\psi_0\rangle
```
ixfoduap, line 153:
```suggestion
#     \hat{H}_{\text{Ising}}(\boldsymbol\mu)} |\psi_0\rangle.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+#
+# by feeding one of the copies of :math:`|\psi_0\rangle` into a quantum circuit
```
ixfoduap, line 155:
```suggestion
# This is done by feeding one of the copies of :math:`|\psi_0\rangle` into a quantum circuit
```
Otherwise sentence is too long

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# with the QGRNN ansatz, with some "guessed" set of parameters :math:`\boldsymbol\mu`
+# and a guessed interaction graph, :math:`G'`.
+# We then use a classical optimizer to maximize the average
+# "similarity" between the time-evolved states and the states prepared
+# with the QGRNN.
+#
+# As the QGRNN states becomes more "similar" to
```
ixfoduap, line 162:
```suggestion
# As the QGRNN states becomes more similar to
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+
+######################################################################
+# Learning an Ising Model with the QGRNN
+# ---------------------------------------
+#
+
+
+######################################################################
+# We now attempt to use the QGRNN to learn the parameters corresponding
+# to an arbitrary transverse field Ising model Hamiltonian.
```
ixfoduap, line 183:
```suggestion
# to an arbitrary transverse-field Ising model Hamiltonian.
```
And should hyphenate transverse-field everywhere else

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# the simulation.
+#
+
+
+qubit_number = 4
+qubits = range(qubit_number)
+
+
+######################################################################
+# In order to use the QGRNN, we need access to quantum data. In this
```
ixfoduap, line 216:
```suggestion
# In this
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+#
+#     \hat{H}_{\text{Ising}}(\boldsymbol\theta) \ = \ \displaystyle\sum_{(i, j) \in E}
+#     \theta_{ij}^{(1)} Z_{i} Z_{j} \ + \ \displaystyle\sum_{i} \theta_{i}^{(2)} Z_{i} \ + \
+#     \displaystyle\sum_{i} X_{i},
+#
+# where :math:`E` is the set of edges in the interaction graph, and
+# :math:`X_i` and :math:`Z_i` are the Pauli-X and Pauli-Z on the
+# :math:`i`-th qubit.
+#
+# For this tutorial, we choose the target parameters manually, by sampling from
```
ixfoduap, line 248:
```suggestion
# For this tutorial, we choose the target parameters by sampling from
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+
+reg1 = list(range(qubit_number))  # First qubit register
+reg2 = list(range(qubit_number, 2 * qubit_number))  # Second qubit register
+
+control = 2 * qubit_number  # Index of control qubit
+trotter_step = 0.01  # Trotter step size
+
+# Defines the interaction graph for the new qubit system
+
+new_ising_graph = nx.Graph()
```
ixfoduap, line 455: you should be able to use `nx.complete_graph()`, right?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+new_ising_graph.add_edges_from([(4, 5), (5, 6), (6, 7), (4, 6), (7, 4), (5, 7)])
+
+print(f"Edges: {new_ising_graph.edges}")
+nx.draw(new_ising_graph)
+
+
+######################################################################
+# Part of the idea behind the QGRNN is that
+# we don’t know the interaction graph, and it has to be learned. In this case, the graph
+# is learned **automatically** as the target parameters are optimized. The
```
ixfoduap, line 466:
```suggestion
# is learned *automatically* as the target parameters are optimized. The
```
This suggestion is optional

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# :math:`\boldsymbol\mu` parameters that correspond to edges that don't exist in
```
ixfoduap, line 467: This is probably better said above, as I suggested

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# With this done, we implement the QGRNN circuit for some given time value:
+#
+
+
+def qgrnn(params1, params2, time=None):
+
+    # Prepares the low energy state in the two registers
+    qml.QubitStateVector(np.kron(low_energy_state, low_energy_state), wires=reg1+reg2)
+
+    # Evolves the first qubit register with the time-evolution circuit, to prepare a piece of quantum data
```
ixfoduap, line 479:
```suggestion
    # Evolves the first qubit register with the time-evolution circuit to prepare a piece of quantum data
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+    # Applies the SWAP test between the registers
+    swap_test(control, reg1, reg2)
+
+    # Returns the results of the SWAP test
+    return qml.expval(qml.PauliZ(control))
+
+
+######################################################################
+# We have the full QGRNN circuit, but we still need to define a cost
+# function to minimize. We know that
```
ixfoduap, line 496:
```suggestion
# We have the full QGRNN circuit, but we still need to define a cost.
#We know that
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# :math:`| \langle \psi(t) | U_{H}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle |^2`
+# approaches :math:`1` as the states become more similar and approaches
+# :math:`0` as the states become orthogonal. Thus, we choose
+# to minimize the quantity
+# :math:`-| \langle \psi(t) | U_{H}(\boldsymbol\mu, \ \Delta) |\psi_0\rangle |^2`.
+# Since we are interested in calculating this value for many different
+# pieces of quantum data, the final cost function is the **average
```
ixfoduap, line 503: consider italics instead of bold

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+#     \langle \psi(t_i) | \ U_{H}(\boldsymbol\mu, \ \Delta) \ |\psi_0\rangle |^2,
+#
+# where we use :math:`N` pieces of quantum data.
+#
+# Before creating the cost function, we must define a few more fixed
+# variables:
+#
+
+N = 15  # The number of pieces of quantum data that are used for each step
+max_time = 0.1  # The maximum value of time that can be used for quantum data
```
ixfoduap, line 518: This seems small. Just curious: does training not work or take longer if this is increased?

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+axes[1].matshow(ham_matrix)
+axes[1].set_title("Target Hamiltonian")
+
+fig.tight_layout()
+
+plt.show()
+
+
+######################################################################
+# We conclude this demonstration by verifying that the QGRNN found the
```
ixfoduap, line 604:
```suggestion
# We conclude by verifying that the QGRNN found the
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+# target parameters by looking at the
+# exact values of the target and learned parameters. Recall how the target
```
ixfoduap, line 606: Isn't this verified already above where we compare the Hamiltonians? It seems to me that learning the Hamiltonian is the actual goal, not necessarily the parameters

ixfoduap: So maybe don't talk about verification, but about looking in more detail at what the model was able to learn

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,646 @@
...
+
+target_params = matrix_params[0] + matrix_params[1]
+
+print(f"Target parameters: {target_params}")
+print(f"Learned parameters: {qgrnn_params}")
+print(f"Non-Existing Edge Parameters: {zero_weights}")
+
+
+######################################################################
+# As can be seen here, the weights of edges :math:`(1, 3)` and :math:`(2, 0)`
```
ixfoduap:
```suggestion
# The weights of edges :math:`(1, 3)` and :math:`(2, 0)`
```

demonstrations/tutorial_qgrnn.py
Lucaman99: @ixfoduap Yeah, I thought the same thing, but the algorithm really doesn't converge well if it is higher. I can only assume this is due to the fact that for something like 15 samples, for a larger max_time, the variance of times sampled would be larger, making it difficult to approximate the value of the cost function.

demonstrations/tutorial_qgrnn.py
Lucaman99: Unfortunately no, because the vertices have to be the indices of the qubits, which don't begin at 0

demonstrations/tutorial_qgrnn.py
```diff
@@ -11,12 +11,12 @@
...
 ######################################################################
-# In this demonstration, we investigate quantum graph
+# This demonstration investigates quantum graph
 # recurrent neural networks (QGRNN), which are the quantum analogue of a
 # classical graph recurrent neural network, and a subclass of the more
 # general quantum graph
-# neural network ansatz. Both the QGNN and QGRNN were introduced by
-# `Verdon et al. (2019) <https://arxiv.org/abs/1909.12264>`__.
+# neural network ansatz. Both the QGNN and QGRNN were introduced in
+# `this paper by Verdon et al. (2019) <https://arxiv.org/abs/1909.12264>`__.
```
ixfoduap, line 19: I prefer not to use author names when making references to previous work. It is usually dismissive of co-authors that get sent to the et al. and generally perpetuates a culture of ego that is not particularly healthy.

demonstrations/tutorial_qgrnn.py
```diff
@@ -135,31 +134,31 @@
...
 #
-# by feeding one of the copies of :math:`|\psi_0\rangle` into a quantum circuit
-# with the QGRNN ansatz, with some "guessed" set of parameters :math:`\boldsymbol\mu`
+# This is done by feeding one of the copies of :math:`|\psi_0\rangle` into a quantum circuit
+# with the QGRNN ansatz, with some guessed set of parameters :math:`\boldsymbol\mu`
 # and a guessed interaction graph, :math:`G'`.
 # We then use a classical optimizer to maximize the average
 # "similarity" between the time-evolved states and the states prepared
-# with the QGRNN.
+# with the QGRNN, by updating the guessed parameters.
```
ixfoduap, line 159: I would suggest removing 'by updating the guessed parameters'; this is clear from context

demonstrations/tutorial_qgrnn.py
```diff
@@ -324,40 +322,38 @@ def create_hamiltonian_matrix(n, graph, params):
...
 # Essentially, we choose a
 # VQE ansatz such that the circuit cannot learn the exact ground state,
 # but it can get fairly close. Another way to arrive at the same result is 
 # to perform VQE with a reasonable ansatz, but to terminate the algorithm 
-# before it converges to the ground state. If we did use the exact ground state as
-# :math:`|\psi_0\rangle`, then we would have trivial time-dependence
-# and our data would be useless.
+# before it converges to the ground state. If we used the exact ground state
+# :math:`|\psi_0\rangle`, the time-dependence would be trivial and the
+# data would be useless.
```
ixfoduap, line 333:
```suggestion
# data would not provide enough information about the Hamiltonian parameters.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -428,19 +424,22 @@ def swap_test(control, register1, register2):
...
+# After performing this procedure, the value returned from a measurement of the circuit is
 # :math:`\langle Z \rangle`, with respect to the ancilla qubit that is used to perform
-# the SWAP test. When a SWAP
-# test is performed, the probability of measuring the :math:`|0\rangle` state
-# in the control qubit is related to the fidelity between registers. In addition,
-# the probability of measuring :math:`|0\rangle` is also related to :math:`\langle Z \rangle`.
+# the SWAP test. The probability of measuring the :math:`|0\rangle` state
+# in the control qubit is related to both the fidelity 
+# between registers and :math:`\langle Z \rangle`.
 # From a bit of algebra, we find that :math:`\langle Z \rangle` is equal to the fidelity.
```
ixfoduap: This reads a bit awkwardly. Consider revising

demonstrations/tutorial_qgrnn.py
```diff
@@ -428,19 +424,22 @@ def swap_test(control, register1, register2):
...
 #
 # Before creating the full QGRNN and the cost function, we
 # define a few more fixed values. Among these is a "guessed"
-# interaction graph, for which we will use a complete graph. This choice
+# interaction graph, which we choose to be a complete graph. This choice
```
ixfoduap, line 436:
```suggestion
# interaction graph, which we set to be a complete graph. This choice
```
Maybe there's a better term than 'set', but should avoid repeated use of 'choose'

demonstrations/tutorial_qgrnn.py
```diff
@@ -590,21 +584,24 @@ def cost_function(params):
...
 fig.tight_layout()
 
 plt.show()
 
 
 ######################################################################
-# We conclude this demonstration by verifying that the QGRNN found the
-# target parameters by looking at the
-# exact values of the target and learned parameters. Recall how the target
+# These images look very similar, indicating the QGRNN has done a good job
```
ixfoduap, line 599:
```suggestion
# These images look very similar, indicating that the QGRNN has done a good job 
```

PR #113, opened 2020-08-05
Update qonn.py

co9olguy:
Some final bits of polishing

### DIFF
--- a/demonstrations/qonn.py
+++ b/demonstrations/qonn.py
@@ -8,20 +8,21 @@ Optimizing a quantum optical neural network
     :property="og:description": Optimizing a quantum optical neural network using PennyLane.
     :property="og:image": https://pennylane.ai/qml/_images/qonn_thumbnail.png
 
-This tutorial is based on the paper from `Steinbrecher et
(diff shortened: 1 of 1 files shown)


demonstrations/qonn.py
```diff
@@ -79,12 +80,12 @@
...
-# (or wires) as you want your quantum optical neural network to have. 4
-# modes are used for this demonstration due to the use of a dual-rail encoding. The
-# cutoff dimension should be set to the same value as the number of wires (a
+# We create a Strawberry Fields simulator device with as many quantum modes
+# (or wires) as we want our quantum-optical neural network to have. Four
+# modes are used for this demonstration, due to the use of a dual-rail encoding. The
+# cutoff dimension is set to the same value as the number of wires (a
 # lower cutoff value will cause loss of information, while a higher value
-# will only use unnecessary resources without any improvement).
+# might use unnecessary resources without any improvement).
```
co9olguy, line 88: Since the Kerr gate changes energy, the assumption that photon-number is preserved (and hence the cutoff subspace is never left) is not correct

PR #116, opened 2020-08-16
Pasqal demo

co9olguy:


demonstrations.rst
```diff
@@ -223,9 +223,15 @@ algorithms using PennyLane and near-term quantum hardware.
  
 .. customgalleryitem::
     :tooltip: Compare the parameter-shift rule with backpropagation
-    :figure: demonstrations/tutorial_backprop_thumbnail.png
+    :figure: demonstrations/backprop/tutorial_backprop_thumbnail.png
```
co9olguy, line 226: Noticed that that tutorial was the only one that didn't have its own folder for images

josh146: I think this was intentional. I noticed that the demonstrations were getting crowded, and that it is easy to browse if everything is in the same directory, with images satisfying the naming convention `tutorial_name_image_name.png`. This way images will appear directly beside the tutorial file.

I think I was testing it out to see if it worked in practice!

co9olguy:
@josh146 Now that this functionality has been merged into Cirq, I'd like to push out this demo. Will also require us to release a new version of pennylane-cirq, which I assume is why the build is failing :smile:

co9olguy:
Until that branch is merged, the tutorial will have to be built locall (10mins runtime, sorry :frowning:)

josh146:
@co9olguy I changed the base branch to `dev`: this branch builds against GitHub master of PL, and GitHub master of all the plugins, so should work well.

(Except that the build will likely fail due to the QChem demos and QChem master diverging, I'll get @agran2018 to see if we can prioritize those qchem demo fixes)

demonstrations.rst
```diff
@@ -184,6 +184,7 @@ Demos
 
         </section>
     </div>
+>>>>>>> master
```
josh146, line 187:
```suggestion
```

agran2018:
> (Except that the build will likely fail due to the QChem demos and QChem master diverging, I'll get @agran2018 to see if we can prioritize those qchem demo fixes)

Hi @josh146, all QChem demos are already updated. They are approved PRs that are pending to be merged.

co9olguy:
@josh146 working on merging in that branch to pennylane-cirq now

co9olguy:
@josh146 Everything looks like it's working now, ready for review :+1:

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+    
+    Neutral atoms (green dots) arranged in various configurations. Image originally from [#barredo2017]_.
+    
+The startup `Pasqal <https://pasqal.io/>`_ is one of the companies working to bring 
+neutral-atom quantum computing devices to the world. To support this new class of devices, 
+Pasqal has contributed some new features to the quantum software library `Cirq <https://cirq.readthedocs.io/en/stable/>`_.
+ 
+In this demo, we will use PennyLane, Cirq, and TensorFlow to show off the unique abilities of 
+neutral atom devices, leveraging them to make a variational quantum circuit which has a
+very unique topology: *the Eiffel tower*.
```
glassnotes, line 34: This is a really nice introduction, the language is clear and the graphic goes a long way towards making the reader excited about what's coming next.

One point I got stuck on was that I didn't immediately make the connection between quantum circuits and 3-dimensional objects. What properties or aspects of the circuit are translated into the points? Is it that the gates are the laser pulses that move the atoms around?

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+import matplotlib.pyplot as plt
+
+##############################################################################
+# Building the Eiffel tower
+# -------------------------
+#
+# Our first step will be to load and visualize the data for the Eiffel tower 
+# configuration, which was generously provided by the team at Pasqal.
+
+coords = np.loadtxt("pasqal/Eiffel_tower_data.dat")
```
glassnotes, line 54: Is this file available somewhere for the reader to download?

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+fig = plt.figure()
+ax = fig.add_subplot(111, projection='3d')
+ax.view_init(40, 15)
+fig.subplots_adjust(left=0, right=1, bottom=0, top=1)
+ax.set_xlim(-20, 20)
+ax.set_ylim(-20, 20)
+ax.set_zlim(-40, 10)
+plt.axis('off')
+ax.scatter(xs, ys, zs, c='g',alpha=0.3)
+plt.show();
```
glassnotes, line 68: This would be a cool plot to make interactive in the future. We could give the reader the ability to rotate the tower around, zoom, etc.

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+ax = fig.add_subplot(111, projection='3d')
+ax.view_init(40, 15)
+fig.subplots_adjust(left=0, right=1, bottom=0, top=1)
+ax.set_xlim(-20, 20)
+ax.set_ylim(-20, 20)
+ax.set_zlim(-40, 10)
+plt.axis('off')
+ax.scatter(xs, ys, zs, c='g', alpha=0.3)
+
+base_mask = [3, 7, 11, 15]
```
glassnotes, line 88: What is the base mask? These points aren't used in the plot below.

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+# Converting to Cirq qubits
+# -------------------------
+#
+# Our next step will be to convert these datapoints into objects that 
+# Cirq understands as qubits. For neutral-atom devices in Cirq, we can use the
+# ``ThreeDQubit`` class, which carries information about the three-dimensional
+# arrangement of the qubits.
+#
+# Now, neutral atom devices come with some physical restrictions. 
+# Specifically, qubits in a particular configuration can't be arbitrarily
```
glassnotes, line 109: The phrasing in this sentence is a bit awkward. Maybe, "qubits in a particular configuration can't be forced to interact with each other arbitrarily"?

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+#
+# Now, neutral atom devices come with some physical restrictions. 
+# Specifically, qubits in a particular configuration can't be arbitrarily  
+# interacted with one another. Instead, there is the notion of a 
+# *control radius;* any atoms which are within the system's control radius
+# can interact with one another. Qubits separated by a distance larger than
+# the control radius cannot interact.
+# 
+# In order to allow our Eiffel tower qubits to interact with 
+# one another more easily, we will artificially scale some dimensions
```
glassnotes, line 116: Could be helpful to have a side-by-side plot here of what things look like before and after scaling. 

Another idea is to plot the control radii of a few qubits before and after to show why it's necessary.

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+dev = qml.device("cirq.pasqal", control_radius=control_radius, 
+                 qubits=qubits, wires=num_wires)
+
+##############################################################################
+# Creating a quantum circuit
+# --------------------------
+# 
+# We will now make a variational circuit out of the Eiffel tower configuration
+# from above. Each of the 9 qubits we are using can be thought of
+# as a single wire in a quantum circuit. We will interact these qubits through
```
glassnotes, line 143:
```suggestion
# as a single wire in a quantum circuit. We will cause these qubits to interact  by applying

```

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+# a sequence of two-qubit gates. Specifically, the circuit consists of several
+# stages:
+#
+# i. Input classical data is converted into quantum information at the
+#    first vertical level of qubits, by using single-qubit bit flips
```
glassnotes, line 148:
```suggestion
#    first (lowest) vertical level of qubits, by using single-qubit bit flips
```

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+    ax.plot(xs=[third_x[corner], third_x[(corner + 1) % 4]],
+            ys=[third_y[corner], third_y[(corner + 1) % 4]],
+            zs=[third_z[corner], third_z[(corner + 1) % 4]],
+            c='grey', linestyle='--');
+
+plt.show();
+
+##############################################################################
+# In this figure, the red dots represent the specific qubits we will use in
+# our circuit (the green dots are not used in thie demo).
```
glassnotes, line 222:
```suggestion
# our circuit (the green dots are not used in this demo).

```

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+# Creating a quantum circuit
+# --------------------------
+# 
+# We will now make a variational circuit out of the Eiffel tower configuration
+# from above. Each of the 9 qubits we are using can be thought of
+# as a single wire in a quantum circuit. We will interact these qubits through
+# a sequence of two-qubit gates. Specifically, the circuit consists of several
+# stages:
+#
+# i. Input classical data is converted into quantum information at the
```
glassnotes, line 147: What does the data correspond to?

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+    for idx, wire in enumerate(range(4, 8)):
+        controlled_rotation(weights[idx], wires=[wire, peak_qubit])
+        
+    return qml.expval(qml.PauliZ(wires=peak_qubit))
+    
+    
+##############################################################################
+# Training the circuit
+# --------------------
+#
```
glassnotes, line 265: At this point I got lost about the relationship between the tower configuration and the variational circuit. I think in the last sentence in the introduction, before building the tower, it would be beneficial to  explain that we will not only make a variational circuit, but we'll also use it as a classifier to solve a simple problem. Then in this section, refer back to that initial goal (/learning outcome) before training the circuit.

glassnotes: (This also relates to my above question about 'what is the classical data'.)

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+print("Final cost value: {}".format(cost()))
+
+##############################################################################
+# Success! The circuit has learned to transfer the state of the first qubit
+# to the state of the last qubit, while ignoring the state of all other input
+# qubits.
+#
+# The programmable three-dimensional configurations of neutral atom quantum
+# computers provide a special tool that is hard to replicate in other
+# platforms. What possibilities could this open up for quantum computing,
```
glassnotes, line 305: I like the idea of closing with a question. I would provide just a bit more scaffolding to help the reader in formulating an answer. Highlight concrete points to think about like "how do you think the *shape* of the qubits affects the outcome", or  "what are the limitations of superconducting/ion trap platforms that make neutral atoms more attractive for these applications" (connectivity, etc.).

glassnotes (commented):
This is really neat! I had just a few questions, and some suggestions to improve the logical flow (and ideas for graphics, but these are not necessary to address!).

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+Let's get to it!
+"""
+
+# First we load some necessary libraries and functions
+from itertools import combinations
+import pennylane as qml
+import numpy as np
+import tensorflow as tf
+from mpl_toolkits.mplot3d import Axes3D
+import matplotlib.pyplot as plt
```
josh146, line 45: I have to admit, I'm not so much a fan of this structure in demos. While it's good Python practice to include imports at the top of the file, I find it more pedagogically useful to include some of the more specific imports later on, when that functionality is needed and can be better understood from context. E.g., `combinations`, `plt`, `Axes3D`.

Otherwise, I find it a bit overwhelming to parse (you go straight from a great introduction to a block of code with little explanation).

demonstrations/tutorial_pasqal.py
josh146: Would be good to add a link directly in the ReST block above!

demonstrations/tutorial_pasqal.py
josh146: We could actually do that now @co9olguy if you wanted to use plotly?

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+    qml.RY(-phi, wires=wires[1])
+    qml.CNOT(wires=wires)
+
+@qml.qnode(dev, interface="tf")
+def circuit(weights, data):
+    
+    # Input classical data loaded into qubits at second level
+    for idx in range(4):
+        if data[idx]:
+            qml.PauliX(wires=idx)
```
josh146, line 249: Since data is not yet defined, it's a bit difficult to tell what is happening here. I am assuming `data` is a flat list of integers, and the if statement takes advantage of the fact that `bool(1) is True`?

I would make this explicit, and add a comment.

josh146: Also, is there a reason for not using [`qml.templates.BasisEmbedding`](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.embeddings.BasisEmbedding.html)? That would make it easier to follow

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+# Training the circuit
+# --------------------
+# 
+# In order to train our circuit, we will need a cost function to analyze. For
+# the purposes of this simple demo, we will consider a very simple classifier:
+# if the first input qubit is in state :math:`\vert 0 \rangle`, the model
+# should make the prediction "0", and if that qubit is in state
+# :math:`\vert 1 \rangle`,
+# the model should predict "1" (independent of what the other qubit 
+# states are. In other words, the idealized trained model should learn an
```
josh146, line 272: Suggest making this a bullet point list for readability:
```rest
* if the first input qubit is in state :math:`\vert 0 \rangle`, the model
  should make the prediction "0", and

* if that qubit is in state :math:`\vert 1 \rangle`,
  the model should predict "1" (independent of what the other qubit states are.
```

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,318 @@
...
+    label = data[0]
+    output = (-circuit(weights, data) + 1) / 2
+    return tf.abs(output - label) ** 2
+
+opt = tf.keras.optimizers.Adam(learning_rate=0.1)
+
+for step in range(100):
+    opt.minimize(cost, weights)
+    if step % 5 == 0:
+        print("Step {}: cost={}".format(step, cost()))
```
josh146, line 294: I'm not used to seeing a keras-style optimization! No gradient tape!

Although I do find it surprising initially that the weights are updated 'in place'.

demos_research.rst
```diff
@@ -93,6 +105,12 @@ IBM, Google, Rigetti, and many more—or on real hardware devices.
     :figure: demonstrations/expressivity_fourier_series/expressivity_thumbnail.png
     :description: :doc:`demos/tutorial_expressivity_fourier_series`
     :tags: autograd
+    
+.. customgalleryitem::
+    :tooltip: Making a quantum machine learning model using neutral atoms
+    :figure: demonstrations/pasqal/pasqal_thumbnail.png
```
josh146, line 111: the thumbnail looks great

josh146 (commented):
image

😆 I'm still happy with 16 minutes though.

demonstrations/tutorial_pasqal.py
co9olguy: luckily the two that you mentioned are no longer needed :laughing:

co9olguy: I tend to agree with you about imports. I've moved everything closer to where it's needed :+1:

demonstrations/tutorial_pasqal.py
co9olguy: these plots took a long time to get right, not only with the datapoints, but also the viewing angles. I'd much prefer to leave this idea in the backlog for now :laughing:

demonstrations/tutorial_pasqal.py
co9olguy: Thanks for spotting. It's needed for a future plot, so I've added a comment for clarity

demonstrations/tutorial_pasqal.py
co9olguy: This is actually a bit finicky. There are some physical constraints on the control radius that are not intuitive (e.g., something like "the max control radius must be no greater than 3x the smallest distance between qubits"), and which you don't really find out about until you do something wrong. Rather that going on a digression, I'd just leave this without further elaboration

demonstrations/tutorial_pasqal.py
co9olguy: I try to avoid the templates :laughing: 

I've added a comment much earlier explaining the form of the expected classical data.

demonstrations/tutorial_pasqal.py
co9olguy: Thanks, made some modifications both above and here to make the narrative more clear.

demonstrations/tutorial_pasqal.py
co9olguy: Thanks, just made a link in the rst to its location on github :+1:

co9olguy:
Thanks @glassnotes, @josh146, I've incorporated your feedback and the tutorial is better for it.

Looking to publish and advertise on Tuesday this week.

josh146 (approved):

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,344 @@
...
+# Success! The circuit has learned to transfer the state of the first qubit
+# to the state of the last qubit, while ignoring the state of all other input
+# qubits.
+#
+# The programmable three-dimensional configurations of neutral-atom quantum
+# computers provide a special tool that is hard to replicate in other
+# platforms. Could the physical
+# arrangement of qubits, in particular the third dimension, be leveraged to
+# make quantum algorithms more sparse or efficient? Could neutral-atom
+# systems—with their unique programmability over the geometry—allow us to
```
glassnotes, line 328:
```suggestion
# systems—with their unique programmability of the geometry—allow us to

```

demonstrations/tutorial_pasqal.py
```diff
@@ -0,0 +1,344 @@
...
+import tensorflow as tf
+np.random.seed(143)
+init_weights = np.pi * np.random.rand(4)
+
+weights = tf.Variable(init_weights, dtype=tf.float64)
+
+data = np.random.randint(0, 2, size=4)
+
+def cost():
+    data = np.random.randint(0, 2, size=4)
```
glassnotes, line 304: Conceptual question - I've never seen random sampling used in a cost function like this. Isn't it possible that if the sampling is biased in some way, the model won't learn properly (like, maybe some subset of values gets sampled more than others, so it learns those really well, but fails with the rest)? Or are there some theoretical results that guarantee it can still learn even in the presence of randomness like this?

glassnotes (approved):
@co9olguy awesome, looks like it's good to go!

PR #119, opened 2020-08-18
Update tutorial_quantum_chemistry.py

antalszava:
The method has been renamed: https://pennylane.readthedocs.io/en/stable/code/api/pennylane_qchem.qchem.convert_observable.html
https://github.com/PennyLaneAI/pennylane/blob/master/qchem/CHANGELOG.md#improvements-2


antalszava:
It was more of a manual find (related to a user question). Hmm, that could perhaps involve some sort of parsing of sphinx directives (e.g. `:func:`) and checking if their names appear in the codebase as a definition? :thinking: Not sure if there would be a more sophisticated way of doing it :smile:

PR #126, opened 2020-08-27
Quantum embeddings and metric learning pre-processing

antalszava:
**Summary:**
Adds the pre-processing script used in the Quantum embeddings and metric learning tutorial that was used to obtain the `X_antbees.txt` file. This process is the same as the one used in the Quantum transfer learning tutorial and described in the corresponding paper.


josh146 (approved):
💪

mariaschuld (commented):
Thanks so much Antal, looks good!

PR #142, opened 2020-09-30
Make VQC tutorial plot colouring consistent 

glassnotes:
**Title:** Make VQC tutorial plot colouring consistent 

**Summary:** In the Iris data example, the first 3 plots of the data colour class 1 red and class -1 blue, whereas the contour plot at the end does the opposite. I modified the first plots to match the contour plot for consistency.

**Relevant references:** None

**Possible Drawbacks:** None


josh146 (approved):
good catch!

PR #145, opened 2020-10-13
Fix stochastic parameter shift tutorial

thisac:
Fixes the failing stochastic parameter shift tutorial by extracting the variable from a wrapping np.ndarray.

Only changes `theta1.val` to `theta1.item().val` in the `Generator` function.

### DIFF
--- a/demonstrations/tutorial_stochastic_parameter_shift.py
+++ b/demonstrations/tutorial_stochastic_parameter_shift.py
@@ -7,52 +7,52 @@ The Stochastic Parameter-Shift Rule
     :property="og:image": https://pennylane.ai/qml/_images/stochastic_parameter_shift_thumbnail.png
 
 We demonstrate how the stochastic parameter-shift rule, discovered by Banchi and Crooks [#banchi2020]_,
-can be used to differentiate arbitrary qubit gates, generalizing the original 
-:doc:`parameter-shift rule </glossary/parameter_shift>`, which applies only for gates of a particular 
+can be used to differentiate arbitrary qubit gates, generalizing the original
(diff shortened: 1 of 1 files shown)


antalszava (approved):
Looks good to me! :100:

demonstrations/tutorial_stochastic_parameter_shift.py
```diff
@@ -274,22 +274,22 @@ def param_shift(theta):
 Z = np.array([[1, 0], [0, -1]])
 
 def Generator(theta1, theta2, theta3):
-    # the inputs will show up as Pennylane Variable objects;
-    # we have to extract their numerical values
-    G = theta1.val * np.kron(X, I) - \
+    # the inputs will show up as Pennylane Variable objects, with theta1 being
+    # wrapped in a np.ndarray; we have to extract their numerical values
+    G = theta1.item().val * np.kron(X, I) - \
```
co9olguy, line 279: This was initially just a hack because the needed objects were PL variables. With the tape refactor, the variable class is gone (at least when running a tape computation). @josh146 Is it already possible to just treat these are numpy arrays, or do we still have to wait until tape mode is the primary interface?

thisac: Good point @co9olguy. Much nicer to be able to skip the `.val`s and `.item()`s (just checked and this demo works well with the new quantum tape).

josh146: Yep, with `qml.enable_tape`, the `.val` and `.items()` should be able to be removed 🙂

PR #147, opened 2020-10-14
Adds related demos in sidebar 

glassnotes:
**Title:** Adds related demos in sidebar

**Summary:** This PR adds a new component in the sidebar called "Related demos" so that we can link related tutorials together.  

**Relevant references:** None

**Possible Drawbacks:** None

### DIFF
--- a/demonstrations/pytorch_noise.py
+++ b/demonstrations/pytorch_noise.py
@@ -9,6 +9,10 @@ PyTorch and noisy devices
         by using it to optimize a noisy quantum hardware device.
     :property="og:image": https://pennylane.ai/qml/_images/bloch.gif
 
+.. related::
+
+   tutorial_noisy_circuit_optimization Optimizing noisy circuits with Cirq
+
 Let's revisit the original :ref:`qubit rotation <qubit_rotation>` tutorial, but instead of
 using the default NumPy/autograd QNode interface, we'll use the :doc:`introduction/interfaces/torch`.
 We'll also replace the ``default.qubit`` device with a noisy ``forest.qvm`` device, to
--- a/demonstrations/tutorial_backprop.py
+++ b/demonstrations/tutorial_backprop.py
@@ -9,7 +9,6 @@ Quantum gradients with backpropagation
 
 .. related::
 
-   parameter_shift Parameter-shift rules
    tutorial_quantum_natural_gradient Quantum natural gradient
 
 In PennyLane, any quantum device, whether a hardware device or a simulator, can be
--- a/demonstrations/tutorial_backprop.py
+++ b/demonstrations/tutorial_backprop.py
@@ -7,6 +7,11 @@ Quantum gradients with backpropagation
 
     :property="og:image": https://pennylane.ai/qml/_images/sphx_glr_tutorial_backprop_002.png
 
+.. related::
+
+   parameter_shift Parameter-shift rules
+   tutorial_quantum_natural_gradient Quantum natural gradient
+
 In PennyLane, any quantum device, whether a hardware device or a simulator, can be
 trained using the :doc:`parameter-shift rule </glossary/parameter_shift>` to compute quantum
 gradients. Indeed, the parameter-shift rule is ideally suited to hardware devices, as it does
--- a/demonstrations/tutorial_data_reuploading_classifier.py
+++ b/demonstrations/tutorial_data_reuploading_classifier.py
@@ -9,6 +9,11 @@ Data-reuploading classifier
    :property="og:description": Implement a single-qubit universal quantum classifier using PennyLane.
    :property="og:image": https://pennylane.ai/qml/_images/universal_dnn1.png
 
+.. related::
+
+   tutorial_variational_classifier Variational quantum classifier
+   tutorial_multiclass_classification Multiclass margin classifier
+
 A single-qubit quantum circuit which can implement arbitrary unitary
 operations can be used as a universal classifier much like a single
 hidden-layered Neural Network. As surprising as it sounds,
--- a/demonstrations/tutorial_data_reuploading_classifier.py
+++ b/demonstrations/tutorial_data_reuploading_classifier.py
@@ -13,6 +13,7 @@ Data-reuploading classifier
 
    tutorial_variational_classifier Variational quantum classifier
    tutorial_multiclass_classification Multiclass margin classifier
+   tutorial_expressivity_fourier_series Quantum models as Fourier series
 
 A single-qubit quantum circuit which can implement arbitrary unitary
 operations can be used as a universal classifier much like a single
--- a/demonstrations/tutorial_data_reuploading_classifier.py
+++ b/demonstrations/tutorial_data_reuploading_classifier.py
@@ -1,8 +1,8 @@
 r"""
 .. _data_reuploading_classifier:
 
-Data-reuploading classifer
-==========================
+Data-reuploading classifier
+===========================
 *Author: Shahnawaz Ahmed (shahnawaz.ahmed95@gmail.com)*
 
 .. meta::
--- a/demonstrations/tutorial_doubly_stochastic.py
+++ b/demonstrations/tutorial_doubly_stochastic.py
@@ -7,6 +7,11 @@ Doubly stochastic gradient descent
         strategy with doubly stochastic gradient descent.
     :property="og:image": https://pennylane.ai/qml/_images/single_shot.png
 
+.. related::
+
+   tutorial_backprop Quantum gradients with backprop
+   tutorial_quantum_natural_gradient Quantum natural gradient
+
 In this tutorial we investigate and implement the doubly stochastic gradient descent
 paper from `Ryan Sweke et al. (2019) <https://arxiv.org/abs/1910.01155>`__. In this paper,
 it is shown that quantum gradient descent, where a finite number of measurement samples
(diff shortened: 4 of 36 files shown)


glassnotes:
The demos have been updated with links to each other - below is a list of less well-connected ones. Let me know here if there are any suggestions or relationships I've missed.

Don't link to anything:
 - Basic tutorial: qubit rotation
 - Gaussian transformation
 - Optimizing noisy circuits with Cirq
 - Ensemble classification with Forest and Qiskit

Nothing links to them:
 - 3-qubit Ising model with PyTorch
 - Plugins and hybrid computation
 - QAOA maxcut (will soon be linked with Intro QAOA tutorial)
 - Quantum chemistry with PennyLane
 - Variationally optimizing measurement protocols
 - Frugal shot optimization with Rosalin
 - Quantum circuit structure learning
   
Orphans:
 - QGRNN
 - QGANs
 - Quantum models as Fourier series
 - Quanvolutional NNs
 - Quantum transfer learning
 - Quantum computation with neutral atoms

demonstrations/pytorch_noise.py
```diff
@@ -9,6 +9,10 @@
         by using it to optimize a noisy quantum hardware device.
     :property="og:image": https://pennylane.ai/qml/_images/bloch.gif
 
+.. related::
+
+   tutorial_noisy_circuit_optimization Optimizing noisy circuits with Cirq
```
josh146, line 14: Sidenote: it's a shame the title can't be automatically extracted from the filename. I think I spent a while trying to do this, and couldn't get it to work.

josh146: I think it's possible, just requires some advanced knowledge of how `docutils` works, in order to extract the titles from the parser inside the `RelatedDirective` class :)

josh146, line 14: How come the related links aren't bi-directional? For example, I would expect "Optimizing noisy circuits with Cirq" to link back to here?

demonstrations/tutorial_backprop.py
```diff
@@ -7,6 +7,11 @@
 
     :property="og:image": https://pennylane.ai/qml/_images/sphx_glr_tutorial_backprop_002.png
 
+.. related::
+
+   parameter_shift Parameter-shift rules
+   tutorial_quantum_natural_gradient Quantum natural gradient
```
josh146, line 13: This is great!

demonstrations/tutorial_data_reuploading_classifier.py
```diff
@@ -1,14 +1,19 @@
 r"""
 .. _data_reuploading_classifier:
 
-Data-reuploading classifer
-==========================
+Data-reuploading classifier
```
josh146, line 4: Oops. Good catch. This is one of our most popular pages as well

demonstrations/tutorial_plugins_hybrid.py
```diff
@@ -10,6 +10,11 @@
         containing both photonic and qubit devices.
     :property="og:image": https://pennylane.ai/qml/_images/photon_redirection.png
 
+.. related::
+
+   tutorial_qubit_rotation Basic tutorial: qubit rotation
```
josh146, line 15: Does it make sense to have Qubit Rotation also link to these tutorials? In this case, the 'Related tutorials' section acts as a sort of 'Where to go next' cue for the user.

I've had feedback before from users that sometimes they end up on qubit rotation (since we link to it in the 'Getting Started' section), and don't really know where to go next.

josh146 (approved):
thanks for doing this, it looks great @glassnotes 🌮 

This is an example of a relatively minor addition which I think leads to a significant navigation improvement for users!

josh146:
That was meant to be a 🎉 emoji, but I'm leaving in the accidental 🌮

mariaschuld:
Great, just two thoughts:

Frugal shot optimization with Rosalin
Quantum circuit structure learning
-> These could link to and get links from other "optimiser" tutorials like quantum natural gradient (as well as each other)?

Quantum transfer learning
-> When the embeddings tutorial is back live (currently under repair) we could link from it

demonstrations/tutorial_backprop.py
```diff
@@ -7,6 +7,11 @@
 
     :property="og:image": https://pennylane.ai/qml/_images/sphx_glr_tutorial_backprop_002.png
 
+.. related::
+
+   parameter_shift Parameter-shift rules
```
mariaschuld, line 12: I think this link is not working? At least one of the parameter-shift links gave me an "unknown artefact" error...maybe double check if you can reproduce.

mariaschuld (approved):
Very cool feature indeed...it's so much fun to click from one to the other!

demonstrations/tutorial_data_reuploading_classifier.py
```diff
@@ -1,14 +1,19 @@
...
 *Author: Shahnawaz Ahmed (shahnawaz.ahmed95@gmail.com)*
 
 .. meta::
    :property="og:description": Implement a single-qubit universal quantum classifier using PennyLane.
    :property="og:image": https://pennylane.ai/qml/_images/universal_dnn1.png
 
+.. related::
+
+   tutorial_variational_classifier Variational quantum classifier
+   tutorial_multiclass_classification Multiclass margin classifier
```
mariaschuld, line 15: It's not obvious, but the Fourier series tutorial can be linked here (it explains some of the magic behind data re-uploading).

demonstrations/tutorial_doubly_stochastic.py
```diff
@@ -7,6 +7,11 @@
         strategy with doubly stochastic gradient descent.
     :property="og:image": https://pennylane.ai/qml/_images/single_shot.png
 
+.. related::
+
+   tutorial_backprop Quantum gradients with backprop
+   tutorial_quantum_natural_gradient Quantum natural gradient
```
mariaschuld, line 13: I'd add the frugal shot optimiser tutorial (there is one, right?). The papers overlap quite a lot.

glassnotes:
[ch1048]

demonstrations/tutorial_backprop.py
glassnotes: Yup, I thought it was a tutorial being referenced lower down, but it's actually just the glossary page so I'm removing this link.

demonstrations/tutorial_plugins_hybrid.py
glassnotes: I've linked it to a few other basic tutorials now. But that brings up an interesting point, maybe we should try and design some tutorials that are "in between" the very basics, and applications/research-level stuff.

demonstrations/pytorch_noise.py
glassnotes: I was trying to think from the standpoint of someone working sequentially through the tutorials, so a number of the links are one-directional. That said, there are probably multiple paths so maybe we should aim for bi-directionality.

co9olguy: I was thinking to have the bi-directional links to related tutorials, then somehow have a separate page/mechanism which curate a specific recommended curriculum & order?

demonstrations/tutorial_plugins_hybrid.py
co9olguy: Absolutely!

co9olguy:
This is awesome to have, thanks @glassnotes!

I wonder if the sidebar is a bit hidden? e.g., maybe in the future we could think about having 2-3 related tutorials at the top or bottom in more noticeable locations?

demonstrations/tutorial_isingmodel_PyTorch.py
```diff
@@ -9,6 +9,11 @@
         PennyLane to optimize a 3-qubit Ising model.
     :property="og:image": https://pennylane.ai/qml/_images/isingspins.png
 
+.. related::
+
+   tutorial_state_preparation Training a quantum circuit with PyTorch
+   pytorch_noise PyTorch noise
```
co9olguy, line 15: Note that the "caption" for this tutorial is different in different places. Is there a way to automatically ensure that the title is the same no matter which tutorial it appears in?

co9olguy: e.g., elsewhere it's called "Pytorch and noisy devices"

co9olguy:
Just did a fly-by, won't review in detail, but this is a great addition

demonstrations/tutorial_isingmodel_PyTorch.py
glassnotes: Oh good catch, that one is just my mistake :)

glassnotes: (@josh146 thinks automating this would be possible, but challenging.)

glassnotes:
> This is awesome to have, thanks @glassnotes!
> 
> I wonder if the sidebar is a bit hidden? e.g., maybe in the future we could think about having 2-3 related tutorials at the top or bottom in more noticeable locations?

I do like the idea of having them at the bottom because it shows you where to go next once you're done reading. 

On a related note, the bug with the footer eating the sidebar stuff sometimes hides the whole section for tutorials where the table of contents is very long. In principle the sidebar should be scrollable, I've noticed this doesn't become possible until you scroll all the way to the bottom of the page. So once that is fixed, they should be more visible.

demonstrations/pytorch_noise.py
glassnotes: Yeah that would be awesome! I think I am going to save that for a future PR; my next planned work on the demos is focused more on the content, so that will be a better time to decide on an order (and figure out how to fill in the gaps).

PR #148, opened 2020-10-16
[WIP] Adds interactivity to demos

glassnotes:
**Title:** Adds interactivity to demos.

**Summary:** If all goes as planned, this PR will enable us to make fully interactive demos an option for users.

**Relevant references:** None

**Possible Drawbacks:** Currently, a major hurdle is getting the binder kernels set up properly, and launching quickly enough. Interactivity won't be enjoyable if it takes too long for these to load.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -189,8 +189,8 @@ htmlhelp_basename = "QMLdoc"
 
 # Options for thebe interactivity
 thebe_config = {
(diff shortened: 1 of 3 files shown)


glassnotes:
This was good for proof-of-concept but I think it is not the direction we want to take with the demos, so I'm closing this PR.

PR #150, opened 2020-10-21
No horizontal scrollbars for code on demos

glassnotes:
**Title:** No horizontal scrollbars on demos

**Summary:** Reformatted the code in a number of demos so that there are no horizontal scrollbars, except in special cases (links, circuits, etc.) 

**Relevant references:** None

**Possible Drawbacks:** None


co9olguy:
:pray: :pray: :pray:

demonstrations/qgrnn.py
```diff
@@ -1614,8 +1614,10 @@ def cost_function(params):
 
 target_params = matrix_params[0] + matrix_params[1]
 
-print(f"Target parameters: {target_params}")
-print(f"Learned parameters: {qgrnn_params}")
+print("Target parameters: \t Learned parameters")
+for x in range(len(target_params)):
+    print(f"{target_params[x]}\t{qgrnn_params[x]}")
+print()
```
josh146, line 1620: (by the way, any demo that doesn't start with a filename `tutorial_` is not executed by Sphinx! So you'll need to modify the output code block as well to match the change, on line 1629)

glassnotes: Aha! I was wondering what was happening there. Thanks :+1:

demonstrations.rst
```diff
@@ -126,7 +126,7 @@ Demos
                         <div class="card">
                           <img class="card-img-top img-fluid" src="_static/thumbs/pes_h2.png" alt="Card image cap">
                           <div class="card-body">
-                            <h4 class="card-title">Doubly stochastic gradient descent</h4>
+                            <h4 class="card-title">A brief overview of VQE</h4>
```
co9olguy, line 129: :+1:

demonstrations/qgrnn.py
```diff
@@ -1626,8 +1628,16 @@ def cost_function(params):
...
+#      Target parameters  Learned parameters
+#      0.56               0.5958460805811168
+#      1.24               1.33650946569539
+#      1.67               1.7835767870051
+#      -0.79              -0.8428389553658877
+#      -1.44              -1.4068803387384392
+#      -1.43              -1.3495612547209628
+#      1.18               1.03363436004853
+#      -0.93              -1.0618808945949776
+#
```
co9olguy, line 1640: Would be great if the decimal places were the anchor for alignment of the numbers rather than the starting character (which can be a number of a - sign)

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -259,7 +261,9 @@ def training(features, Y):
         Y_train_batch = Y_train[batch_index]
 
         optimizer.zero_grad()
-        curr_cost = multiclass_svm_loss(q_circuits, params, feat_vecs_train_batch, Y_train_batch)
+        curr_cost = multiclass_svm_loss(
+            q_circuits, params, feat_vecs_train_batch, Y_train_batch
+        )
         curr_cost.backward()
         optimizer.step()
```
co9olguy, line 269: The final output cell of this demo is *just* large enough to have a scrollbar. If we could save 1-2 characters, even that one could be fixed. It looks like there is some unecessary white space after "Iter:" that we could remove

demonstrations/tutorial_quantum_chemistry.py
```diff
@@ -250,7 +250,9 @@
 # [#jensenbook]_ describing the fermionic Hamiltonian are retrieved from the
 # previously generated file ``'./pyscf/sto-3g/water.hdf5'``.
 
-qubit_hamiltonian = qchem.decompose(hf_file, mapping="jordan_wigner", core=core, active=active)
+qubit_hamiltonian = qchem.decompose(
+    hf_file, mapping="jordan_wigner", core=core, active=active
+)
 print("Electronic Hamiltonian of the water molecule represented in the Pauli basis")
 print(qubit_hamiltonian)
```
co9olguy, line 258: It's too bad that the line `print(geometry)` relies on the internal data representation, it is a bit too long. What about using "\n".join(m for m in geometry) or something similar?

demonstrations/tutorial_rosalin.py
```diff
@@ -561,7 +561,9 @@ def step(self, params):
     params = opt.step(params)
     cost_rosalin.append(cost_analytic(params))
     shots_rosalin.append(opt.shots_used)
-    print("Step {}: cost = {} shots_used = {}".format(i, cost_rosalin[-1], shots_rosalin[-1]))
+    print(
+        "Step {}: cost = {} shots_used = {}".format(i, cost_rosalin[-1], shots_rosalin[-1])
```
co9olguy, line 565:
```suggestion
        "Step {}: cost = {}, shots_used = {}".format(i, cost_rosalin[-1], shots_rosalin[-1])

```

co9olguy (approved):
Thanks for this @glassnotes, always bothered me :laughing:

demonstrations/qgrnn.py
glassnotes: I did this one as shown because that's what the actual output would look like running the code snippet. But I have no objection to lining things up nicely in the manually-added output. :)

demonstrations/tutorial_multiclass_classification.py
glassnotes: Managed to fix this one :+1:

co9olguy: :100:

demonstrations/tutorial_quantum_chemistry.py
josh146: I don't know, just personal preference here, but I would prefer the clearer `print(geometry)` + a horizontal scroll bar, over a more complicated print statement.

Nice to get rid of horizontal scrollbars, but not if they lead to more complicated code?

demonstrations/tutorial_rosalin.py
```diff
@@ -561,7 +561,10 @@ def step(self, params):
     params = opt.step(params)
     cost_rosalin.append(cost_analytic(params))
     shots_rosalin.append(opt.shots_used)
-    print("Step {}: cost = {} shots_used = {}".format(i, cost_rosalin[-1], shots_rosalin[-1]))
+    print(
+        "Step {}: cost = {}, shots_used = {}".format(i, cost_rosalin[-1], shots_rosalin[-1])
+
+    )
```
josh146, line 567: GitHub added in an extra line!

Although, we could use f-strings to make the line shorter.
```suggestion
    print(f"Step {i}: cost = {cost_rosalin[-1]}, shots_used = {shots_rosalin[-1]}")
```

demonstrations/qgrnn.py
```diff
@@ -1614,8 +1614,10 @@ def cost_function(params):
 
 target_params = matrix_params[0] + matrix_params[1]
 
-print(f"Target parameters: {target_params}")
-print(f"Learned parameters: {qgrnn_params}")
+print("Target parameters \tLearned parameters")
+for x in range(len(target_params)):
+    print(f"{target_params[x]}\t\t\t{qgrnn_params[x]}")
+print()
 print(f"Non-Existing Edge Parameters: {zero_weights}")
```
josh146, line 1621:
```suggestion
print(f"\nNon-Existing Edge Parameters: {zero_weights}")
```

demonstrations/tutorial_backprop.py
```diff
@@ -277,7 +277,9 @@ def circuit(params):
 
 repeat = 3
 number = 10
```
josh146, line 279: Maybe it's worth shortening these variables (`num`, `rep`), rather than splitting the `timeit` lines onto multiple lines? Looks a bit ugly to me :/

josh146 (approved):

demonstrations/tutorial_backprop.py
glassnotes: I will try and see how it looks, but use `reps` instead of `rep` if it still fits :+1:

glassnotes: Unfortunately even with `rep` some of the later ones still do have to be split. During this PR I noticed that the column width is only about 93-94 characters, as opposed to the 100 that is in the instructions for `black`. Is there a way to just slightly increase the width of the white rectangle the tutorial contents live in?

josh146: Oh good idea. How does this look?
image

I did this by decreasing the left and right margin of the code-blocks by 45px:

image

josh146: (This margin change would have to be undone for smaller screen resolutions)

glassnotes: I was thinking like this width, and then increasing the width of the contents correspondingly.

image

josh146: That works too!

co9olguy: Be careful taking @josh146's suggestions about what looks good on a screen, he's famous for looking at everything in super-maximized multi-monitor resolutions :laughing:

glassnotes: I have a very similar set up :sweat_smile:

glassnotes: I slightly increased the width of the content, everything looks okay on both my iPad and desktop. But probably good for someone else to give it a go and make sure nothing else broke unexpectedly.

josh146: @glassnotes this is very pedantic (apologies!) but the reduction in 10px in the margins doesn't look so good to me.

Before:

image

After:

image

Could we instead make the `max-width` of the middle column itself 10px wider, rather than reducing the column margins?

glassnotes: Either is fine with me - my CSS skills are not great, I increased the margin because I couldn't figure out how to increase the column width :disappointed:

josh146: Ah! What you want is
```css
#right-column {
    position: relative;
    padding: 20px 0px 100px 0px;
    margin-left: auto;
    background-color: #fff;
    max-width: 900px;
```
on line 47 of `xanadu.css_t`. Increasing `max-width` to 910px should do it!

glassnotes: 910px wasn't quite enough to get the lines in backprop - neither is 914px, but past 910 things on other pages start to get a bit wonky (see below). I think maybe the margin change might be a better solution in this case; I can try reducing them a bit less than the previous try.
image

glassnotes: Okay, check the version of [fc0fa25](https://github.com/PennyLaneAI/qml/pull/150/commits/fc0fa2543cf5fd503525b5598451375dba5b5e2f). Both width and margins are slightly larger and smaller respectively, but in such a way that objects on other pages are not adversely affected. I don't think it's worth to spend any more time on this now given I'm not a CSS expert.

Issue #83, opened 2020-05-13
Add a banner on the home page, with 'Start here' links


josh146:
With no organization on the 'Demos' page, it is difficult for new users to identify exactly where they should start. There are a couple of potential solutions.

* Add a banner on the home page, with 'Start here' links. Something similar to the following?

  ![image](https://user-images.githubusercontent.com/2959003/81773990-bbc98c80-9528-11ea-926b-41a1289c2e02.png)

* Start organizing the demos?

* Keep the demos as is, but 'Feature' some demos at the top of the page (similar to the home page carousel).

glassnotes:
Is this still something we want? I can add it to the [demos entry page](https://pennylane.ai/qml/demonstrations.html) directly above (or below) the "basic", "research", and "community" cards.

josh146:
Good question. This issue was made before we (re)categorized the demos page into separate 'basics' and 'research' pages.

The example in the top comment is taken from https://strawberryfields.ai/photonics/. Should we do something similar here still?

glassnotes:
I think it would be nice to have somewhere near the top of the main demo page to get peoples attention, but I also wouldn't want it to be too obtrusive since there are is the community banner and featured demo carousel. Maybe it would be good as a dropdown sphinx element?

glassnotes:
We've decided to close this as per discussion in #159  as it is no longer relevant after redesign.

PR #159, opened 2020-11-13
[WIP] Add "start here" banner on main demo page

glassnotes:
**Title:** Add "start here" banner on main demo page

**Summary:** A simple implementation using a dropdown admonition that suggests a starting point for new users. 

**Relevant references:** None

**Possible Drawbacks:** Not as relevant after categorization during the re-design, but still nice to have for people who land on the demo page for the first time. Also we'll need to tweak the CSS if we decide to keep this. 

**Related GitHub Issues:** #83

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -52,6 +52,7 @@ extensions = [
     "sphinx.ext.ifconfig",
     "sphinx_gallery.gen_gallery",
     "sphinx_sitemap",
+    "sphinx_togglebutton"
 ]
 
 
--- a/demonstrations.rst
+++ b/demonstrations.rst
@@ -1,3 +1,4 @@
+
 .. raw:: html
 
     <style>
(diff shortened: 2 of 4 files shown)


josh146:
Thanks @glassnotes! Since this is still marked as a draft, I've left some general high-level comments.

* I really like the contents, especially the link to installation!

* Should we also include a link to the 'quickstart guides'? E.g., "Begin with our PennyLane quickstart guides, then start by training a quantum circuit to rotate a qubit , or tune a beamsplitter!"

* The style and color look a bit out of place to me compared to the rest of the website 🤔 

* I thought the toggling was a good idea when we discussed it previously, but now that I can play around with it in practice, it seems a bit distracting. Since it's only a single sentence, better to keep it visible?

* I'm also wondering if it makes the demo page a bit too busy. What do you think about instead placing it on the index page directly? e.g.,

  image

  With the new demo landing page, the index page is a bit emptier now, and so maybe this is a better place for it

glassnotes:
Oh I like that version much better! It is more likely that this is the first page people see, rather than the demo page, so I think it makes sense to have it up front.

As for the toggle, I found it is not very smooth, the text and links don't appear at the same time and the toggle button is not aligned with the text. I'm sure that can all be fixed through CSS sorcery, but if it goes on the front page like your example it doesn't need to be toggleable.

glassnotes:
So the more I tinker with this, the less I think it's necessary. Someone who goes to the main page will see "what is QML" up front, which naturally leads people to the demos through a link at the bottom. Most of the stuff on "key concepts" also links to a demo. And the demo page itself now clearly highlights the basic demos and says "begin your journey...". @josh146 if you're cool with it maybe we just close down #83 and this PR for now.

josh146:
> @josh146 if you're cool with it maybe we just close down #83 and this PR for now.

Yep convinced :+1:

The original feedback ("it is hard to find where to start") dates from before the (a) demo landing page, and (b) demo categorization.

Perhaps we can even argue that the demo landing page satisfies/fixes #83.

glassnotes:
If it's just "where do I start", then I think the demo landing page takes care of #83. 

However this leads directly to the question  "where do I go next". This is partially addressed by the related demo links in the sidebar, but eventually it would be nice to have some recommended paths laid out. We can make a separate issue for that and think about where to put it - maybe it is a similar light green box with a couple ordered suggestions, sitting on the "basic demos" page.

josh146:
:+1:

Another suggestion is to have navigation hints at the bottom of the demo. This could be as simple as 'Previous/Next demo', or more complicated (Enjoyed this demo? Check out these related demos <thumbnail> <thumbnail> <thumbnail>)

glassnotes:
Yeah for sure! I'm gonna close these two things then, and set up a new issue for the new ideas.

PR #160, opened 2020-11-13
qsim supremacy demo

thisac:
**Title:**
Quantum supremacy using qsim

**Summary:**
In the paper Quantum supremacy using a programmable superconducting processor Google showed that their quantum computer could complete a task that would take a classical computer potentially thousands of years. For their simulation benchmarks they used a simulator called qsim, which we will also use here, provided via our PennyLane-Cirq plugin.

In this demonstarion, we will walk you through how their benchmarks were constructed and run, and provide an example of what their simulations looked like. We will be using PennyLane along with the PennyLane-Cirq plugin and the qsim-device, running via the Cirq backend.

**Relevant references:**
[Quantum supremacy using a programmable superconducting processor](https://www.nature.com/articles/s41586-019-1666-5)

[Supplementary information for “Quantum supremacy using a programmable superconducting processor”](https://arxiv.org/abs/1910.11333)

[Dataset for “Quantum supremacy using a programmable superconducting processor”](https://datadryad.org/stash/dataset/doi:10.5061/dryad.k6t1rj8)

[Supplementary material for “Characterizing Quantum Supremacy in Near-Term Devices”](https://www.nature.com/articles/s41567-018-0124-x#Sec7)


josh146:
Thanks @thisac!

Could you modify `requirements.txt` to include `qsimcirq` when you get the chance? Hopefully it will then build 🤞

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# ``dev.generate_samples`` and similarly, sample random bitstrings from a
+# normal distribution by generating all basis states, along with their
+# corresponding bitstrings, and sample directly from them using NumPy.
+#
+
+seed=np.random.randint(0, 42424242)
+probs = circuit(seed=seed)
+
+# sampling from the circuit's probability distribution should give a fidelity close to 1
+samples = dev.generate_samples()
```
josh146, line 341: I have to admit, I'm not so keen on using a `QubitDevice` method directly in the tutorial 🤔 I know you were looking into this before, but is there no way to use `return qml.sample(..)`? Or do we need to allow `return qml.sample()` to sample from the computational basis in PennyLane?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+print("Normal distribution:", f"{f_normal:.7f}".rjust(15))
+
+##############################################################################
+# .. rst-class:: sphx-glr-script-out
+#
+#  Out:
+#
+#  .. code-block:: none
+#
+#    Circuit's distribution:    1.0016549
```
josh146, line 358: How come the fidelity is above 1 here? I would have expected it to _approach_ 1, but be bound <= 1 🤔

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+print("Theoretical:", f"{theoretical_value:.7f}\n".rjust(24))
+
+##############################################################################
+# .. rst-class:: sphx-glr-script-out
+#
+#  Out:
+#
+#  .. code-block:: none
+#
+#     Theoretical:               0.9995118
```
josh146, line 381: I'm also a bit surprised that the theoretical fidelity is < the sampled fidelity!

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+#
+
+######################################################################
+# Quantum supremacy
+# -----------------
+#
+# Why are we calculating this specific fidelity, and what does it actually
+# mean if we get a cross-entropy benchmarking fidelity close to 1? This is
+# an important question, containing one of the main arguments behind why
+# this experiment is able to show quantum supremacy.
```
josh146, line 434: A nice way to finish off the demo :)

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# readout errors and simulating a noisy circuit, comparing it to the
+# hardware device output, and then extrapolating the time it would take to
+# run a larger circuit, the paper is able to predict the 10,000 years it
+# would take to classically reach the same fidelity for the particular
+# circuit as they do with the Sycamore chip (see Fig. 4 in `[1]
+# <https://www.nature.com/articles/s41586-019-1666-5>`__).
+#
+# For more reading on this, the original Nature paper [1] is highly
+# recommended, along with the suplementary information if you want to dive
+# deeper into the maths and physics of the experiment. The blogpost in [5],
```
josh146, line 463:
```suggestion
# deeper into the maths and physics of the experiment. The blog post in [5],
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# along with the accompanying GitHub repo, also provides a nice introduction
+# to the cross-entropy benchmark fidelity, along with some calculations
+# including the effects of added noise models.
```
josh146, line 466:
```suggestion
# along with the accompanying GitHub repo, also provides a nice introduction
# to the cross-entropy benchmark fidelity, and includes calculations
# highlighting the effects of added noise models.
```
(to avoid the repeat of 'along')

demos_research.rst
```diff
@@ -123,6 +123,12 @@ plateaus in quantum neural networks.
 
 :html:`<div class="gallery-grid row">`
 
+.. customgalleryitem::
+    :tooltip: Construct a supremacy circuit with qsim.
+    :figure: demonstrations/qsim_supremacy/qsim.png
+    :description: :doc:`demos/qsim_supremacy`
+    :tags: cirq qsim
```
josh146, line 130: For some reason, this thumbnail does not render. Perhaps because the thumbnail image has not been committed?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# We also need to define the number of shots per circuit instance to be
+# used. This corresponds to the number of times that the circuit is
+# sampled. This will be used later when calculating the cross-entropy
+# benchmark fidelity. The more shots, the more accurate the results will
+# be. We will use 500.000 shots; the same number of samples that are used
+# in the supremacy paper, but feel free to change this to whichever value
+# you wish (depending on your own hardware restrictions).
+#
+
+shots = 500000
```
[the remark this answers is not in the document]
co9olguy: :100:

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# this demo, is called the linear cross-entropy benchmarking fidelity of
+# the circuit. It's defined as
+#
+# .. math::
+#
+#    F_{XEB} = 2^{n}\left<P(x_i)\right>_i - 1
+#
+# where :math:``n`` is the number of qubits, :math:`P(x_i)` is the
+# probability of bitstring :math:`x_i` computed for the ideal quantum
+# circuit, and the average is over the observed bitstrings.
```
co9olguy, line 280: It read a bit weird to have the index `i` appearing in the RHS in a quantity that does not depend on the index. Are you going for an implicit Einstein summation with the two `i`s used this way? :thinking:

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+#
+#  Out:
+#
+#  .. code-block:: none
+#
+#     Theoretical:               0.9995118
+#
+
+######################################################################
+# The values above might seem a bit arbitrary. To show that the fidelity
```
co9olguy, line 385: Why are they arbitrary? You've just gone to some length to explain and motivate them :thinking:

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+print("Circuit's distribution:", f"{f_circuit:.7f}".rjust(12))
+
+# sampling from a normal distribution should give a fidelity close to 0
+basis_states = dev.generate_basis_states(wires)
+samples = np.array([basis_states[i] for i in np.random.randint(0, len(basis_states), size=shots)])
+f_normal = fidelity_xeb(samples, probs)
+print("Normal distribution:", f"{f_normal:.7f}".rjust(15))
+
+##############################################################################
+# .. rst-class:: sphx-glr-script-out
```
co9olguy, line 352: Maybe I missed something, but why are the outputs hand-coded as mock code cells? Is it because we won't be running this demo in CI?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# cross-entropy benchmark fidelities and then calculate the mean fidelity.
+# This value should get closer to the theoretical value the more
+# evaluations we do.
+#
+# .. note::
+#
+#    The following mean fidelity calculations can be interesting to play
+#    around with. You can change the qubit grid at the top of this demo
+#    using, e.g., 8 or 4 qubits; change the number of shots used; as well
+#    as the number of circuit evaluations below. Running the following code
```
co9olguy, line 397: The second sentence here could use another round of polish, read somewhat awkwardly

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+f_circuit = []
+num_of_evaluations = 100
+for i in range(num_of_evaluations):
+    seed=np.random.randint(0, 42424242)
+
+    probs = circuit(seed=seed)
+    samples = dev.generate_samples()
+
+    f_circuit.append(fidelity_xeb(samples, probs))
+    print(f"\r{i + 1:4d} / {num_of_evaluations:4d}{' ':17}{np.mean(f_circuit):.7f}", end="")
```
co9olguy, line 413: Line length causing horizontal scrollbars

demonstrations/qsim_supremacy.py
co9olguy: I agree, but it might be useful to have some of this up front as well, to make sure readers are interested enough to read through the entire demo

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# Quantum supremacy
+# -----------------
+#
+# Why are we calculating this specific fidelity, and what does it actually
+# mean if we get a cross-entropy benchmarking fidelity close to 1? This is
+# an important question, containing one of the main arguments behind why
+# this experiment is able to show quantum supremacy.
+#
+# The idea behind this fidelity is that it should be impossible to
+# effectively get a value close to 1 by any classical sampling methods.
```
co9olguy, line 437:
```suggestion
# It should be effectively impossible to get a value close to 1 for this
# fidelity by any classical sampling methods.
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# random quantum circuits follow, which are hard to simulate classically.
+# On the other hand, a quantum device, running a circuit as the one
+# constructed above, should be able to sample from such a distribution
+# without much overhead. Thus, by showing that a quantum device can produce
+# a high enough fidelity value for a large enough circuit, quantum
+# supremacy can be claimed. This is exactly what the paper discussed in
+# this demonstration has done.
+#
+# There's still one issue that hasn't been touched on yet, and will be left
+# for a future demonstration: the addition of noise in quantum hardware.
```
co9olguy, line 448: let's not make any promises :stuck_out_tongue_winking_eye:

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# Simply put, this noise will lower the cross-entropy benchmark fidelity
```
co9olguy, line 449:
```suggestion
# Simply put, this noise will lower the cross-entropy benchmark fidelity,
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# getting it closer to 0, as would also be the case for any classically
+# tractable sampling from probability distributions. The larger the
```
co9olguy, line 451: What does it mean "classically tractable sampling" here? I didn't understand :thinking:

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# circuit, the more noise there will also be, and thus the lower the
+# fidelity will be. By calculating the specific single-qubit, two-qubit and
```
co9olguy, line 453:
```suggestion
# fidelity will be. By calculating the specific single-qubit, two-qubit, and
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# readout errors and simulating a noisy circuit, comparing it to the
+# hardware device output, and then extrapolating the time it would take to
+# run a larger circuit, the paper is able to predict the 10,000 years it
+# would take to classically reach the same fidelity for the particular
+# circuit as they do with the Sycamore chip (see Fig. 4 in `[1]
```
co9olguy, line 458: run-on sentence; recommend splitting up or rephrasing

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# <https://www.nature.com/articles/s41586-019-1666-5>`__).
+#
+# For more reading on this, the original Nature paper [1] is highly
```
co9olguy, line 461:
```suggestion
# For more reading on this, the original paper [1] is highly
```

demonstrations/qsim_supremacy.py
co9olguy, line 463:
```suggestion
# recommended (along with the suplementary information if you want to dive
# deeper into the math and physics of the experiment). The blog post in [5],
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+#
+# [3]: `Dataset for "Quantum supremacy using a programmable superconducting
+# processor" <https://datadryad.org/stash/dataset/doi:10.5061/dryad.k6t1rj8>`__
+#
+# [4]: `Supplementary material for "Characterizing Quantum Supremacy in
+# Near-Term Devices" <https://www.nature.com/articles/s41567-018-0124-x#Sec7>`__
+#
+# [5]: `Unpacking the Quantum Supremacy Benchmark with Python
+# <https://medium.com/@sohaib.alam/unpacking-the-quantum-supremacy-benchmark-with-python-67a46709d>`__
+#
```
co9olguy, line 487:
```suggestion
#

```

co9olguy (changes requested):
Thanks for taking a deep dive into this demo @thisac! It's not a straightforward one to do! :tada: 

I agree with Josh that some images would help bring the reader along. My other major recommendation would be to bring some more of the motivations for the quantum supremacy experiment up front before diving deeply into the code implementation. 

As a minor suggestion, I'd recommend trying to be a bit less verbose in the text, just to make things easier on the reader. A good test is to see if you can look at a sentence and remove words from it without any sacrifice to the clarity.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+#       0 & 0 & 1 & 0 \\
+#       0 & 0 & 0 & e^{-i\phi}
+#    \end{bmatrix}
+#
+# These two gates have already been made accesible via the Cirq plugin.
+#
+
+
+######################################################################
+# Here comes one of the tricky parts. The way the paper decides which
```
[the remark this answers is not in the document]
josh146: Makes the paper sound sentient!

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+classical computer potentially thousands of years. For their simulation
+benchmarks they used a simulator called `qsim
+<https://github.com/quantumlib/qsim>`__, which we will also use here,
+provided via our `PennyLane-Cirq plugin
+<https://pennylane-cirq.readthedocs.io/en/latest/>`__.
+
+To construct a task that was able to both be run on the Sycamore chip and
+simulated classically, a pseudo-random circuit was constructed by
+alternating single-qubit and two-qubit gates in a specific, semi-random
+pattern. This is basically a way of building a random unitary that is also
```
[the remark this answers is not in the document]
thisac: Yes, they have a very nice figure, but I was a bit unsure if we can use images directly from the paper (not sure if this is something we've done for other demos). Would you say that this is alright?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+compatible with their hardware. The circuit output could then be sampled a
+number of times, producing a set of bitstrings corresponing to the
+measurements of the circuit. The more qubits there are, and the deeper the
+circuit is, the more difficult it becomes to simulate and sample this
+probability distribution, from which these bitstrings come from,
+classically. By comparing run-times for the classical simulations along
+with the outputs of the physical quantum chip for smaller circuits, and
+then extrapolating classical run-times for a larger circuits, this is
+explained as a suffient measure of quantum supremacy. [1]
```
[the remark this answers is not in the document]
thisac: This is much better. I just added these references early, without thinking twice about how references should be implemented. :smile: Thanks!

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# each qubit following the paper's suplementary dataset [3]. We also create
+# a mapping between the wire number and the Cirq qubit to easier reference
+# specific qubits later. Feel free to play around with different grids and
+# number of qubits. Just keep in mind that the grid needs to stay
+# connected. You could, for example, remove the final row (last four qubits
+# in the list) to simulate an 8 qubit system.
+#
+
+qubits = sorted([
+    cirq.GridQubit(3, 3),
```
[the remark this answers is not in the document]
thisac: This is just taken from their own code (they used these qubits). I don't know exactly why, but I could probably change it to 0 to 3 instead. It makes more sense.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+    cirq.GridQubit(5, 3),
+    cirq.GridQubit(5, 4),
+    cirq.GridQubit(5, 5),
+    cirq.GridQubit(5, 6),
+])
+
+wires = len(qubits)
+
+# create a mapping between wire number and Cirq qubit
+qb2wire = {i: j for i, j in zip(qubits, range(wires))}
```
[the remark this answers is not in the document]
thisac: It doesn't seem to work to pass a `cirq.qubit` as a wire for a gate (should it?), so that is why this `qb2wire` is defined. E.g. when I later apply `ops.ISWAP(wires=(qb_1, qb_2))`, `qb_1` and `qb_2` must be the indices of the the qubits (and not the `cirq.GridQubit`s themselves).

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+######################################################################
+# Now let's create the qsim-device, available via the Cirq plugin, together
+# with the ``wires`` and ``qubits`` keywords that we defined above. qsim is
```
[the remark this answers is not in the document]
thisac: I'm not sure. Both the paper and the GitHub repo uses "qsim" (and not `qsim`). Once in the paper it's referred to as "Qsim" (at the start of a sentence). qsim isn't really used in code at all though, since the module were using is called `qsimcirq`. What do you think is best? I'm happy with whichever. :slightly_smiling_face:

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# a Schrödinger full state-vector simulator that was used for the cross
+# entropy benchmarking in Google's supremacy experiment [1].
+#
+# We also need to define the number of shots per circuit instance to be
+# used. This corresponds to the number of times that the circuit is
+# sampled. This will be used later when calculating the cross-entropy
+# benchmark fidelity. The more shots, the more accurate the results will
+# be. We will use 500.000 shots; the same number of samples that are used
```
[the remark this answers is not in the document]
thisac: Very true!

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# .. math::
+#
+#    \frac{1}{\sqrt{2}}
+#    \begin{bmatrix}
+#       1 & \sqrt{i}  \\
+#       \sqrt{-i} & 1 \\
+#    \end{bmatrix}
+#
+
+sqrtXgate = lambda wires: qml.RX(np.pi / 2, wires=wires)
```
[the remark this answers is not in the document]
thisac: Oh, I completely missed that. Nice!

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# 2).
+#
+# While we can define any patterns we'd like, the two gate sequences below
+# are the ones that are used in the supremacy paper. The shorter one is
+# used for their classically verifiable benchmarking, while the slightly
+# longer sequence, which is much harder to simulate classically, is used
+# for estimating the cross-entropy fidelity in what they call the supremacy
+# regime. We will use the shorter gate sequence for the following
+# demonstration, although feel free to play around with other combinations
+# if you wish.
```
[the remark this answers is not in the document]
thisac: Thanks, I think that's good. Sometimes I have a tendency to over-complicate my sentences.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+
+# gate_sequence = np.resize(["A", "B", "C", "D", "C", "D", "A", "B"], m)
+gate_sequence = np.resize(["A", "B", "C", "D"], m)
+
+
+######################################################################
+# Finally, we can define the circuit itself and create a QNode that we will
+# use for circuit evaluation with the qsim device.
+#
+# Each circuit-loop consists of alternating layers of single-qubit gates
```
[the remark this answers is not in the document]
thisac: Bad wording on my side. It's just each gate sequence step. "A" corresponds to one "circuit-loop" with random single-qubit gates applied on all wires and two-qubit gates on all wires connected by "A". It's the "main" loop in the circuit.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+# are randomly selected and applied to each qubit in the circuit, while the
+# two-qubit gates are applied to the qubits connected by A, B, C or D as
+# defined above. The circuit finally ends with a half-cycle, consisting of
+# only a layer of single-qubit gates. Note that the last half-cycle is only
+# applied once after the gate sequence is completed.
+#
+# We define the circuit, letting it return the state probabilities, and
+# decorate it with the QNode decorator, binding it to the qsim simulator
+# device. Later, we will also extract samples directly from the device
+# without needing to add it as a return statement in the circuit.
```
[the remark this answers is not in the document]
thisac: Yes, I can update this. I just thought it looked fairly nice this way. It also avoids having to update, and explain, the sample values: -1 to 1 and 1 to 0. Would you say it's OK to simply return samples, and then use `dev.probability` to get the probabilities instead (I still need them for the fidelity calculations), or is it nicer if I have an if-statement in the circuit, like this:

```python
if return_probs:
    return qml.probs(wires=range(wires))
else:
    return [qml.sample(qml.PauliZ(w)) for w in range(wires)]
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+#
+
+@qml.qnode(dev)
+def circuit(seed=42):
+    np.random.seed(seed)
+
+    # m full cycle - single-qubit gates & two-qubit gate
+    for gs in gate_sequence:
+        for w in range(wires):  # TODO: avoid same gate twice on a qubit
+            np.random.choice(single_qubit_gates)(wires=w)
```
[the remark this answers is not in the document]
thisac: I almost forgot about this. I should attempt to fix it. The same single-qubit gate should never be applied to the same wire twice in a row (with potential two-qubit gates in-between). That's possible as it's currently implemented.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+
+        for qb_1, qb_2 in d[gs]:
+            ops.ISWAP(wires=(qb_1, qb_2))
+            ops.CPhase(-np.pi/6, wires=(qb_1, qb_2))
+
+    # one half-cycle - single-qubit gates only
+    for w in range(wires):
+        np.random.choice(single_qubit_gates)(wires=w)
+
+    return qml.probs(wires=range(wires))
```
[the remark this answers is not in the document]
thisac: We're interested in both. We need to sample from the circuit and then use the sampled bitstrings ideal probability (calculated from `qml.probs`) to calculate the fidelity. Either way we need to get one from the circuit output and one from `dev`, or both from the circuit.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,487 @@
...
+
+######################################################################
+# We set a random seed and use it to calculated the probability for all the
+# possible 12 qubit states. We can then sample from the device by calling
+# ``dev.generate_samples`` and similarly, sample random bitstrings from a
+# normal distribution by generating all basis states, along with their
+# corresponding bitstrings, and sample directly from them using NumPy.
+#
+
+seed=np.random.randint(0, 42424242)
```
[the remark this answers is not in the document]
thisac: The range is completely cherry-picked by me (although the seed will be fairly random). Is there a nice way of getting a random seed or do I have to randomize an integer, like I'm doing here? :thinking:

demonstrations/qsim_supremacy.py
thisac: As I also mentioned in a comment to Nathan above, this could easily be done, but I would have to change -1 to 1 and 1 to 0 for all samples, and either way I would need to get the probabilities later, either via an if statement in the circuit or by using `dev.probabilities`. I guess the later wouldn't really solve your issue, so maybe the former is the best way to go.

demonstrations/qsim_supremacy.py
thisac: `qsimcirq` didn't want to get installed via the CI (not sure why, but it was complaining about not having `cmake` installed, and when I installed it it complained some more, so I gave up and hard-coded the output). The script, as it is now, takes some time to finish, although it would be easy to lower the number of runs and/or samples to get it done quicker.

demonstrations/qsim_supremacy.py
thisac: That's shouldn't actually be too surprising, since this cross-entropy benchmarking fidelity isn't strictly bound between 0 and 1 unless using an infinite amount of shots/trials. 

For examples, following the definition F = 2^n * <P(x)> - 1, where <P(x)> is the mean of the simulated probabilities of the measured bitstrings, if we measure only one bitstring with P(x) = 0.0005, then F = 2^12 * 0.0005 - 1 = 1.048 > 1 for 12 qubits. Same goes for the negative values.

demonstrations/qsim_supremacy.py
thisac: Same reasoning as above, that's why I added the next part, which (when run) shows clearly how the fidelity gets closer to the theoretical value the more runs there are.

demonstrations/qsim_supremacy.py
thisac: Oh, maybe arbitrary is the wrong word to use here. I mean that they aren't particularly close to the theoretical values, and they are, as you both noted, both >1 and <0. :smile: The code snippet below shows that the calculated fidelity actually gets closer to the theoretical value the more times it's run.

demonstrations/qsim_supremacy.py
thisac: I just mean that sampling _effectively_ from any distribution classically (e.g. without using a quantum circuit) would always produce a value close to zero. Although, I might need to reword this a bit, since Haoyu commented that this might not always be the case.

demos_research.rst
thisac: There is no image yet. I still need to create one. :pensive:

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
josh146: The if statement approach is my preference 👍

demonstrations/qsim_supremacy.py
josh146: Yep - no binary wheel is provided by `qsimcirq`, so when you pip install it it will be compiled from source.

Which requires setting up CircleCI to install all the C++ toolchain required for compilation, which is non-trivial. I suggested best to handcode the demo output for now so that we are not blocked. We can always look back into installing `qsimcirq` on Circle later if we have time.

demonstrations/qsim_supremacy.py
josh146: Ah! Got it.

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
co9olguy: Yes, IIRC you can't pass the qubits directly, you need to pass their index. In this case, since the qubits are already `sorted`, then you can get away with `range(wires)` as your index set.

After thinking about it, I think you've got the nicest approach here. Otherwise, you'd have to manually remember which qubit was, e.g., the fifth in the `sorted` list, e.g., wire `5`

demonstrations/qsim_supremacy.py
co9olguy: Doesn't sound too hard?
```
samples = [0, 1, 1, 0 ..., 0]
ints = np.array(samples) * 2 - 1
```

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
co9olguy: I think it's ok if you give attribution

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
co9olguy: I like `qsim`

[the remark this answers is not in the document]
co9olguy: Honestly it feels weird to refer to a proper noun/name without a capital. Putting it in a "code" format gives us a plausible reason for keeping lowercase

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
co9olguy: You mean you want a random seed that changes every time?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+        return qml.probs(wires=range(wires))
+    else:
+        return [qml.sample(qml.PauliZ(i)) for i in range(wires)]
+
+
+######################################################################
+# The cross-entropy benchmarking fidelity
+# ---------------------------------------
+#
+# The performance metric that is used in the paper, and the one that we
```
co9olguy, line 330:
```suggestion
# The performance metric that is used in the supremacy experiment, and the one that we
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+#
+#    F_{XEB} = 2^{n}\left<P(x_i)\right> - 1,
+#
+# where :math:`n` is the number of qubits, :math:`P(x_i)` is the
+# probability of bitstring :math:`x_i` computed for the ideal quantum
+# circuit, and the average is over the observed bitstrings.
+#
+# The idea behind using this fidelity is that it will be close to 1 for
+# samples obtained from random quantum circuits, such as the one we defined
+# above, and close to zero for a normal probability distribution, that can
```
co9olguy, line 344: Unless you specifically mean a "Normal distribution", I think it's best to avoid the word "normal" here

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+# be effectively sampled from classically. Let us briefly calculate the
+# expected theoretical value for the fidelity based on the number of qubits
+# in the circuit. Sampling a bitstring from a random quantum circuit would
```
co9olguy, line 347:
```suggestion
# be effectively sampled from classically. 
# Sampling a bitstring from a random quantum circuit would
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+#
+# which leads to the theoretical fidelity
+#
+# .. math::
+#
+#    F_{XEB} = 2^{n}\left<P(x_i)\right> - 1 = \frac{2N}{N+1} - 1.
+#
+# We implement this fidelity as the function below, where ``samples`` is a
+# list sampled bitstrings, and ``probs`` is a list with corresponding
+# sampling probabilities for the same noise-less circuit.
```
co9olguy, line 370:
```suggestion
# sampling probabilities for the same noiseless circuit.
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+# We set a random seed and use it to calculate the probability for all the
+# possible bitstrings. We can then sample from the same circuit, transform
+# the Pauli-Z eigenvalue samples into their correponding bitstrings, and
+# use them to calulate the cross-entropy benchmarking fidelity.
+#
+
+seed = np.random.randint(0, 42424242)
+probs = circuit(seed=seed, return_probs=True)
+
+samples = circuit(seed=seed).T
```
co9olguy, line 395: How come we need to transpose? :thinking: Just so we can iterate over axis 0 in the `for` loop below?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+
+# take the eigenvalues and transform -1 to 1 and 1 to 0,
+# and then join each sample together to form a bitstring
+# (-(array([-1, 1] - 1) / 2 = array([1, 0])))
+bitstring_samples = []
+for sam in samples:
+    new_sam = -(sam - 1) // 2
```
co9olguy, line 402: Instead of explaining the math in the comment, I think we can just give more clear names, and the reader should be able to follow along.
e.g., `samples`->`circuit_samples` or `pauliZ_samples`
and `new_sam` = `bittring_sample`

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+    bitstring_samples.append("".join(str(bs) for bs in new_sam))
+
+f_circuit = fidelity_xeb(bitstring_samples, probs)
+
+######################################################################
+# Similarly, we can sample random bitstrings from a normal distribution by
```
co9olguy, line 408: Again, do you really mean *the* Normal distribution here? :thinking:. What does a normal distribution over bitstrings mean?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+basis_states = dev.generate_basis_states(wires)
+random_integers = np.random.randint(0, len(basis_states), size=shots)
+bitstring_samples = []
+for i in random_integers:
+    bitstring_samples.append("".join(str(bs) for bs in basis_states[i]))
+
+f_normal = fidelity_xeb(bitstring_samples, probs)
+
+######################################################################
+# Finally, let us print the two different values. Sampling from the
```
co9olguy, line 422:
```suggestion
# Finally, let's compare the two different values. Sampling from the
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+print("Theoretical:", f"{theoretical_value:.7f}\n".rjust(24))
+
+##############################################################################
+# .. rst-class:: sphx-glr-script-out
+#
+#  Out:
+#
+#  .. code-block:: none
+#
+#     Theoretical:               0.9995118
```
co9olguy, line 469: It was a bit weird to have only the theoretical value printed here, but not the sampling value. Later the theoretical value is printed again, so I'd recommend just merging this block with the next one, and dropping one of the `print`s

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+#
+
+######################################################################
+# To show that the fidelity from the circuit sampling actually tends
+# towards the theoretical value we can run several different random
+# circuits, calculate their respective cross-entropy benchmark fidelities
+# and then calculate the mean fidelity of all the runs. The more
+# evaluations we do, the closer to the theoretical value we should get.
+#
+# In the supremacy paper, they typically calculate each of their
```
co9olguy, line 479:
```suggestion
# In the supremacy experiment, they typically calculate each of their
```
Just felt that the demo refers too much to the paper itself, as if it is the primary agent, not the overall supremacy undertaking

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+# this experiment is used to demonstrate quantum supremacy.
+#
+# Much is due to the Porter-Thompson probability distribution that the
+# random quantum circuits follow, which are hard to simulate classically.
+# On the other hand, a quantum device, running a circuit as the one
+# constructed above, should be able to sample from such a distribution
+# without much overhead. Thus, by showing that a quantum device can produce
+# a high enough fidelity value for a large enough circuit, quantum
+# supremacy can be claimed. This is exactly what Google's supremacy paper
+# has done.
```
co9olguy, line 541:
```suggestion
# supremacy can be claimed. This is exactly what Google's supremacy experiment
# has done.
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+#
+# There's still one issue that hasn't been touched on yet: the addition of
+# noise in quantum hardware. Simply put, this noise will lower the
+# cross-entropy benchmark fidelity, getting it closer to 0. The larger the
+# circuit, the more noise there will be, and thus the lower the fidelity.
+# By calculating the specific single-qubit, two-qubit, and readout errors
+# of the Sycamore chip, using them to simulate a noisy circuit, the Google
```
co9olguy, line 548:
```suggestion
# of the Sycamore chip, and using them to simulate a noisy circuit, the Google
```

co9olguy (changes requested):
Thanks @thisac! Amazing work.

I've left a number of suggestions, but they are all in the spirit of polishing, clarifying, or simplifying as much as we can.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+#
+# .. figure:: ../demonstrations/qsim_supremacy/supremacy_circuit.png
+#     :align: center
+#     :width: 90%
+#
+#     **Image taken from Arute, F., Arya, K., Babbush, R. et al.** [#Arute2019]_
+#
+# The logic below iterates through all connections and returns a dictionary
+# ``gate_order`` with lists of tuples as keys. Each tuple contains two
+# neighbouring qubits with the key as their connection label. We will use
```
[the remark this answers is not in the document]
thisac: My bad. Should have said:

The logic below iterates through all connections and returns a dictionary
``gate_order`` with lists of tuples as **values**. Each tuple contains two
neighbouring qubits with the key as their connection label.

I'll rewrite this.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,595 @@
...
+# The two-qubit gates are applied to the qubits connected by A, B,
+# C, or D as defined above. The circuit finally ends with a half-cycle,
+# consisting of only a layer of single-qubit gates.
+#
+# We need both the probabilities of the circuit as well as being able to
+# sample from it, as we will see in the next section. To do this, we add a
+# keyword argument to our circuit allowing us to switch between the two
+# returns. We sample from the Pauli-Z observable on all wires, which will
+# give us the eigenvalues :math:`\pm 1` of the observable, corresponding to
+# the states :math:`\left|0\right>` and :math:`\left|1\right>`.
```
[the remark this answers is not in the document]
thisac: It didn't render properly due to a missing space after the first math part earlier for me, though I think it's working now.

demonstrations/qsim_supremacy.py
thisac: I specifically meant a "Normal distribution", although when thinking about it, it should be "uniform distribution". I'll update this. Thanks for spotting it!

demonstrations/qsim_supremacy.py
thisac: Yes. I think it makes more sense to have the samples with shape `(shots, modes)` although, due to the list comprehension loop to return the samples, they are actually returned with shape `(modes, shots)`.

demonstrations/qsim_supremacy.py
thisac: I'll clarify.

demonstrations/qsim_supremacy.py
thisac: I think you're right here! :+1:

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
thisac: I think it's a valid way of choosing a seed for the demonstration, but it would be equally fine to set a specific seed here as well. In the second place that this is done, the seed needs to be different each time, so there it is needed (unless I use a set of predefined, different seeds, but I think setting a random one makes more sense). The important thing is that we use the same seed to calculate both the probabilities and the samples for each run (so that the circuits are identical).

demonstrations/qsim_supremacy.py
```diff
@@ -287,16 +277,18 @@ def generate_single_qubit_gate_list():
...
+# single-qubit gates.
 #
-# We need both the probabilities of the circuit as well as being able to
-# sample from it, as we will see in the next section. To do this, we add a
-# keyword argument to our circuit allowing us to switch between the two
-# returns. We sample from the Pauli-Z observable on all wires, which will
-# give us the eigenvalues :math:`\pm 1` of the observable, corresponding to
-# the states :math:`\left|0\right>` and :math:`\left|1\right>`.
+# From the QNode, we also need both the probabilities of the measurement
+# results, as well raw samples. To facilitate this, we add a keyword
```
co9olguy, line 287:
```suggestion
# results, as well as raw samples. To facilitate this, we add a keyword
```

demonstrations/qsim_supremacy.py
```diff
@@ -287,16 +277,18 @@ def generate_single_qubit_gate_list():
...
+# The circuit ends with a half-cycle, consisting of only a layer of
+# single-qubit gates.
 #
-# We need both the probabilities of the circuit as well as being able to
-# sample from it, as we will see in the next section. To do this, we add a
-# keyword argument to our circuit allowing us to switch between the two
-# returns. We sample from the Pauli-Z observable on all wires, which will
-# give us the eigenvalues :math:`\pm 1` of the observable, corresponding to
-# the states :math:`\left|0\right>` and :math:`\left|1\right>`.
+# From the QNode, we also need both the probabilities of the measurement
```
co9olguy, line 286:
```suggestion
# From the QNode, we need both the probabilities of the measurement
```

demonstrations/qsim_supremacy.py
```diff
@@ -341,10 +333,9 @@ def circuit(seed=42, return_probs=False):
 #
 # The idea behind using this fidelity is that it will be close to 1 for
 # samples obtained from random quantum circuits, such as the one we defined
-# above, and close to zero for a normal probability distribution, that can
-# be effectively sampled from classically. Let us briefly calculate the
-# expected theoretical value for the fidelity based on the number of qubits
-# in the circuit. Sampling a bitstring from a random quantum circuit would
+# above, and close to zero for a uniform probability distribution, that can
```
co9olguy, line 336:
```suggestion
# above, and close to zero for a uniform probability distribution, which can
```

co9olguy (approved):

demos_research.rst
```diff
@@ -123,6 +123,12 @@ plateaus in quantum neural networks.
 
 :html:`<div class="gallery-grid row">`
 
+.. customgalleryitem::
```
co9olguy, line 126: :thinking: We're due for a new subsection of the website, covering general quantum computing (quantum volume I think can go in there too)

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+# are applied, as well as the cycles, each consisting of a layer of
+# single-qubit gates and a pair of two-qubit gates. Note that each coloured
+# two-qubit gate represented in the image is implemented as the two
+# consecutive gates iSWAP and CPhase in this demo.
+#
+# .. figure:: ../demonstrations/qsim_supremacy/supremacy_circuit.png
+#     :align: center
+#     :width: 90%
+#
+#     **Image taken from Arute, F., Arya, K., Babbush, R. et al.** [#Arute2019]_
```
josh146, line 199: g18

Had a go re-doing this one (and tried to remove some of the overlap), feel free to use it if you would like :)

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+# longer sequence, which is much harder to simulate classically, is used
+# for estimating the cross-entropy fidelity in what they call the "supremacy
+# regime". We will use the shorter gate sequence for the following
+# demonstration; feel free to play around with other combinations.
+#
+
+m = 14  # number of cycles
+
+# gate_sequence = np.resize(["A", "B", "C", "D", "C", "D", "A", "B"], m)
+gate_sequence = np.resize(["A", "B", "C", "D"], m)
```
josh146, line 248: It might look slighter nice to leave both uncommented, e.g.,
```suggestion
gate_sequence_longer = np.resize(["A", "B", "C", "D", "C", "D", "A", "B"], m)
gate_sequence = np.resize(["A", "B", "C", "D"], m)
```

josh146: (even though `gate_sequence_longer` isn't used)

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+
+
+######################################################################
+# The single-qubit gates are randomly selected and applied to each qubit in
+# the circuit, while avoiding the same gate being applied to the same wire
+# twice in a row. We do this by creating a helper function that
+# specifies the order in which the single-qubit
```
josh146, line 255:
```suggestion
# twice in a row. We do this by creating a helper function ``generate_single_qubit_gate_list()`` that
# specifies the order in which the single-qubit
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+# gates should be applied. We can use this list within the
+# circuit to know which gate to apply when.
+#
+
+def generate_single_qubit_gate_list():
+    # create the first list by randomly selecting indices
+    # from `single_qubit_gates`
```
josh146, line 262:
```suggestion
    # from single_qubit_gates
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+            ops.CPhase(-np.pi/6, wires=(qb_1, qb_2))
+
+    # one half-cycle - single-qubit gates only
+    for w in range(wires):
+        single_qubit_gates[gate_idx[-1][w]](wires=w)
+
+    if return_probs:
+        return qml.probs(wires=range(wires))
+    else:
+        return [qml.sample(qml.PauliZ(i)) for i in range(wires)]
```
josh146, line 315: I like this!

josh146: Do we want to make it PEP8 compliant, e.g.,
```suggestion
    if return_probs:
        return qml.probs(wires=range(wires))

    return [qml.sample(qml.PauliZ(i)) for i in range(wires)]
```

I'm in two minds, because an explicit if-else makes it easier for readers to follow.

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+
+        # retrieve the corresponding probability for the bitstring
+        sampled_probs.append(probs[bitstring_idx])
+
+    return 2 ** len(samples[0]) * np.mean(sampled_probs) - 1
+
+
+######################################################################
+# We set a random seed and use it to calculate the probability for all the
+# possible bitstrings. We can then sample from the same circuit, transform
```
josh146, line 378: Maybe worth restructuring this to avoid both sentences beginning with "We"?

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+probs = circuit(seed=seed, return_probs=True)
+
+# transpose the samples to get the shape `(shots, wires)`
+circuit_samples = circuit(seed=seed).T
+
+# take the eigenvalues and transform -1 to 1 and 1 to 0
+bitstring_samples = []
+for sam in circuit_samples:
+    bitstring_sample = -(sam - 1) // 2
+    bitstring_samples.append("".join(str(bs) for bs in bitstring_sample))
```
josh146, line 393: Perhaps worth explaining more explicitly why this is needed, for those unsure (e.g., sampling Z returns -1, 1, but we need computational basis samples)

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+# close to 0.
+#
+# .. note::
+#
+#     The cross-entropy benchmarking fidelity may output
+#     values that are negative or that are larger than 1, for any finite
+#     number of samples. This is due to the random nature of the sampling.
+#     For an infinite amount of samples, or circuit runs, the observed
+#     values will tend towards the theoretical ones, and will then always
+#     lie in the 0-to-1 interval.
```
josh146, line 424: Nice!

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+#
+print("Circuit's distribution:", f"{f_circuit:.7f}".rjust(12))
+print("Uniform distribution:", f"{f_uniform:.7f}".rjust(14))
```
josh146, line 427: I like the alignment of the digits 😆

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+# constructed above, should be able to sample from such a distribution
+# without much overhead. Thus, by showing that a quantum device can produce
+# a high enough fidelity value for a large enough circuit, quantum
+# supremacy can be claimed. This is exactly what Google's supremacy experiment
+# has done.
+#
+# There's still one issue that hasn't been touched on yet: the addition of
+# noise in quantum hardware. Simply put, this noise will lower the
+# cross-entropy benchmark fidelity, getting it closer to 0. The larger the
+# circuit, the more noise there will be, and thus the lower the fidelity.
```
josh146, line 518:
```suggestion
# noise in quantum hardware. Simply put, this noise will lower the
# cross-entropy benchmark fidelity---the larger the
# circuit, the more noise there will be, and thus the lower the fidelity, with the
# fidelity approaching 0 as the noise increases.
```

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+# thus proclaimed quantum supremacy (see Fig. 4 in [#Arute2019]_).
+#
+# .. note::
+#
+#     For more reading on this, the original paper [#Arute2019]_ is highly
+#     recommended (along with the suplementary information [#Arute2019sup]_ if you want
+#     to dive deeper into the math and physics of the experiment). The blog
+#     post in [#Sohaib2019]_, along with the accompanying GitHub repo, also provides
+#     a nice introduction to the cross-entropy benchmark fidelity, and
+#     includes calculations highlighting the effects of added noise models.
```
josh146, line 533: I find this a really strong conclusion 💪

josh146 (commented):
Looking great @thisac! Huge improvements across the board.

demonstrations/qsim_supremacy.py
co9olguy: @josh146 you're a legend

demonstrations/qsim_supremacy.py
co9olguy: I vote no. I actually find explicit `if/else` more readable

josh146: I think I agree

demonstrations/qsim_supremacy.py
```diff
@@ -0,0 +1,567 @@
...
+
+    tutorial_quantum_metrology Variationally optimizing measurement protocols
+    tutorial_noisy_circuit_optimization Optimizing noisy circuits with Cirq
+
+In the paper `Quantum supremacy using a programmable superconducting
+processor <https://www.nature.com/articles/s41586-019-1666-5>`__, the
+Google AI Quantum team and collaborators showed that the Sycamore quantum processor could
+complete a task that would take a classical computer potentially thousands
+of years. They faced their quantum chip off against JEWEL—one of the
+world's most powerful supercomputers—using a classical statevector simulator called
```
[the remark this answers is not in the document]
thisac: Thanks @josh146. I thought so too, but I think it got changed to `-` during code-review, so I was a bit unsure (and just went with the suggestion). Is `---` preferred? :slightly_smiling_face:

demonstrations/qsim_supremacy.py
thisac: This is awesome @josh146! Thanks a lot for drawing this. :tada:

demonstrations/qsim_supremacy.py
thisac: Thanks. I've started following this PEP8 suggestion recently (and I've actually started to like it), although I don't mind either way to be honest. I'll keep it like this then since there are 2 no and 1 abstain. :laughing:

demos_research.rst
thisac: I was thinking that as well. This demo didn't really fit into neither Optimization nor QML. :thinking:

demonstrations/qsim_supremacy.py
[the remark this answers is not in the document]
josh146: Oops. If that was me, it was accidental. I prefer `---`, but it is personal preference!

josh146 (approved):
🏆

PR #163, opened 2020-11-19
Adds demo on quantum volume

glassnotes:
**Title:** Adds demo on quantum volume

**Summary:** Adds a demo that discusses motivation and theory behind the quantum volume metric, and walks a user through how to calculate it with PennyLane on a simulated hardware device.

**Relevant references:** References in tutorial. 

**Possible Drawbacks:** Still have a couple points to address.
 - [x] build currently fails when trying to access hardware noise model; need to find solution (maybe save a copy locally?)
 - [x] decide how to categorize (currently under "basic", but should be moved since the methodology actually ended up being pretty tricky) (edit: moved under 'applications') 
 - [x] discuss addition of permutations and random unitaries to PL (edit: will do after demo release, and update demo at later date)
 - [x] improve discussion of statistical assumptions
 - [x] flesh out conclusion
 - [x] elaborate on limitations and needing classical simulation of circuits
 - [x] figure out how to increase spacing between images and captions
 - [x] remove warning that gets output when loading the device  


demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+with tape:
+    qml.probs(wires=range(m))
+
+output_probs = tape.execute(dev_ideal).reshape(
+    2 ** m,
+)
+heavy_outputs, prob_heavy_output = heavy_output_set(m, output_probs)
+
+print(f"Probability of a heavy output is {prob_heavy_output}")
+print(f"Heavy outputs are {heavy_outputs}")
```
ixfoduap, line 433: Awesome!

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+
+
+##############################################################################
+#
+# Step 2: run the circuits
```
ixfoduap, line 438: Here I'm confused: are we running on a noisy simulator?

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# First things first, let's set up a hardware device. We'll make a 5-qubit
```
ixfoduap, line 441: This is the only section so far that I think needs work. I find its current version hard to follow. Also, I'm still not sure why we are using qiskit.

I know that compilation is an important factor in quantum volume, but couldn't we work with a pre-compiled circuit of our choice that makes things simpler? We could also use a nice image like the ones in the beginning to illustrate the connectivity instead of an nx graph

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+        "coupling_map": coupling_map,
+        "layout_method": "sabre",
+        "routing_method": "sabre",
+    }
+)
+
+
+##############################################################################
+#
+# It's time to run the protocol. We'll start with the smallest circuits on 2
```
ixfoduap, line 519: Isn't this a very intense computation? Also, couldn't we use analytic probabilities instead of shots for the sake of the tutorial? It may also help with statistical noise

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+        probs_ideal[m - min_m, trial] = prob_heavy_output
+        probs_noisy[m - min_m, trial] = fraction_device_heavy_output
+
+##############################################################################
+#
+# Step 3: perform a statistical analysis
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Having run our experiments, we can now get to the heart of the quantum volume
+# protocol - what *is* the largest square circuit that our processor can run?
```
ixfoduap, line 570:
```suggestion
# protocol: what *is* the largest square circuit that our processor can run?
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+probs_mean_ideal = np.mean(probs_ideal, axis=1)
+probs_mean_noisy = np.mean(probs_noisy, axis=1)
+
+print(f"Ideal mean probabilities: {probs_mean_ideal}")
+print(f"Device mean probabilities: {probs_mean_noisy}")
+
+##############################################################################
+#
+# So we see that for the ideal probabilities, we're well over 2/3. In fact,
+# we're closing in on 0.85, which is the expected value as mentioned in the box
```
ixfoduap, line 584: add here that it's 0.85 = (1+ln 2)/2

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# Having run our experiments, we can now get to the heart of the quantum volume
+# protocol - what *is* the largest square circuit that our processor can run?
+#
+# Let's first check out the means, and see how much higher they are than 2/3.
+
+
+probs_mean_ideal = np.mean(probs_ideal, axis=1)
+probs_mean_noisy = np.mean(probs_noisy, axis=1)
+
+print(f"Ideal mean probabilities: {probs_mean_ideal}")
```
ixfoduap, line 578: This is arguably the most important part of the tutorial, but I have a hard time interpreting the output. It would be best to clearly indicate the value of `m` and whether we meet the threshold or not. Maybe something like

```python
for i, p in enumerate(probs_mean_noisy):
    print('m = ', i+2')
    print('p = ', p)
    print('Threshold met = ', p>2/3)
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+#
+# So we see that for the ideal probabilities, we're well over 2/3. In fact,
+# we're closing in on 0.85, which is the expected value as mentioned in the box
+# above. For the device probabilities, however, we see that already by the
+# 4-qubit case, we're very close to the threshold. Since it's below the
+# threshold at :math:`m=5`, the highest volume this processor can have is
+# :math:`\log_2 V_Q = 4`. But isn't enough to just see the *mean* of the heavy
+# output probabilities be greater than 2/3. Since we're dealing with randomness,
+# we also want to be confident that these results were not just a fluke!
+# Try playing around with the code yourself - are there any parameters you can
```
ixfoduap, line 654:
```suggestion
# Try playing around with the code yourself: are there any parameters you can
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# change to improve the volume with this same noise model? What happens if we
+# don't specify a high level of optimization and transpilation? Furthermore, how
```
ixfoduap, line 656: I think transpilation is not defined and may be confusing

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# do the results change with a more complex noise model from an actual device?
+#
+# Concluding thoughts
```
ixfoduap, line 659: This is a bit random to end. But good to mention! Maybe somewhere in the beginning?

ixfoduap (commented):
Really great and useful tutorial!

I think the first part explaining the basics of quantum volume is excellent. I mostly left minor suggestions.

The second part where the protocol is implemented was a bit confusing and the code a bit heavy. I wonder if we can polish it to be simpler and flow better? I'm also not a huge fan of writing so much qiskit code, specially since we can now simulate noisy circuits in pennylane

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# Equipped with our definition of quantum volume, it's time to actually try and
+# compute it ourselves. We'll use the `PennyLane-Qiskit
+# <https://pennylaneqiskit.readthedocs.io/en/latest/>`_ plugin to determine the
+# volume of a noisy toy device.
+#
+# Loosely, the protocol for quantum volume consists of three steps:
+#
+# 1. Construct random square circuits of increasing size
+#
+# 2. Run those circuits on both a simulator, and on a noisy hardware device
```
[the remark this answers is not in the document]
glassnotes: The ideal device is needed in order to compute the heavy outputs. Based on your other comments I think I need to look into and explain better the classical requirements of the protocol, and the disadvantages of this.

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+#
+# First we write a function that randomly permutes qubits. We'll do this by
+# using numpy to generate a permutation, and then apply it with SWAP gates.
+
+# For reproducibility
+np.random.seed(42)
+rng = np.random.default_rng()
+
+
+def permute_qubits(num_qubits):
```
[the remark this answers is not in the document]
glassnotes: I was thinking this is something that may be worth adding to PennyLane, since permutations are a fairly common occurrence. (And that way, we can just call it without showing any implementation :) )

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# generate a circuit with 3 qubits. To help with the visualization, we'll also apply
+# an RZ gate to each qubit at the beginning to make sure all 5 qubits are there, and
+# in order.
+
+num_qubits = 5
+dev_ideal = qml.device("default.qubit", analytic=True, wires=num_qubits)
+
+m = 3
+
+with qml.tape.QuantumTape() as tape:
```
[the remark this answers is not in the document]
glassnotes: I *think* so, but not 100% sure. The reason I used it is that we're generating random circuits that have to be executed identically over two devices. So having it as a tape that then executes separately on each device seemed more straightforward than trying to make two separate qnodes with the same contents.

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# The median of this probability distribution is 0.208, and the heavy bit
+# strings are '00', and '10', because these are the two probabilities above the
+# median. If we were to run this circuit, the probability of obtaining one of
+# the heavy outputs is 0.792, and so we expect that we will see one of these two
+# most of the time.
+#
+# Each circuit in a circuit family has its own heavy output probability. If our
+# quantum computer is of high quality, then we should expect to see heavy
+# outputs quite often across all the circuits. On the other hand, if it's of
+# poor quality and everything is depolarized, we will end up with output
```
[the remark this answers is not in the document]
glassnotes: I often mix up "depolarized" and "decohered". Decohered is what I was going for here, thanks for catching this!

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+
+##############################################################################
+#
+# Computing the quantum volume
+# ----------------------------
+#
+#
+# Equipped with our definition of quantum volume, it's time to actually try and
+# compute it ourselves. We'll use the `PennyLane-Qiskit
+# <https://pennylaneqiskit.readthedocs.io/en/latest/>`_ plugin to determine the
```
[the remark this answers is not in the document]
glassnotes: When I was writing this locally, I was running with simulated noise models of the actual hardware, but was thinking the same thing about TOS so I switched to just constructing a noise model manually. The other reason for using PL-Qiskit here though is that the QV protocol encourages heavy compilation of the circuits, which is accessible through the Qiskit transpiler options.

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+import numpy as np
+
+###############################################################################
+#
+# First we write a function that randomly permutes qubits. We'll do this by
+# using numpy to generate a permutation, and then apply it with SWAP gates.
+
+# For reproducibility
+np.random.seed(42)
+rng = np.random.default_rng()
```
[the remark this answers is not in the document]
glassnotes: This is the new numpy way of random number generation (as of v1.17). I will add a note about that here, this is the first time I've actually used it, and could be the first time people are seeing it.

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
josh146: (Personal opinion) I would find it quite cool if the demo used simulated noise models of the actual hardware. It makes the demonstration feel more 'tangible', and less theoretical.

demonstrations/tutorial_quantum_volume.py
glassnotes: Great suggestion, thanks! :star: I've implemented something like this and it really helps with interpreting the results.

demos_basics.rst
```diff
@@ -52,8 +52,14 @@ photonic quantum computing.
     :tooltip: Compare the parameter-shift rule with backpropagation.
     :figure: demonstrations/tutorial_backprop_thumbnail.png
     :description: :doc:`demos/tutorial_backprop`
-    :tags: tensorflow autograd:html:`</div></div><div style='clear:both'>`
```
josh146, line 55: oops

demos_basics.rst
```diff
@@ -52,8 +52,14 @@ photonic quantum computing.
...
+    :tags: tensorflow autograd 
 
+.. customgalleryitem::
+   :tooltip: Learn how to compute the quantum volume of a quantum processor.
+   :figure: demonstrations/quantum_volume/quantum_volume_thumbnail.png
```
josh146, line 59: great thumbnail 😆

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+Twice per year, a project called the TOP500 [#top500]_ releases a list of the
+500 most powerful supercomputing systems in the world. However, there is a large
+amount of variation in how supercomputers are built. `Some
+<https://en.wikipedia.org/wiki/Fugaku_(supercomputer)>`_ use 48-core processors,
+while `others <https://en.wikipedia.org/wiki/Sunway_TaihuLight>`_ use processors
+with up to 260 cores. The speed of processors will differ, and they may be
+connected in different ways. We can't rank them by simply counting the number of
+processors. In order to make a fair comparison, we need benchmarking standards
+that give us a holistic view of their performance. To that end, the TOP500
```
josh146, line 19:
```suggestion
that give us a holistic view of their performance.

To that end, the TOP500
```

Suggest making this a new paragraph (since it's quite a big block of text when rendered!).

Starting sentence of the new paragraph might need some modification though.

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+rankings are based on something called the LINPACK benchmark [#linpack]_. The
+task of the supercomputers is to solve a dense system of linear equations, and
+the metric of interest is the rate at which they perform floating-point
+operations (FLOPS). Today's top machines reach speeds well into the regime of
```
josh146, line 23:
```suggestion
the metric of interest is the rate at which they `perform floating-point
operations (FLOPS) <https://en.wikipedia.org/wiki/FLOPS>`__.
Today's top machines reach speeds well into the regime of
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+A similar problem is emerging with quantum computers: we
+can't judge quantum computers on the number of qubits alone. Present-day devices
+have a number of limitations, an important one being the fairly high gate error
+rates. Typically the qubits on a chip are not all connected to each other, so it
+may not be possible to perform operations on arbitrary pairs of
+them. Considering this, can we tell if a machine with 20 noisy qubits is better
+than one with 5 very high-quality qubits? Or if a machine with 8 fully-connected
+qubits is better than one with 16 qubits of comparable error rate, but arranged in
+a square lattice?  Furthermore, how can we make comparisons between different
+types of qubits?
```
josh146, line 37: 'Considering this..' could also be the beginning of a new paragraph?

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+qubits on which it can reliably run circuits of a prescribed type.
+
+You can think of it loosely like the quantum analogue of the LINPACK
+Benchmark. Different quantum computers are tasked with solving the same problem,
+and the success will be a function of many properties: error rates, qubit
+connectivity, even the quality of the software stack. Like LINPACK, a single
+number won't tell us everything about a quantum computer, but it does establish
+a framework for comparing them.
+
+After working through this tutorial, you'll be able to define quantum volume,
```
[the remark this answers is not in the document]
josh146: 🧊

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# that contribute to the successful execution of a computation. Therefore, we
+# must be very explicit about what exactly we are benchmarking, and what is our
+# measure of success. In general, to set up a benchmark for a quantum computer
+# we need to decide on a number of things [#robin]_:
+#
+#  1. A family of circuits with a well-defined structure and variable size
+#  2. A set of rules detailing how the circuits can be compiled
+#  3. A measure of success for individual circuits
+#  4. A measure of success for the family of circuits
+#  5. (Optional) An experimental design specifying how the circuits are to be run
```
josh146, line 79:
```suggestion
# 1. A family of circuits with a well-defined structure and variable size
# 2. A set of rules detailing how the circuits can be compiled
# 3. A measure of success for individual circuits
# 4. A measure of success for the family of circuits
# 5. (Optional) An experimental design specifying how the circuits are to be run
```
(the indentation causes Sphinx to render it as a block quote)

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+# .. figure:: ../demonstrations/quantum_volume/model_circuit_cross.png
+#     :align: center
+#     :width: 60%
+#
+#     ..
+#
+#     A schematic of the random circuit structure used in the quantum volume protocol.
+#     Image source: [#cross]_.
+#
+# Specifically, the circuits consist of :math:`d` sequential layers acting on
```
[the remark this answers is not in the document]
josh146: agree!

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# measurement outcomes may not be easy to sample using classical
+# techniques.
+#
+# A distribution that is theorized to fulfill this property is the distribution
+# of *heavy* output bit strings. Heavy bit strings are those whose outcome
+# probabilities are above the median of the distribution. For example, suppose
+# we run a two-qubit circuit, and find that the measurement probabilities for
+# the output states are as follows:
+
+measurement_probs = {"00": 0.558, "01": 0.182, "10": 0.234, "11": 0.026}
```
josh146, line 142: Might be nice to explicitly show the median calculation?
```suggestion
import numpy as np
measurement_probs = {"00": 0.558, "01": 0.182, "10": 0.234, "11": 0.026}
prob_array = np.fromiter(measurement_probs.values(), dtype=np.float)
np.median(prob_array)
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+##############################################################################
+#
+# The median of this probability distribution is 0.208. The heavy bit
+# strings are '00' and '10', because these are the two probabilities above the
+# median. If we were to run this circuit, the probability of obtaining one of
+# the heavy outputs is 0.792, and so we expect that we will see one of these two
+# most of the time.
```
josh146, line 150: It might also be nice to show this explicitly, e.g.,
```python
np.sum(prob_array[prob_array > np.median(prob_array)])
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+#
+# As an example, if we have a 20-qubit device, and find that we get heavy
+# outputs reliably for up to depth-4 circuits on 4 qubits, then the quantum
+# volume is :math:`\log_2 V_Q = 4`. Quantum volume is incrementally, as shown
+# below - we gradually work our way up to larger circuits, until we find
+# something we can't do.  Very loosely, quantum volume is like an effective
+# number of qubits. Even if we have those 20 qubits, only 4 of them work well
+# enough together to sample from distributions that would be considered hard
+#
+# .. figure:: ../demonstrations/quantum_volume/qv_square_circuits.svg
```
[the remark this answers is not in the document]
josh146: Same here!

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+#
+# .. note::
+#
+#    In many sources, the quantum volume of processors is reported as
+#    :math:`V_Q` explicitly, rather than :math:`\log_2 V_Q` as is the
+#    convention in this demo. As such, IonQs processor has the potential for a
+#    quantum volume of :math:`2^{32} > 4000000`. Here we use the :math:`\log`
+#    because it is more straightforward to understand that they have 32
+#    high-quality, well-connected qubits than to extract this at first glance from the
+#    explicit value of the volume.
```
josh146, line 245: My comments are becoming sparser because the previous two sections have been really good

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+##############################################################################
+#
+# Computing the quantum volume
+# ----------------------------
+#
+# Equipped with our definition of quantum volume, it's time to compute it
+# ourselves!. We'll use the `PennyLane-Qiskit
+# <https://pennylaneqiskit.readthedocs.io/en/latest/>`_ plugin to compute the
+# volume of one of the IBM processors, since their properties are easily
+# accessible through this interface.
```
josh146, line 258:
```suggestion
# volume of one of the IBM processors, since their properties are easily
# accessible through this plugin.
```
?

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
josh146: Woah.

I also did not know about it!

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
josh146: > I was thinking this is something that may be worth adding to PennyLane

:100:

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+    working_order = list(range(num_qubits))
+
+    # We will permute by iterating through the permutation and swapping
+    # things back to their proper place.
+    for idx_here in range(num_qubits):
+        if working_order[idx_here] != perm_order[idx_here]:
+            # Where do we need to send the qubit at this location?
+            idx_there = working_order.index(perm_order[idx_here])
+            qml.SWAP(wires=[idx_here, idx_there])
```
josh146, line 304:
```suggestion
            idx_there = working_order.index(perm_order[idx_here])

            qml.SWAP(wires=[idx_here, idx_there])
```

(only because it took me a while to spot the swap, I couldn't work out what the function was returning!

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,661 @@
...
+                working_order[idx_there],
+                working_order[idx_here],
+            )
+
+
+##############################################################################
+#
+# Next, we need to apply SU(4) gates to pairs of qubits. PennyLane doesn't have
+# built-in functionality to generate these random matrices, however its cousin
+# `StrawberryFields <https://strawberryfields.ai/>`_ does! We will use the
```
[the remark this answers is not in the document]
josh146: Another thing we should port to PL 😆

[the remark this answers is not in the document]
josh146:
```suggestion
# `Strawberry Fields <https://strawberryfields.ai/>`_ does! We will use the
```

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
josh146: Yeah, I think this might the case... I imagine, since the QNode is a function, calling it twice will result in different circuits

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+dev_ideal = qml.device("default.qubit", analytic=True, wires=num_qubits)
+
+m = 3  # number of qubits
+
+with qml.tape.QuantumTape() as tape:
+    for qubit in range(num_qubits):
+        qml.RZ(0, wires=qubit)
+    qml.templates.layer(qv_circuit_layer, m, num_qubits=m)
+
+print(tape.draw())
```
josh146, line 361: (out of curiosity, do you want to suppress the matrix printing here as well? If so, we could add it as a keyword argument)

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+num_qubits = 5
+dev_ideal = qml.device("default.qubit", analytic=True, wires=num_qubits)
+
+m = 3  # number of qubits
+
+with qml.tape.QuantumTape() as tape:
+    for qubit in range(num_qubits):
+        qml.RZ(0, wires=qubit)
+    qml.templates.layer(qv_circuit_layer, m, num_qubits=m)
```
josh146, line 359: 🔁

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+
+##############################################################################
+#
+# Next, we need to apply SU(4) gates to pairs of qubits. PennyLane doesn't have
+# built-in functionality to generate these random matrices, however its cousin
+# `StrawberryFields <https://strawberryfields.ai/>`_ does! We will use the
+# ``random_interferometer`` method, which can generate unitary matrices uniformly
+# at random. (This function actually generates elements of U(4), but they are
+# essentially equivalent up to a global phase).
```
josh146, line 320: I wonder if [ArbitraryUnitary](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.subroutines.ArbitraryUnitary.html) can be made Haar random? Perhaps (maybe) by how the parameters are chosen??

josh146: This might reduce the classical processing required to decompose the matrix as the number of qubits scales up (since `ArbitraryUnitary` is written in terms of Pauli rotations)

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# point is that this circuit with 3 layers actually has depth much greater than
+# 3, since each layer has both SWAPs and SU(4) operations that are further
+# decomposed into elementary gates when run on the actual processor.
+#
+# One last thing we'll need before running our circuits is the machinery to
+# determine the heavy outputs. This is quite an interesting aspect of the
+# protocol --- we're required to compute the heavy outputs classically in order
+# to get the results! As a consequence, it will only be possible to calculate
+# quantum volume for processors up to a certain point before they become too
+# large. That said, classical simulators are always improving, and can simulate
```
josh146, line 376: Paragraph break?
```suggestion
# large.
#
# That said, classical simulators are always improving, and can simulate
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# The heavy outputs can be retrieved from a classically-obtained probability
+# distribution as follows:
+#
+
+def heavy_output_set(m, probs):
+    # Sort the probabilities
+    probs_ascending_order = np.argsort(probs)
+    sorted_probs = probs[probs_ascending_order]
+
+    median_prob = np.median(sorted_probs)
```
josh146, line 392: The sorting might be unnecessary, it looks like `np.median` takes this into account?
```python
In [54]: np.median([3, 1, 0, 2, 4, 5, 6])
Out[54]: 3.0
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# need a supercomputer to do so). Furthermore, the designers of the proposal
+# don't expect this to be an issue until gate error rates decrease below
+# :math:`\approx 10^{-4}`, after which we may need to adjust the protocol to not
+# require classical simulation, or even consider new volume metrics [#cross]_.
+#
+# The heavy outputs can be retrieved from a classically-obtained probability
+# distribution as follows:
+#
+
+def heavy_output_set(m, probs):
```
josh146, line 387: Might be good to specify what `m` is here. Would also be helpful to know what `probs` is (e.g., a list? an array? a dictionary like at the very top?)

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+    # Sort the probabilities
+    probs_ascending_order = np.argsort(probs)
+    sorted_probs = probs[probs_ascending_order]
+
+    median_prob = np.median(sorted_probs)
+
+    # Heavy outputs are the bit strings above the median
+    heavy_outputs = [
+        format(x, f"#0{m+2}b")[2:] for x in list(probs_ascending_order[2 ** (m - 1) :])
+    ]
```
josh146, line 397: It's a bit hard to parse this line 🤔 

Shouldn't `median_prob` be used somewhere here?

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+    # Probability of a heavy output
+    prob_heavy_output = np.sum(sorted_probs[np.where(sorted_probs > median_prob)])
```
josh146, line 400: I think the `np.where` is superfluous here

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+
+##############################################################################
+#
+# As an example, let's compute the heavy outputs and probability for our circuit
+# above.
+#
+
+# Adds a measurement of first m qubits to the previous circuit
+with tape:
+    qml.probs(wires=range(m))
```
josh146, line 413: I have to admit, I find it pretty cool you can keep working with the same tape

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# First, we can take a look at the arrangement of the qubits on the processor
+# by plotting its hardware graph.
+
+ourense_hardware_graph = nx.Graph(dev_ourense.backend.configuration().coupling_map)
+
+nx.draw_networkx(
+    ourense_hardware_graph,
+    node_color="cyan",
+    labels={x: x for x in range(dev_ourense.num_wires)},
+)
```
josh146, line 450: Could we have this plotted without the black border?

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+        with qml.tape.QuantumTape() as tape:
+            qml.templates.layer(qv_circuit_layer, m, num_qubits=m)
+            qml.probs(wires=range(m))
+
+        output_probs = tape.execute(dev_ideal).reshape(2 ** m, )
+        heavy_outputs, prob_heavy_output = heavy_output_set(m, output_probs)
+
+        # Execute circuit on the noisy device
+        tape.execute(dev_noisy)
+
```
josh146, line 518: I suppose, here, you _could_ use a QNode, since the tape construction and execution happens together, and only once.

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,732 @@
...
+# To actually perform the simulations, we'll need to access a copy of the
+# Ourense noise model. Again, we won't be running on Ourense directly - rather
+# we'll set up a local device to simulate its behaviour.
+#
+
+noise_model = noise.NoiseModel.from_backend(dev_ourense.backend.properties())
+
+dev_noisy = qml.device(
+    "qiskit.aer", wires=dev_ourense.num_wires, shots=1000, noise_model=noise_model
+)
```
josh146, line 466: This is awesome - if not already documented, it might be nice to add to the PL-Qiskit docs!

Thinking a bit more, perhaps we should even have demo thumbnails for all demos that use Qiskit on the PL-Qiskit docs?

josh146 (commented):
This is an awesome demo @glassnotes 🧊

demonstrations/tutorial_quantum_volume.py
glassnotes: I've gone back and forth with this a lot with the intention of trying to make the intro more compact. However I agree it should be separated into more paragraphs. While the cost is that it looks slightly longer, I think it also makes it more digestible.

demonstrations/tutorial_quantum_volume.py
glassnotes: Good call!

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
glassnotes: I can work on adding this to the development branch. However I also don't want to delay the release of the demo for that, so I'll see how quickly I can get this done. Right now this function actually performs the inverse permutation of what is specified, which doesn't matter in the context of doing random permutations, but does matter if the user wants to specify a permutation explicitly.

demonstrations/tutorial_quantum_volume.py
glassnotes: It is possible to do it for like a general parameterized unitary by selecting the parameters from their distributions in the Haar measure (see. eqs 16-17 [here](https://arxiv.org/pdf/1708.00735.pdf) for SU(4)). But this is more for application in the qumode case, and you'd still have to decompose into qubit operations. I'm not sure how to do it using the Pauli words explicitly, would have to look into that. (tagging @nquesada, maybe he knows?)

demonstrations/tutorial_quantum_volume.py
glassnotes: What does that emoji mean? 

(Also is there a way to get the thing to print all wires in order without resorting to this?)

demonstrations/tutorial_quantum_volume.py
glassnotes: I'm indifferent. It's kinda ugly, but it also gives a more complete picture of the random circuit.

demonstrations/tutorial_quantum_volume.py
glassnotes: This whole method can definitely be simplified for a few reasons you point out, I will try to streamline it :+1: 

The sorting is used to get the explicit heavy output basis states, and this line here is turning them from integers in the sorted probs array to the binary strings that we need for later. Do you know of a cleaner way to convert the ints to properly-padded binary strings?

demonstrations/tutorial_quantum_volume.py
glassnotes: Good idea :)

demonstrations/tutorial_quantum_volume.py
glassnotes: I got this to work, but there is some additional output I don't know know how to suppress - any idea how to remove that? 

image

demonstrations/tutorial_quantum_volume.py
josh146: I find more and shorter paragraphs generally better for web content (news articles -- sometimes paragraphs can be 1-2 sentences long! -- blogs, tutorials) and longer paragraphs better suited to 'heftier' mediums (papers, manuscripts, books)

demonstrations/tutorial_quantum_volume.py
josh146: oh, weird 🤔

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
josh146: I agree. Demos are good place for working out what is missing/needs to be added to PennyLane. But we can always update the demo later once the feature is added!

This also avoids there being pressure to get the feature in, which can tempt us to cut corners e.g., in code review.

demonstrations/tutorial_quantum_volume.py
josh146: Oh, it was a subtle joke at the long discussion over the alternative name for this function (`qml.repeat`). Although I realize it was maybe too subtle now!

> Also is there a way to get the thing to print all wires in order without resorting to this?

I'm not sure I follow?

demonstrations/tutorial_quantum_volume.py
josh146: Unfortunately, this is the only way I know of doing it 🤔 

```python
>>> p = probs_ascending_order[2 ** (m - 1) :]
>>> bin_array = np.int64((p[:, None] & (1 << np.arange(4))) > 0)[:, ::-1]
>>> heavy_outputs = ["".join(row.astype(np.str)) for row in bin_array]
```

which is even _more_ obfuscated/esoteric.

demonstrations/tutorial_quantum_volume.py
glassnotes: So since the permutations are generated randomly, it is sometimes the case that, e.g. wires 0 and 2 are permuted first, and wire 1 doesn't enter the picture until later. When this happens, the wires in the tape drawing are ordered [0, 2, 1] instead of [0, 1, 2]. Also if wires 3 and 4 aren't used, they also don't show up in the drawing. So to get around this I've been "doing nothing" to the qubits in order.

josh146: Oh right, hence the `qml.RZ(0, wires=qubit)`.

This sounds like a bug in the PL circuit drawer, maybe we should make an issue 🤔

demonstrations/tutorial_quantum_volume.py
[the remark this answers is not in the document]
glassnotes: Awesome, let's leave it here for now and I'll work on the proper implementation afterwards.

demonstrations/tutorial_quantum_volume.py
glassnotes: I don't think it's necessarily a bug, I think it might just be a feature of tape mode because you don't have to specify a device right away, and you can also label the wires however you like. So if you write a circuit like:

```
with qml.tape.QuantumTape() as tape:
    qml.RX(0.3, wires=2)
    qml.RY(0.5, wires="a")
    qml.RZ(0.2, wires=0)
```
the drawer gives this output:
```
 2: ──RX(0.3)──┤  
 a: ──RY(0.5)──┤  
 0: ──RZ(0.2)──┤  
```
and just orders the wires as it receives them.

josh146: Good point, I have a bad habit of forgetting that non-integer wire labels are allowed. Hmm 🤔 I can't think of another more elegant solution

josh146: Maybe a better solution would be to have draw accept the wire order?
```python
>>> tape.draw(wire_order=[0, 1, 2, 'a', 'other', 0.65])
```

glassnotes: Yeah that would work! I think the most common use case is integers, so maybe we could have that as the default? (So like if no order is passed, just like check if all labels were integers, and then sort them)

demonstrations/tutorial_quantum_volume.py
glassnotes: Made an [issue](https://github.com/PennyLaneAI/pennylane-qiskit/issues/119) for this, will tackle it once this demo is live.

josh146: Awesome! 💯

demonstrations/tutorial_quantum_volume.py
```diff
@@ -176,16 +192,19 @@
...
-# distribution*.) We can integrate this distribution to find that the median
-# sits at :math:`\ln 2`.  We can further compute the expectation value of
-# obtaining something greater than the median by integrating the distribution
-# from :math:`\ln 2` to infinity, to obtain :math:`(1 + \ln 2)/2`. This is the
-# expected heavy output probability! Numerically it is around 0.85, and in fact
-# we will observe this later on.
+# :math:`1/2^m`, where :math:`m` is the number of qubits. Taking the modulus
+# squared of such numbers and making a histogram of them yields a distribution
+# of probabilities with the form :math:`\hbox{Pr}(p) \sim 2^m e^{-2^m p}.` This
+# is also known as the *Porter-Thomas distribution*.
```
josh146, line 198: Oh! The Porter-Thomas distribution is also discussed in the just-merged qsim demo: https://pennylane.ai/qml/demos/qsim_supremacy.html

Should we link to it?

demonstrations/tutorial_quantum_volume.py
```diff
@@ -385,19 +407,22 @@ def qv_circuit_layer(num_qubits):
 #
 
 def heavy_output_set(m, probs):
-    # Sort the probabilities
+    # Compute heavy outputs of an m-qubit circuit with measurement outcome
+    # probabilities given by probs. probs is an array with the probabilities
+    # ordered as '000', '001', ... '111'.
```
josh146, line 412: This makes the function much easier to interpret 💪

demonstrations/tutorial_quantum_volume.py
```diff
@@ -385,19 +407,22 @@ def qv_circuit_layer(num_qubits):
...
+    # Sort the probabilities so that those above the median are in the second half
     probs_ascending_order = np.argsort(probs)
     sorted_probs = probs[probs_ascending_order]
 
-    median_prob = np.median(sorted_probs)
-
     # Heavy outputs are the bit strings above the median
     heavy_outputs = [
+        # Convert integer indices to m-bit binary strings
         format(x, f"#0{m+2}b")[2:] for x in list(probs_ascending_order[2 ** (m - 1) :])
```
josh146: Funnily enough, just adding this comment makes it more accessible to me. Previously, I had a feeling that this line was in general doing some binary conversion, but I felt the need to interpret it statement-by-statement to be sure.

With the comment, I no longer feel the need to interpret it statement-by-statement.

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,764 @@
...
+#
+# .. note::
+#
+#    In many sources, the quantum volume of processors is reported as
+#    :math:`V_Q` explicitly, rather than :math:`\log_2 V_Q` as is the
+#    convention in this demo. As such, IonQ's processor has the potential for a
+#    quantum volume of :math:`2^{32} > 4000000`. Here we use the :math:`\log`
+#    because it is more straightforward to understand that they have 32
+#    high-quality, well-connected qubits than to extract this at first glance from the
+#    explicit value of the volume.
```
josh146, line 264: I've seen this definition referred to as 'quantum logume' on twitter

josh146: https://twitter.com/CraigGidney/status/1311791052999847937

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,764 @@
...
+# in order.
+
+num_qubits = 5
+dev_ideal = qml.device("default.qubit", analytic=True, wires=num_qubits)
+
+m = 3  # number of qubits
+
+with qml.tape.QuantumTape() as tape:
+    for qubit in range(num_qubits):
+        qml.RZ(0, wires=qubit)
```
josh146, line 378: In the meantime, might be good to have a comment here briefly explaining why identities are being added to the circuit?

demonstrations/tutorial_quantum_volume.py
```diff
@@ -425,22 +457,22 @@ def heavy_output_set(m, probs):
 # ~~~~~~~~~~~~~~~~~~~~~~~~
 #
 # Now it's time to run the protocol. First, let's set up our hardware
-# device. We'll use a simulated version of the 5-qubit IBM Ourense as a example
+# device. We'll use a simulated version of the 5-qubit IBM Ourense as an example
 # - the reported quantum volume according to IBM is :math:`V_Q=8`, so we
```
josh146, line 461:
```suggestion
# --- the reported quantum volume according to IBM is :math:`V_Q=8`, so we
```

josh146 (approved):
Looks and reads really great to me @glassnotes!

My only remaining (and relatively minor) concern was whether we want to publicly use the tape in demonstrations. Mostly just due to how new the tape is.

But I think I've come around to showcasing the tape! Since

1. I agree that there are advantages to using it here rather than a QNode (t o avoid randomness issues)

2. To an extent, it also makes the code cleaner since you use the same tape throughout, adding to it and executing it on additional devices.

3. We've been advertising the tape anyway e.g., on twitter.

demos_basics.rst
```diff
@@ -156,6 +150,12 @@ hardware.
     :description: :doc:`demos/tutorial_qnn_module_torch`
     :tags: pytorch
 
+.. customgalleryitem::
+   :tooltip: Learn how to compute the quantum volume of a quantum processor.
+   :figure: demonstrations/quantum_volume/quantum_volume_thumbnail.png
+   :description: :doc:`demos/tutorial_quantum_volume`
+   :tags: characterization qiskit
```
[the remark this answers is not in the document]
glassnotes: I didn't put it on the research page because it is currently divided into "Optimization" and "QML", and this one doesn't really fit into either category. We could make a new category?

demonstrations/tutorial_quantum_volume.py
glassnotes: Ah! We should definitely put this one in the "related demos" in the side bar. Also, @thisac are you willing to give this a quick review since you've worked recently with similar math?

demos_basics.rst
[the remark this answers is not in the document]
co9olguy: Yes we should. The qsim demo is also out of place on the "Optimization" section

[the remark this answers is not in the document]
co9olguy: For now, let's name it something like "Quantum Computing". Slightly longer-term, we can make a separate subpage for this

demonstrations/tutorial_quantum_volume.py
```diff
@@ -10,27 +10,32 @@
...
+A similar problem is emerging with quantum computers: we can't judge quantum
+computers on the number of qubits alone. Present-day devices have a number of
+limitations, an important one being the fairly high gate error rates. Typically
+the qubits on a chip are not all connected to each other, so it may not be
+possible to perform operations on arbitrary pairs of them.
+
+Considering this, can we tell if a machine with 20 noisy qubits is better
 than one with 5 very high-quality qubits? Or if a machine with 8 fully-connected
 qubits is better than one with 16 qubits of comparable error rate, but arranged in
 a square lattice?  Furthermore, how can we make comparisons between different
```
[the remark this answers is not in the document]
glassnotes: I gave that a go and it weirdly adds too much space (even when I remove the padding I added). Will keep as is for now.

demonstrations/tutorial_quantum_volume.py
glassnotes: Good idea - added a link to tape mode as well and mentioned it's a new feature.

demonstrations/tutorial_quantum_volume.py
thisac: Sure, I'll have a look at it now!

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,773 @@
...
+with up to 260 cores. The speed of processors will differ, and they may be
+connected in different ways. We can't rank them by simply counting the number of
+processors!
+
+In order to make a fair comparison, we need benchmarking standards that give us
+a holistic view of their performance. To that end, the TOP500 rankings are based
+on something called the LINPACK benchmark [#linpack]_. The task of the
+supercomputers is to solve a dense system of linear equations, and the metric of
+interest is the rate at which they the metric of interest is the rate at which
+they perform `floating-point operations (FLOPS)
```
thisac, line 26:
```suggestion
supercomputers is to solve a dense system of linear equations, and the metric of
interest is the rate at which they perform `floating-point operations (FLOPS)
```

demonstrations/tutorial_quantum_volume.py
```diff
@@ -0,0 +1,773 @@
...
+<https://en.wikipedia.org/wiki/FLOPS>`__. Today's top machines reach speeds well
+into the regime of hundreds of peta FLOPs! While a single number certainly
```
thisac, line 28:
```suggestion
into the regime of hundreds of petaFLOPS! While a single number certainly
```
:thinking: ?

demonstrations/tutorial_quantum_volume.py
glassnotes: Now that the demo is non-executable, I was able to put in a static image without the border :+1:

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+supercomputers is to solve a dense system of linear equations, and the metric of
+interest is the rate at which they perform `floating-point operations (FLOPS)
+<https://en.wikipedia.org/wiki/FLOPS>`__. Today's top machines reach speeds well
+into the regime of hundreds of petaFLOPS! While a single number certainly
+cannot tell the whole story, it still gives us insight into the quality of the
+machines, and provides a standard so we can compare them.
+
+A similar problem is emerging with quantum computers: we can't judge quantum
+computers on the number of qubits alone. Present-day devices have a number of
+limitations, an important one being the fairly high gate error rates. Typically
```
ixfoduap, line 37:
```suggestion
limitations, an important one being gate error rates. Typically
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+the qubits on a chip are not all connected to each other, so it may not be
+possible to perform operations on arbitrary pairs of them.
+
+Considering this, can we tell if a machine with 20 noisy qubits is better
+than one with 5 very high-quality qubits? Or if a machine with 8 fully-connected
+qubits is better than one with 16 qubits of comparable error rate, but arranged in
+a square lattice?  Furthermore, how can we make comparisons between different
```
ixfoduap, line 44:
```suggestion
a square lattice?  How can we make comparisons between different
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+#     Heavy output probability = 0.792
+#
+
+##############################################################################
+#
+# Each circuit in a circuit family has its own heavy output probability. If our
+# quantum computer is of high quality, then we should expect to see heavy
+# outputs quite often across all the circuits. On the other hand, if it's of
+# poor quality and everything is totally decohered, we will end up with output
+# probabilities that are roughly all the same, as decoherence will reduce the
```
ixfoduap, line 195:
```suggestion
# probabilities that are roughly all the same, as noise will reduce the
```
to avoid repeated use of decoherence

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+# square circuits scramble things up enough so that the effective operation
+# looks like a Haar-random unitary :math:`U`. Since in the circuits we are
+# applying :math:`U` to the all-zero ket, the measurement outcome probabilities
+# will be the moduli squared of the entries in the first column of :math:`U`.
+#
+# Now if :math:`U` is Haar-random, we can say something about the form of these
+# entries. In particular, they are complex numbers for which both the real and
+# imaginary parts are normally distributed with mean 0 and variance
+# :math:`1/2^m`, where :math:`m` is the number of qubits. Taking the modulus
+# squared of such numbers and making a histogram of them yields a distribution
```
ixfoduap, line 219:
```suggestion
# squared of such numbers and making a histogram yields a distribution
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+# is also known as the *Porter-Thomas distribution*.
+#
+# By looking at the form of the underlying probability distribution, the
+# exponential distribution :math:`\hbox{Pr}(x) = e^{-x}`, we can calculate some
+# properties of the heavy output probabilities. First, we can integrate the exponential
+# distribution to find that the median sits at :math:`\ln 2`.  We can further
+# compute the expectation value of obtaining something greater than the median
+# by integrating :math:`x e^{-x}` from :math:`\ln 2` to :math:`\infty` to obtain
+# :math:`(1 + \ln 2)/2`. This is the expected heavy output probability!
+# Numerically it is around 0.85, and in fact we will observe this later on in our results.
```
ixfoduap, line 230:
```suggestion
# Numerically it is around 0.85, as we will observe later in our results.
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+# <https://pennylaneqiskit.readthedocs.io/en/latest/>`_ plugin to compute the
+# volume of one of the IBM processors, since their properties are easily
+# accessible through this plugin.
+#
+#
+# Loosely, the protocol for quantum volume consists of three steps:
+#
+# 1. Construct random square circuits of increasing size
+#
+# 2. Run those circuits on both a simulator, and on a noisy hardware device
```
ixfoduap, line 307:
```suggestion
# 2. Run those circuits on both a simulator and on a noisy hardware device
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+
+
+##############################################################################
+#
+# Next, we need to apply SU(4) gates to pairs of qubits. PennyLane doesn't have
+# built-in functionality to generate these random matrices, however its cousin
+# `Strawberry Fields <https://strawberryfields.ai/>`_ does! We will use the
+# ``random_interferometer`` method, which can generate unitary matrices uniformly
+# at random. (This function actually generates elements of U(4), but they are
+# essentially equivalent up to a global phase).
```
ixfoduap, line 361:
```suggestion
# at random. This function actually generates elements of U(4), but they are
# essentially equivalent up to a global phase.
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+# :math:`\approx 10^{-4}`, after which we may need to make adjustments to remove
+# the classical simulation, or even consider new volume metrics [#cross]_.
+#
+# The heavy outputs can be retrieved from a classically-obtained probability
+# distribution as follows:
+#
+
+def heavy_output_set(m, probs):
+    # Compute heavy outputs of an m-qubit circuit with measurement outcome
+    # probabilities given by probs. probs is an array with the probabilities
```
ixfoduap, line 484:
```suggestion
    # probabilities given by probs, which is an array with the probabilities
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+    return heavy_outputs, prob_heavy_output
+
+
+##############################################################################
+#
+# As an example, let's compute the heavy outputs and probability for our circuit
+# above.
+#
+
+# Adds a measurement of first m qubits to the previous circuit
```
ixfoduap, line 509:
```suggestion
# Adds a measurement of the first m qubits to the previous circuit
```

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+with tape:
+    qml.probs(wires=range(m))
+
+# Run the circuit, compute heavy outputs, and print results
+output_probs = tape.execute(dev_ideal).reshape(2 ** m, )
+heavy_outputs, prob_heavy_output = heavy_output_set(m, output_probs)
+
+print("State\tProbability")
```
ixfoduap, line 517: Thanks Olivia for inadvertently convincing me to start using f-strings!

demonstrations/quantum_volume.py
```diff
@@ -0,0 +1,908 @@
...
+# :math:`2\sigma` of the mean.)
+#
+# At this point, we're going to do some statistical sorcery and make some
+# assumptions about our distributions. Whether or not a circuit is successful
+# (in the sense that it produces heavy outputs more the 2/3 of the time) is a
+# binary outcome. When we sample many circuits, it is almost like we are
+# sampling from a *binomial distribution* where the outcome probability is
+# equivalent to the heavy output probability. In the limit of a large number of
+# samples (in this case 200 circuits), a binomial distribution starts to look
+# pretty normal. If we make this approximation, we can compute the standard
```
ixfoduap, line 733:
```suggestion
# like a normal distribution. If we make this approximation, we can compute the standard
```

ixfoduap (approved):
This is one of the best PennyLane tutorials out there. Amazing work! :star_struck: :rocket: :guitar: 

All my comments are nit picks and completely optional

conf.py
```diff
@@ -89,6 +89,9 @@
 warnings.filterwarnings(
     "ignore", category=UserWarning, message=r"Matplotlib is currently using agg"
 )
+warnings.filterwarnings(
+    "ignore", category=UserWarning, message=r"Timestamps in IBMQ backend"
+)
```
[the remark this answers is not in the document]
glassnotes: Since the demo isn't executable anymore this is no longer needed - however I'm going to leave it because it may be useful to have for future demos that use PennyLane-Qiskit.

PR #166, opened 2020-11-24
Update requirements to have pyscf==1.7.2 and no scipy pinning

antalszava:
**Summary:**
Following up on https://github.com/PennyLaneAI/pennylane/pull/926 in PennyLane-QChem, we would like to use similar requirements for the `pyscf` and the `scipy` packages.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ pillow==6.2.0
 pip==20.1
 pyscf==1.7.3
 git+https://github.com/PennyLaneAI/pennylane.git
-git+https://github.com/PennyLaneAI/pennylane.git@pin_down_pyscf_scipy_qchem#subdirectory=qchem
+git+https://github.com/PennyLaneAI/pennylane.git#subdirectory=qchem
 git+https://github.com/PennyLaneAI/pennylane-sf.git
 git+https://github.com/rigetti/pennylane-forest.git
 git+https://github.com/PennyLaneAI/pennylane-cirq.git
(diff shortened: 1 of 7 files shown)


josh146:
Thanks @antalszava! If all demos pass (with the exception of `tutorial_vqe_parallel`, which we know has a bug), then feel free to close this PR, and update the other PR in the PL repo with the result.

antalszava:
Added the link to the other PR and reverted incrementing the version numbers for the cache names in the build.

Would we still like to change the requirements of `scipy` and `pyscf` here? Removed the `scipy` pinning and pinned  `pyscf==1.7.3` (would be pulled as latest for `pyscf<1.7.4`) for now (can close though if I missed something).

josh146:
> Would we still like to change the requirements of scipy and pyscf here?

If everything passes, we might as well. I don't see a drawback to incrementing versions here if we know it doesn't cause any issues.

requirements.txt
```diff
@@ -10,7 +10,7 @@ strawberryfields==0.15
 sympy==1.6.2
 pillow==6.2.0
 pip==20.1
-pyscf==1.7.1
+pyscf==1.7.3
```
antalszava, line 13:
```suggestion
pyscf==1.7.2
```

antalszava:
The latest failed check contained the known problem with `tutorial_vqe_parallel`:
https://app.circleci.com/pipelines/github/PennyLaneAI/qml/1947/workflows/aedf67f3-7718-47a2-b9ef-78b560d57ec1/jobs/2018

Reverting the caching numbers.

PR #174, opened 2020-12-01
Update qsim demo + title

thisac:
I realised that I somewhat incorrectly stated that the distribution that the circuit followed is a Porter-Thomas distribution while, in fact, it only converges to such a distribution for many qubits. Clarified this section a bit.

Also updated the title to "Beyond classical computing with qsim", including in metadata.


demonstrations/qsim_supremacy.py
```diff
@@ -341,17 +341,18 @@ def circuit(seed=42, return_probs=False):
...
 #
 # .. math::
 #
 #    Pr(p) = (N - 1)(1- p)^{N-2},
 #
-# where :math:`N = 2^n` is the number of possible bitstrings.
-# From this we can then calculate the expectation value
+# where :math:`N = 2^n` is the number of possible bitstrings [#Boixo2018]_.
+# This distribution is approximated well by the Porter-Thomas distribution,
+# given by :math:`Pr(p) = Ne^{-Np}`; a characteristic of chaotic quantum
```
glassnotes, line 354:
```suggestion
# given by :math:`Pr(p) = Ne^{-Np}`, a characteristic of chaotic quantum
```

demonstrations/qsim_supremacy.py
```diff
@@ -2,11 +2,11 @@
...
 .. _quantum_neural_net:
 
-Quantum supremacy using qsim
-============================
+Beyond classical computing with qsim
+====================================
 
 .. meta::
-    :property="og:description": Discover quantum supremacy using Google's qsim simulator, and recreate their benchmarks and circuits.
+    :property="og:description": Use Google's qsim simulator to go beyond classical computing, and recreate their benchmarks and circuits.
```
co9olguy, line 9:
```suggestion
    :property="og:description": Use Google's qsim simulator to explore the barriers between quantum and classical computing, and recreate their benchmarks and circuits.
```
still think it sounds weird to use a classical computer program to go beyond classical computing :stuck_out_tongue_closed_eyes:

demos_research.rst
```diff
@@ -124,7 +124,7 @@ plateaus in quantum neural networks.
 :html:`<div class="gallery-grid row">`
 
 .. customgalleryitem::
-    :tooltip: Quantum supremacy with qsim.
+    :tooltip: Beyond classical computing with qsim.
     :figure: demonstrations/qsim_supremacy/sycamore.png
     :description: :doc:`demos/qsim_supremacy`
```
glassnotes, line 129: Title of the files/directories should also be changed to match

demonstrations/qsim_supremacy.py
co9olguy, line 354:
```suggestion
# given by :math:`Pr(p) = Ne^{-Np}`, a characteristic property of chaotic quantum
```

co9olguy (approved):

PR #179, opened 2020-12-05
Corrects a typo in the definition of hafnian

nquesada:
Minor typo is fixed confusing perfect matching permutations with the set of all permutations.

The hafnian can be defined in two equivalent ways:

image

In the current version in `master` we are using the prefactor from the first definition but the summation from the second.

### DIFF
--- a/demonstrations/tutorial_gbs.py
+++ b/demonstrations/tutorial_gbs.py
@@ -226,9 +226,9 @@ for i in measure_states:
 #
 #     The hafnian of a matrix is defined by
 #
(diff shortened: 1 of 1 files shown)


co9olguy:
Other than the typo where an `=` should be an `\in`, I believe the equation as-is, is correct (according to the first equation in the image you posted), but the text should say "set of all permutations", correct? This is what was written previously, but the change was indicated by @ixfoduap during code review

josh146:
@co9olguy @nquesada would the best fix simply to be to revert to saying ‘set of all permutations’ in the text?

co9olguy:
That's my suggestion as well, but I'd like @ixfoduap to give his thoughts, since he requested the change originally

ixfoduap (approved):

PR #182, opened 2020-12-10
Adds tutorial on measurement optimization and grouping

josh146:
- [x] Revisit VQE and describe the problem
- [x] Discuss simultaneous measurement of commuting observables
- [x] Describe qubit-wise commuting Pauli words, and show how they can be simultaneously measured using PennyLane.
- [x] Introduce qml.grouping, and also show that PL QNodes have built-in ability to measure QWC observables
- [x] Explain the problem of grouping observables into QWC groups (min-clique graph image)
- [x] Show an example using the grouping module, and mention how it is built in to `ExpvalCost`
- [x] Conclusion: mention other approaches (unitary partitioning, full commutation), and also that this is useful beyond just VQE.


demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#         |\langle \phi_n|\psi\rangle|^2,\\
+#     \langle\hat{B}\rangle &= \langle \psi | \hat{B} | \psi \rangle = \langle \psi | \left( \sum_n
+#         \lambda_{B, n} |\phi_n\rangle\langle \phi_n| \right) | \psi \rangle = \sum_n \lambda_{B,n}
+#         |\langle \phi_n|\psi\rangle|^2.
+#
+# So, assuming we know the eigenvalues of the commuting observables in advance, if we perform a
+# measurement in their shared eigenbasis, we only need to perform a **single measurement** of the
+# probabilities :math:`|\langle \phi_n|\psi\rangle|^2` in order to recover both expectation values! 😍
+#
+# Fantastic! But, can we use this to reduce the number of measurements we need to perform in VQE?
```
glassnotes, line 268:
```suggestion
# Fantastic! But, can we use this to reduce the number of measurements we need to perform in the VQE?

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#   observables.
+#
+# ..
+#
+# - If :math:`\hat{A}` and :math:`\hat{B}` **do** commute (:math:`[\hat{A},
+#   \hat{B}] = 0`), then :math:`\sigma_A^2
+#   \sigma_B^2 \geq 0` and there exists a measurement basis where we can **simultaneously measure** the
+#   expectation value of both observables on the same state.
+#
+# .. admonition:: Aside: commutativity and shared eigenbases
```
glassnotes, line 211: I thought this section was quite important for understanding the previous statements; I suggest to include it in the main text rather than in an aside.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# some recent techniques we can harness to address both.
+
+##############################################################################
+# Qubit-wise commuting Pauli terms
+# --------------------------------
+#
+# Back when we summarized VQE, we saw that each term of the Hamiltonian is generally represented
+# as a tensor product of Pauli terms:
+#
+# .. math:: h_i = \prod_{n=0}^{N} \sigma_n.
```
glassnotes, line 285:
```suggestion
# .. math:: h_i = \bigotimes_{n=0}^{N} \sigma_n.

```
(I think that will give the right symbol?)

glassnotes: Also double-check the indices on the product here and wherever else it appears; I think as written this would give $N+1$ qubits because the top of the product is used inclusively (as opposed to in Python where it goes only to N - 1).

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# Luckily, this allows us to take a bit of a shortcut. Rather than consider **full commutativity**,
```
glassnotes, line 287: The transition here seems abrupt, emphasize that it is the tensor product structure that allows for the shortcut.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# we can consider a subset known as **qubit-wise commutativity** (QWC).
```
glassnotes, line 288:
```suggestion
# we can consider a slightly less strict condition known as **qubit-wise commutativity** (QWC).

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# To start with, let's consider single Pauli operators. We know that the Pauli operators
```
glassnotes, line 290:
```suggestion
# To start with, let's consider single-qubit Pauli operators. We know that the Pauli operators

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# commute with themselves as well as the identity, but they do *not* commute with
+# each other:
+#
+# .. math::
+#
+#     [\sigma_i, I] = 0, ~~~ [\sigma_i, \sigma_i] = 0, ~~~ [\sigma_i, \sigma_j] = c \sigma_k \delta_{ij}.
```
glassnotes, line 296: There is a clash of notation here and with the expression of the $h_i$as tensor products of Paulis. In the previous expressions, the subscript $n$ is used to index the qubits, but here it is being used to indicate $\sigma_x, \sigma_y, \sigma_z$.  It might be better here to express the commutation relations in terms of $X, Y, Z$ to fix this, and tie better into the next example.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# .. math::
+#
+#     \begin{array}{ | *1c | *1c | *1c | *1c | *1c |}
+#       X &\otimes &Y &\otimes &I\\
+#       X &\otimes &I &\otimes &Z
+#     \end{array} ~~~~ \Rightarrow ~~~~ [X, X] = 0, ~~~ [Y, I] = 0, ~~~ [I, Z] = 0.
+#
+# As a consequence, both terms must commute:
+#
+# .. math:: [X\otimes Y \otimes I, X\otimes I \otimes Z] = 0.
```
glassnotes, line 311: Awesome, this is a really clear example! The vertical barriers are very helpful for delineating between the subsystems.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# .. important::
+#
+#     Qubit-wise commutativity is a **sufficient** but not **necessary** condition
+#     for full commutativity. For example, the two Pauli terms :math:`Y\otimes Y` and
+#     :math:`X\otimes X` are not qubit-wise commuting, but do commute (have a go verifying this!).
+#
+# Once we have identified a qubit-wise commuting pair of Pauli terms, it is also straightforward to
+# find the gates to rotate the circuit into the shared eigenbasis. To do so, we simply rotate
+# each wire one-by-one depending on the Pauli operator we are measuring on that wire:
```
glassnotes, line 321:
```suggestion
# each qubit one-by-one depending on the Pauli operator we are measuring on that wire:

```
(?)

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# .. raw:: html
+#
+#     </div>
+#
+# Therefore, in this particular example:
+#
+# * Wire 0: we are measuring both terms in the :math:`X` basis, apply the Hadamard gate
+# * Wire 1: we are measuring both terms in the :math:`Y` basis, apply the :math:`H S^{-1}` gates
+# * Wire 2: we are measuring both terms in the :math:`Z` basis, no gate needs to be applied.
```
glassnotes, line 360:
```suggestion
# * Wire 2: we are measuring both terms in the :math:`Z` basis (the computational basis), no gate needs to be applied.

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+
+
+obs = [
+    qml.PauliX(0) @ qml.PauliY(1),
+    qml.PauliX(0) @ qml.PauliZ(2)
+]
+
+
+##############################################################################
+# First, lets naively use two separate circuit evaluations to measure
```
glassnotes, line 372:
```suggestion
# First, let's naively use two separate circuit evaluations to measure

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+    return qml.probs(wires=range(3))
+
+
+rotated_probs = circuit_qwc(weights)
+print(rotated_probs)
+
+
+##############################################################################
+# We're not quite there yet; we have only calculated the probabilities of the variational circuit
+# rotated into the shared eigenbasis; :math:`|\langle \phi_n |\psi\rangle|^2`. To recover the
```
glassnotes, line 422:
```suggestion
# rotated into the shared eigenbasis --- the :math:`|\langle \phi_n |\psi\rangle|^2`. To recover the

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# *expectation values* of the two QWC observables from the probabilities, recall that we need one
+# final piece of information; their eigenvalues :math:`\lambda_{A, n}` and :math:`\lambda_{B, n}`.
```
glassnotes, line 424:
```suggestion
# final piece of information: their eigenvalues :math:`\lambda_{A, n}` and :math:`\lambda_{B, n}`.

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# We know that the Pauli operators have eigenvalues :math:`(1, -1)`, while the identity
+# operator has eigenvalues :math:`(1, 1)`; we can make use of ``np.kron`` to quickly
+# generate the probabilities of the full Pauli terms.
+
+eigenvalues_XYI = np.kron(np.kron([1, -1], [1, -1]), [1, 1])
+eigenvalues_XIZ = np.kron(np.kron([1, -1], [1, 1]), [1, -1])
+
+# Taking the linear combination of the eigenvalues and the probabilities
+print("Expectation value of XYI = ", np.dot(eigenvalues_XYI, rotated_probs))
```
glassnotes, line 434: Sign differs from previous result.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# We're not quite there yet; we have only calculated the probabilities of the variational circuit
+# rotated into the shared eigenbasis; :math:`|\langle \phi_n |\psi\rangle|^2`. To recover the
+# *expectation values* of the two QWC observables from the probabilities, recall that we need one
+# final piece of information; their eigenvalues :math:`\lambda_{A, n}` and :math:`\lambda_{B, n}`.
+#
+# We know that the Pauli operators have eigenvalues :math:`(1, -1)`, while the identity
+# operator has eigenvalues :math:`(1, 1)`; we can make use of ``np.kron`` to quickly
+# generate the probabilities of the full Pauli terms.
+
+eigenvalues_XYI = np.kron(np.kron([1, -1], [1, -1]), [1, 1])
```
glassnotes, line 430: How do we know the eigenvalues generated in this way will be in the correct order?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+print(rotated_probs)
+
+
+##############################################################################
+# We're not quite there yet; we have only calculated the probabilities of the variational circuit
+# rotated into the shared eigenbasis; :math:`|\langle \phi_n |\psi\rangle|^2`. To recover the
+# *expectation values* of the two QWC observables from the probabilities, recall that we need one
+# final piece of information; their eigenvalues :math:`\lambda_{A, n}` and :math:`\lambda_{B, n}`.
+#
+# We know that the Pauli operators have eigenvalues :math:`(1, -1)`, while the identity
```
glassnotes, line 426:
```suggestion
# We know that the single-qubit Pauli operators each have eigenvalues :math:`(1, -1)`, while the identity

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# operator has eigenvalues :math:`(1, 1)`; we can make use of ``np.kron`` to quickly
```
glassnotes, line 427:
```suggestion
# operator has eigenvalues :math:`(1, 1)`. We can make use of ``np.kron`` to quickly

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# Taking the linear combination of the eigenvalues and the probabilities
+print("Expectation value of XYI = ", np.dot(eigenvalues_XYI, rotated_probs))
+print("Expectation value of XIZ = ", np.dot(eigenvalues_XIZ, rotated_probs))
+
+
+##############################################################################
+# Compare this to the result when we used two circuit evaluations. We have successfully used a
+# single circuit evaluation to recover both expectation values!
+#
+# Luckily, PennyLane automatically performs this QWC under-the-hood. We simply
```
glassnotes, line 442:
```suggestion
# Luckily, PennyLane automatically performs this QWC  grouping under-the-hood. We simply

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+##############################################################################
+# Behind the scenes, PennyLane is making use of our built-in
+# :mod:`qml.grouping <pennylane.grouping>` module, which contains functions for diagonalizing QWC
+# terms:
+
+rotations, new_obs = qml.grouping.diagonalize_qwc_pauli_words(obs)
+
+print(rotations)
+print(new_obs)
+
```
glassnotes, line 466: Remind the reader that the first line corresponds to the basis rotations that were discussed above; they're just re-expressed here as RX and RY (on that note, why is RY used on the first qubit here rather than H?)

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+
+##############################################################################
+# Check out the :mod:`qml.grouping <pennylane.grouping>` documentation for more details on its
+# provided functionality and how it works.
+#
+# What happens, though, if we (in a moment of reckless abandonment!) ask a QNode to simultaneously
```
glassnotes, line 472:
```suggestion
# What happens, though, if we (in a moment of reckless abandon!) ask a QNode to simultaneously

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# measure two observables that *aren't* qubit-wise commuting? For example, lets consider
```
glassnotes, line 473:
```suggestion
# measure two observables that *aren't* qubit-wise commuting? For example, let's consider
```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#         ]
+#
+# .. rst-class:: sphx-glr-script-out
+#
+#  Out:
+#
+#  .. code-block:: none
+#
+#     pennylane.qnodes.base.QuantumFunctionError: Only observables that are qubit-wise commuting
+#     Pauli words can be returned on the same wire
```
glassnotes, line 493: "words" has not been used to describe the Pauli strings in the main text, just in `diagonalize_qwc_pauli_words` and here, is it worth briefly mentioning what this means?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# The QNode has detected that the two observables are not qubit-wise commuting, and
```
glassnotes, line 495: Great example!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# has raised an error.
+#
+# So, a strategy begins to take shape: given a Hamiltonian containing a large number of Pauli terms,
+# can we somehow partition the terms into **fewest** number of QWC groups, to minimize the measurements
```
glassnotes, line 499:
```suggestion
# can we somehow partition the terms into **fewest** number of QWC groups to minimize the number of measurements

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#     :width: 70%
+#     :align: center
+#
+# In the above graph, every node represents an individual Pauli term of the Hamiltonian, with
+# edges connecting terms that are qubit-wise commuting. Groups of qubit-wise commuting terms are
+# represented as **complete subgraphs**. Straight away, we can make an observation:
+# there is no unique solution for partitioning the Hamiltonian into groups of qubit-wise commuting
+# terms! In fact, there are several solutions:
+#
+# .. figure:: /demonstrations/measurement_optimize/graph2.png
```
glassnotes, line 524: I really like this drawing!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#     :width: 90%
+#     :align: center
+#
+# Of course, of the potential solutions above, there is one that is more optimal than the others;
```
glassnotes, line 528:
```suggestion
# Of course, of the potential solutions above, there is one that is more optimal than the others ---

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# on the bottom left, we have partitioned the graph into *two* complete subgraphs, as opposed to the
+# other solutions that require three complete subgraphs. If we were to go with this solution,
+# we would be able to measure the expectation value of the Hamiltonian using two circuit evaluations.
+#
+# This problem---finding the minimum number of complete subgraphs of a graph---is actually quite well
+# known in graph theory, where it is referred to as the `minimum clique cover problem
+# <https://en.wikipedia.org/wiki/Clique_cover>`__ (with 'clique' being another term for a complete subgraph).
+#
+# Unfortunately, that's where our good fortune ends---the minimum clique cover is known to
```
glassnotes, line 537:
```suggestion
# Unfortunately, that's where our good fortune ends---the minimum clique cover problem is known to

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+#
+# Many of these heuristic approaches have roots in another graph problem known as `graph
+# colouring <https://en.wikipedia.org/wiki/Graph_coloring>`__; the assignment of colours to
+# the graph's vertices such that no adjacent vertices have the same colour. How is this related
+# to the minimum clique cover problem, though? If we take our QWC graph above, and generate the
+# `complement graph <https://en.wikipedia.org/wiki/Complement_graph>`__ by drawing edges
+# between all *non*-adjacent nodes,
+#
+# .. figure:: /demonstrations/measurement_optimize/graph3.png
+#     :width: 100%
```
glassnotes, line 554: This is picky, but the positioning of nodes and shape of the graph is different in the complement graph. Is there a problem-specific reason for this?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+num_groups = len(set(groups.values()))
+
+print("Minimum number of QWC groupings found:", num_groups)
+
+for i in range(num_groups):
+    print(f"\nGroup {i}:")
+
+    for term, group_id in groups.items():
+        if group_id == i:
+            print(term)
```
glassnotes, line 625: Is it a bit hard to parse the output here and relate it back to the Pauli terms. Can these be easily re-expressed as Pauli words like YXX, YYXX, or Y0 X2 X3, etc. instead, to make it more concise?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+    [terms[1], terms[3]],  # Z0 Z1, Z0 Z1 Z2 Z3
+    [terms[0], terms[4]],  # Z0, X2 X3
+    [terms[1], terms[4]],  # Z0 Z1, X2 X3
+    [terms[4], terms[5]],  # X2 X3, Y0 X2 X3
+    [terms[4], terms[6]],  # X2 X3, Y0 Y1 X2 X3
+    [terms[5], terms[6]],  # Y0 X2 X3, Y0 Y1 X2 X3
+])
+
+
+nx.draw(G)
```
glassnotes, line 602: It would be helpful to have the nodes labeled by their Pauli words - I'm not sure if that is straightforward to do in `nx`, though.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# 2. Construct the complement QWC graph.
+#
+# 3. Use a graph colouring heuristic algorithm to determine a graph colouring
+#    with a minimum number of colours. Each coloured vertex set corresponds to a
+#    qubit-wise commuting group of Hamiltonian terms.
+#
+# 4. Generate and evaluate the circuit ansatz (with additional rotations) per
+#    QWC grouping, extracting probability distributions.
+#
+# 5. Finally, post-process the probability distributions with the observable eigenvalues
```
glassnotes, line 646: Great summary!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+    return [qml.expval(o) for o in group]
+
+weights = qml.init.strong_ent_layers_normal(n_layers=3, n_wires=4)
+result = [circuit(weights, group=g) for g in obs_groupings]
+
+print("Term expectation values:")
+for group, expvals in enumerate(result):
+    print(f"Group {group} expectation values:", expvals)
+
+print("<H> = ", np.sum(np.hstack(result)))
```
glassnotes, line 687: What happened to the coefficients of the Pauli terms, do they get absorbed somehow into the expectation value calculations?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+
+H, num_qubits = qml.qchem.molecular_hamiltonian("h2o", "h2o.xyz")
+print("Number of Hamiltonian terms/required measurements:", len(H.ops))
+
+# grouping
+groups = qml.grouping.group_observables(H.ops, grouping_type='qwc', method='rlf')
+print("Number of required measurements after optimization:", len(groups))
+
+##############################################################################
+# We went from 2052 required measurements/circuit evaluations to 523 (just over *two thousand*
```
glassnotes, line 717:
```suggestion
# We went from 2050 required measurements/circuit evaluations to 523 (just over *two thousand*

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# the savings that can be made by considering *full* commutativity [#yen2020]_, unitary
+# partitioning [#izmaylov2019]_, and Fermionic basis rotation grouping [#huggins2019]_.
+# Work has also been performed to reduce the classical overhead associated with measurement
+# optimization, allowing the classical measurement grouping to be performed in linear time
+# [#gokhale2020]_. For example, recall that qubit-wise commutativity is only a subset of
+# full commutativity; if we consider full commutativity instead, we can further reduce the
+# number of groups required.
+#
+# Finally, it is worth pointing out that, as the field of variational quantum algorithms
+# grows, this problem of measurement optimization no longer just applies to VQE (the algorithm
```
glassnotes, line 733:
```suggestion
# grows, this problem of measurement optimization no longer just applies to the VQE (the algorithm

```

glassnotes (commented):
@josh146  this is an awesome tutorial! It walks through the process very clearly and all of the figures and examples contribute to understanding the content. (I was already familiar with the topic, but I understand it way better now having read this, especially the graph colouring part :star2:)

Comments are mostly typos / phrasing suggestions, but there are a few points where some additional clarification would be helpful.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+# representation, but the number of terms in the Hamiltonian scales like
+# :math:`\mathcal{O}(N^4)`! 😱😱😱
+#
+# We can mitigate this somewhat by choosing smaller `basis sets
+# <https://en.wikipedia.org/wiki/Basis_set_(chemistry)>`__ to represent the electronic structure
+# wavefunction, however this comes with an accuracy cost, and doesn't reduce the number of
+# measurements significantly enough to allow us to scale to classically intractable problems.
+#
+# .. figure:: /demonstrations/measurement_optimize/n4.png
+#     :width: 70%
```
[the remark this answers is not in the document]
josh146: Yep, oops! Thanks for catching this

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+
+The study of :doc:`variational quantum algorithms </glossary/variational_circuit>` was spearheaded
+by the introduction of the :doc:`variational quantum eigensolver <tutorial_vqe>` (VQE) algorithm in
+2014 [#peruzzo2014]_. While classical variational techniques have been known for decades to estimate
+the ground state energy of a molecule, VQE allowed this variational technique to be applied using
+quantum computers. Since then, the field of variational quantum algorithms has evolved
+significantly, with larger and more complex models being proposed (such as
+:doc:`quantum neural networks </demos/quantum_neural_net>`, :doc:`QGANs </demos/tutorial_QGAN>`, and
+:doc:`variational classifiers </demos/tutorial_variational_classifier>`). However, quantum chemistry
+remains one of the flagship uses-cases for variational quantum algorithms, and VQE the standard-bearer.
```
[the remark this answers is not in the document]
co9olguy: Yep, I knew exactly what was meant here. And it ties together with "flagship" nicely

demonstrations/tutorial_measurement_optimize.py
josh146: As long as the kronecker product order matches the tensor product order, the result will be identical 🙂

demonstrations/tutorial_measurement_optimize.py
josh146: Yep, this is because of a bug with `qml.S.inv()` that was fixed in master but not in release! This will have the correct sign when built against `master`.

Perhaps I will remove the `S.inv()` gate (it can be replaced by `RY(-pi/4)`) so this has the correct sign with the release version.

demonstrations/tutorial_measurement_optimize.py
josh146: Oops, my english failed me!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+The variational quantum eigensolver (VQE) is the OG variational quantum algorithm. Harnessing
+near-term quantum hardware to solve for the electronic structure of molecules, VQE is *the*
+algorithm that sparked the variational circuit craze of the last 5 years, and holds the greatest
+promise for showcasing a quantum advantage on near-term quantum hardware. It has also inspired
+other quantum algorithms such as the :doc:`Quantum Approximate Optimization Algorithm (QAOA)
+</demos/tutorial_qaoa_intro>`.
+
+To scale VQE beyond the regime of classical computation, however, we need to use it to solve for the
+ground state of excessively larger and larger molecules. A side effect is that the number of
+measurements we need to make on the quantum hardware also grows polynomially---a huge bottleneck,
```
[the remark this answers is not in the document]
josh146: Ah, good point. I think the subtlety here is that the O(N^4) growth in measurement terms leads to a quantum resource bottleneck, rather than a strict 'efficiency' bottleneck per se

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,781 @@
...
+The appeal of VQE lies almost within its simplicity. A circuit ansatz :math:`U(\theta)` is chosen
+(typically the Unitary Coupled-Cluster Singles and Doubles
+(:func:`~pennylane.templates.subroutines.UCCSD`) ansatz), and the qubit representation of the
+molecular Hamiltonian is computed:
+
+.. math:: H = \sum_i c_i h_i,
+
+where :math:`h_i` are the terms of the Hamiltonian written as a product of Pauli operators :math:`\sigma_n`:
+
+.. math:: h_i = \prod_{n=0}^{N} \sigma_n.
```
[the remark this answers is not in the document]
josh146:
```suggestion
.. math:: h_i = \bigotimes_{n=0}^{N} \sigma_n.
```
Oops. How did I not spot that 🤦

demonstrations/tutorial_measurement_optimize.py
josh146: I tried for about an hour to make the vertical barriers red and/or dashed, but failed 😆

demonstrations/tutorial_measurement_optimize.py
josh146: :+1:

Re: RY --- this is because `HS^{-1} = RY(-pi/4)`. I might re-write the above to replace all instances of HS^{-1} with RY. This will also look 'better' as well, since the rotations will only ever add O(1) additional depth to the circuit.

demonstrations/tutorial_measurement_optimize.py
josh146: Ah, this is because for the example I chose

```
H = Z_0 + Z_0 Z_1 + Z_0 Z_1 Z_2 + Z_0 Z_1 Z_2 Z_3 + X_2 X_3 + Y_0 X_2 X_3 + Y_0 Y_1 X_2 X_3,
```

has all ones coefficients which led to an easier example 🤔 

If this wasn't the case, you would need to:

* Keep track of the order of the coefficients after the grouping
* Take the dot product of the coefficients and expvals, rather than the sum as I do above.

I *almost* added distinct coefficients, but thought it would add too much complexity?

demonstrations/tutorial_measurement_optimize.py
josh146: Maybe it's too late at night, but I'm drawing a blank at how to express them using X, Y, Z while preserving the index relations

demonstrations/tutorial_measurement_optimize.py
josh146: Done!

demonstrations/tutorial_measurement_optimize.py
josh146: Yeah, I was a bit annoyed 😆 I put in a lot of effort to avoiding using 'words' in the demo

demonstrations/tutorial_measurement_optimize.py
josh146: Yep; I originally drew the nodes identical, but the number of overlapping lines was very difficult to parse.

This is a case where it would work better as a gif:

1. First you see the edges change to form the complement graph
2. The nodes re-arrange to make the graph easier to view

demonstrations/tutorial_measurement_optimize.py
glassnotes: I was originally thinking to just not write it compactly and show [X, Y] = Z, [Y, Z] = X, etc. However a better option could be to replace \sigma_n in the products by P_n, say all P_n \in {\sigma_x, \sigma_y, \sigma_z}, and then \sigma with the lower index can be used in the relations as written.

demonstrations/tutorial_measurement_optimize.py
josh146: I agree, but it might require a change to how PL prints out tensors? 🙁

demonstrations/tutorial_measurement_optimize.py
josh146: Oh, I tried to do this! Unfortunately not so straightforward, as the `nx.complement` function does not seem to preserve node labels (at least, I couldn't get it to keep the labels)

demonstrations/tutorial_measurement_optimize.py
josh146: done!

demonstrations/tutorial_measurement_optimize.py
glassnotes: Ah! I see, yes this is best as is. (I just forgot what the Hamiltonian looked like, my bad!)  This might just be something to make a note of in the code comments, that you'd have to make some adjustments when coefficients are not all 1.

demonstrations/tutorial_measurement_optimize.py
josh146: Drawing the complement with the nodes re-arranged like this makes the colouring easier to see as well

demonstrations/tutorial_measurement_optimize.py
josh146: Good idea!

josh146: Done

demonstrations/tutorial_measurement_optimize.py
josh146: Have added a note here.

demonstrations/tutorial_measurement_optimize.py
josh146: Have added a note at the top, where the Pauli terms are defined!

```rest
A circuit ansatz :math:`U(\theta)` is chosen
(typically the Unitary Coupled-Cluster Singles and Doubles
(:func:`~pennylane.templates.subroutines.UCCSD`) ansatz), and the qubit representation of the
molecular Hamiltonian is computed:

.. math:: H = \sum_i c_i h_i,

where :math:`h_i` are the terms of the Hamiltonian written as a tensor product of Pauli operators
acting on wire :math:`n`, :math:`P_n \in \{I, \sigma_x, \sigma_y, \sigma_z\}`:

.. math:: h_i = \bigotimes_{n=0}^{N-1} P_n.

(This product of Pauli terms is often referred to as a 'Pauli word' in the literature.) 
```

demonstrations/tutorial_measurement_optimize.py
josh146: I would love to turn the SVGs into a gif, but not sure what software to use 🤔

demonstrations/tutorial_measurement_optimize.py
josh146: I've update this line to
```python
print(" ".join([f"{t.name[-1]}{t.wires.tolist()[0]}" for t in term.obs]))
```

which prints out Pauli words of the form `X0 Y2` etc. I guess it's a tradeoff between output readability and code readability? 🤔

demonstrations/tutorial_measurement_optimize.py
josh146: I've added labels! They are a bit ugly, but I didn't think it was worth having a huge paragraph of code to make it look better, you have to get very manual with respect to node/label coordinates to get a nice look 🤔 Function over style? Let me know what you think!

demonstrations/tutorial_measurement_optimize.py
josh146: Have added a comment!

josh146:
Thanks for the very thorough review @glassnotes! I've implemented all suggestions, and the demo is better off for it.

demonstrations/tutorial_measurement_optimize.py
glassnotes: If it's like < 20 frames, we could use a combination of Inkscape to make them, and then ImageMagick's command line tool `convert` to stitch them together, 

```
convert -loop 0 -delay 20 frames{00...xx}.png  frames-as-gif.gif
```

demonstrations/tutorial_measurement_optimize.py
glassnotes: I think this is a good compromise, it's super helpful to have them and the code isn't unwieldy. (Also since you commented it it's clear what it does even if a reader doesn't want to read it in detail).

(The right most node of the first graph has its label cut off, I have a feeling  that this is not something that happens every time since the `nx` graph printing is somewhat random?)

josh146: Yep, I have no idea how to fix that :/ Maybe there is a way to extend the figure size without importing matplotlib?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,828 @@
...
+
+.. figure:: /demonstrations/measurement_optimize/grouping.png
+    :width: 90%
+    :align: center
+
+In this demonstration, we revisit the VQE algorithm, see first-hand how the required number of
+measurements scales as molecule size increases, and finally use these measurement optimization
+strategies to minimize the number of measurements we need to make. These techniques are valuable
+beyond just VQE---allowing you to add measurement optimization to your toolkit of techniques to
+perform variational algorithms more efficiently.
```
glassnotes, line 41: Nice! :tada:

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,828 @@
...
+H = qml.Hamiltonian(coeffs=np.ones(len(terms)), observables=terms)
+cost_fn = qml.ExpvalCost(qml.templates.StronglyEntanglingLayers, H, dev, optimize=True)
+print(cost_fn(weights))
+
+##############################################################################
+# Beyond VQE
+# ----------
+#
+# Wait, hang on. We dived so deeply into measurement grouping and optimization, we forgot to check
+# how this effects the number of measurements required to perform VQE on :math:`\text{H}_2 \text{O}`!
```
glassnotes, line 754:
```suggestion
# how this affects the number of measurements required to perform the VQE on :math:`\text{H}_2 \text{O}`!

```

demonstrations/tutorial_measurement_optimize.py
josh146: Pushed a fix!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,828 @@
...
+#     | :math:`I`        | :math:`I`                     |
+#     +------------------+-------------------------------+
+#
+# .. raw:: html
+#
+#     </div>
+#
+# Therefore, in this particular example:
+#
+# * Wire 0: we are measuring both terms in the :math:`X` basis, apply the Hadamard gate
```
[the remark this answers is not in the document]
josh146: Done!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+
+##############################################################################
+# We're not quite there yet; we have only calculated the probabilities of the variational circuit
+# rotated into the shared eigenbasis---the :math:`|\langle \phi_n |\psi\rangle|^2`. To recover the
+# *expectation values* of the two QWC observables from the probabilities, recall that we need one
+# final piece of information: their eigenvalues :math:`\lambda_{A, n}` and :math:`\lambda_{B, n}`.
+#
+# We know that the single-qubit Pauli operators each have eigenvalues :math:`(1, -1)`, while the identity
+# operator has eigenvalues :math:`(1, 1)`. We can make use of ``np.kron`` to quickly
+# generate the probabilities of the full Pauli terms, making sure that the order
```
antalszava, line 428: >  We can make use of ``np.kron`` to quickly
>  generate the probabilities of the full Pauli terms

This seems a bit confusing considering that ``np.kron``  is used with eigenvalues, maybe worth rephrasing?

Also maybe worth describing why taking the eigenvalues of each term and using the Kronecker product we can make up the eigenvalues of the product.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+#
+#     pennylane.qnodes.base.QuantumFunctionError: Only observables that are qubit-wise commuting
+#     Pauli words can be returned on the same wire
+#
+# The QNode has detected that the two observables are not qubit-wise commuting, and
+# has raised an error.
+#
+# So, a strategy begins to take shape: given a Hamiltonian containing a large number of Pauli terms,
+# can we somehow partition the terms into **fewest** number of QWC groups to minimize the number of measurements
+# we need to take?
```
antalszava, line 502: The mention of choosing from QWC groups seems a bit sudden as the previous description was mainly related to using QWC to carry out the measurement rather than the lack of a unique QWC grouping.

Maybe worth mentioning that

> QWC relation is not the equivalence relation in the algebraic sense

as in II. A of the last ref. to motivate the next section.

Edit: as this is actually explained a bit later, maybe phrasing the question in a way to mention finding a single QWC grouping could work? And then the next section dives deeper and also uncovers that actually such a grouping is not unique.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+# 5. Finally, post-process the probability distributions with the observable eigenvalues
+#    to recover the Hamiltonian expectation value.
+#
+# Luckily, the PennyLane ``grouping`` module makes this relatively easy. Let's walk through
+# the entire process using the provided grouping functions.
+#
+# Steps 1-3 (finding and grouping QWC terms in the Hamiltonian) can be done via the
+# :func:`qml.grouping.group_observables <pennylane.grouping.group_observables>` function:
+
+obs_groupings = qml.grouping.group_observables(terms, grouping_type='qwc', method='rlf')
```
antalszava, line 702: Worth adding that `rlf` refers to recursive largest first, a variant of largest first.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+weights = qml.init.strong_ent_layers_normal(n_layers=3, n_wires=4)
+result = [circuit(weights, group=g) for g in obs_groupings]
+
+print("Term expectation values:")
+for group, expvals in enumerate(result):
+    print(f"Group {group} expectation values:", expvals)
+
+# Since all the coefficients of the Hamiltonian are unity,
+# we can simply sum the expectation values.
+print("<H> = ", np.sum(np.hstack(result)))
```
antalszava, line 736: :heart_eyes: :heart_eyes: :heart_eyes:

demos_research.rst
```diff
@@ -177,6 +177,12 @@ plateaus in quantum neural networks.
     :description: :doc:`demos/tutorial_vqe_qng`
     :tags: chemistry
 
+.. customgalleryitem::
+    :tooltip: Optimize and reduce the number of measurements required to evaluate a variational algorithm cost function.
+    :figure: demonstrations/measurement_optimize/meas_optimize_thumbnail.png
```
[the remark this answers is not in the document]
antalszava: Also like it a lot! :heart_eyes:

Came to mind that having an  or  in the picture could refer to "observable grouping", (not sure if that would be too much extra work at this point).

antalszava (commented):
@josh146 it's a great description, very detailed and also learnt a lot! :1st_place_medal: Left a couple of comments, curious on your input, but had no major blockers.

Was wondering if it could also be categorized as "quantum computing" :thinking: Just because the "trick" itself "comes from there", though think the optimization aspect is also important :thinking:

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+
+.. math:: \text{cost}(\theta) = \langle 0 | U(\theta)^\dagger H U(\theta) | 0 \rangle.
+
+By using a classical optimizer to *minimize* this quantity, we can estimate
+the ground state energy of the Hamiltonian :math:`H`:
+
+.. math:: H U(\theta_{min}) |0\rangle = E_{min} U(\theta_{min}) |0\rangle.
+
+In practice, when we are using quantum hardware to compute these expectation values we expand out
+the Hamiltonian as its summation, resulting in separate expectation values that need to be calculated for each term:
```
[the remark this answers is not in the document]
josh146: I'm not sure that this is so much due to the restriction of the Z basis! Even if we had a device that could measure in arbitrary bases, we would probably still measure independent terms of the Hamiltonian, as there is less classical overhead compared to attempting to diagonalize an arbitrary NxN Hamiltonian

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+# number of electrons
+electrons = 2
+
+# Define the Hartree-Fock initial state for our variational circuit
+initial_state = qml.qchem.hf_state(electrons, num_qubits)
+
+# Construct the UCCSD ansatz
+singles, doubles = qml.qchem.excitations(electrons, num_qubits)
+s_wires, d_wires = qml.qchem.excitations_to_wires(singles, doubles)
+ansatz = functools.partial(
```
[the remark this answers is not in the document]
josh146: This is because `UCCSD` requires more than one argument (e.g., to set the initial state and the electron occupation), but these only need to be set once; during optimization, we only want to pass the weights

demonstrations/tutorial_measurement_optimize.py
josh146:
```suggestion
# operator has eigenvalues :math:`(1, 1)`. We can make use of ``np.kron`` to quickly
# generate the eigenvalues of the full Pauli terms, making sure that the order
```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,829 @@
...
+# Steps 1-3 (finding and grouping QWC terms in the Hamiltonian) can be done via the
+# :func:`qml.grouping.group_observables <pennylane.grouping.group_observables>` function:
+
+obs_groupings = qml.grouping.group_observables(terms, grouping_type='qwc', method='rlf')
+
+
+##############################################################################
+# The ``grouping_type`` argument allows us to choose how the commuting terms
+# are determined (more on that later!) whereas ``method`` determines the colouring
+# heuristic.
```
josh146, line 708:
```suggestion
# heuristic (in this case, ``"rlf"`` refers to recursive largest first, a variant of largest first colouring heuristic).
```

demonstrations/tutorial_measurement_optimize.py
josh146:
```suggestion
# So, a strategy begins to take shape: given a Hamiltonian containing a large number of Pauli terms,
# there is a high likelihood of there being a significant number of terms that qubit-wise commute. Can
# we somehow partition the terms into **fewest** number of QWC groups to minimize the number of measurements
# we need to take?
```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+   tutorial_qaoa_intro Intro to QAOA
+
+The variational quantum eigensolver (VQE) is the OG variational quantum algorithm. Harnessing
+near-term quantum hardware to solve for the electronic structure of molecules, VQE is *the*
+algorithm that sparked the variational circuit craze of the last 5 years, and holds the greatest
+promise for showcasing a quantum advantage on near-term quantum hardware. It has also inspired
+other quantum algorithms such as the :doc:`Quantum Approximate Optimization Algorithm (QAOA)
+</demos/tutorial_qaoa_intro>`.
+
+To scale VQE beyond the regime of classical computation, however, we need to use it to solve for the
```
albi3ro, line 22:
```suggestion
To scale VQE beyond the regime of classical computation, however, we need to solve for the

```
I feel like "to use it" can be assumed and removed to make a more concise sentence.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+ground state of increasingly larger molecules. A side effect is that the number of
```
albi3ro, line 23:
```suggestion
ground state of increasingly larger molecules. Unfortunately, the number of

```
More dynamic.  Now the main verb is "grows" instead of "is". Gets to the point a bit better.

Also, I was getting hung up on something about the words "side effect".  I think a growing number of measurements is expected and fundamental to the nature of the problem.  

Also, "side effect" of what?

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+measurements we need to make on the quantum hardware also grows polynomially---a huge bottleneck,
+especially when quantum hardware access is limited and expensive.
+
+To mitigate this 'measurement problem', a plethora of recent research dropped over the course of
+2019 and 2020 [#yen2020]_ [#izmaylov2019]_ [#huggins2019]_ [#gokhale2020]_ [#verteletskyi2020]_ ,
+exploring potential strategies to minimize the number of measurements required. In fact, by grouping
```
albi3ro, line 29:
```suggestion
explored potential strategies to minimize the number of measurements required. In fact, by grouping

```

josh146: I might keep this one as is - I've read this sentence outloud several times, and present tense seems to make more grammatical sense

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+#         |\langle \phi_n|\psi\rangle|^2.
+#
+# So, assuming we know the eigenvalues of the commuting observables in advance, if we perform a
+# measurement in their shared eigenbasis (the :math:`|\phi_n\rangle`), we only need to perform a **single measurement** of the
+# probabilities :math:`|\langle \phi_n|\psi\rangle|^2` in order to recover both expectation values! 😍
+#
+# Fantastic! But, can we use this to reduce the number of measurements we need to perform in the VQE algorithm?
+# To do so, we must find the answer to two questions:
+#
+# 1. How do we determine which terms of the cost Hamiltonian are commuting?
```
glassnotes, line 271:
```suggestion
# 1. How do we determine which terms of the cost Hamiltonian commute?

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+#
+# The answers to these questions aren't necessarily easy nor straightforward. Thankfully, there are
+# some recent techniques we can harness to address both.
+
+##############################################################################
+# Qubit-wise commuting Pauli terms
+# --------------------------------
+#
+# Back when we summarized the VQE algorithm, we saw that each term of the Hamiltonian is generally represented
+# as a tensor product of Pauli terms:
```
glassnotes, line 283:
```suggestion
# as a tensor product of Pauli operators:

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+#
+# .. math:: h_i = \bigotimes_{n=0}^{N-1} P_n.
+#
+# Luckily, this tensor product structure allows us to take a bit of a shortcut. Rather than consider
+# **full commutativity**, we can consider a slightly less strict condition known as **qubit-wise
+# commutativity** (QWC).
+#
+# To start with, let's consider single-qubit Pauli operators with the identity. We know that the Pauli operators
```
glassnotes, line 291:
```suggestion
# To start with, let's consider single-qubit Pauli operators and the identity. We know that the Pauli operators

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+# Taking the linear combination of the eigenvalues and the probabilities
+print("Expectation value of XYI = ", np.dot(eigenvalues_XYI, rotated_probs))
+print("Expectation value of XIZ = ", np.dot(eigenvalues_XIZ, rotated_probs))
+
+
+##############################################################################
+# Compare this to the result when we used two circuit evaluations. We have successfully used a
+# single circuit evaluation to recover both expectation values!
+#
+# Luckily, PennyLane automatically performs this QWC grouping under-the-hood. We simply
```
glassnotes, line 443:
```suggestion
# Luckily, PennyLane automatically performs this QWC grouping under the hood. We simply

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+# be `NP-hard <https://en.wikipedia.org/wiki/NP-hardness>`__, meaning there is no known (classical)
+# solution to finding the optimum/minimum clique cover in polynomial time.
+#
+# Thankfully, there is a silver lining: we know of polynomial-time algorithms for finding
+# *approximate* solutions to the minimum clique cover problem. These heuristic approaches, while
+# not guaranteed to find the optimum solution, scale quadratically with the number of nodes in the
+# graph/terms in the Hamiltonian [#yen2020]_, so work reasonably well in practice.
+#
+# Many of these heuristic approaches have roots in another graph problem known as `graph
+# colouring <https://en.wikipedia.org/wiki/Graph_coloring>`__; the assignment of colours to
```
glassnotes, line 551:
```suggestion
# colouring <https://en.wikipedia.org/wiki/Graph_coloring>`__, the assignment of colours to
```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+#
+# So, we now have a strategy for minimizing the number of measurements we need to perform
+# for our VQE problem:
+#
+# 1. Determine which terms of the Hamiltonian are qubit-wise commuting, and use
+#    this to construct a graph representing the QWC relationship.
+#
+# 2. Construct the complement QWC graph.
+#
+# 3. Use a graph colouring heuristic algorithm to determine a graph colouring
```
glassnotes, line 687:
```suggestion
# 3. Use a graph colouring heuristic algorithm to determine a graph colouring for the complement graph

```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+
+H = qml.Hamiltonian(coeffs=np.ones(len(terms)), observables=terms)
+cost_fn = qml.ExpvalCost(qml.templates.StronglyEntanglingLayers, H, dev, optimize=True)
+print(cost_fn(weights))
+
+##############################################################################
+# Beyond VQE
+# ----------
+#
+# Wait, hang on. We dived so deeply into measurement grouping and optimization, we forgot to check
```
glassnotes, line 755:
```suggestion
# Wait, hang on. We dove so deeply into measurement grouping and optimization, we forgot to check

```

(I  think?)

glassnotes (approved):
Looks great @josh146 ! Just caught a couple typos / grammatical things but otherwise I think this is good to go :+1:

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+
+.. related::
+
+   tutorial_vqe Variational quantum eigensolver
+   tutorial_quantum_chemistry Quantum chemistry with PennyLane
+   tutorial_qaoa_intro Intro to QAOA
+
+The variational quantum eigensolver (VQE) is the OG variational quantum algorithm. Harnessing
+near-term quantum hardware to solve for the electronic structure of molecules, VQE is *the*
+algorithm that sparked the variational circuit craze of the last 5 years, and holds the greatest
```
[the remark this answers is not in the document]
josh146:
```suggestion
algorithm that sparked the variational circuit craze of the last 5 years, and holds great
```

demonstrations/tutorial_measurement_optimize.py
josh146: Isn't this an okay usage of the semicolon? here, the semicolon is taking the place of a conjunctive adverbs and acting as transitional phrase. E.g., instead of saying

> ... a graph problem known as graph colouring[, which is] the assignment...

demonstrations/tutorial_measurement_optimize.py
josh146: Oops, yep definitely!

demonstrations/tutorial_measurement_optimize.py
josh146: You're right, this is a consequence, not a side effect!
```suggestion
ground state of increasingly larger molecules. A consequence is that the number of
```

demonstrations/tutorial_measurement_optimize.py
glassnotes: Ahh, it probably is, it just didn't feel quite right when I was reading it. No need to incorporate the suggestion :+1:

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,830 @@
...
+.. meta::
+    :property="og:description": Optimize and reduce the number of measurements required to evaluate a variational algorithm cost function.
+    :property="og:image": https://pennylane.ai/qml/_images/grouping.png
+
+.. related::
+
+   tutorial_vqe Variational quantum eigensolver
+   tutorial_quantum_chemistry Quantum chemistry with PennyLane
+   tutorial_qaoa_intro Intro to QAOA
+
```
glassnotes, line 14: Do you want to add the first byline here so we can iron out the format before updating the earlier demos?

josh146: @glassnotes what format do you suggest?

glassnotes: Let's do something like:

*Author: PennyLane dev team (or something like this). Posted: 18 Jan 2021. Last updated: XX YYY ZZZZ. (where applicable)*

josh146:
```suggestion

*Author: PennyLane dev team. Posted: 18 Jan 2021. Last updated: 18 Jan 2021.*
```

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -0,0 +1,831 @@
...
+*Author: PennyLane dev team. Posted: 18 Jan 2021. Last updated: 18 Jan 2021.*
```
josh146, line 15:
```suggestion
*Author: PennyLane dev team. Posted: 18 Jan 2021. Last updated: 18 Jan 2021.*

```

PR #184, opened 2020-12-14
Adds quantum volume demo to gallery

glassnotes:
Displays the demo in the "quantum computing" section, and fixes the incorrect metadata image from the previous PR.

### DIFF
--- a/demonstrations/quantum_volume.py
+++ b/demonstrations/quantum_volume.py
@@ -6,7 +6,7 @@ Quantum volume
 .. meta::
     :property="og:description": Learn about quantum volume, and how to 
         compute it.
-    :property="og:image": https://pennylane.ai/qml/_images/bloch.png
+    :property="og:image": https://pennylane.ai/qml/_images/quantum_volume_thumbnail.png
 
 .. related::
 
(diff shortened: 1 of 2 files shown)


demonstrations/quantum_volume.py
```diff
@@ -6,7 +6,7 @@
 .. meta::
     :property="og:description": Learn about quantum volume, and how to 
         compute it.
-    :property="og:image": https://pennylane.ai/qml/_images/bloch.png
+    :property="og:image": https://pennylane.ai/qml/_images/quantum_volume_thumbnail.png
```
co9olguy, line 9: Doesn't appear to exist? :thinking:

co9olguy (approved):
Thanks @glassnotes!
Can you confirm the address of the image property is working correctly? That URL didn't show up for me, but maybe it's because the tutorial needs to actually be published first?

glassnotes:
> Thanks @glassnotes!
> Can you confirm the address of the image property is working correctly? That URL didn't show up for me, but maybe it's because the tutorial needs to actually be published first?

Yes it won't show up at that address until it's published. In the artifact it shows up if you adjust the end of the url to `_build/html/_images/quantum_volume_thumbnail.png` :+1:

PR #188, opened 2020-12-17
Adds tutorial on noisy circuits

ixfoduap:
This is a short tutorial introducing PennyLane's native support for simulating noisy computations. It highlights our ability to compute channel gradients, including analytical gradients in some cases.

This is a relatively polished version, but it's still missing images and figures that Josh is helping me prepare :art:


demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+for i in range(steps):
+    x = opt.step(cost, x)
+
+p = sigmoid(x)
+print(f"QNode output = {damping_circuit(x):.4f}")
+print(f"Observed expectation value = {ev}")
+print(f"Optimized noise parameter = {p:.4f}")
+
+######################################################################
+# Voilà! We've trained the noisy channel to reproduce the experimental observation. 😎
```
josh146, line 211: This is awesome!

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# help you design, simulate, and optimize noisy quantum circuits. We look forward to seeing what
+# the quantum community can achieve with them! 🚀
+#
+# References
+# ----------
+#
+# .. [#johannes]
+#
+# Meyer, Johannes Jakob, Johannes Borregaard, and Jens Eisert, "A variational toolbox for quantum
+# multi-parameter estimation." `arXiv:2006.06303 (2020) <https://arxiv.org/abs/2006.06303>`__.
```
josh146, line 224:
```suggestion
#     Johannes Jakob Meyer, Johannes Borregaard, and Jens Eisert, "A variational toolbox for quantum
#     multi-parameter estimation." `arXiv:2006.06303 (2020) <https://arxiv.org/abs/2006.06303>`__.
```

josh146 (commented):
Looking awesome @ixfoduap!

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# :math:`\sum_i K_{i}^{\dagger} K_i = \mathcal{I}`. From an initial state :math:`\rho`, the output
+# state after the action of a channel :math:`E` is:
+#
+# .. math::
+#
+#     \rho' = E(\rho) = \sum_i K_i \rho K_{i}^{\dagger}.
+#
+# The action of a quantum channel can thus be interpreted as applying the transformation
+# :math:`K_i` with probability :math:`p_i = \text{Tr}[K_i \rho K_{i}^{\dagger}]`. For
+# example, consider the bit flip channel. It describes a transformation that flips the state of
```
glassnotes, line 84: [This is totally optional] To help with both interpreting / constructing Kraus operators, maybe it would be neat to discuss both "directions." First, start with something like "Consider the two Kraus operators below." Show them and explain how they satisfy the criteria. Then relate to the sentence above discussing K_i applied with probability p_i, and how this is the bit flip channel (applies an X gate (K_1) with probability p, and the identity (K_0) otherwise). Then you can show another example going in the other direction, e.g., knowing that you want to implement a phase flip channel, show how you set up the Kraus operators (send K_1 -> Z). I found Kraus operators really abstract when I first saw them, and remember a lot of exercises like "construct a set that does X", or "what does this set do" so having examples of both is useful.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+    qml.BitFlip(p, wires=0)
+    qml.BitFlip(p, wires=1)
+    return qml.expval(ZZ)
+
+
+ps = [0.001, 0.01, 0.1, 0.2]
+for p in ps:
+    print(f"QNode output = {bitflip_circuit(p):.4f}")
+
+print(f"Output state = {np.real(dev.state)}")
```
glassnotes, line 111:
```suggestion
print(f"Output state = \n{np.real(dev.state)}")
```

(Helps with readability)

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# Besides the bit flip channel, PennyLane supports several other noisy channels that are commonly
+# used to describe experimental imperfections: :func:`~.pennylane.PhaseFlip`,
+# :func:`~.pennylane.AmplitudeDamping`, # :func:`~.pennylane.GeneralizedAmplitudeDamping`,
+# :func:`~.pennylane.PhaseDamping`, and the :func:`~.pennylane.DepolarizingChannel`. You can also
+# build your own custom channel using the operation :func:`~.pennylane.QubitChannel` by
+# specifying its Kraus matrices.
+#
+# For example, the depolarizing channel is a generalization of
+# the bit flip and phase flip channels, where each of the three possible Pauli errors can be
+# applied to a single qubit. A circuit modelling the effect of depolarizing noise in
```
glassnotes, line 126: Show the mathematical expression and what the Kraus operators are.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+    qml.CNOT(wires=[0, 1])
+    qml.DepolarizingChannel(p, wires=0)
+    qml.DepolarizingChannel(p, wires=1)
+    return qml.expval(ZZ)
+
+
+ps = [0.001, 0.01, 0.1, 0.2]
+for p in ps:
+    print(f"QNode output = {depolarizing_circuit(p):.4f}")
+
```
glassnotes, line 142: It may seem obvious but add a sentence to help interpret the output (expectation value gets further and further from ideal value, reinforces that noise is bad).

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+######################################################################
+# Modelling the noise that occurs in real experiments requires careful consideration. PennyLane
+# offers the flexibility to experiment with different combinations of noisy channels to mimic the
+# performance of quantum algorithms when deployed on real devices.
+#
+# Channel gradients
+# -----------------
+#
+# The ability to compute gradients of any operation is an essential ingredient of differentiable
+# quantum programming. In PennyLane, it is possible to
```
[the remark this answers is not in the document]
glassnotes: Link will be here: https://pennylane.ai/qml/glossary/quantum_differentiable_programming.html

(Will merge the page later today)

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# 2}}(|00\rangle + |11\rangle)`.
+
+print(np.real(dev.state))
+
+######################################################################
+# Just as mixed states are represented by density matrices, incoherent noise is modelled by
+# *quantum channels*, which are linear, completely positive, and trace-preserving (`CPTP
+# <https://www.quantiki.org/wiki/channel-cp-map>`__) maps. A convenient strategy for representing
+# quantum channels is to employ Kraus operators :math:`K = \{K_i\}` satisfying the condition
+# :math:`\sum_i K_{i}^{\dagger} K_i = \mathcal{I}`. From an initial state :math:`\rho`, the output
```
glassnotes, line 75: In the example later on, there is explicit mention of analytical gradients for Kraus ops proportional to unitaries. When introducing Kraus operators here may be useful to discuss other properties of the K_i - can they be any general linear transformations, and it is just the combined criteria they must satisfy? Presumably them being unitary is a special case?

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
+r"""
+
+Noisy circuits
+==============
+
+.. meta::
+    :property="og:description": Learn how to simulate nosiy quantum circuits
+    :property="og:image":
+
+.. related::
```
glassnotes, line 10: I didn't notice the new demo in the gallery, is it going to be categorized the same as these two under 'basic/applications'?

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+
+    tutorial_noisy_circuit_optimization Optimizing noisy circuits with Cirq
+    pytorch_noise PyTorch and noisy devices
+
+In this tutorial, you'll learn how to simulate noisy circuits using built-in functionality in
+PennyLane. We'll cover the basics of noisy channels and density matrices, then use example code to
+simulate noisy circuits. PennyLane, the library for differentiable quantum computations, has the
```
glassnotes, line 17:
```suggestion
simulate noisy circuits. PennyLane, the library for differentiable quantum computations, has
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+unique feature that we can compute *gradients* of noisy channels. So you'll also learn how to employ
```
glassnotes, line 18:
```suggestion
unique features that enable us to compute *gradients* of noisy channels. So you'll also learn how to employ
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+channel gradients to optimize noise parameters in a circuit. We're putting the N in NISQ. 🇳
```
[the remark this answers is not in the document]
glassnotes: Whoa is this a 3D drawing??

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# gradients exist for channels whose Kraus operators are proportional to unitary
+# matrices [#johannes]. In other cases, gradients are evaluated using finite difference.
+#
+# To illustrate this property, we'll consider an elementary example. We'll aim to train the noise
+# parameters of a circuit in order to reproduce an observed expectation value. So suppose that we
+# run the circuit to prepare a Bell state
+# on a hardware device and observe that the expectation value of :math:`Z_0\otimes Z_1` is
+# not equal to one, as would occur with an ideal device. In the experiment, it is known that the
+# major source of noise is amplitude damping, as would happen for example due to photon loss.
+# This process projects a state to :math:`|0\rangle` with probability :math:`p`. What damping
```
glassnotes, line 163: Show the Kraus operators?

demonstrations/tutorial_noisy_circuits.py
glassnotes, line 19: :laughing:  I love this

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+print(f"Observed expectation value = {ev}")
+print(f"Optimized noise parameter = {p:.4f}")
+
+######################################################################
+# Voilà! We've trained the noisy channel to reproduce the experimental observation. 😎
+#
+# Developing quantum algorithms that leverage the power of NISQ devices requires serious
+# consideration of the effects of noise. With PennyLane, you have access to tools that can
+# help you design, simulate, and optimize noisy quantum circuits. We look forward to seeing what
+# the quantum community can achieve with them! 🚀
```
glassnotes, line 216: Great conclusion! The emoji add a lot here :rocket:

glassnotes (commented):
Nice demo @ixfoduap! The code and set of examples make it very clear how to use this functionality.

My comments relate mostly to assumed background knowledge. I think some more details on Kraus operators would be helpful, as well as more background on pure vs. mixed states. There is some technical terminology used later on that may not be familiar to someone unless they have already studied this stuff. That said - the length of this demo is very nice right now, and I'd be hesitant to clutter it with additional background knowledge. Rather than implementing some of the suggestions, it may be better to just specify some prerequisite knowledge at the top, and explain that the demo shows the reader how to implement that in PennyLane.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# :math:`K_i` with probability :math:`p_i = \text{Tr}[K_i \rho K_{i}^{\dagger}]`. For
+# example, consider the bit flip channel. It describes a transformation that flips the state of
+# a qubit (applies an X gate) with probability :math:`p`. Its Kraus operators are
+#
+# .. math::
+#
+#     K_0 = (1-p)\begin{pmatrix}1 & 0\\ 0 & 1\end{pmatrix}
+#     K_1 = p\begin{pmatrix}0 & 1\\ 1 & 0\end{pmatrix}.
+#
+# This channel can be implemented in PennyLane using the :func:`~.pennylane.BitFlip` operation.
```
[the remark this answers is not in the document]
glassnotes: I am running into this in another PR - I think it might need to be `:class:` instead of `:func:` here, since the channels are classes.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+# Let's see what happens when we simulate this type of noise acting on
+# both qubits in the circuit. We'll evaluate the QNode for different bit flip probabilitites.
```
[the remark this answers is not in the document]
ixfoduap: Hahaha probabilitites is such a funny word :rofl:

demonstrations/tutorial_noisy_circuits.py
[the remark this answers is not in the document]
josh146: oops, yep you're right - `:class:`.

demonstrations/tutorial_noisy_circuits.py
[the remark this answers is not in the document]
josh146: I think the way to link to it is

```
:doc:`/glossary/quantum_differentiable_programming`
```

demonstrations/tutorial_noisy_circuits.py
ixfoduap: I didn't know about `\n`! Thanks for the suggestion and lesson!

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+
+print(f"QNode output = {circuit():.4f}")
+
+######################################################################
+# The device stores the output state as a density matrix. In this case, since we haven't
+# introduced any noise, the density matrix is equal to the rank-one
+# projector :math:`|\psi\rangle\langle\psi|, where :math:`|\psi\rangle=\frac{1}{\sqrt{
+# 2}}(|00\rangle + |11\rangle)`.
+
+print(np.real(dev.state))
```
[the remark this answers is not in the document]
ixfoduap: I know where you're coming from, but I've always thought that density matrices are the real quantum states! So I'm happy to keep as is, and hint to the reader that density matrices are the states when working with mixed-state computations

demonstrations/tutorial_noisy_circuits.py
ixfoduap: Thanks for the suggestion Olivia! I'm afraid that this may break the flow and turn the demo into a tutorial on Kraus operators :grimacing: 
Based on your suggestions, I've strategically added some additional sentences that I hope can give a solid intuition for these concepts. We can revisit if you think they could still use some work

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+ps = [0.001, 0.01, 0.1, 0.2]
+for p in ps:
+    print(f"QNode output = {depolarizing_circuit(p):.4f}")
+
+######################################################################
+# Modelling the noise that occurs in real experiments requires careful consideration. PennyLane
+# offers the flexibility to experiment with different combinations of noisy channels to mimic the
+# performance of quantum algorithms when deployed on real devices.
```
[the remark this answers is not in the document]
ixfoduap: I would go out on a limb and say that there is no reason to simulate a noise channel on hardware; they have enough noise as it is. But you're right that we can simulate all kinds of channels, and this would be interesting in hardware, for instance to simulate open quantum systems

demos_basics.rst
```diff
@@ -180,4 +186,5 @@ hardware.
     demos/tutorial_qaoa_intro
     demos/tutorial_qnn_module_tf
     demos/tutorial_qnn_module_torch
+    demos/tutorial_noisy_circuits.py
```
glassnotes, line 189:
```suggestion
    demos/tutorial_noisy_circuits
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+print(f"Experimental expectation value = {ev}")
+print(f"Optimized noise parameter p = {p:.4f}")
+
+######################################################################
+# Voilà! We've trained the noisy channel to reproduce the experimental observation. 😎
+#
+# Developing quantum algorithms that leverage the power of NISQ devices requires serious
+# consideration of the effects of noise. With PennyLane, you have access to tools that can
+# help you design, simulate, and optimize noisy quantum circuits. We look forward to seeing what
+# the quantum community can achieve with them! 🚀 🎉 😸
```
glassnotes, line 270: :laughing:

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+#     E(\rho) = \sum_i K_i \rho K_{i}^{\dagger}.
+#
+# Like pure states, which are special cases of mixed states, unitary
+# transformations are special cases of a quantum channels. They have a single Kraus operator,
+# the unitary :math:`U`, and they transform a state as
+# :math:`U\rho U^\dagger`.
+#
+# More generally, the action of a quantum channel can be interpreted as applying the
+# transformation
+# :math:`\frac{1}{p_i}K_i\rho K_i^\dagger` with probability :math:`p_i = \text{Tr}[K_i \rho K_{i}^{
```
glassnotes, line 105: Can this be expressed only verbally, for example "applying each transformation K_i with some associated probability".

glassnotes (commented):
@ixfoduap updated version is looking great! I found that adding the extra examples of Kraus operators was very helpful. Just caught a few typos, and have a couple more suggestions for phrasing.

demonstrations/tutorial_noisy_circuits.py
josh146: image

Seems to be a good location!

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+
+.. related::
+
+    tutorial_noisy_circuit_optimization Optimizing noisy circuits with Cirq
+    pytorch_noise PyTorch and noisy devices
+
+In this demonstration, you'll learn how to simulate noisy circuits using built-in functionality in
+PennyLane. We'll cover the basics of noisy channels and density matrices, then use example code to
+simulate noisy circuits. PennyLane, the library for differentiable quantum computations, has
+unique features that enable us to compute gradients of noisy channels. So you'll also learn how
```
josh146, line 18: Minor suggestion: the 'So you'll learn' sounds a bit repetitive as it was used to open the paragraph. Maybe find a way to rephrase - 'we'll also explore how to employ...'

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+
+    ..
+"""
+
+##############################################################################
+#
+# Noisy operations
+# ----------------
+# Noise is any unwanted transformation that corrupts the intended
+# output of a quantum computation. It can be separated into two categories. Coherent
```
josh146, line 35: Minor recommendation: this paragraph might be easier to parse if the Coherent and Incoherent sections were split into bullet points, to provide visual cues to the statement 'separated into two categories':

> Noise is any unwanted transformation that corrupts the intended output of a quantum computation. It can be separated into two categories:
>
> * **Coherent noise** is described by unitary operations that maintain the purity of the output quantum state. A common source are systematic errors originating from imperfectly-calibrated devices that do not exactly apply the desired gates, e.g, applying a rotation by an angle ϕ+ϵ instead of ϕ. 
>
> * **Incoherent noise** is more problematic: it originates from a quantum computer becoming entangled with the environment, resulting in mixed states — probability distributions over different pure states. Incoherent thus noise leads to outputs that are always random, regardless of what basis we measure in.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+# instead of :math:`\phi`.
+# Incoherent noise is more problematic: it originates from a
+# quantum computer becoming entangled with the environment, resulting in
+# mixed states --- probability distributions
+# over different pure states. Incoherent thus noise leads to
+# outputs that are always random, regardless of what basis we measure in.
+#
+# Mixed states are described by `density matrices
+# <https://en.wikipedia.org/wiki/Density_matrices>`__.
+# A density matrix is a more general method of describing quantum states that elegantly
```
josh146, line 48: Minor: might be better to say

```suggestion
# Density matrices provide a more general method of describing quantum states that elegantly
```

josh146: (to make a better distinction between the density matrix itself and the purpose of density matrices)

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+# encodes a distribution over pure states (a mixed state) in a single mathematical object.
+# Mixed states are the most general description of a quantum state, of which pure
+# states are a special case.
+#
+# Compared to pure states that
+# are described as vectors, a different method is needed for simulating mixed-state quantum
+# computations. This is the purpose of PennyLane's ``default.mixed`` device, which provides native
+# support for mixed states and for simulating noisy computations. Let's use ``default.mixed`` to
+# simulate a simple circuit for preparing the
+# Bell state :math:`|\psi\rangle=\frac{1}{\sqrt{2}}(|00\rangle+|11\rangle)`. We ask the QNode to
```
josh146, line 58: I love this, it sounds so polite 😆

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+print(f"Output state is = \n{np.real(dev.state)}")
+
+######################################################################
+# Just as mixed states are represented by density matrices, incoherent noise is modelled by
+# quantum channels. Mathematically, a quantum channel is are linear, completely positive,
+# and trace-preserving (`CPTP
+# <https://www.quantiki.org/wiki/channel-cp-map>`__) map. A convenient strategy for representing
+# quantum channels is to employ Kraus operators :math:`\{K_i\}` satisfying the condition
```
josh146, line 90:
```suggestion
# quantum channels is to employ `Kraus operators <https://en.wikipedia.org/wiki/Quantum_operation#Kraus_operators>`__
:math:`\{K_i\}` satisfying the condition
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+######################################################################
+# The device stores the output state as a density matrix. In this case, the density matrix is
+# equal to :math:`|\psi\rangle\langle\psi|`,
+# where :math:`|\psi\rangle=\frac{1}{\sqrt{2}}(|00\rangle + |11\rangle)`.
+
+
+print(f"Output state is = \n{np.real(dev.state)}")
+
+######################################################################
+# Just as mixed states are represented by density matrices, incoherent noise is modelled by
```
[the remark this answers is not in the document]
josh146: I agree, I think this is missing a short discussion on unitary operations/matrices

demonstrations/tutorial_noisy_circuits.py
josh146: > turn the demo into a tutorial on Kraus operators 😬

I honestly wouldn't mind 😆

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+
+print(f"Output state for bit flip probability {p} is = \n{np.real(dev.state)}")
+
+######################################################################
+# Besides the bit flip channel, PennyLane supports several other noisy channels that are commonly
+# used to describe experimental imperfections: :class:`~.pennylane.PhaseFlip`,
+# :class:`~.pennylane.AmplitudeDamping`, :class:`~.pennylane.GeneralizedAmplitudeDamping`,
+# :class:`~.pennylane.PhaseDamping`, and the :class:`~.pennylane.DepolarizingChannel`. You can also
+# build your own custom channel using the operation :class:`~.pennylane.QubitChannel` by
+# specifying its Kraus matrices, or even submit a pull request introducing a new channel.
```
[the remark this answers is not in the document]
josh146: Maybe good to include a link to the developer hub here? https://pennylane.readthedocs.io/en/stable/development/guide.html

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,281 @@
...
+# All that remains is to optimize the parameter. We use a straightforward gradient descent
+# method.
+
+
+opt = qml.GradientDescentOptimizer(stepsize=10)
+steps = 50
+x = 0.0
+
+for i in range(steps):
+    x = opt.step(cost, x)
```
josh146, line 257: Very very minor: I always like it when you can see the change during training. You could do the following:

```python
for i in range(steps):
    x, cost_value = opt.step_and_cost(cost, x)
    if i % 10 == 0:
        print(f"Step: {step}    Cost: {cost_val}")
```

josh146 (commented):
All my suggestions are relatively minor!

demonstrations/tutorial_noisy_circuits.py
ixfoduap: Love this!

demonstrations/tutorial_noisy_circuits.py
ixfoduap: Excuse Sir Mr. QNode, would you be so kind as to indulge me in providing this expectation value?

demonstrations/tutorial_noisy_circuits.py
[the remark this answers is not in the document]
ixfoduap: My point is that density matrices are generalizations of pure states, and quantum channels are generalizations of unitary transformations. 

I'll revise and try to explain these concepts better

demonstrations/tutorial_noisy_circuits.py
ixfoduap: Great suggestion! We can do both :grin:

demonstrations/tutorial_noisy_circuits.py
josh146: hahaha

josh146: After ruminating, I support making this change

demonstrations/tutorial_noisy_circuits.py
ixfoduap: Be careful what you wish for!

trbromley:
[ch1184]

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+# <https://www.quantiki.org/wiki/channel-cp-map>`__) map. A convenient strategy for representing
+# quantum channels is to employ `Kraus operators
+# <https://en.wikipedia.org/wiki/Quantum_operation#Kraus_operators>`__
+# :math:`\{K_i\}` satisfying the condition
+# :math:`\sum_i K_{i}^{\dagger} K_i = I`. For an initial state :math:`\rho`, the output
+# state after the action of a channel :math:`E` is:
+#
+# .. math::
+#
+#     E(\rho) = \sum_i K_i \rho K_{i}^{\dagger}.
```
albi3ro, line 97: So does the channel need to be named $E$?

Seeing $E(\rho)$ makes me think we are taking the energy, and it takes my brain a second to adjust.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+#
+# Like pure states, which are special cases of mixed states, unitary
```
albi3ro, line 99:
```suggestion
# Like pure states are special cases of mixed states, unitary
```
More concise. 
Also, makes the two phrases parallel each other.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+# transformations are special cases of a quantum channels. They have a single Kraus operator,
```
albi3ro, line 100: What "they" refers to is a bit unclear. "Unitary transformations" have a single Kraus operator?

"have" is confusing me too.  Maybe, "A single Kraus operator represents a Unitary transformation", or "applies", even "is"

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+    qml.AmplitudeDamping(sigmoid(x), wires=0)  # p = sigmoid(x)
+    qml.AmplitudeDamping(sigmoid(x), wires=1)
+    return qml.expval(qml.PauliZ(0) @ qml.PauliZ(1))
+
+######################################################################
+# We optimize the circuit with respect to a simple cost function that attains its minimum when
+# the output of the QNode is equal to the experimental value:
+
+
+def cost(x):
```
albi3ro, line 249: This may be a good place to show off the new functionality of the optimizers handling general cost functions.  Using global variables inside a function generally makes me a bit uneasy.

```python
ev = np.tensor([ev], requires_grad=False)

def cost(x, target):
    return (damping_circuit(x) - target[0])**2
```

Then in the optimizer step
```python
(x, ev), cost_val = opt.step_and_cost(cost, x, ev)
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+######################################################################
+# Incoherent noise is modelled by
+# quantum channels. Mathematically, a quantum channel is a linear, completely positive,
+# and trace-preserving (`CPTP
+# <https://www.quantiki.org/wiki/channel-cp-map>`__) map. A convenient strategy for representing
+# quantum channels is to employ `Kraus operators
+# <https://en.wikipedia.org/wiki/Quantum_operation#Kraus_operators>`__
+# :math:`\{K_i\}` satisfying the condition
+# :math:`\sum_i K_{i}^{\dagger} K_i = I`. For an initial state :math:`\rho`, the output
+# state after the action of a channel :math:`E` is:
```
ixfoduap, line 93:
```suggestion
# state after the action of a channel :math:`\Phi` is:
```

demonstrations/tutorial_noisy_circuits.py
ixfoduap, line 97:
```suggestion
#     \Phi(\rho) = \sum_i K_i \rho K_{i}^{\dagger}.
```

ixfoduap: Changed to Phi!

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+    :width: 20%
+
+    ..
+"""
+
+##############################################################################
+#
+# Noisy operations
+# ----------------
+# Noise is any unwanted transformation that corrupts the intended
```
josh146, line 34: corrupts is a great word in this context

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+#
+# * **Incoherent noise** is more problematic: it originates from a quantum computer
+#   becoming entangled with the environment, resulting in mixed states --- probability
+#   distributions over different pure states. Incoherent noise thus leads to outputs that are
+#   always random, regardless of what basis we measure in.
+#
+# Mixed states are described by `density matrices
+# <https://en.wikipedia.org/wiki/Density_matrices>`__.
+# They provide a more general method of describing quantum states that elegantly
+# encodes a distribution over pure states (a mixed state) in a single mathematical object.
```
josh146, line 50: I feel like the `(a mixed state)` parenthesis here is not needed? The sentence reads fine to me without it.

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+# The purpose of PennyLane's ``default.mixed`` device is to provide native
+# support for mixed states and for simulating noisy computations. Let's use ``default.mixed`` to
+# simulate a simple circuit for preparing the
+# Bell state :math:`|\psi\rangle=\frac{1}{\sqrt{2}}(|00\rangle+|11\rangle)`. We ask the QNode to
+# return the expectation value of :math:`Z_0\otimes Z_1`:
+#
+import pennylane as qml
+from pennylane import numpy as np
+
+qml.enable_tape()
```
josh146, line 63:
```suggestion
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+# :math:`\{K_i\}` satisfying the condition
+# :math:`\sum_i K_{i}^{\dagger} K_i = I`. For an initial state :math:`\rho`, the output
+# state after the action of a channel :math:`\Phi` is:
+#
+# .. math::
+#
+#     \Phi(\rho) = \sum_i K_i \rho K_{i}^{\dagger}.
+#
+# Just like pure states are special cases of mixed states, unitary
+# transformations are special cases of quantum channels. Unitary transformations are represnted
```
josh146, line 100:
```suggestion
# transformations are special cases of quantum channels. Unitary transformations are represented
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+#
+# More generally, the action of a quantum channel can be interpreted as applying a
+# transformation corresponding to the Kraus operator :math:`K_i` with some associated
+# probability. More precisely, the channel applies the
+# transformation
+# :math:`\frac{1}{p_i}K_i\rho K_i^\dagger` with probability :math:`p_i = \text{Tr}[K_i \rho K_{i}^{
+# \dagger}]`. Quantum
+# channels therefore represent a probability distribution over different possible
+# transformations on a quantum state. For
+# example, consider the bit flip channel. It describes a transformation that flips the state of
```
josh146, line 113: Suggest a line break for clarity:
```suggestion
# transformations on a quantum state.
#
# For example, consider the bit flip channel. It describes a transformation that flips the state of
```

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -0,0 +1,288 @@
...
+#
+# .. math::
+#
+#     K_0 &= \sqrt{1-p}\begin{pmatrix}1 & 0\\ 0 & 1\end{pmatrix}, \\
+#     K_1 &= \sqrt{p}\begin{pmatrix}0 & 1\\ 1 & 0\end{pmatrix}.
+#
+# This channel can be implemented in PennyLane using the :class:`qml.BitFlip <pennylane.BitFlip>`
+# operation.
+# Let's see what happens when we simulate this type of noise acting on
+# both qubits in the circuit. We'll evaluate the QNode for different bit flip probabilities.
```
josh146, line 125: Suggest a line break for clarity
```suggestion
# operation.
#
# Let's see what happens when we simulate this type of noise acting on
# both qubits in the circuit. We'll evaluate the QNode for different bit flip probabilities.
```

josh146 (approved):
🔕 📣

demonstrations/tutorial_noisy_circuits.py
josh146: wait, this one is not optional lol

PR #193, opened 2020-12-21
Adds new content to empty glossary entries

glassnotes:
**Title:** Adds new content to empty glossary entries

**Summary:** This PR increases the utility of the glossary by adding basic content pages to a handful of entries that either don't already have a page, or don't link to a relevant tutorial. There are also two (relevant) structural changes to the glossary:
 - the entry for quantum Boltzmann machines was removed as I don't feel it is sufficiently related to the other entries, and it is unlikely we will be making a demo in the near term that would be linked to the entry
 - the entries for automatic differentiation and quantum differentiable programming have been merged into a single entry for 'differentiable quantum programming'

**Relevant references:** None

**Possible Drawbacks:** None

TODO:
 - [X] quantum CNNs (might eventually make a neat full-length demo)
~~-[ ] quantum Boltzmann machines~~
 - [x] quantum differentiable programming
 - [x] quantum feature map
 - [x] quantum node


glossary.rst
```diff
@@ -279,6 +254,10 @@ Key Concepts
         specific subsystems. The architecture defines which algorithms a variational circuit can implement by
         fixing the trainable parameters. A circuit ansatz is analogous to the architecture of a neural network.
 
+    :doc:`Differentiable Quantum Programming </glossary/differentiable_quantum_programming>`
+        The paradigm of making quantum algorithms differentiable, and thereby trainable. See also
+        :doc:`quantum gradient </glossary/quantum_gradient>`.
```
josh146, line 259: 😍

glossary.rst
```diff
@@ -76,16 +76,6 @@ Key Concepts
...
-            {
-                "@type": "Question", 
-                "name" : "What is automatic differentiation?",
-                "acceptedAnswer" :
-                {
-                    "@type": "Answer",
-                    "text": "The process of automatically computing derivatives of the steps of computer programs."
-                }
-            },
-
```
josh146, line 88: Does this need to be replaced with differentiable quantum programming?

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+.. _glossary_differentiable_quantum_programming:
+
+Differentiable Quantum Programming
+==================================
+
+`Differentiable programming
+<https://en.wikipedia.org/wiki/Differentiable_programming>`_ is a style of
+programming that uses `automatic differentiation
+<https://en.wikipedia.org/wiki/Automatic_differentiation>`_ to compute the
+derivatives of functions with respect to program inputs. In quantum computing,
```
josh146, line 13: Maybe this can be expanded to explain how the concept of differentiable programming differs from standard deep learning? It's more of a conceptual shift rather than a technical shift, if anything.

* Treats the entire classical program/code as differentiable. You can include classical control such as for loops, if statements, etc, and the output of the program remains differentiable.

* This concept evolved from the 'define-by-run' approach of PyTorch, TF2, which evolved from the earlier 'define-and-run'

* Unlike standard neural networks, there is no underlying assumption to the structure of the model. It simply consists of composable parametrized function blocks.

* The program structure *itself* is dynamic and can change with input, while remaining trainable/differentiable.

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+parameters. Differentiable quantum programming is a paradigm that leverages this
+to make quantum algorithms differentiable, and thereby trainable.
+
+
+Automatic differentiation
+-------------------------
+
+Derivatives and gradients are ubiquitous throughout science and engineering.
+In recent years, automatic differentiation has become a key feature in many numerical software libraries,
+in particular for machine learning (e.g., Theano_, Autograd_, Tensorflow_, or Pytorch_).
```
josh146, line 25: Might as well include Jax.

Also, it's little known, but [Chainer](https://docs.chainer.org/en/stable/index.html) was the library that first brought define-by-run to greater attention!

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+algorithm for :math:`\nabla h(x)` with the same degree of complexity as the original function.
+
+*Automatic* differentiation should be distinguished from other forms of differentiation:
+
+* *Symbolic differentiation*, where the actual equation of the derivative function is computed and
+  evaluated; has a limited scope since it requires "hand-written" support for new functions.
+* *Numerical differentiation*, such as the finite-difference
+  method familiar from high-school calculus, where the derivative of a function is approximated by
+  numerically evaluating the function at two infinitesimally separated points. However, the approximation can be
+  unstable, and in many settings an exact gradient is preferred.
```
josh146, line 49: Oh! This question has come up a lot in the forum. C.f. https://discuss.pennylane.ai/t/parameter-shift-rules/496/4

It would be great to have trivial examples of each of the three, as well as some FAQs/when each is most applicable.

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+    :target: javascript:void(0);
+
+:html:`<br>`
+
+For example, in PennyLane parameterized quantum operations carry information
+about their parameters and their domains, and specify a "recipe" that details
+how to automatically compute gradients. Many operations make use of
+:doc:`parameter-shift rules </glossary/parameter_shift>` for this purpose (see,
+for example, the arbitrary unitary rotation `qml.Rot
+<https://pennylane.readthedocs.io/en/stable/code/api/pennylane.Rot.html>`_ which
```
josh146, line 73: Can be replaced by

```
:class:`qml.Rot <pennylane.Rot>`
```

(could be `pennylane.ops.Rot`, not sure off the top of my head). This is preferable, since it is independent of the base URL.

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. figure:: ../_static/concepts/autodiff_quantum.png
+    :align: center
+    :width: 30%
+    :target: javascript:void(0);
+
+:html:`<br>`
+
+For example, in PennyLane parameterized quantum operations carry information
+about their parameters and their domains, and specify a "recipe" that details
```
josh146, line 69: maybe the mention of domains is not needed?

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+how to automatically compute gradients. Many operations make use of
+:doc:`parameter-shift rules </glossary/parameter_shift>` for this purpose (see,
+for example, the arbitrary unitary rotation `qml.Rot
+<https://pennylane.readthedocs.io/en/stable/code/api/pennylane.Rot.html>`_ which
+uses parameter-shift rules to compute the derivative with respect to each of its
+three parameters). In this way, the gradient of arbitrary sequences of
+parameterized gates can be computed (if the parameter is used more than once,
+the product rule for derivatives can be applied). Once evaluated the gradients
```
josh146, line 77: the product rule is no longer performed by PL in tape mode :) It is delegated to the ML framework

glossary/qcnn.rst
```diff
@@ -0,0 +1,67 @@
...
+<https://en.wikipedia.org/wiki/Convolutional_neural_network>`_ (CNNs) are a type
+of classical machine learning model often used in computer vision and image
+processing applications. The structure of CNNs consists of applying alternating
+*convolutional layers* (plus an activation function) and *pooling layers* to an
+input array, typically followed by some fully connected layers before the
+output.
+
+:html:`<br>`
+
+.. figure:: ../_static/concepts/cnn.svg
```
josh146, line 19: These images are awesome

glossary/quantum_feature_map.rst
```diff
@@ -0,0 +1,50 @@
...
+true for many data sets. To work around this it may be possible to transform the
+data into a new space where it is linear by way of a *feature map*.
+
+More formally, let :math:`\cal{X}` be a set of input data. A feature map
+:math:`\phi` is a function that acts as :math:`\phi : \cal{X} \rightarrow
+\cal{F}` where :math:`\cal{F}` is the feature space. The outputs of the map on
+the individual data points, :math:`\phi(x)` for all :math:`x \in \cal{X}`, are
+called feature vectors.
+
+.. figure:: ../_static/concepts/feature_map_svm.svg
```
josh146, line 24: So nice!

josh146: I can't really explain it, but there is something about hand-drawn diagrams that makes me less likely to just 'scan over' them. I think reading papers/wikipedia sometimes desensitizes me to scientific diagrams, but the charm + lack of straight lines in these illustrations captures my attention.

glossary/quantum_node.rst
```diff
@@ -0,0 +1,33 @@
...
+    :width: 80%
+    :target: javascript:void(0);
+
+
+.. seealso::
+
+    In PennyLane, quantum nodes can be constructed by using either the
+    ``qnode`` `decorator
+    <https://pennylane.readthedocs.io/en/stable/code/api/pennylane.qnode.html>`_,
+    or the ``QNode`` `constructor <https://pennylane.readthedocs.io/en/stable/code/api/pennylane.QNode.html>`_.
```
josh146, line 33: Same comment here re: using

```
:class:
```

and 

```
:func:
```

josh146 (commented):
Looks great @glassnotes! I think my only major comment was expanding out *what* differentiable programming means. Apart from that, everything else was minor.

glossary/differentiable_quantum_programming.rst
glassnotes: Oh cool I didn't know about that, thanks! :smile:

glossary.rst
glassnotes: Good catch, I forgot to change the metadata for that entry!

glassnotes:
> Looks great @glassnotes! I think my only major comment was expanding out _what_ differentiable programming means. Apart from that, everything else was minor.

Thanks, @josh146 !  I'll work on incorporating your suggestions for this, it'll be good to have a more comprehensive resource for that on our site that we can point people to.

co9olguy:
Thanks @glassnotes! I'm just leaving a fly-by comment (didn't review), but can we use the term "quantum differentiable programming" rather than "differentiable quantum programming"? We have used both in the past interchangably, but I think it's more beneficial to use the former as people might find it easier from a search engine or connect easier to the classical idea (since the phrase "differentiable programming" remains intact)

glossary/differentiable_quantum_programming.rst
mariaschuld: To keep the first sentences limited (so they pop up in searches) one could add a new section with this?

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+way, the gradient of arbitrary sequences of parameterized gates can be
+computed. Once evaluated the gradients can be fed forward into subsequent parts
+of a larger hybrid computation.
+
+:html:`<br>`
+
+.. figure:: ../_static/concepts/autodiff_quantum_circuit.svg
+    :align: center
+    :width: 60%
+    :target: javascript:void(0);
```
mariaschuld, line 84: I like the picture, but would it be possible to optically emphasize which textbox corresponds to which gate?

glossary/differentiable_quantum_programming.rst
```diff
@@ -0,0 +1,92 @@
...
+:html:`<br>`
+
+.. figure:: ../_static/concepts/autodiff_quantum.png
+    :align: center
+    :width: 30%
+    :target: javascript:void(0);
+
+:html:`<br>`
+
+For example, in PennyLane parameterized quantum operations carry information
```
mariaschuld, line 69: Hmm, I am a bit confused how this paragraph matches the header?

glossary/qcnn.rst
mariaschuld: I agree. The article is also great!

glossary/quantum_embedding.rst
```diff
@@ -6,15 +6,16 @@
...
+A *quantum embedding* represents classical data as quantum states in a Hilbert
+space via a :doc:`quantum feature map </glossary/quantum_feature_map>`. It takes a
+classical datapoint :math:`x` and translates it into a set of gate parameters in
+a quantum circuit, creating a quantum state :math:`| \psi_x \rangle`. This
+process is a crucial part of designing quantum algorithms and affects their
+computational power—for more details, see `Schuld & Petruccione (2018)
+<https://www.springer.com/gp/book/9783319964232>`_, but also `Havlicek et
+al. (2018) <https://arxiv.org/abs/1804.11326>`_, `Schuld & Killoran (2018)
+<https://arxiv.org/abs/1803.07128>`_, `Lloyd et al. (2020)
+<https://arxiv.org/abs/2001.03622>`_.
```
mariaschuld, line 18: Shall we remove the reference to our book in the note and important block below? I think Aroosa wrote this using the book, but it sounds a bit like self-citing :)

glossary/quantum_feature_map.rst
```diff
@@ -0,0 +1,50 @@
...
+
+.. figure:: ../_static/concepts/quantum_feature_map.svg
+    :align: center
+    :width: 70%
+    :target: javascript:void(0);
+
+
+For some more detailed examples of quantum feature maps, see the entry for
+:doc:`quantum embeddings </glossary/quantum_embedding>`, and the key reference by `Schuld &
+Killoran (2018) <https://arxiv.org/abs/1803.07128>`_.
```
mariaschuld, line 47: I think we need to add the Havlicek paper here too...

glossary/quantum_node.rst
```diff
@@ -0,0 +1,30 @@
+.. role:: html(raw)
+   :format: html
+
+.. _glossary_quantum_node:
+
+Quantum Node
+============
```
mariaschuld, line 7: Nice!

mariaschuld (approved):
The articles are great and all my comments more optional.

glassnotes:
> Thanks @glassnotes! I'm just leaving a fly-by comment (didn't review), but can we use the term "quantum differentiable programming" rather than "differentiable quantum programming"? We have used both in the past interchangably, but I think it's more beneficial to use the former as people might find it easier from a search engine or connect easier to the classical idea (since the phrase "differentiable programming" remains intact)

Sure, I swapped the order because while I was doing research I saw that both Yao and [a paper about it](https://arxiv.org/abs/2004.01122) call it "differentiable quantum programming". I can swap it back though, no strong feelings either way.

glossary/quantum_embedding.rst
glassnotes: I think we should keep it - I have also used that book as a reference on this topic multiple times :)

glossary/quantum_feature_map.rst
glassnotes: Done, just let me know if it is not the one your were thinking of.

glossary/differentiable_quantum_programming.rst
glassnotes: Yeah good call, it's all kinda squashed together.  I tried to separate it a bit and add the association as suggested.

glossary/differentiable_quantum_programming.rst
glassnotes: I tried to incorporate this in the introductory paragraphs while keeping the first few lines the same.  I'm not sure how well it fits though, the transition into the next part feels a bit choppy (this is also my first time really learning what differentiable programming is, so hopefully there is nothing super wrong in there!).

glossary/quantum_differentiable_programming.rst
```diff
@@ -0,0 +1,242 @@
...
+
+
+Automatic differentiation
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you write an algorithm to compute some function :math:`f(x, y)` (which may
+include mathematical expressions, but also control flow statements like
+:code:`if`, :code:`for`, etc.), then automatic differentiation provides an
+algorithm for :math:`\nabla f(x, y)` with the same degree of complexity as the
+original function.
```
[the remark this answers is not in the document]
josh146: But yeah, maybe in this case it is safer to simply specify

```
:math:`f(x,y): R^2 \rightarrow R`
```

and "(reverse-mode) autodifferentiation"?

Since I don't want readers to assume that parameter-shift = autodifferentiation = same complexity as original function.

glossary/quantum_differentiable_programming.rst
```diff
@@ -0,0 +1,242 @@
...
+To perform automatic differentiation, functions are first translated into
+computational graphs in terms of elementary operations that have known
+derivatives, such as addition and multiplication.
+
+:html:`<br>`
+
+.. figure:: ../_static/concepts/autodiff_classical.png
+    :align: center
+    :width: 30%
+    :target: javascript:void(0);
```
josh146, line 135: This. is. so. awesome.

josh146 (approved):
@glassnotes I love the changes. And the pictures/animations are so awesome!!

I think it strikes a great balances of explaining the concepts to the reader without getting bogged down in complicated mathematics.

glossary/quantum_differentiable_programming.rst
```diff
@@ -0,0 +1,242 @@
...
+Computer algebra systems such as Mathematica perform symbolic differentiation
+--- if you ask it for the derivative of :math:`\sin(x)`, it will return to you
+explicitly the function :math:`\cos(x)`, and not a numerical value. Under the
+hood, a set of differentiation rules are implemented and followed. This includes
+things like how to differentiate constants, polynomials, sums, and chain rules,
+as well as derivatives of common functions (e.g., trigonometric functions). This
+is a very powerful tool because once the set of rules is implemented, we can
+symbolically differentiate arbitrary functions that are encompassed by
+them. However the scope of this method can be limited since it requires
+"hand-written" support for new functions.
```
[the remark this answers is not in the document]
glassnotes: This is actually something I was wondering about while writing it :) Thanks for the clarification here!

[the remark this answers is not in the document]
josh146: Wait I think I am wrong on the latter part of my sentence - "potentially an analytic solution may
not exist!"

This is only true for integrals right? For derivatives, if `f(x)` has a closed form solution, than so does `f'(x)` I think oops 🤦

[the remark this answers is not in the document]
glassnotes: For piecewise/continuous functions there may not be a derivative everywhere. But for simplicity we can just remove the line and leave it at "convoluted" :+1:

glossary/quantum_differentiable_programming.rst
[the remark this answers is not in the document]
glassnotes: I could have sworn I left a comment on this earlier... anyways, I adjusted things around this in the "notes" box actually, and thanked you for the clarification :) It is a good idea to point out that parameter shift is forward-mode autodiff.

PR #194, opened 2020-12-22
Add html anchor to community demo cards

josh146:
Adds HTML anchors/IDs to the community demo cards.

The IDs are of the form `DD-MM-YY-lastname`, where `lastname` is the last author name to appear on the card.

Please give it a go and let me know if it works! Simply append `#id` to the end of the URL (@co9olguy @doctorperceptron). For example:

https://2240-214003948-gh.circle-artifacts.com/0/_build/html/demos_community.html#26-11-2020-willems


demos_community.rst
```diff
@@ -182,3 +185,19 @@ Community
 .. toctree::
     :maxdepth: 2
     :hidden:
+
+.. raw:: html
+
+    <script type="text/javascript">
```
josh146, line 191: This javascript is to account for the navbar height, which sphinx is unaware of

demos_community.rst
```diff
@@ -182,3 +185,19 @@ Community
...
+        var divId;
+        divId = $(location).attr('hash');
+
+        $(window).on('load', function() {
+            if (divId) {
+                $(divId).addClass("active");
+                setTimeout(function(){
+                    $('html, body').animate({scrollTop: $(divId).offset().top - 60}, 0);
+                }, 500);
```
josh146, line 200: For some reason, the sphinx scrolling is very slow :( So I've had to add 500ms timeout here, so that the corrective navbar scrolling occurs _after_ the sphinx scrolling

glassnotes:
Thanks @josh146!  :tada:
 
Caught a few bugs while testing it out:
 - the top line of the title and card are cut off from the position of the anchor
 - I noticed that a green box appears around the selected demo when I first go to the page; if I then change the URL to a new demo while still on the page, the green box doesn't move, until I manually refresh the page
 - it doesn't seem to take duplicates into account; for example we have four demos on the same day from the same person, can we make it tack on numbers, or the first word of the title, if newly generated IDs are not unique? 
 - for two-word last names, it grabs only the last part of it (this not really important, but is something I personally run into quite often!)

josh146:
> the top line of the title and card are cut off from the position of the anchor

Oh, the corrective scrolling javascript didn't fix this on your end? 🤔 I'm not too sure... could you send a gif?

> I noticed that a green box appears around the selected demo when I first go to the page; if I then change the URL to a new demo while still on the page, the green box doesn't move, until I manually refresh the page

Yes, I noticed this as well, but I couldn't work out how to fix this 😆 There must be a jquery function that activates on URL changes.

> * it doesn't seem to take duplicates into account; for example we have four demos on the same day from the same person, can we make it tack on numbers, or the first word of the title, if newly generated IDs are not unique?
> * for two-word last names, it grabs only the last part of it (this not really important, but is something I personally run into quite often!)

These are both easy fixes, we can change the ID format. I suppose date and title is safer than date and author name?

glassnotes:
> Oh, the corrective scrolling javascript didn't fix this on your end? thinking I'm not too sure... could you send a gif?

community-card

> These are both easy fixes, we can change the ID format. I suppose date and title is safer than date and author name? 

As long as they don't all start with "quantum", should be okay!

josh146:
I can't seem to recreate it, I thought it might be a browser issue but it acts the same for me in firefox and chrome:

ezgif-6-7d3acd308fb1

co9olguy:
When I tested, topmost card text was hidden under navbar :frowning_face:

josh146:
CSS my old nemesis ☹️ 😿

PR #195, opened 2020-12-24
Adds tutorial on using the SPSA optimizer

antalszava:


trbromley:
[ch1031]

demonstrations/spsa.py
```diff
@@ -0,0 +1,822 @@
...
+# bounces around the optimum after a certain point.
+#
+# Of interest now are the number of device executions. Gradients need
+# to be computed for each of the 4 qubit rotations with 3 parameters. The
+# gradient of each parameter requires 2 evaluations, and this must be done over
+# all 15 terms. Factoring in the number of iteration steps (20), we expect the
+# total number of device executions
+#
+max_dev_execs = params.size * len(h2_ham.terms[0]) * 2 * max_iterations
+print("Expected device executions = {max_dev_execs}")
```
antalszava, line 578:
```suggestion
print(f"Expected device executions = {max_dev_execs}")
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,852 @@
...
+#
+# Our next step will be to run the same VQE simulation on our simulated version
+# of the Melbourne hardware. The entire process and cost function remain the
+# same; we'll simply swap out the device.
+#
+
+from qiskit import IBMQ
+from qiskit.providers.aer import noise
+
+# Note: do not run the simulation on this device, as it will send it to a real hardware
```
antalszava, line 619:
```suggestion
# Note: you will need to be authenticated to IBMQ to run the following code.
# Do not run the simulation on this device, as it will send it to a real hardware
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+plt.legend(fontsize=14)
+plt.show()
+
+##############################################################################
+#
+# .. figure:: ../demonstrations/spsa/first_comparison.png
+#     :align: center
+#     :width: 75%
+#
+# It seems that SPSA performs great and it does with a significant saving when
```
glassnotes, line 375:
```suggestion
# It seems that SPSA performs great and it does so with a significant savings when
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+# compared to gradient descent!
+#
+# Let's take a deeper dive to see how much better it is actually.
```
glassnotes, line 378:
```suggestion
# Let's take a deeper dive to see how much better it actually is.
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+##############################################################################
+#
+# We can see that as the optimization progresses, we approach the true value.
+# However, due to shot noise it bounces around the optimum.
+#
+# Of interest now are the number of device executions. Gradients need
+# to be computed for each of the 4 qubit rotations with 3 parameters. The
+# gradient of each parameter requires 2 evaluations, and this must be done over
+# all 15 terms. Factoring in the number of iteration steps (20), we expect the
+# total number of device executions
```
glassnotes, line 586:
```suggestion
# all 15 terms. Factoring in the number of iteration steps (20), we estimate the
# total number of device executions as follows:
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+# Note from the plot that the total number of executions is actually less than
+# this. This is because PennyLane is clever about taking the gradient, and will
+# not do so in cases where there is no dependence on the parameters. For example,
+# no gradients need to be computed for the Hamiltonian term that is simply `I`,
+# and there may be shortcuts for other Pauli terms as well.
+#
+# Gradient descent on simulated hardware
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Our next step will be to run the same VQE simulation on our simulated version
```
glassnotes, line 612:
```suggestion
# Our next step will be to run the same VQE simulation on a simulated version
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+#
+
+##############################################################################
+# Conclusion
+# ----------
+#
+# SPSA is a useful optimization technique that may be particularly beneficial on
+# near-term quantum hardware. It uses significantly fewer iterations to achieve
+# comparable result quality as gradient-based methods, giving it the potential
+# to save significant time and resources. It can be a good alternative to
```
glassnotes, line 815:
```suggestion
# to save time and resources. It can be a good alternative to
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+# gradient-based methods when the optimization problem involves executing
+# quantum circuits with many free parameters.
+#
+# There are also extensions to SPSA that could be interesting to explore in
+# this context. One, in particular, uses an adaptive technique to approximate
+# the *Hessian* matrix during optimization to effectively increase the
+# convergence rate of SPSA. [#spall_overview]_ The proposed technique can also
```
glassnotes, line 822:
```suggestion
# convergence rate of SPSA [#spall_overview]_. The proposed technique can also
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+
+##############################################################################
+# References
+# ----------
+#
+# .. [#spall_overview]
+#
+#    James C. Spall, "An Overview of the Simultaneous Perturbation Method
+#    for Efficient Optimization."
+#    `<https://www.jhuapl.edu/SPSA/PDF-SPSA/Spall_An_Overview.PDF>`__, 1998
```
glassnotes, line 836:
```suggestion
#    James C. Spall, "`An Overview of the Simultaneous Perturbation Method
#    for Efficient Optimization <https://www.jhuapl.edu/SPSA/PDF-SPSA/Spall_An_Overview.PDF>`__", 1998
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,853 @@
...
+As there would be two circuit evaluation for each free parameter, the number of
+overall quantum circuit executions for computing a quantum gradient scales
+linearly with the number of free parameters, i.e., :math:`O(p)` with :math:`p` being
+the number of free parameters. This scaling can be very costly for optimization
+tasks with many free parameters. For the overall optimization this scaling means
+we need :math:`O(pn)` quantum circuit evaluations, where :math:`n` is the number of
+optimization steps taken.
+
+There are, however, certain optimization techniques that are gradient-free and
+hence offer an alternative approach for computing the gradients of quantum circuits.
```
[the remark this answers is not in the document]
antalszava: Thanks! Rephrased this sentence.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+
+.. figure:: ../demonstrations/spsa/spsa_opt.png
+   :align: center
+   :width: 60%
+
+   ..
+
+   A schematic of the search paths used by gradient descent with
+   parameter-shift and SPSA in a low-noise setting.
+   Image source: [#spall_overview]_
```
josh146, line 118: If you would like, I can contribute a hand drawn image for this one!

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+
+    To run this demo locally, you'll need to install the `noisyopt
+    <https://github.com/andim/noisyopt>`_ library. This library contains a
+    straightforward implementation of SPSA that can be used in the same way as the
+    optimizers available in `SciPy's minimize method
+    <https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html>`_.
+
+First, let's consider a simple quantum circuit on a sampling device. For this,
+we'll be using a device from the `PennyLane-Qiskit plugin
+<https://pennylaneqiskit.readthedocs.io/en/latest/>`_ that samples quantum
```
josh146, line 137: How come you choose pl-qiskit here?

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+.. note::
+
+    Just with other PennyLane devices, the number of samples taken for a device
+    execution can be specified using the ``shots`` keyword argument of the
+    device.
+
+Once we have a device selected, we just need a couple of other ingredients for
+the pieces of an example optimization to come together:
+
+* a template: :func:`~.pennylane.templates.layers.StronglyEntanglingLayers`,
```
josh146, line 150: Maybe good to instead say 'circuit ansatz' here?

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+
+dev_sampler = qml.device("qiskit.aer", wires=num_wires, shots=1000)
+
+##############################################################################
+# We seed so that we can simulate the same circuit every time.
+np.random.seed(50)
+
+all_pauliz_tensor_prod = qml.operation.Tensor(*[qml.PauliZ(i) for i in range(num_wires)])
+
+
```
josh146, line 170: I know that this is added by Black, but I find that the double line breaks are a bit wasted on rendered websites...

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+def circuit(params):
+    qml.templates.StronglyEntanglingLayers(params, wires=list(range(num_wires)))
+    return qml.expval(all_pauliz_tensor_prod)
+
+
+##############################################################################
+# After this, we'll initialize the parameters in a way that is compatible with
+# the ``noisyopt`` package. We are flattening our parameters, which will be very
+# convenient later on when using the SPSA optimizer. Just keep in mind that
+# this is done for compatibility.
```
josh146, line 181:
```suggestion
# After this, we'll initialize the parameters in a way that is compatible with
# the ``noisyopt`` package. The ``noisyopt`` package requires the trainable parameters
# be a flattened array. As a result, our cost function must accept a flat array of parameters
# to be optimized.
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+##############################################################################
+# Once we have defined each piece of the optimization, there's only one
+# remaining component required for the optimization: the *SPSA optimizer*.
+# We'll use the SPSA optimizer provided by the ``noisyopt`` package. Once
+# imported, we can initialize parts of the optimization such as the number of
+# iterations, a collection to store the cost values, and a callback function.
+# Once the optimization has concluded, we save the number of device executions
+# required for completion using the callback function. This will be an
+# interesting quantity!
+from noisyopt import minimizeSPSA
```
josh146, line 200: It's definitely cool showing off using `noisyopt` with PL, but a part of me was slightly disappointed, as I thought you were going to show how to code up SPSA 😆 

I wonder if it is worth having a very basic SPSA implementation at the end of the "Simultaneous perturbation stochastic approximation (SPSA)" section, but then say 'now lets use ``noisyopt``...' in this section.

josh146: In fact, having a basic version of SPSA coded up in the previous section (even to optimize a basic classical function) would be very nice to help understand the SPSA math?

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+
+niter_spsa = 200
+
+# Evaluate the initial cost
+cost_store_spsa = [cost(init_params)]
+device_execs_spsa = [0]
+dev_sampler._num_executions = 0
```
josh146, line 207: Shouldn't this already be 0?

josh146: I hesitate a bit having a private attribute being set

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+
+
+##############################################################################
+# Choosing the hyperparameters
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# The ``noisyopt`` package allows us to choose the initial value of two
+# hyperparameters for SPSA: the :math:`c` and :math:`a` coefficients. Recall
+# from above that the :math:`c` values control the amount of random shift when
+# evaluating the cost function, while the :math:`a` is analogous to a learning
```
josh146, line 229:
```suggestion
# evaluating the cost function, while the :math:`a` coefficients are analogous to a learning
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+#     Iteration = 10, Cost = -0.998
+#     Iteration = 11, Cost = -0.992
+#     Iteration = 12, Cost = -0.994
+#     Iteration = 13, Cost = -1.0
+#     Iteration = 14, Cost = -0.996
+#     Iteration = 15, Cost = -0.996
+#     Iteration = 16, Cost = -0.998
+#     Iteration = 17, Cost = -0.996
+#     Iteration = 18, Cost = -0.996
+#     Iteration = 19, Cost = -0.996
```
josh146, line 342: It might be useful to add `Quantum evaluations:` to this list, since, at first though, I thought that GD was doing better! Since it converges to -0.996 after 19 iterations, but SPSA takes 200.

But then I remembered that 19 iterations for grad descent -> a _very_ large number of parameter shifts

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+# not do so in cases where there is no dependence on the parameters. For example,
+# no gradients need to be computed for the Hamiltonian term that is simply `I`,
+# and there may be shortcuts for other Pauli terms as well.
+#
+# Gradient descent on simulated hardware
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Our next step will be to run the same VQE simulation on a simulated version
+# of the Melbourne hardware. The entire process and cost function remain the
+# same; we'll simply swap out the device.
```
josh146, line 612: I worry it feels like this section (and the previous one) feels like a bit of a divergence --- it suddenly feels like a VQE demo, and the focus is no longer on SPSA. Is it possible to combine the two sections and reduce the explanations where it makes sense (instead linking to other demos), so that VQE + SPSA can be emphasized?

josh146 (commented):
A really awesome tutorial @antalszava!

I left comments throughout, but I think writing an awesome introduction, and shortening the VQE section, would really help elevate the demo 🙂

demonstrations/spsa.py
antalszava: Sure, if it would help with putting the theory into context. The main aim of the demo was to show that SPSA can be useful, so using `noisyopt` was a convenient way to meet the optimizer and to offer an introduction to it. But if it brings value to have a custom implementation, then will swap it.

demonstrations/spsa.py
antalszava: We use noise models and we can simulate a real device later using PL-Qiskit.

demonstrations/spsa.py
antalszava: That would be awesome, thank you! 🙂

demonstrations/spsa.py
antalszava: We used `dev_sampler` for gradient descent too, so the number of executions was recorded. A clear way can be just to create a new device.

antalszava:
Thank you for the comments @josh146! :)

demonstrations/spsa.py
josh146: This is definitely a subjective opinion, but I definitely expected to see SPSA coded up based on the intro! It might be quite a bit of extra work, I'm curious what @glassnotes thinks?

glassnotes: It would take some effort but it is not too challenging to implement (plus, it would save us a dependency where we had to do a bit of tweaking to get the parameter shapes to fit right, I can't remember the details). We can also add it afterwards to the selection of optimizers in PL. The biggest concern is just time, since we were wanting to get the demo out ASAP.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+vector :math:`\hat{\theta}_{0}`. After :math:`k` iterations, the :math:`(k+1)` th
+parameter iterates can be obtained as
+
+.. math:: \hat{\theta}_{k+1} = \hat{\theta}_{k} - a_{k}\hat{g}_{k}(\hat{\theta}_{k}),
+
+where :math:`\hat{g}_{k}` is the estimate of the gradient :math:`g(u) = \frac{
+\partial L}{\partial \theta}` at the iterate :math:`\hat{\theta}_{k}` based on
+prior measurements of the cost function, and :math:`a_{k}` is a positive number.
+
+As previously mentioned, SPSA is robust to any noise that may occur
```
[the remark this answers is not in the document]
antalszava: It was hinted at on line 48, will rephrase though to make it less ambiguous.

demos_research.rst
```diff
@@ -195,6 +195,13 @@ plateaus in quantum neural networks.
     :description: :doc:`demos/tutorial_measurement_optimize`
     :tags: chemistry
 
+.. customgalleryitem::
+    :tooltip: Reduce the number of device executions by using a stochastic approximation optimization.
+    :figure: demonstrations/spsa/spsa_thumbnail.png
```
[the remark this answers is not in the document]
antalszava: Updated!

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+
+As there would be two circuit evaluations for each free parameter, the number of
+overall quantum circuit executions for computing a quantum gradient scales
+linearly with the number of free parameters, i.e., :math:`O(p)` with :math:`p` being
+the number of free parameters. This scaling can be very costly for optimization
+tasks with many free parameters. For the overall optimization this scaling means
+we need :math:`O(pn)` quantum circuit evaluations, where :math:`n` is the number of
+optimization steps taken.
+
+Fortunately, there are certain optimization techniques that offer an
```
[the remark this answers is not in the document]
antalszava: Added an intro.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+h2_ham, num_qubits = qchem.molecular_hamiltonian(
+    name,
+    geometry,
+    charge=charge,
+    mult=multiplicity,
+    basis=basis_set,
+    active_electrons=2,
+    active_orbitals=2,
+    mapping="jordan_wigner",
+)
```
agran2018, line 428: Here is another tutorial to update in terms of the new ``molecular_hamiltonian`` function :-)). Also, if we want to avoid deviating the attention of the user to the qchem specifics I would suggest to keep it as simple as possible. For example:
```python
h2_ham, num_qubits = qchem.molecular_hamiltonian(name, geometry) 
```
It can be done later when we update the tutorial.

demonstrations/spsa.py
```diff
@@ -0,0 +1,848 @@
...
+# with a cost function that computes the expectation value of each Hamiltonian
+# term after running the ansatz circuit. We'll also set the maximum number of
+# gradient descent iterations to 20.
+#
+# To set the hyperparameters, we refer back to the `intro VQE demo
+# </demos/tutorial_vqe>`_, which specifies a random seed to choose variational
+# parameters that will converge well. It is common to have to run the VQE
+# multiple times, with different sets of initial parameters, to ensure
+# convergence. If you try out the code yourself, you'll find that there is a
+# local minimum around :math:`-0.47` Hartree that the gradient descent
```
agran2018, line 489: This is tangential. This is the lowest excited state of H2 with S=1, S_z=0 :-)). See [HERE](https://pennylane.ai/qml/demos/tutorial_vqe_uccsd_obs.html)

agran2018 (changes requested):
Thanks @antalszava for the nice demo!. I have left few comments/suggestions. Happy to give it a second round when is ready.

demonstrations/spsa.py
antalszava: Some things can be cut down when loading H2, though the biggest gain would be omitting optimization with the ideal (non-noisy device). Optimization using that was already showcased in the VQE intro tutorial.

Just double-checking that that's something valuable to leave out, also tagging @glassnotes.

demonstrations/spsa.py
josh146: > The biggest concern is just time, since we were wanting to get the demo out ASAP.

yep not a big issue at all, happy to keep the demo as is :+1:

Perhaps it's just a matter of reframing the intro, which made me think 'this demo will show me how to code SPSA' instead of 'this demo will showcase how using SPSA can benefit optimization'

demonstrations/spsa.py
josh146: > Optimization using that was already showcased in the VQE intro tutorial.

If that's the case, could we just link to that demo very prominently, and then go straight into the noisy optimization?

demonstrations/spsa.py
josh146: image

Gave it a shot, let me know if you want any changes!

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+Background
+----------
+
+In PennyLane, quantum gradients on hardware are commonly computed using parameter-shift rules.  For
+quantum circuits that have multiple free parameters, using these parameter-shift
+rules to compute quantum gradients involves computing the partial derivatives of
+the quantum function with respect to every free parameter. These partial
+derivatives are then used to apply the product rule when computing the quantum
+gradient (see `parameter-shift rules
+<https://pennylane.ai/qml/glossary/parameter_shift.html>`_). For qubit
```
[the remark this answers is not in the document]
antalszava: Thanks! Based on your suggestion, updating to:
>Computing quantum gradients involves evaluating the partial derivative of the
quantum function with respect to every free parameter. The partial derivatives
are used to apply the product rule to compute the gradient of the quantum
circuit.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+operations that are generated by one of the Pauli matrices, each partial
+derivative computation will involve two quantum circuit evaluations with a
+positive and a negative shift in the parameter values.
+
+As there are two circuit evaluations for each free parameter, the number of
+overall quantum circuit executions for computing a quantum gradient scales
+linearly with the number of free parameters, i.e., :math:`O(p)` with :math:`p` being
```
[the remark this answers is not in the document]
antalszava: Thanks rephrased this part.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+the number of free parameters. This scaling can be very costly for optimization
+tasks with many free parameters. For the overall optimization this scaling means
+we need :math:`O(pn)` quantum circuit evaluations, where :math:`n` is the number of
+optimization steps taken.
+
+Fortunately, there are certain optimization techniques that offer an
+alternative to computing the gradients of quantum circuits. One such technique
```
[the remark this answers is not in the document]
antalszava: Not quite sure about this suggestion.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+is called the Simultaneous Perturbation Stochastic Approximation (SPSA)
+algorithm. SPSA is an optimization method that involves *approximating* the
```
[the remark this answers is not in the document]
antalszava: Added.

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+:math:`\theta` --- this is exactly the problem that we'd consider when optimizing
+quantum functions!
+
+Just like with gradient-based methods, SPSA starts with an initial parameter
+vector :math:`\hat{\theta}_{0}`. After :math:`k` iterations, the :math:`(k+1)` th
+parameter iterates can be obtained as
+
+.. math:: \hat{\theta}_{k+1} = \hat{\theta}_{k} - a_{k}\hat{g}_{k}(\hat{\theta}_{k}),
+
+where :math:`\hat{g}_{k}` is the estimate of the gradient :math:`g(u) = \frac{
```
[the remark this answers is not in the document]
antalszava: Good catch! Should be `\theta`. Thanks

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+# hyperparameters for SPSA: the :math:`c` and :math:`a` coefficients. Recall
+# from above that the :math:`c` values control the amount of random shift when
+# evaluating the cost function, while the :math:`a` coefficients are analogous to a learning
+# rate and affects the degree to which the parameters change at each update
+# step.
+#
+# With stochastic approximation, specifying such hyperparameters significantly
+# influences the convergence of the optimization for a given problem. Although
+# there is no universal recipe for selecting these values (as they are greatly
+# dependent on the specific problem), [#spall_implementation]_ includes
```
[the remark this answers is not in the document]
antalszava: Choosing the hyperparameters for the optimizers (e.g., step size) is problem dependent and for stochastic approximation methods, it is, even so, the case.

demonstrations/spsa.py
antalszava: Good idea, thanks! Just double-checking:
```
    active_electrons=2, active_orbitals=2
```
would be default?

Also, is the `molecular_hamiltonian` function due to be changed? (not sure about this)

demonstrations/spsa.py
```diff
@@ -0,0 +1,850 @@
...
+In this demo, you'll learn how the SPSA algorithm works, and how to apply it in
+PennyLane to compute gradients of quantum circuits. You'll also see it in action
+using noisy quantum data!
+
+Simultaneous perturbation stochastic approximation (SPSA)
+---------------------------------------------------------
+
+SPSA is a general method for minimizing differentiable multivariate functions.
+It is particularly useful for functions for which evaluating the gradient is not
+possible, or too resource intensive. SPSA provides a stochastic method for approximating the gradient of a
```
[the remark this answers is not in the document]
antalszava: We know that quantum functions are differentiable, yet their gradients are not necessarily available. The parameter-shift rules are neat tricks that help us get the gradients, but they are not completely general. Therefore there are cases indeed, where we know that a function is differentiable, can evaluate the function itself (running the circuit on a quantum computer) but we have no information about its gradient (cannot run a quantum experiment that yields the gradient of the quantum function).

On another note, the demo is more focused on the other example hinted on:
>  evaluating the gradient is not possible, **or too resource intensive**.

Let me know if it's good to incorporate parts of this description in the text.

demonstrations/spsa.py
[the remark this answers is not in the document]
agran2018: No problem Antal, this is a matter of style, totally optional.

demonstrations/spsa.py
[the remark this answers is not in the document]
agran2018: oh yes. Sorry Antal for my ambigous comment. What I wanted to suggest was to rephrase the sentence within parenthesis, also optional.

agran2018 (commented):
Thanks @antalszava!, looks good to me.

demonstrations/spsa.py
agran2018: The default for ``active_electrons`` and ``active_orbitals`` is ``None`` which means to consider all electrons and orbitals as active. In the case of H2 in a minimal basis they are both equal to ``2``.

demonstrations/spsa.py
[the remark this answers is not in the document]
agran2018: I see. Thanks @antalszava for the explanation.

demonstrations/spsa.py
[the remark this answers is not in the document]
antalszava: Oh got it! Rephrased

demonstrations/spsa.py
co9olguy: I don't know what it is, this optimization landscape frightens me :fearful:

antalszava: Oh gosh @josh146, completely missed this thread on my first pass today and only found it after our earlier chat from today! 😲 Thank you for the drawing!!! 😊 🎊

-----
It would look as follows, perhaps making the two optimization paths, the dots for each data point and the font bigger could help with identifying what the picture showcases?


image

demonstrations/spsa.py
antalszava: Thanks! Added

demonstrations/spsa.py
antalszava: Removed the occurrences of `dev_sampler._num_executions`

demonstrations/spsa.py
antalszava: Got it! 🙂  Maybe it's better to keep it as is for now and could update later if we'd like.

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+
+.. related::
+
+   tutorial_vqe A brief overview of VQE
+   tutorial_vqe_qng Accelerating VQE with the QNG
+
+*Author: PennyLane dev team. Posted: 19 Mar 2021. Last updated: 19 Mar 2021.*
+
+In this tutorial, we investigate using a gradient-free optimizer called
+Simultaneous Perturbation Stochastic Approximation (SPSA) to optimize quantum
```
glassnotes, line 19:
```suggestion
the Simultaneous Perturbation Stochastic Approximation (SPSA) algorithm to optimize quantum
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+Just like with gradient-based methods, SPSA starts with an initial parameter
+vector :math:`\hat{\theta}_{0}`. After :math:`k` iterations, the :math:`(k+1)` th
+parameter iterates can be obtained as
+
+.. math:: \hat{\theta}_{k+1} = \hat{\theta}_{k} - a_{k}\hat{g}_{k}(\hat{\theta}_{k}),
+
+where :math:`\hat{g}_{k}` is the estimate of the gradient :math:`g(\theta) =
+\frac{ \partial L}{\partial \theta}` at the iterate :math:`\hat{\theta}_{k}`
+based on prior measurements of the cost function, and :math:`a_{k}` is a
+positive number. [#spall_overview]_
```
glassnotes, line 99:
```suggestion
positive number [#spall_overview]_.
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+
+First, let's consider a simple quantum circuit on a sampling device. For this,
+we'll be using a device from the `PennyLane-Qiskit plugin
+<https://pennylaneqiskit.readthedocs.io/en/latest/>`_ that samples quantum
+circuits to get measurement outcomes and later post-processes these outcomes to
+compute statistics like expectation values.
+
+.. note::
+
+    Just with other PennyLane devices, the number of samples taken for a device
```
glassnotes, line 156:
```suggestion
    Just as with other PennyLane devices, the number of samples taken for a device
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+
+dev_sampler_spsa = qml.device("qiskit.aer", wires=num_wires, shots=1000)
+
+##############################################################################
+# We seed so that we can simulate the same circuit every time.
+np.random.seed(50)
+
+all_pauliz_tensor_prod = qml.operation.Tensor(
+    *[qml.PauliZ(i) for i in range(num_wires)]
+)
```
glassnotes, line 183:
```suggestion
all_pauliz_tensor_prod = qml.operation.Tensor(*[qml.PauliZ(i) for i in range(num_wires)])
```

This got expanded when I ran the code through Black (with a conservative number of lines), so we can either keep as is or put back to the original - it doesn't cause a horizontal scrollbar.

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+qnode_spsa = qml.QNode(circuit, dev_sampler_spsa)
+
+
+def cost_spsa(params):
+    return qnode_spsa(params.reshape(num_layers, num_wires, 3))
+
+
+##############################################################################
+# Once we have defined each piece of the optimization, there's only one
+# remaining component required for the optimization: the *SPSA optimizer*.
```
glassnotes, line 211:
```suggestion
# remaining component required: the *SPSA optimizer*.
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+
+##############################################################################
+# Choosing the hyperparameters
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# The ``noisyopt`` package allows us to choose the initial value of two
+# hyperparameters for SPSA: the :math:`c` and :math:`a` coefficients. Recall
+# from above that the :math:`c` values control the amount of random shift when
+# evaluating the cost function, while the :math:`a` coefficients are analogous to a learning
+# rate and affects the degree to which the parameters change at each update
```
glassnotes, line 252:
```suggestion
# rate and affect the degree to which the parameters change at each update
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+#     Iteration = 170, Number of device executions = 254, Cost = -0.974
+#     Iteration = 180, Number of device executions = 269, Cost = -0.962
+#     Iteration = 190, Number of device executions = 284, Cost = -0.988
+#     Iteration = 200, Number of device executions = 299, Cost = -0.964
+
+##############################################################################
+#
+# Now let's perform the same optimization using gradient descent. We set the
+# step size according to a favourable value found after grid search for fast
+# convergence. Note that we also reset the number of executions of the device.
```
glassnotes, line 311:
```suggestion
# convergence. Note that we also create a new device in order to reset the execution count to 0. 
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+#     Iteration = 17, Number of device executions = 2178, Cost = -0.996
+#     Iteration = 18, Number of device executions = 2299, Cost = -0.996
+#     Iteration = 19, Number of device executions = 2420, Cost = -0.996
+#
+
+##############################################################################
+# SPSA and gradient descent comparison
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# At this point, nothing else remains, but to check which of these approaches did
```
glassnotes, line 377:
```suggestion
# At this point, nothing else remains but to check which of these approaches did
```

demonstrations/spsa.py
```diff
@@ -0,0 +1,734 @@
...
+#     Iteration = 15, Number of device executions = 5328,  Energy = -0.99592924 Ha
+#
+#     Final estimated value of the ground-state energy = -0.98134253 Ha
+#     Accuracy with respect to the true energy: 0.15484692 Ha
+#
+
+
+plt.figure(figsize=(10, 6))
+
+plt.scatter(
```
glassnotes, line 530:
```suggestion
plt.plot(
```

demonstrations/spsa.py
antalszava: Thank you for checking that! The one-liner seems good

demonstrations/spsa.py
```diff
@@ -0,0 +1,732 @@
...
+
+This demonstration shows how the SPSA optimizer performs:
+
+1. A simple task on a sampling device,
+2. The variational quantum eigensolver on a simulated hardware device.
+
+Throughout the demo, we show results obtained with SPSA and with gradient
+descent and also compare the number of device executions required to complete
+each optimization.
+
```
agran2018, line 31: Nice condensed intro!

demonstrations/spsa.py
```diff
@@ -0,0 +1,732 @@
...
+h2_spsa_device_executions_melbourne = [0]
+h2_spsa_energies_melbourne = [wrapped_cost(params)]
+
+
+def callback_fn(xk):
+    cost_val = wrapped_cost(xk)
+    h2_spsa_energies_melbourne.append(cost_val)
+
+    # TODO: update comment if verified
```
agran2018, line 587: should this comment be updated/deleted?

agran2018 (approved):
Nice tutorial!, thank you @antalszava, @glassnotes !. I just left a comment regarding line 587

``# TODO: update comment if verified``

demonstrations/spsa.py
antalszava: Good catch, thanks!

PR #196, opened 2020-12-26
Modify the multiclass classifier demo to allow it to run with an arbitrary number of wires

YumiA:
The modifications aim to make the algorithm generic that it can deal with any number of features, hence any number of qubits.

**Title:** Make the code more generic

**Summary:** To work with any number of features without the need of changing the code.

**Relevant references:** 

**Possible Drawbacks:** The evaluation of the circuit output may not be the optimal one.

**Related GitHub Issues:**


demonstrations/tutorial_multiclass_classification.py
```diff
@@ -90,14 +91,21 @@ def layer(W):
...
-qnode1 = qml.QNode(circuit, dev).to_torch()
-qnode2 = qml.QNode(circuit, dev).to_torch()
-qnode3 = qml.QNode(circuit, dev).to_torch()
+    even_wirelist = [i for i in range(num_qubits) if i%2 == 0]
+    PZ = qml.PauliZ(0)
+    for wi in even_wirelist[1:]:
+        PZ = PZ@qml.PauliZ(wi)
+    if(num_qubits%2!=0):
+        PZ = PZ@qml.PauliZ(num_qubits-1)
+    return qml.expval(PZ)
```
antalszava, line 104: Might be worth running `black -l 100 tutorial_multiclass_classification.py` on this file to pick up some formatting details here like adding whitespace (might change some further parts of the script, which is okay).

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -76,9 +76,10 @@
 # entangling/CNOT gates
 
 def layer(W):
-    qml.Rot(W[0, 0], W[0, 1], W[0, 2], wires=0)
-    qml.Rot(W[1, 0], W[1, 1], W[1, 2], wires=1)
-    qml.CNOT(wires=[0, 1])
+    for i in range(num_qubits):
+        qml.Rot(W[i, 0], W[i, 1], W[i, 2], wires=i)
+    for j in range(int(num_qubits/2)):
+        qml.CNOT(wires=[j*2, j*2+1])
```
antalszava, line 82: This approach will pairwise entangle qubits that is qubits 0-1, qubits 2-3, etc will be entangled. Would that be what we'd aim for here?

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -199,7 +207,7 @@ def accuracy(labels, hard_predictions):
 
 def load_and_process_data():
     data = np.loadtxt("multiclass_classification/iris.csv", delimiter=",")
-    X = torch.tensor(data[:, 0:feature_size])
+    X = torch.tensor(data[:, 1:feature_size+1])
```
antalszava, line 210: How come this was changed? It seems like now we'd classify features 1-4 instead of 0-3.

antalszava (changes requested):
Hi @YumiA :slightly_smiling_face: Thank you so much for looking into this!

Left a couple of suggestions and questions for my own understanding. One of the key things to consider would be how exactly the architecture generalizes for more than 4 features.

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -90,14 +91,21 @@ def layer(W):
...
-    qml.templates.embeddings.AmplitudeEmbedding(feat, [0, 1], pad=0.0, normalize=True)
+    qml.templates.embeddings.AmplitudeEmbedding(feat, range(0,num_qubits), pad=(2**num_qubits-feature_size), normalize=True)
     for W in weights:
         layer(W)
-    return qml.expval(qml.PauliZ(0))
 
-qnode1 = qml.QNode(circuit, dev).to_torch()
-qnode2 = qml.QNode(circuit, dev).to_torch()
-qnode3 = qml.QNode(circuit, dev).to_torch()
+    even_wirelist = [i for i in range(num_qubits) if i%2 == 0]
```
albi3ro, line 98: Thanks for the pull request!

You might want to add some commenting here to explain what is going on.

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -90,14 +91,21 @@ def layer(W):
 # Data is embedded in each circuit using amplitude embedding:
 
 def circuit(weights, feat=None):
-    qml.templates.embeddings.AmplitudeEmbedding(feat, [0, 1], pad=0.0, normalize=True)
+    qml.templates.embeddings.AmplitudeEmbedding(feat, range(0,num_qubits), pad=(2**num_qubits-feature_size), normalize=True)
```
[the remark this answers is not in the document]
YumiA: Hi @antalszava :) and thanks for the review and the suggestions.
Yes there is a reason for using the formula ((2**num_qubits-feature_size); as per my understanding the AmplitudeEmbedding function needs to have exactly 2^n numbers (n = # of qubits) to encode them and the pad is used to complement the features in case their size was less than 2^n.

demonstrations/tutorial_multiclass_classification.py
YumiA: This is was done by mistake as I used it to run one some dataset whose first column was just a sequence number.
```suggestion
    X = torch.tensor(data[:, 0:feature_size])
```

demonstrations/tutorial_multiclass_classification.py
YumiA: I'm not sure how to do that, maybe because I'm new here!

demonstrations/tutorial_multiclass_classification.py
YumiA: When I ran the code as is after increasing the number of qubits to fit my dataset, I got bad accuracy. After I applied the above modifications, I got 98% accuracy. Could you please suggest and explain the right approach?

demonstrations/tutorial_multiclass_classification.py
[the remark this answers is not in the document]
antalszava: Sure thing! :slightly_smiling_face: 

>  as per my understanding the AmplitudeEmbedding function needs to have exactly 2^n numbers (n = # of qubits) to encode them 

Correct!

> the pad is used to complement the features in case their size was less than 2^n.

Padding indeed involves making up for the case when `m<2^n` where `m` is the number of features and `n` is the number qubits. The `pad` argument, however, defines the constant to pad with rather than the remaining values that could be embedded (it does not stand for the difference `d` of `m+d==2^n`, but rather the padding value). See the `pad` and `pad_with` arguments in the [latest version of the `AmplitudeEmbedding` documentation](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.templates.embeddings.AmplitudeEmbedding.html) for more details.

demonstrations/tutorial_multiclass_classification.py
antalszava: No worries there! `black` is a formatting tool for Python that reformats Python code such that the result conforms with PEP8 coding standards. We use that for developing PennyLane. In this case, it could help with adding some whitespace in the code.

When using `pip`, `black` can be installed by running `pip install black` from the command line. Once installed, the command `black` is available to be used. `black -l 100 tutorial_multiclass_classification.py` can then be run in the folder where the `tutorial_multiclass_classification.py` file is located.

You can find more details on submitting a Pull Request [here](https://pennylane.readthedocs.io/en/latest/development/guide/pullrequests.html).

demonstrations/tutorial_multiclass_classification.py
antalszava, line 210: Got it! Maybe it could be reverted then. The reason is that the demo continues to use the Iris dataset, which could also potentially have `feature_size==5` with `data[:, 0:feature_size]`, whereas `data[:, 1:feature_size+1]` would result in an `IndexError` as the first features were skipped.

```suggestion
    X = torch.tensor(data[:, 0:feature_size])
```

antalszava: Got it! Reverted this then. The reason is that the demo continues to use the Iris dataset, which could also potentially have `feature_size==5` with `data[:, 0:feature_size]`, whereas `data[:, 1:feature_size+1]` would result in an `IndexError` as the first features were skipped.

demonstrations/tutorial_multiclass_classification.py
antalszava: > When I ran the code as is after increasing the number of qubits to fit my dataset, I got bad accuracy.

Would that mean only having `qml.CNOT(wires=[0, 1])`?

> Could you please suggest and explain the right approach?

Sure! It's great that the accuracy was already such. Another step forward would be to apply even more `CNOT` operations. In the case of having two qubits, applying a single `CNOT` can create entanglement between qubits 0 and 1.

However, as the number of qubits increases, the ability to create pairwise entanglement between any two qubits of the entire system requires applying a CNOT operation between every qubit. That is, for three qubits, we'll need to apply three `CNOT`s, for four qubits four, etc.

This is demonstrated in the [`StronglyEntanglingLayers` template](https://pennylane.readthedocs.io/en/stable/code/api/pennylane.templates.layers.StronglyEntanglingLayers.html?highlight=strongly#pennylane.templates.layers.StronglyEntanglingLayers). This template consists:
1. Applying arbitrary single-qubit rotations on each qubit
2. Pairwise entangling every qubit

As a matter of fact, this code snippet actually corresponds to using `StronglyEntanglingLayers` with `W` weights! Therefore, having a logic similar as [here](https://github.com/PennyLaneAI/pennylane/blob/master/pennylane/templates/layers/strongly_entangling.py#L100) could work: applying CNOTs in a ring structure between every two qubit.

antalszava:
Hi @YumiA, left a couple of comments, hope they help! The two checks fail due to something unrelated to the changes made in this pull request, we are looking into solving that.

demonstrations/tutorial_multiclass_classification.py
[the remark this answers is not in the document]
YumiA: Got it, thank you. But wouldn't be better to use something like the following:
```suggestion
    qml.templates.embeddings.AmplitudeEmbedding(feat, range(0,num_qubits), pad=np.mean(feat), normalize=True)
```

demonstrations/tutorial_multiclass_classification.py
YumiA: Thanks again for the explanation. I ran the tool and here is the result:
```suggestion
    even_wirelist = [i for i in range(num_qubits) if i % 2 == 0]
    PZ = qml.PauliZ(0)
    for wi in even_wirelist[1:]:
        PZ = PZ @ qml.PauliZ(wi)
    if num_qubits % 2 != 0:
        PZ = PZ @ qml.PauliZ(num_qubits - 1)
    return qml.expval(PZ)
```
!

demonstrations/tutorial_multiclass_classification.py
YumiA: So this should achieve the required strong entanglement:
```suggestion
    for j in range(int(num_qubits/2)):
        qml.CNOT(wires=[j*2, j*2+1])
```

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -76,9 +76,10 @@
 # entangling/CNOT gates
 
 def layer(W):
-    qml.Rot(W[0, 0], W[0, 1], W[0, 2], wires=0)
-    qml.Rot(W[1, 0], W[1, 1], W[1, 2], wires=1)
-    qml.CNOT(wires=[0, 1])
+    for i in range(num_qubits):
+        qml.Rot(W[i, 0], W[i, 1], W[i, 2], wires=i)
+    for j in range(int(num_qubits - 1)):
+        qml.CNOT(wires=[j, j + 1])
```
antalszava, line 82: 1. As `num_qubits` is an integer, we can get away with not having a type conversion.
2. To complete the "ring" pattern of CNOTs, we need to apply a CNOT such that the last qubit is the control and the first qubit is the target.

(Note: worth trying that the suggestion indeed works correctly, did not run it.)
```suggestion
    for j in range(num_qubits - 1):
        qml.CNOT(wires=[j, j + 1])
    if num_qubits >=2:
        # Apply additional CNOT to complete the ring pattern
        qml.CNOT(wires=[num_qubits - 1, 0])    
```

demonstrations/tutorial_multiclass_classification.py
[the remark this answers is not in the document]
antalszava: Not quite. `np.mean(feat)` could actually be a value mistaken to be a feature. This is something we'd like to avoid by using "non-informative constants", see reasons for this in the paragraph after equation (2) in section 2. A on page 4 from the [second ref](https://arxiv.org/pdf/1804.00633.pdf) of the original tutorial.

demonstrations/tutorial_multiclass_classification.py
antalszava: See this [comment](https://github.com/PennyLaneAI/qml/pull/196#discussion_r553410032).

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -90,14 +94,21 @@ def layer(W):
...
 
-qnode1 = qml.QNode(circuit, dev).to_torch()
-qnode2 = qml.QNode(circuit, dev).to_torch()
-qnode3 = qml.QNode(circuit, dev).to_torch()
+    even_wirelist = [i for i in range(num_qubits) if i % 2 == 0]
+    PZ = qml.PauliZ(0)
+    for wi in even_wirelist[1:]:
+        PZ = PZ @ qml.PauliZ(wi)
+    if num_qubits % 2 != 0:
+        PZ = PZ @ qml.PauliZ(num_qubits - 1)
```
antalszava, line 106: One more question here: how come only even wires were measured here? (And additionally the last wire `if num_qubits % 2 != 0`) @mariaschuld might also have insights here

antalszava:
@YumiA thanks for the changes! This is looking great, only had one last question from my side. After no more modifications are being done, it might be worth running `black` for another time.

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -76,9 +76,13 @@
...
 def layer(W):
-    qml.Rot(W[0, 0], W[0, 1], W[0, 2], wires=0)
-    qml.Rot(W[1, 0], W[1, 1], W[1, 2], wires=1)
-    qml.CNOT(wires=[0, 1])
+    for i in range(num_qubits):
+        qml.Rot(W[i, 0], W[i, 1], W[i, 2], wires=i)
+    for j in range(num_qubits - 1):
+        qml.CNOT(wires=[j, j + 1])
+    if num_qubits >=2:
+        # Apply additional CNOT to complete the ring pattern
```
mariaschuld, line 84: Since this is just an example circuit, is this really needed? Does it change the performance of the example?

And if so, maybe better to be explicit?
```suggestion
        # Apply additional CNOT to entangle the last with the first qubit
```

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -90,14 +94,21 @@ def layer(W):
 # Data is embedded in each circuit using amplitude embedding:
 
 def circuit(weights, feat=None):
-    qml.templates.embeddings.AmplitudeEmbedding(feat, [0, 1], pad=0.0, normalize=True)
+    qml.templates.embeddings.AmplitudeEmbedding(feat, range(0,num_qubits), pad=0.0, normalize=True)
```
mariaschuld, line 97: Should do the same thing :)

```suggestion
    qml.templates.embeddings.AmplitudeEmbedding(feat, range(num_qubits), pad=0.0, normalize=True)
```

demonstrations/tutorial_multiclass_classification.py
mariaschuld: To be honest, I am not sure I understand this piece of code. What would you want to do?

mariaschuld (commented):
Hey @YumiA, thanks so much for the contribution, it's great to generalise this demo. 

Like Antal, I was only wondering about the measurement. Maybe you could explain, and add a comment so other users understand it too?

demonstrations/tutorial_multiclass_classification.py
YumiA: Thanks Dr. @mariaschuld for the review.

demonstrations/tutorial_multiclass_classification.py
YumiA: :)
yes that's better.

demonstrations/tutorial_multiclass_classification.py
YumiA: Thanks Dr. @mariaschuld .
I wanted to measure one qubit of each two and in case of odd number of qubits the last qubit will be measured as well. I just thought that I follow the same way of the original code:
` return qml.expval(qml.PauliZ(0))`

It measured only one qubit out of two.

mariaschuld: Ah, I see, thanks for explaining! 

Hmm, I haven't written the demo, but I strongly suspect that the idea is to just measure the first qubit no matter how many classes there are. This is a mere design decision: your generalisation is surely not wrong (and I cannot tell if one would work better than the other), but just leaving the ``qml.expval(qml.PauliZ(0))`` could be much simpler to understand for readers? 

So I suggest, unless your solution has a significant performance gain in the optimisation, just measure the expval of the first qubit?

YumiA: Actually, this change came out due to the failure of the original code to classify my dataset. Once I applied the above change I got the expected accuracy (98%). I took the dataset from an experiment that used the classical neural network with 200 hidden nodes  to achieve 99% accuracy.
However, I rerun the new code after the changes we committed here and unfortunately the accuracy went down to 85%.
I'm doing a last run with nonzero padding and let's see how will be the result!

YumiA: Here are the results:
1- Applying all the changes approved here:
Figure_6
2- Same but with non-zero padding
`qml.templates.embeddings.AmplitudeEmbedding(feat, range(num_qubits), pad=float((2**num_qubits)-feature_size), normalize=True)`
Figure_7

I have no explanation, I hope you may help me.

mariaschuld: Thanks so much for the plots! 

Wait, I am a bit confused now. Aren't your changes supposed to reproduce the result of the demo, but generalise the code only? So shouldn't the single-qubit measurement work exactly as before, if you choose the features so that the number of qubits is 2?

But overall this question of yours is always tricky, is the demo there to show a simple example that may not work super well, or is it better to add some more complex design decisions? I personally find clarity more important than performance, because ultimately we would need a full-blown systematic benchmark analysis to see which measurement/template setting works best, and even then it will only be true for the particular dataset. So I find that as long as a demo has reasonable results (i.e. it learns something) then it should be fine.

I would suggest the following: 
* If you can use `PauliZ(wires=0)` and make the generalised demo reproduce the earlier results, then I'd go for that. 
* If for some reason (maybe because I understood things wrong) you want to use a more "global" measurement constructed from tensor products, then maybe would one constructed from all qubits (`for wi in range(num_qubits)`) do the job? Or from the first half  (`for wi in range(num_qubits//2)`)?
* If none of the above works for you, then I'd explain your choice in a comment in the code. I'd still have a question about the logic though. When I print what wires you use in the tensor product, I get this:
``` python
num_qubits = 3

even_wirelist = [i for i in range(num_qubits) if i % 2 == 0]
for wi in even_wirelist[1:]:
    print(wi)
if num_qubits % 2 != 0:
    print(num_qubits - 1)
# 2
# 2
```
Surely that is not what you want? Or am I overlooking something?

YumiA: Well, let me tell you the story. I'm studying Quantum Machine Learning and I'm trying to understand how things work. (I'm a senior high school student)

So, no I didn't make those changes for the sake of the demo. Actually, I got a dataset from an experiment on the Azure ML gallery: https://gallery.azure.ai/Experiment/Breast-cancer-dataset
The experiment achieved very high accuracy so it was a good practice for me to try to achieve the same using this tutorial here. Accordingly, the changes were meant to really generalize the tutorial for any dataset could be used.

YumiA: So according to the plots above, I would say that we may need a better way of padding the data!!

YumiA:
> Hey @YumiA, thanks so much for the contribution, it's great to generalise this demo.
> 
> Like Antal, I was only wondering about the measurement. Maybe you could explain, and add a comment so other users understand it too?

I put some comments and rerun the black tool again.

demonstrations/tutorial_multiclass_classification.py
mariaschuld: Hey @YumiA. Thanks for the clarification. 

If you don't mind, let us do the following, to merge your PR as fast as possible:

We leave the old measurement, so instead of 

``` python
    even_wirelist = [i for i in range(num_qubits) if i % 2 == 0]
    PZ = qml.PauliZ(0)
    for wi in even_wirelist[1:]:
        PZ = PZ @ qml.PauliZ(wi)
    if num_qubits % 2 != 0:
        PZ = PZ @ qml.PauliZ(num_qubits - 1)
    return qml.expval(PZ)
```
you revert to
 
```python
   return qml.expval(qml.PauliZ(0))
```
I checked myself, and the results are the same as in the current demo (as expected, since the experiment is the same, it just uses more general code).

We add a note stating that global or more complicated measurements could be beneficial, so every user can make their own choices.

I will make the suggestions in the code, if you want you can just accept them and we can proceed to merging.

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -60,7 +60,9 @@
 batch_size = 10
 lr_adam = 0.01
 train_split = 0.75
-num_qubits = 2
+# The number of the needed qubits is calculated based on the number of features
+# since the data should be encoded into the circuit
```
mariaschuld, line 64:
```suggestion
# the number of the required qubits is calculated from the number of features
```

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -75,36 +77,53 @@
...
-qnode3 = qml.QNode(circuit, dev).to_torch()
+    # Measuring only the even nodes
+    even_wirelist = [i for i in range(num_qubits) if i % 2 == 0]
+    PZ = qml.PauliZ(0)
+    for wi in even_wirelist[1:]:
+        PZ = PZ @ qml.PauliZ(wi)
+    # If the number of qubits is odd, measure also the last qubit.
+    if num_qubits % 2 != 0:
+        PZ = PZ @ qml.PauliZ(num_qubits - 1)
+    return qml.expval(PZ)
```
mariaschuld, line 113:
```suggestion
    return qml.expval(qml.PauliZ(0))
```

demonstrations/tutorial_multiclass_classification.py
```diff
@@ -75,36 +77,53 @@
...
 
 
 #################################################################################
 # We now define the quantum nodes that will be used. As we are implementing our
 # multiclass classifier as multiple one-vs-all classifiers, we will use 3 QNodes,
 # each representing one such classifier. That is, ``circuit1`` classifies if a
 # sample belongs to class 1 or not, and so on. The circuit architecture for all
-# 3 nodes are the same. We use the PyTorch interface for the QNodes.
+# nodes are the same. We use the PyTorch interface for the QNodes.
 # Data is embedded in each circuit using amplitude embedding:
```
mariaschuld, line 97:
```suggestion
# Data is embedded in each circuit using amplitude embedding.
# .. note::
#     For demonstration purposes we are using a very simple circuit here. 
#     You may find that other choices, for example more 
#     elaborate measurements, increase  the power of the classifier.      
```

antalszava (approved):
Looks good to me! :100: Thank you so much for adjusting it along the way @YumiA and for the nice contribution! :blush:

YumiA:
Thanks @antalszava for the approval. Is this done now? Should I close the request?

antalszava:
Hi @YumiA, we'll merge in the pull request at a convenient time and it will also close it automatically. :)

PR #199, opened 2021-01-11
change deprecated VQECost to ExpvalCost

albi3ro:
<b>Intro to QAOA</b> currently uses the deprecated `VQECost` instead of `ExpvalCost`.  

I thus updated the function.

### DIFF
--- a/demonstrations/tutorial_qaoa_intro.py
+++ b/demonstrations/tutorial_qaoa_intro.py
@@ -313,7 +313,7 @@ def circuit(params, **kwargs):
 # step is PennyLane's specialty: optimizing the circuit parameters.
 #
 # The cost function is the expectation value of :math:`H_C`, which we want to minimize. The
-# function :func:`~.pennylane.VQECost` is designed for this purpose: it returns the
+# function :func:`~.pennylane.ExpvalCost` is designed for this purpose: it returns the
 # expectation value of an input Hamiltonian with respect to the circuit's output state.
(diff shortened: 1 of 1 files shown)


demonstrations/tutorial_qaoa_intro.py
```diff
@@ -313,15 +313,15 @@ def circuit(params, **kwargs):
 # step is PennyLane's specialty: optimizing the circuit parameters.
 #
 # The cost function is the expectation value of :math:`H_C`, which we want to minimize. The
-# function :func:`~.pennylane.VQECost` is designed for this purpose: it returns the
+# function :func:`~.pennylane.ExpvalCost` is designed for this purpose: it returns the
```
glassnotes, line 316: Just noticed that this link (and a couple others) don't render - I think it needs to be `:class:` instead of `:func:` here.

glassnotes (approved):
Nice catch, thanks @albi3ro !

demonstrations/tutorial_qaoa_intro.py
glassnotes: (I can't seem to make suggested changes on the unchanged stuff, but same occurs in line 330 with the `GradientDescentOptimizer` - could you please adjust that one as well with the next update?)

albi3ro: I'm going ahead with merging this, and then creating a new branch and pull request to get the links working.

PR #203, opened 2021-01-15
[WIP] Fix demos in tape mode

thisac:
Contains fixes and tweaks to make all demos fully work with tape mode.


demonstrations/tutorial_vqe.py
```diff
@@ -191,9 +190,9 @@ def circuit(params, wires):
 max_iterations = 200
 conv_tol = 1e-06
 
-prev_energy = cost_fn(params)
+
 for n in range(max_iterations):
-    params = opt.step(cost_fn, params)
+    params, prev_energy = opt.step_and_cost(cost_fn, params)
```
glassnotes, line 195: I switched this to `step_and_cost`, but because it is looking for convergence, it still needs to compute the cost function again with the new parameters to determine whether or not the convergence criteria is reached.

demonstrations/tutorial_noisy_circuit_optimization.py
```diff
@@ -322,14 +322,14 @@ def noisy_cost(x):
...
     # we can optimize both in the same training loop
-    params = opt.step(cost, params)
-    noisy_circuit_params = opt.step(noisy_cost, noisy_circuit_params)
-
-    if (i + 1) % 5 == 0:
-        print("Step {:5d}. Cost: {: .7f}; Noisy Cost: {: .7f}"
-              .format(i + 1, 
-                      cost(params), 
-                      noisy_cost(noisy_circuit_params)))
+    params, _cost = opt.step_and_cost(cost, params)
```
glassnotes, line 325: I updated this to `step_and_cost` but I'm not 100% happy with how it looks (had to add an extra print statement to catch the output after the final step). Depending on peoples thoughts we can revert this back to the original.

thisac: I think it's fine to use cost as well (i.e. we don't have to use `step_and_cost` everywhere just because it's there), but if there's a noticeable improvement/speedup then it's nice to switch. Not sure how to best solve the last-cost-not-being-printed issue to be honest. I think both how it looks now and how it looked before are equally fine, and just showcase two different ways of doing this. :slightly_smiling_face:

glassnotes: Thanks! I think I will revert it back to keep things simple then. The demo takes 5s to build so there is no significant time savings, and the new version looks more complicated

josh146:
I've updated this PR to point to the `make-tape-default` branch on PennyLane 🤞

glassnotes:
> I've updated this PR to point to the `make-tape-default` branch on PennyLane crossed_fingers

Looking at the output of the failing tests, it is still seeing some of the same errors I see when I run things in non-tape mode (e.g., the stochastic parameter shift demo gets `unsupported operand type(s) for -: 'Variable' and 'Variable`.

josh146:
> Looking at the output of the failing tests, it is still seeing some of the same errors I see when I run things in non-tape mode (e.g., the stochastic parameter shift demo gets unsupported operand type(s) for -: 'Variable' and 'Variable).

For some reason, it's not using the correct python package. Maybe we need to bump the cache versions?

josh146 (approved):

Issue #205, opened 2021-01-20
[DEMO] Quantum-Enhanced Transformer


rdisipio:
#### General information

**Name**
Riccardo Di Sipio

**Affiliation** (optional)
Ideal.com (O5 Systems)

--------------------------------------------------------------------------------

#### Demo information

**Title**
A Quantum-Enhanced Transformer

**Abstract**
The Transformer neural network architecture revolutionized the analysis of text. Here we show an example of a Transformer with quantum-enhanced multi-headed attention. In the quantumized version, dense layers are replaced by simple Variational Quantum Circuits. An implementation based on PennyLane and TensorFlow-2.x illustrates the basic concept.

**Relevant links**
Code: https://github.com/rdisipio/qtransformer
Blog: https://towardsdatascience.com/toward-a-quantum-transformer-a51566ed42c2

glassnotes:
Thanks, @rdisipio ! We'll take a look as soon as we can.

glassnotes:
Hi @rdisipio , your demo is ready to go in PR #206.  I have made one minor edit to the abstract - please check it out, and if you're happy with it let me know and we'll merge it in :+1:

rdisipio:
Thanks, it looks good to me.

Best,
Riccardo

glassnotes:
Excellent, I've merged it in and you should see it live within the next day or so. Thanks again!

PR #206, opened 2021-01-22
Adds new community demo

glassnotes:
Adds community demo detailed in #205.


josh146 (approved):
Nice! Looks good, and all links work.

Minor suggestion, but might suggest removing/merging some sentences in the middle, as the card is getting quite large (especially on mobile view).

Issue #201, opened 2021-01-13
Make as many non-executable demos executable as possible


josh146:
Currently, several demos are not executed when the website is deployed, for a couple of reasons:

* Execution time: `qgrnn`, `qonn`, `quantum_neural_net`

* Requires access to hardware/external services or APIs: `pytorch_noise`, `quantum_volume`

* Tricky to install dependency on CircleCI: `qsim_beyond_classical`

For the ones with currently a slow execution time, perhaps porting them to use backpropagation (rather than parameter-shift) will be beneficial, and allow them to be executable.

Note: for `quantum_neural_net`, it should be updated to use `strawberryfields.tf` and the CVQNN layers (rather than hand-coding the layers manually)

KetpuntoG:
These are the three demos that we can currently try to make executable:

- function_fitting_qsp.py
- learning2learn.py (pending of a bug fix)
- qnspsa.py

PR #207, opened 2021-01-25
Qgrnn updates

albi3ro:
Some minor updates to the QGRNN tutorial.  The updates include

* separating weights and bias into two separate variables
* use of numpy.rng
* use of `nx.complete_graph`
* Move optimization printing out of cost function
* Use `step_and_cost` instead of just `step`
* renaming some variables for readability

More complete work could transform the tutorial to using `ApproxTimeEvolution`, instead of a custom implementation of it.


josh146:
Thanks @albi3ro! Do the changes allow the demo to run in tape-mode?

albi3ro:
It ran in tape mode with no changes and still does.

But executing it got me reading it. And reading it gave me ideas of things to change.

josh146:
Awesome 🙂 I assumed it was a required change for the demo to build with the new PL version. Since it is not required for supporting the new PL version, no need to prioritize review until after release -- I'll remove the requests for review I made.

glassnotes:
@albi3ro out of curiosity did you record execution times for tape mode vs. non-tape mode? This is one of the demos that's currently non-executable because of its long runtime, however I noticed while testing some of the other demos that using tape mode resulted in some speedup.

albi3ro:
Ok! That was *significantly* faster in tape mode.  Basically an order of magnitude faster.

I ran for ten steps and timed using `/usr/bin/time` from the command line.  

Non-tape mode was:
real 242.65
user 243.89
sys 30.96

Tape mode:
real 19.06
user 22.03
sys 11.96

glassnotes:
> Ok! That was _significantly_ faster in tape mode. Basically an order of magnitude faster.

Incredible!! Thanks for looking into that. When we tackle #201 then we may be able to make this one executable :tada:

glassnotes:
@albi3ro if you have the bandwidth, did you want to take a crack at making this one executable? (I am also happy to do so)

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
+"""
+The Quantum Graph Recurrent Neural Network
+===========================================
+
+.. meta::
+    :property="og:description": Using a quantum graph recurrent neural network to learn quantum dynamics.
+    :property="og:image": https://pennylane.ai/qml/_images/qgrnn_thumbnail.png
+
+*Author: Jack Ceroni. Posted: 27 July 2020. Last updated: 26 Oct 2020.*
```
glassnotes, line 9: Reminder to change the "last updated" to the day this gets merged in :)

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+ising_graph = nx.cycle_graph(qubit_number)
+
+print(f"Edges: {ising_graph.edges}")
+nx.draw(ising_graph)
+
+#######################################################################
+# .. figure:: ../demonstrations/qgrnn/graph1.png
+#      :width: 70%
+#      :align: center
+#
```
glassnotes, line 235:
```suggestion

```
I think this is no longer needed because the plot now gets output when the code runs.

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+
+target_weights = [0.56, 1.24, 1.67, -0.79]
+target_bias = [-1.44, -1.43, 1.18, -0.93]
+
+
+######################################################################
+# In theory, these parameters can
+# be any value we want, provided they are reasonably small enough that the QGRNN can reach them
+# in a tractable number of optimization steps.
+# In `matrix_params`, the first list represents the :math:`ZZ` interaction parameters and
```
glassnotes, line 267:
```suggestion
# In ``matrix_params``, the first list represents the :math:`ZZ` interaction parameters and
```
(I think it's two ` to make it come up in code font)

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+# the second list represents the single-qubit `Z` parameters.
```
glassnotes, line 268:
```suggestion
# the second list represents the single-qubit :math:`Z` parameters.
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+            else:
+                interaction_term = np.kron(interaction_term, np.identity(2))
+        full_matrix += weights[i] * interaction_term
+
+    # Creates the bias components of the matrix
+    for i in range(0, n_qubits):
+        z_term = x_term = 1
+        for j in range(0, n_qubits):
+            if j == i:
+                z_term = np.kron(z_term, qml.PauliZ.matrix)
```
glassnotes, line 295: Nice, renaming the variables in this part makes things much clearer!

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+ham_matrix = create_hamiltonian_matrix(qubit_number, ising_graph, target_weights, target_bias)
+plt.matshow(ham_matrix, cmap="hot")
+plt.show()
+
+
+######################################################################
+# .. figure:: ../demonstrations/qgrnn/ising_hamiltonian.png
+#      :width: 50%
+#      :align: center
+#
```
glassnotes, line 315:
```suggestion

```
Same as above, explicit display of figure can be removed.

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+
+plt.subplots_adjust(wspace=0.3, hspace=0.3)
+plt.show()
+
+
+######################################################################
+# .. figure:: ../demonstrations/qgrnn/hamiltonian_comparison.png
+#      :width: 100%
+#      :align: center
+#
```
glassnotes, line 628:
```suggestion

```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,687 @@
...
+        weights_edge.append(weights[ii])
+    else:
+        weights_noedge.append(weights[ii])
+
+######################################################################
+# Then, we print all of the weights:
+#
+
+print("Target parameters \tLearned parameters")
+print("\t Weights:")
```
glassnotes, line 660: Would remove the `\t` here and left-justify, then add a row of `------` and some spacing underneath to help with separating the two types of parameters. Something like this maybe:

```
Target parameters       Learned parameters

Weights
------------------------------------------------
0.56                    0.5782895244479018
1.24                    1.335028329676283
1.67                    1.804480439985849
-0.79                   -0.8395497395039528

Bias
------------------------------------------------
-1.44                   -1.352958693194635
-1.43                   -1.313891802560214
1.18                    0.9811173767093425
-0.93                   -1.0579331212003393
```

glassnotes (commented):
@albi3ro changes look great, what an amazing speedup! I just left a few minor comments, but otherwise this looks good to go.

demonstrations/tutorial_qgrnn.py
albi3ro: Thanks!

demonstrations/tutorial_qgrnn.py
albi3ro: Added a bar too "|"

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,674 @@
+"""
+The Quantum Graph Recurrent Neural Network
+===========================================
+
+.. meta::
+    :property="og:description": Using a quantum graph recurrent neural network to learn quantum dynamics.
+    :property="og:image": https://pennylane.ai/qml/_images/qgrnn_thumbnail.png
+
+*Author: Jack Ceroni. Posted: 27 July 2020. Last updated: 25 March 2021.*
```
glassnotes, line 9:
```suggestion
*Author: Jack Ceroni. Posted: 27 July 2020. Last updated: 25 Feb 2021.*
```

demonstrations/tutorial_qgrnn.py
```diff
@@ -0,0 +1,674 @@
...
+axes[0].set_title("Target", y=1.13)
+
+axes[1].matshow(init_ham, vmin=-7, vmax=7, cmap="hot")
+axes[1].set_title("Initial", y=1.13)
+
+axes[2].matshow(new_ham_matrix, vmin=-7, vmax=7, cmap="hot")
+axes[2].set_title("Learned", y=1.13)
+
+plt.subplots_adjust(wspace=0.3, hspace=0.3)
+plt.show()
```
glassnotes, line 611: The rendered plot here is quite small, any way to increase it? (Sorry didn't catch this earlier, I wasn't sure if in the previous iteration it was the old static image or the rendered image that was small.)

demonstrations.rst
```diff
@@ -100,7 +100,7 @@ Demos
                     </a>
                 </li>
                 <li>
-                    <a href="demos/qgrnn.html">
+                    <a href="demos/tutorial_qgrnn.html">
```
glassnotes, line 103: Nice catch!

glassnotes (approved):
Just two small additional changes, but once those are done, go ahead and merge :tada:

demonstrations/tutorial_qgrnn.py
albi3ro: I have no idea how that happened...

demonstrations/tutorial_qgrnn.py
albi3ro: This seems to be less an issue with the image and more the amount of padding given to the image.  Is there a way to reduce the padding given to the generated figure? I'm not too familiar with the procedure used to generate the website.

I could also switch it to have "initial" in its own row and then "target" and "learned" sharing a row. That would widen each plot.

demonstrations/tutorial_qgrnn.py
glassnotes: Thinking ahead :grinning:

demonstrations/tutorial_qgrnn.py
glassnotes: Hmm... maybe try increasing the `figsize`, and/or throwing a `plt.tight_layout()` in there rather than `plt.subplots_adjust`? I know it's a pain to test things out because the runtime is too long to iterate quickly. If you can't get something working after a try or two we should ask Josh since he's the most knowledgeable about the demo build process.

josh146: >  If you can't get something working after a try or two we should ask Josh since he's the most knowledgeable about the demo build process.

Unfortunately I don't have any good solutions here! Normally I would download the Jupyter notebook version of the demo (or just copy the demo into a jupyter notebook), run all the cells, and then that allows you to iterate on the final plotting without performing the optimization every time.

The matplotlib image created in the notebook should match exactly the one on the website; Sphinx does not do any post-processing (I believe).

albi3ro: Given any updates to the plot generating code will probably make the code more complicated, and any improvements to sphinx to display it better are out of our technical depth, I think at this point it may be just better to leave as is.  I'm merging it in.

Issue #209, opened 2021-01-26
[DEMO] Layerwise learning for quantum neural networks


fioyarce:
#### General information

**Name**
Felipe Oyarce Andrade.

--------------------------------------------------------------------------------

#### Demo information

**Title**
Layerwise learning for quantum neural networks.

**Abstract**
In this project we’ve implemented a strategy presented by [Skolik et al., 2020](https://arxiv.org/abs/2006.14904) for effectively train quantum neural networks. In layerwise learning the strategy is to gradually increase the number of parameters by adding a few layers and training them while freezing the parameters of previous layers already trained.
An easy way for understanding this technique is to think that we’re dividing the problem into smaller circuits to successfully avoid to fall into Barren Plateaus. Here, we provide a proof-of-concept for the implementation of this technique in Pennylane’s Pytorch interface for binary classification in the MNIST dataset.

**Relevant links**
https://arxiv.org/abs/2006.14904
https://github.com/felipeoyarce/layerwise-learning
https://nbviewer.jupyter.org/github/felipeoyarce/layerwise-learning/blob/master/layerwise_learning/Layerwise%20learning%20-%20Image%20Classification.ipynb

glassnotes:
Thank you @fioyarce , we'll take a look at get back to you as soon as we can!

glassnotes:
Please see #210, abstract text is lightly edited to help it fit better in the card. If you're happy with it, let me know and we'll get it merged in!

glassnotes:
Awesome, thanks, it's been merged and you'll see it on the site within the next day or so. Thanks again for the contribution! :grin:

PR #210, opened 2021-01-26
Add new user demo to community page.

glassnotes:
Adds the user demo in #209 to the community page.

### DIFF
--- a/demonstrations/tutorial_measurement_optimize.py
+++ b/demonstrations/tutorial_measurement_optimize.py
@@ -765,6 +765,8 @@ print("Number of Hamiltonian terms/required measurements:", len(H.ops))
 groups = qml.grouping.group_observables(H.ops, grouping_type='qwc', method='rlf')
 print("Number of required measurements after optimization:", len(groups))
 
+qml.disable_tape()
+
 ##############################################################################
 # We went from 2050 required measurements/circuit evaluations to 523 (just over *two thousand*
 # down to *five hundred* 😱😱😱).
--- a/demos_community.rst
+++ b/demos_community.rst
@@ -191,7 +191,25 @@ Community
     replaced by simple Variational Quantum Circuits. An implementation based on
     PennyLane and TensorFlow-2.x illustrates the basic concept.
 
-:html:`<div class="card hidden-card"></div></div>`
+.. community-card::
+    :title: Layerwise learning for quantum neural networks
+    :author: Felipe Oyarce Andrade
+    :date: 26/01/2021
+    :code: https://nbviewer.jupyter.org/github/felipeoyarce/layerwise-learning/blob/master/layerwise_learning/Layerwise%20learning%20-%20Image%20Classification.ipynb
+    :paper: https://arxiv.org/abs/2006.14904
+    :color: heavy-rain-gradient
+
+    In this project we’ve implemented a strategy presented by <a
+    href="https://arxiv.org/abs/2006.14904" target="_blank">Skolik et al., 2020</a> for
+    effectively training quantum neural networks. In layerwise learning the
+    strategy is to gradually increase the number of parameters by adding a few
+    layers and training them while freezing the parameters of previous layers
+    already trained.  An easy way for understanding this technique is to think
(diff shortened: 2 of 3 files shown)


demos_community.rst
```diff
@@ -191,7 +191,24 @@ Community
     replaced by simple Variational Quantum Circuits. An implementation based on
     PennyLane and TensorFlow-2.x illustrates the basic concept.
 
-:html:`<div class="card hidden-card"></div></div>`
+.. community-card::
+    :title: Layerwise learning for quantum neural networks
+    :author: Felipe Oyarce Andrade
+    :date: 26/01/2021
+    :code: https://nbviewer.jupyter.org/github/felipeoyarce/layerwise-learning/blob/master/layerwise_learning/Layerwise%20learning%20-%20Image%20Classification.ipynb
```
thisac, line 198: Should we perhaps instead link to the repo here (I think it perhaps makes more sense, since there's a readme and a `requirements.txt` file there that might be of interest to people who want to run it)?

Also, how come you removed the link to the paper? :thinking:

thisac: Oh, it's because the demo isn't written by the author(s) of the paper?

demos_community.rst
```diff
@@ -191,7 +191,24 @@ Community
...
+    In this project we’ve implemented a strategy presented by <a
+    href="https://arxiv.org/abs/2006.14904" target="_blank">Skolik et al., 2020</a> for
+    effectively training quantum neural networks. In layerwise learning the
+    strategy is to gradually increase the number of parameters by adding a few
+    layers and training them while freezing the parameters of previous layers
+    already trained.  An easy way for understanding this technique is to think
+    that we’re dividing the problem into smaller circuits to successfully avoid
+    falling into Barren Plateaus. We provide a proof-of-concept
+    implementation of this technique in Pennylane’s Pytorch interface for binary
+    classification in the MNIST dataset.
```
thisac, line 210: Wondering if we should cut down the length of the abstracts (e.g. limit them to 50 words), or add a "read more"/"expand" text-button.

thisac (approved):

demos_community.rst
glassnotes: > Should we perhaps instead link to the repo here

Yes, good call. 

> Oh, it's because the demo isn't written by the author(s) of the paper?

Yes exactly, and the paper it's based on is linked in the abstract.

demos_community.rst
glassnotes: I'll leave it for now, but I like the idea of having a "read more". 

@josh146 tagging you here, I will add this to the running list of CSS improvements we have.

josh146: I think a read more button is a great idea!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -765,6 +765,8 @@ def circuit(weights, group=None, **kwargs):
 groups = qml.grouping.group_observables(H.ops, grouping_type='qwc', method='rlf')
 print("Number of required measurements after optimization:", len(groups))
 
+qml.disable_tape()
```
josh146, line 768: @glassnotes I found the bug, it was my fault :( The optimize grouping demo was enabling the tape, but not disabling it... which was breaking all subsequent demos.

glassnotes: Whoa!! I did not expect that, nice find :+1:

co9olguy: Won't these calls be removed from all demos anyway by next week's release?

glassnotes: Yup :+1: Have just double checked and removed a few that were missed.

josh146: Yep, this is just temporary to make sure the demos and PRs build now. This line can be removed on Tuesday.

co9olguy: but will it be removed?? :laughing:

josh146: Yep, it is already removed in the PR to drop on Tuesday: #203

PR #215, opened 2021-01-30
Add auto-invalidation of CDN to continuous-deployment

### LINKED ISSUE #214: Force browsers to clear cache
Often new demos are not viewable unless the user manually clears the browser cache. This makes it annoying to advertise new tutorials, because some users will think they're "not there"

josh146:
Closes #214


co9olguy:
This is a bit different than #214. This forces the CDN to refresh the cache, not the client browser. Obviously the former is more expensive, so I was thinking more to do the latter :thinking:

josh146:
> so I was thinking more to do the latter 

Oh 🤔 Is that possible?

Maybe it's worth merging this anyway, since the effect of invalidating the CDN is non-negligible

co9olguy:
Seems there are many methods to do it, yes, but constantly shifting: https://stackoverflow.com/questions/1922910/force-browser-to-clear-cache

co9olguy (approved):

PR #225, opened 2021-03-04
Automate layout of the community demos page

josh146:
Automates the building of the `demos_community.rst` page.

The following two files are added:

* `demos_community.yaml`: a yaml file with the community demos data. *This is the file that should be edited from now on.* Note that demos are in the order that they will appear on the page.

* `demos_community.rst_t`: a templated ReST file that is used to build the final ReST file.

When running Sphinx, `demos_community.rst` will be *automatically* generated from these two files. As a result, it has been added to the gitignore so that it does not get committed.


conf.py
```diff
@@ -194,6 +196,25 @@
...
+right_cards = card_data[1::2]
+
+if len(left_cards) > len(right_cards):
+    right_cards.append({})
+
+card_pairs = list(zip(left_cards, right_cards))
+
+loader = FileSystemLoader(".")
+env = Environment(loader=loader)
+template = env.get_template("demos_community.rst_t")
```
co9olguy, line 214: can we call it `demos_community.rst.template` or something similarly direct?

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+:html:`<div class="card hidden-card"></div></div>`
+{% endif %}
+
+:html:`</div></div><br><div style='clear:both'>`
+
+{% endfor %}
+
+.. If the final card deck only has a single card, we insert a 'hidden card'
+   so that the card does not become full-width.
+   :html:`<div class="card hidden-card"></div></div>`
```
co9olguy, line 88: Is this still necessary? From what I could tell, it was covered by above change

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+
+
+.. For two cards per row, create a new card deck every two cards.
+
+{% for card1, card2 in card_pairs %}
+
+:html:`<div class="card-deck">`
+
+.. community-card::
+    :title: {{ card1["title"] }}
```
co9olguy, line 57: Could we use this as an opportunity to introduce direct hyperlink anchors into each card? Or would it be too much extra work?

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+{% endif %}
+
+:html:`</div></div><br><div style='clear:both'>`
+
+{% endfor %}
+
+.. If the final card deck only has a single card, we insert a 'hidden card'
+   so that the card does not become full-width.
+   :html:`<div class="card hidden-card"></div></div>`
+
```
trbromley, line 89:
```suggestion
```

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+    </div>
+
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/mdbootstrap/4.8.10/css/mdb.min.css" rel="stylesheet">
+
+    <hr>
+    <br>
+
+
+.. For two cards per row, create a new card deck every two cards.
+
```
trbromley, line 51:
```suggestion
```

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+{% endfor %}
+
+.. If the final card deck only has a single card, we insert a 'hidden card'
+   so that the card does not become full-width.
+   :html:`<div class="card hidden-card"></div></div>`
+
+.. toctree::
+    :maxdepth: 2
+    :hidden:
```
trbromley, line 92: Thanks, this was quite educational! I haven't used `jinja` but at least now I might be tempted if a similar situation arises in future!

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+    {{ card1["description"] | replace("\n", " ") }}
+
+{% if "title" in card2 %}
+
+.. community-card::
+    :title: {{ card2["title"] }}
+    :author: {{ card2["author"] }}
+    :date: {{ card2["date"] }}
+    {% if "code" in card2 -%}:code: {{ card2["code"] }}{% endif -%}
+    {% if "blog" in card2 -%}:blog: {{ card2["bog"] }}{% endif -%}
```
trbromley:
```suggestion
    {% if "code" in card2 -%}:code: {{ card2["code"] }}{% endif %}
    {% if "blog" in card2 -%}:blog: {{ card2["bog"] }}{% endif %}
```

demos_community.rst_t
```diff
@@ -0,0 +1,92 @@
...
+{% for card1, card2 in card_pairs %}
+
+:html:`<div class="card-deck">`
+
+.. community-card::
+    :title: {{ card1["title"] }}
+    :author: {{ card1["author"] }}
+    :date: {{ card1["date"] }}
+    {% if "code" in card1 -%}:code: {{ card1["code"] }}{% endif -%}
+    {% if "blog" in card1 -%}:blog: {{ card1["bog"] }}{% endif -%}
```
trbromley:
```suggestion
    {% if "code" in card1 -%}:code: {{ card1["code"] }}{% endif %}
    {% if "blog" in card1 -%}:blog: {{ card1["bog"] }}{% endif %}
```
I noticed the paper link wasn't showing up, so not sure if this correction fixes things
image

demos_community.rst_t
```diff
@@ -69,8 +69,8 @@ Community
     :title: {{ card2["title"] }}
     :author: {{ card2["author"] }}
     :date: {{ card2["date"] }}
-    {% if "code" in card2 -%}:code: {{ card2["code"] }}{% endif -%}
-    {% if "blog" in card2 -%}:blog: {{ card2["bog"] }}{% endif -%}
+    {% if "code" in card2 -%}:code: {{ card2["code"] + "\n    " }}{% endif -%}
+    {% if "blog" in card2 -%}:blog: {{ card2["blog"] + "\n    " }}{% endif -%}
```
trbromley, line 73: Have introduced this not very nice looking `+ "\n "`. There's probably a nicer way, but the reason this is here is to ensure a new line. Before, `:paper:` was getting tagged onto the previous line because of the `-%}`. However, without `-%}`, if `:paper:` wasn't present then there would be an extra blank line added.

trbromley (approved):
Thanks! :100:

demos_community.rst_t
trbromley: To show the problem more visually, without this fix, papers and blog links didn't show:
image

co9olguy: could we just use the html `<br>` here?

demos_community.rst_t
josh146: From looking at #194, I would say this is enough work to continue being its own PR. It requires changes here, changes to the card HTML definition (in a different file), as well as changes to CSS and new javascript inserted somewhere.

demos_community.rst_t
josh146: Oh, nice catch Tom. There must be some combination of hyphens that make this work 🤔 Jinja also allows hyphens at the beginning of the template statement, so maybe that will work. I find it very confusing to figure out.

> could we just use the html `<br>` here?

No, this is a spacing issue during the Jinja <--> ReST conversion, not the ReST <---> html conversion. We just need to make sure Jinja generates spacing that creates valid ReST.

josh146: @trbromley I _think_ I fixed this in 2d0b7e6e20154eceb3aa6077d5490645fff4cd4a

trbromley (approved):
Thanks @josh146 !

demos_community.rst_t
trbromley: Nice, looks better than the `\n    `!

PR #234, opened 2021-03-20
Fix inconsistency in QV demo

glassnotes:
Due to fluctuations in the results when running this demo, some of the text did not align with the code outputs when discussing the simulated experimental results. This is now fixed :+1:


josh146 (approved):
Thanks @glassnotes!

PR #237, opened 2021-03-25
Fix text rendering error in multiclass demo

co9olguy:
A "note" box wasn't rendering correctly

Before:
image

After:
image


antalszava (approved):
Nice catch! 💯

glassnotes (approved):

PR #244, opened 2021-04-08
Update networkx version.

glassnotes:
The build pipeline on the `dev` branch is currently failing for some demos due to  the error 
```
networkx.exception.NetworkXError: random_state_index is incorrect
```

It looks like this was a known issue in a recent release and has been fixed in version `2.5.1` (see [here](https://github.com/networkx/networkx/issues/4722) and [here](https://github.com/networkx/networkx/issues/4718)). The `qml` repo specifies `networkx==2.4` so we can try updating this to see if it helps (and hopefully it will not cause other issues in the meantime).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,9 +20,9 @@ jobs:
       - restore_cache:
           keys:
             # when lock file changes, use increasingly general patterns to restore cache
-            - pip-v22f-{{ .Branch }}-{{ checksum "requirements.txt" }}
-            - pip-v22f-{{ .Branch }}-
-            - pip-v22f-
+            - pip-v23a-{{ .Branch }}-{{ checksum "requirements.txt" }}
+            - pip-v23a-{{ .Branch }}-
+            - pip-v23a-
 
       - run:
           name: Install dependencies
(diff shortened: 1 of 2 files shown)


josh146:
It's super weird that NX 2.4, which used to work, suddenly became affected by this bug 🤔 Maybe one of the plugins has a dependency that was updating the NX version?

josh146:
@glassnotes it worked! no more NX errors (other errors remain though). Once the CI is done I'll merge it in.

glassnotes:
> @glassnotes it worked! no more NX errors (other errors remain though). Once the CI is done I'll merge it in.

Thanks for taking a look @josh146 !  I was wondering if it was related to PL itself - its  `requirements.txt` doesn't explicitly specify a version for `networx`, so I thought maybe it was pulling the most recent one and that was being propagated here (?). In any case, looks like remaining errors are just related to PL changes which are being addressed in the other PRs so we're good to go. Feel free to merge in :)

josh146:
> its requirements.txt doesn't explicitly specify a version for networx, so I thought maybe it was pulling the most recent one and that was being propagated here (?)

The pinning in this repo should override any package specific requests though 🤔

PR #247, opened 2021-04-13
Fix qnn module tutorial for TorchLayer

trbromley:
Currently the tutorial is failing due to the issue highlighted here: https://github.com/PennyLaneAI/pennylane/issues/1210.

As a quick fix in this PR, we swap out `BasicEntanglerLayers` for its explicit decomposition.


demonstrations/tutorial_qnn_module_torch.py
```diff
@@ -86,7 +86,10 @@
 @qml.qnode(dev)
 def qnode(inputs, weights):
     qml.templates.AngleEmbedding(inputs, wires=range(n_qubits))
-    qml.templates.BasicEntanglerLayers(weights, wires=range(n_qubits))
+    for weights_layer in weights:
```
albi3ro, line 89:
```suggestion
    # explicit `qml.templates.BasicEngtnglerLayers` until a bug is fixed
    for weights_layer in weights:
```

Maybe just a comment here that there is usually a better way of doing this.

albi3ro (approved):
Hopefully, the bug gets fixed soon and glad it got caught :)

I suggested adding a comment, but that's not particularly critical.  Approve.

demonstrations/tutorial_qnn_module_torch.py
trbromley: :thinking: I see your point, although wonder if it may confuse the reader - unless they also look at the `Keras` tutorial they may not be expecting to see `BasicEntanglerLayers`.

demonstrations/tutorial_qnn_module_torch.py
```diff
@@ -85,8 +85,15 @@
 
 @qml.qnode(dev)
 def qnode(inputs, weights):
-    qml.templates.AngleEmbedding(inputs, wires=range(n_qubits))
-    qml.templates.BasicEntanglerLayers(weights, wires=range(n_qubits))
+    # Embedding
+    qml.RX(inputs[0], wires=0)
+    qml.RX(inputs[1], wires=1)
```
trbromley, line 90: @albi3ro, sorry to rerequest your review! We additionally have to change this, as the bug is also present in `AngleEmbedding`.

albi3ro (approved):
I don't see anymore templates.  Hopefully this time it's good to go :)

PR #249, opened 2021-04-13
update rosalin demo to v0.15.0

albi3ro:
The Rosalin demo now executes without errors.

With the shots refractor, more changes could make this tutorial even better.  PennyLane also now has the Rosalin optimizer as a feature, which didn't exist when this tutorial was written.


demonstrations/tutorial_rosalin.py
```diff
@@ -25,6 +25,11 @@
 * Frugally adapting the number of shots used per parameter update, and
 * Performing a weighted sampling of operators from the cost Hamiltonian.
 
+.. note::
+
+    The Rosalin optimizer is available in PennyLane via the
+    :class:`~.pennylane.ShotAdaptiveOptimizer`.
```
josh146, line 31: > With the shots refractor, more changes could make this tutorial even better. PennyLane also now has the Rosalin optimizer as a feature, which didn't exist when this tutorial was written.

This is a really good point. As a quick fix, I've added this small note here, similarly to what we did with the [Rotoselect demo](https://pennylane.ai/qml/demos/tutorial_rotoselect.html) (which was similar, the demo came _before_ the feature) but we should definitely revisit.

demonstrations/tutorial_rosalin.py
```diff
@@ -197,12 +202,10 @@ def cost(params):
     result = 0
 
     for h, c, p, s in zip(qnodes, coeffs, prob_shots, shots_per_term):
-        # set the number of shots
-        h.device.shots = s
 
         # evaluate the QNode corresponding to
         # the Hamiltonian term, and add it on to our running sum
-        result += c * h(params)
+        result += c * h(params, shots=int(s))
```
josh146, line 208: nice!

josh146 (approved):
Thanks for solving this one @albi3ro! I'll merge it in, since the other failing demos are currently blocking CI.

PR #258, opened 2021-04-23
Temporarily swap the backprop tutorial to use Autograd

### LINKED ISSUE #256: [BUG] Backprop tutorial is producing incorrect benchmarks
The time scaling plot in the backprop tutorial is currently incorrect:

image

Strangely enough, the first section of the demo, comparing single forwards and backwards pass times for parameter-shift vs. backprop, seems to still be returning sensible times.

josh146:
Fixes #256, so that the demo continues to make sense while we determine the cause of the TensorFlow slowdown

image

### DIFF
--- a/demonstrations/tutorial_backprop.py
+++ b/demonstrations/tutorial_backprop.py
@@ -249,7 +249,7 @@ print(2 * forward_time * params.size)
 # designed to support backpropagation.
 #
 # One such device is :class:`default.qubit <pennylane.devices.DefaultQubit>`. It
-# has backends written using both TensorFlow, JAX, and Autograd, so when used with the
+# has backends written using TensorFlow, JAX, and Autograd, so when used with the
 # TensorFlow, JAX, and Autograd interfaces respectively, supports backpropagation.
 # In this demo, we will use the default Autograd interface.
 
@@ -258,7 +258,7 @@ dev = qml.device("default.qubit", wires=4)
 ##############################################################################
 # When defining the QNode, we specify ``diff_method="backprop"`` to ensure that
 # we are using backpropagation mode. Note that this is the *default differentiation
-# mode* for the ``default.qubit``.
+# mode* for the ``default.qubit`` device.
 
 
 @qml.qnode(dev, diff_method="backprop")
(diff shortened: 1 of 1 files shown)


demonstrations/tutorial_backprop.py
```diff
@@ -249,28 +249,26 @@ def circuit(params):
...
+# In this demo, we will use the default Autograd interface.
 
 dev = qml.device("default.qubit", wires=4)
 
 ##############################################################################
 # When defining the QNode, we specify ``diff_method="backprop"`` to ensure that
-# we are using backpropagation mode. Note that this will be the *default differentiation
-# mode* for the ``default.qubit.tf`` device when ``interface="tf"``.
+# we are using backpropagation mode. Note that this is the *default differentiation
+# mode* for the ``default.qubit``.
```
mariaschuld, line 261:
```suggestion
# mode* for the ``default.qubit`` device.
```

mariaschuld: Maybe give user a hint that the autodiff framework is used by default?

mariaschuld (approved):
I actually like it this way, it kind of doesn't suggest that TF is needed for backprop.

There was only one typo to fix.

demonstrations/tutorial_backprop.py
```diff
@@ -249,28 +249,26 @@ def circuit(params):
 # designed to support backpropagation.
 #
 # One such device is :class:`default.qubit <pennylane.devices.DefaultQubit>`. It
-# has backends written using both TensorFlow and Autograd, so when used with the
-# TensorFlow and Autograd interfaces respectively, supports backpropagation.
-# In this demo, we will use the TensorFlow interface.
-
-import tensorflow as tf
+# has backends written using both TensorFlow, JAX, and Autograd, so when used with the
```
mariaschuld, line 252: Not sure if the "both" is a typo?

josh146:
```suggestion
# has backends written using TensorFlow, JAX, and Autograd, so when used with the
```

josh146:
Thanks @mariaschuld!

PR #260, opened 2021-04-26
allow for arbitrary number of layers in qaoa example

peendebak:
**Title:** Allow QAOA maxcut example to use more than 2 layers

**Summary:**

The initial parameters in the original example are defined as `init_params = 0.01 * np.random.rand(2, 2)` which has fixed size and only works for `p=1,2`. This PR replaces it with `init_params = 0.01 * np.random.rand(2, n_layers)`

**Relevant references:**

None

**Possible Drawbacks:**

None

**Related GitHub Issues:**

None


glassnotes (approved):
Hi @peendebak , thanks for the contribution! And good catch, this helps make the function more general. You can go ahead and merge this in at your convenience.

PR #265, opened 2021-05-10
Adds Feedback-Based Quantum Optimization (FALQON) Tutorial

Lucaman99:
**Title:**

Adds a tutorial on Feedback-Based Quantum Optimization (FALQON) to the QML gallery.

**Summary:**

- Discusses the basic theory behind the FALQON algorithm
- Implements the FALQON algorithm to solve the MaxClique problem, in PennyLane
- Discusses benchmarking of FALQON
- Discuss and simulates a hybrid QAOA-FALQON algorithm in PennyLane

**Relevant references:**

https://arxiv.org/abs/2103.08619

**Possible Drawbacks:**

**Related GitHub Issues:**


demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+   tutorial_qaoa_maxcut QAOA for MaxCut
+
+*Authors: David Wakeham and Jack Ceroni. Posted: XXX. Last updated: XXX.*
+
+-----------------------------
+
+While the `Quantum Approximate Optimization Algorithm (QAOA) <https://pennylane.ai/qml/demos/tutorial_qaoa_intro.html>`__
+is one of the best-known processes for solving combinatorial optimization problems with quantum computers,
+it has drawbacks. Since QAOA relies on optimization over a parameter space,
+convergence isn't guaranteed, and the optimization procedure can become "stuck" in local minima.
```
ixfoduap, line 21:
```suggestion
it has drawbacks: convergence isn't guaranteed, and the optimization procedure can become
"stuck" in local minima.
```
It's not accurate that optimization over a parameter space (is there any other typo of optimization?) implies that convergence isn't guaranteed

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+It is similar in spirit to QAOA, but instead uses iterative feedback steps rather than a global optimization
+over parameters, therefore avoiding the issue of local minima. We will show how to implement FALQON in PennyLane
+and test its performance on the MaxClique problem in graph theory!
+
+Theory
+------
+
+To solve combinatorial optimization problems using a quantum computer, a typical strategy is to encode
+the solution to the problem as the ground state of *cost Hamiltonian* :math:`H_c`, and then use some procedure to drive
+the system from an initial state into this ground state. FALQON falls under this broad scheme.
```
ixfoduap, line 38:
```suggestion
the system from an initial state into the ground state of :math:`H_c`. FALQON falls under this broad scheme.
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+Imagine a quantum system governed by a Hamiltonian of the form :math:`H = H_c + \beta(t) H_d`. These kinds of
+Hamiltonians come up quite often in the theory of `quantum control <https://quantiki.org/wiki/quantum-control-theory>`__,
```
ixfoduap, line 41:
```suggestion
Hamiltonians appear often in the theory of `quantum control <https://quantiki.org/wiki/quantum-control-theory>`__,
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+which studies how we may go about driving a quantum system from one state to another. The choice of :math:`\beta(t)` allows
```
ixfoduap, line 42: I suggest removing "which studies how we may go about driving a quantum system from one state to another"

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+us to decide which state we want a system governed by such a Hamiltonian to evolve towards.
```
ixfoduap, line 43: this sentence is awkward; consider revising

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+The time-dependent Schrodinger equation tells us that the dynamics of the system are given by:
```
ixfoduap, line 45:
```suggestion
The time-dependent Schrödinger equation tells us that the dynamics of the system are given by:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+.. math:: i \frac{d}{dt} |\psi(t)\rangle = (H_c + \beta(t) H_d) |\psi(t)\rangle
+
+where we set :math:`\hbar = 1`. Now, let us suppose that the objective of our quantum control experiment is to drive our system
```
ixfoduap, line 49:
```suggestion
where we set :math:`\hbar = 1`. Now suppose that the objective of the quantum control experiment is to drive the system
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+to the state :math:`|\psi\rangle`: the ground state of :math:`H_c`. Phrased differently, we would like to minimize the expectation
+value :math:`\langle H_c\rangle`. Therefore, a reasonable goal is to construct our system such that the expectation decreases with time:
```
ixfoduap, line 51:
```suggestion
value :math:`\langle H_c\rangle`. Therefore, a reasonable goal is to construct the system such that the expectation decreases with time:
```
And look elsewhere for cases where the use of "we" and "our" can be smoothly avoided. And of course, keep it when it the text reads better

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+.. math:: \frac{d}{dt} \langle H_c\rangle_t = \frac{d}{dt} \langle \psi(t)|H_c|\psi(t)\rangle = i \beta(t)\langle [H_d, H_c] \rangle_t \leq 0
```
ixfoduap, line 53:
```suggestion
.. math:: \frac{d}{dt} \langle H_c\rangle_t = \frac{d}{dt} \langle \psi(t)|H_c|\psi(t)\rangle = i \beta(t)\langle [H_d, H_c] \rangle_t \leq 0,
```
and similarly elsewhere make sure to add punctuation to equations too

demonstrations/tutorial_falqon.py
ixfoduap: And make this change throughout

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+where we used the product rule and Schrodinger's equation to derive the above formula. Recall that our control experiment depends on the choice of :math:`\beta(t)`. Thus,
```
ixfoduap, line 55: This line is too long

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+if we pick :math:`\beta(t) = -\langle i[H_d, H_c] \rangle_t`, so that
+
+.. math:: \frac{d}{dt} \langle H_c\rangle_t = -|\langle i[H_d, H_c] \rangle_t|^2 \leq 0
+
+then :math:`\langle H_c \rangle` is guaranteed to strictly decrease, as desired!
+(Note that we bring the :math:`i` into the expectation to give a Hermitian operator.)
+
+Using `techniques from control theory <https://arxiv.org/pdf/1304.3997.pdf>`__, it is possible to rigorously show this choice of :math:`\beta(t)` will
+eventually drive the system into the ground state of :math:`H_c`! Thus, if we evolve some initial state :math:`|\psi_0\rangle` under the
```
ixfoduap, line 64:
```suggestion
eventually drive the system into the ground state of :math:`H_c`. Thus, if we evolve some initial state :math:`|\psi_0\rangle` under the
```

ixfoduap: I love exclamation marks and emojis, but they have to be used with extreme care, otherwise makes the text look like it should not be taken seriously.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+time-evolution operator corresponding to :math:`H`, given by :math:`U(t) = e^{-iHt}`, we will arrive at the ground state of :math:`H_c`.
```
ixfoduap, line 65:
```suggestion
time-evolution operator :math:`U(t) = e^{-iHt}`, we will arrive at the ground state of :math:`H_c`.
```

ixfoduap: I think the explanation so far is good as it is, but in general we have to be careful when evolving under time-dependent hamiltonians, as it's not as simple as applying e^{-iH(t) t} . Ultimately, what you will do is to evolve under time-*independent* hamiltonians, which get changed with each new step, mimicking the process of evolving under a time-dependent one

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+
+import pennylane as qml
+import numpy as np
+from matplotlib import pyplot as plt
+from pennylane import qaoa as qaoa
+import networkx as nx
+
+######################################################################
+# FALQON is an algorithm for combinatorial optimization. Therefore, we must select some kind of combinatorial problem
```
ixfoduap, line 81: I think these first two sentences can be removed

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+edges = [(0, 1), (1, 2), (2, 0), (2, 3), (1, 4)]
+graph = nx.Graph(edges)
+nx.draw(graph, with_labels=True)
+
+######################################################################
+# We must first encode this combinatorial problem into a cost Hamiltonian :math:`H_c`. This ends up being:
+#
+# .. math:: H_c = 3 \sum_{(i, j) \in E(\bar{G})} (Z_i Z_j - Z_i - Z_j) + \displaystyle\sum_{i \in V(G)} Z_i
+#
+# where each qubit is a node in the graph, and the states :math:`|0\rangle` and :math:`|1\rangle` represent if the vertex
```
ixfoduap, line 102:
```suggestion
# where each qubit is a node in the graph, and the states :math:`|0\rangle` and :math:`|1\rangle` represent whether the vertex
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# has been "marked" as part of the clique, as is the case for `most standard QAOA encoding schemes <https://arxiv.org/abs/1709.03489>`__.
+#
+# In addition to defining :math:`H_c`, we also require a driver Hamiltonian :math:`H_d`, which does not commute with :math:`H_c`, and
+# is able to "mix up" our system sufficiently so that it may be driven towards the ground state (similar to the mixer Hamiltonian in QAOA).
```
ixfoduap, line 106:
```suggestion
# In addition to defining :math:`H_c`, we also require a driver Hamiltonian :math:`H_d`, which does not commute with :math:`H_c`. This is similar to the mixer Hamiltonian in QAOA.
```
Pro-tip: don't over-explain! The reader only takes away the important information anyway and the less "noise" there is the easier it becomes to focus on what really matters

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# To keep things simple, we choose a sum over Pauli-X operations on each qubit:
```
ixfoduap, line 107:
```suggestion
# To keep things simple, we choose a sum over Pauli X operations on each qubit:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+cost_h, driver_h = qaoa.max_clique(graph, constrained=False)
+
+print("Cost Hamiltonian")
+print(cost_h)
+print("Driver Hamiltonian")
+print(driver_h)
+
+######################################################################
+# As you may recall from the Theory section, one of the main ingredients in the FALQON algorithm is the operator :math:`i [H_d, H_c]`. In
```
ixfoduap, line 122:
```suggestion
# One of the main ingredients in the FALQON algorithm is the operator :math:`i [H_d, H_c]`. In
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# the case of MaxClique, we can write down the commutator :math:`[H_d, H_c]` explicitly, exploiting the commutator's bilinearity:
```
ixfoduap, line 123:
```suggestion
# the case of MaxClique, we can write down the commutator :math:`[H_d, H_c]` explicitly:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+# .. math:: [H_d, H_c] = 3 \displaystyle\sum_{k \in V(G)} \displaystyle\sum_{(i, j) \in E(\bar{G})} \big( [X_k, Z_i Z_j] - [X_k, Z_i]
+#           - [X_k, Z_j] \big) + 3 \displaystyle\sum_{i \in V(G)} \displaystyle\sum_{j \in V(G)} [X_i, Z_j]
+#
+# Clearly, there are two distinct commutators that we must calculate, :math:`[X_k, Z_j]` and :math:`[X_k, Z_i Z_j]`.
```
ixfoduap, line 128:
```suggestion
# There are two distinct commutators that we must calculate, :math:`[X_k, Z_j]` and :math:`[X_k, Z_i Z_j]`.
```
If it's clear, do we really need to say so? :wink:

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# This is easy, as we know exactly what the
+# `commutators of the Pauli matrices <https://en.wikipedia.org/wiki/Pauli_matrices#Commutation_relations>`__ are.
+# We have:
+#
+# .. math:: [X_k, Z_j] = -2 i \delta_{kj} Y_k \ \ \ \text{and} \ \ \ [X_k, Z_i Z_j] = -2 i \delta_{ik} Y_k Z_j - 2i \delta_{jk} Z_i Y_k
+#
+# where :math:`\delta_{kj}` is the `Kronecker delta <https://en.wikipedia.org/wiki/Kronecker_delta>`__. Therefore, it follows that from substitution
+# into the above equation, and multiplication by :math:`i` that:
```
ixfoduap, line 136:
```suggestion
# where :math:`\delta_{kj}` is the `Kronecker delta <https://en.wikipedia.org/wiki/Kronecker_delta>`__. Therefore it follows from substitution
# into the above equation and multiplication by :math:`i` that:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+# .. math:: i [H_d, H_c] = 6 \displaystyle\sum_{k \in V(G)} \displaystyle\sum_{(i, j) \in E(\bar{G})} \big( \delta_{ki} Y_k Z_j +
+#          \delta_{kj} Z_{i} Y_{k} - \delta_{ki} Y_k - \delta_{kj} Y_k \big) + 6 \displaystyle\sum_{i \in V(G)} Y_{i}
+#
+# This new operator has quite a few terms! Therefore, we write a short method which computes it for us, and return a :class:`~.pennylane.Hamiltonian` object.
+# Note that this method works for any graph:
```
ixfoduap, line 142: Wonderful!

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+        H += 6 * qml.PauliY(k)
+
+    return H
+
+
+print("MaxClique Commutator")
+print(build_commutator(graph))
+
+######################################################################
+# We now must build the actual FALQON algorithm. Our goal is to evolve some initial state under the Hamiltonian :math:`H`, with our
```
ixfoduap, line 169:
```suggestion
# We now build the actual FALQON algorithm. Our goal is to evolve some initial state under the Hamiltonian :math:`H`, with our
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# use the :class:`~.pennylane.templates.ApproxTimeEvolution` template to perform a single layer of Trotterized
+# time-evolution:
+
+def falqon_layer(beta_k, cost_h, driver_h, delta_t):
+    qml.templates.ApproxTimeEvolution(cost_h, delta_t, 1)
+    qml.templates.ApproxTimeEvolution(driver_h, delta_t * beta_k, 1)
+
+######################################################################
+# We then define a method which returns a FALQON ansatz corresponding to a particular cost Hamiltonian, driver
+# Hamiltonian, and :math:`\Delta t`. This simply involves repeating the FALQON layer multiple times. Note that the
```
ixfoduap, line 187:
```suggestion
# Hamiltonian, and :math:`\Delta t`. This involves repeating the FALQON layer multiple times. The
```
To soften the use of easily, simply, etc.
Also, note that using "note that" isn't always necessary or helpful

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+            beta,
+            cost_h=cost_h,
+            driver_h=driver_h,
+            delta_t=delta_t
+        )
+
+    return ansatz
+
+######################################################################
+# Finally, before we put everything together and run the circuit, there is still one issue that must be addressed.
```
ixfoduap, line 207: I think that this should eb explained earlier. You make it sound like it's an issue that has to be dealt with, but I see it as a crucial (and clever!) characteristic of the algorithm. I would prefer to see this explained together with the main idea of the algorithm, before any code is written

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# it is the case that :math:`\beta_k` is dependent on the state of the system at some time, as
+# we have defined:
+#
+# .. math:: \beta(t) = - \langle \psi(t) | i [H_d, H_c] | \psi(t) \rangle
+#
+# Our strategy is to use the value of :math:`A(t) := i\langle [H_d, H_c] \rangle_t` obtained by evaluating the circuit for the previous time-step:
+#
+# .. math:: \beta_{k+1} = -A_k = -A(k\Delta t)
+#
+# This leads immediately to the FALQON algorithm as a recursive process that feeds back into itself.
```
ixfoduap, line 218:
```suggestion
# This leads to the FALQON algorithm as a recursive process that feeds back into itself.
```
Another tip: keep adjectives and adverbs to a minimum; makes for smoother writing

ixfoduap: Also, isn't feeding back into itself the definition of recursive? :thinking:

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# 2. Measure the expectation value :math:`A_k = \langle i[H_c, H_d]\rangle_k`.
+# 3. Set :math:`\beta_{k+1} = -A_k`.
+#
+# We repeat for all :math:`k` from :math:`1` to :math:`n`. At the final step, we can evaluate :math:`\langle H_c \rangle`.
+#
+# .. figure:: ../demonstrations/falqon/falqon.png
+#     :align: center
+#     :width: 80%
+#
+# Implementing this recursive process isn't too difficult, we simply make use of the methods defined above:
```
ixfoduap, line 231:
```suggestion
# Implementing this recursive process can be done using the methods defined above:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+        print("Step {} Done, Cost = {}".format(i + 1, energy))
+
+    return beta, energies
+
+######################################################################
+# Note that we return both the list of :math:`\beta_k` values, as well as the expectation value of the cost Hamiltonian
+# for each step.
+#
+# Without further delay, we can run FALQON for our MaxClique problem! It is important that we choose :math:`\Delta t` small enough
```
ixfoduap, line 261:
```suggestion
# We can now run FALQON for our MaxClique problem! It is important that we choose :math:`\Delta t` small enough
```
Was the above content a delay? I don't think so! It was important and useful information

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# such that our approximate time-evolution is close enough to the real time-evolution, otherwise we the expectation value of
```
ixfoduap, line 262:
```suggestion
# such that the approximate time-evolution is close enough to the real time-evolution, otherwise we the expectation value of
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+n = 40
+beta_1 = 0.0
+delta_t = 0.03
+
+dev = qml.device("default.qubit", wires=graph.nodes) # Creates a device for the simulation
+res_beta, res_energies = max_clique_falqon(graph, n, beta_1, delta_t, dev)
+
+######################################################################
+# As expected, the expectation value of the cost Hamiltonian strictly decreases!
```
ixfoduap, line 273:
```suggestion
# As expected, cost function strictly decreases!
```
This avoids repeated use of expected, expectation

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+
+plt.plot(range(n+1)[1:], res_energies)
+plt.xlabel("Iteration")
+plt.ylabel("Cost Function Value")
+plt.show()
+
+######################################################################
+# To get a better idea of how well
+# FALQON did, we can create a graph showing the probability of measuring each possible bitstring. We create
```
ixfoduap, line 283:
```suggestion
# To get a better understanding of the performance of the FALQON algorithm,
# we can create a graph showing the probability of measuring each possible bitstring. We define
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+
+probs = prob_circuit()
+plt.bar(range(2**len(dev.wires)), probs)
+plt.xlabel("Bitstring")
+plt.ylabel("Measurement Probability")
+plt.show()
+
+######################################################################
+# Clearly, the bitstring occurring with the highest probability is the state :math:`|28\rangle = |11100\rangle`.
```
ixfoduap, line 302:
```suggestion
# The bitstring occurring with the highest probability is the state :math:`|28\rangle = |11100\rangle`.
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# This corresponds to nodes :math:`0`, :math:`1`, and :math:`2`, which is precisely the triangle in our graph,
+# and is clearly the maximum clique. FALQON has solved the MaxClique problem!
```
ixfoduap, line 304:
```suggestion
# This corresponds to nodes :math:`0`, :math:`1`, and :math:`2`, which is precisely the maximum clique.
# FALQON has solved the MaxClique problem!
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+
+######################################################################
+# Benchmarking FALQON
+# -------------------
+#
+# Now that we have seen how FALQON works, it is worth noting how well FALQON performs according to a set of benchmarking
+# criteria. Let's see how FALQON performs on a batch of graphs! We generate graphs randomly using the
+# `Erdos-Renyi model <https://en.wikipedia.org/wiki/Erd%C5%91s%E2%80%93R%C3%A9nyi_model>`__, where we start with
+# the complete graph on :math:`n` vertices and then keep each edge with probability :math:`p`. We then find the maximum cliques on
```
ixfoduap, line 314: Funny, I always viewed it as starting from an empty graph and adding nodes with probability p. Worth double-checking what your code actually does

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# these graphs, classically, using the `Bron-Kerbosch algorithm <https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm>`__.
```
ixfoduap, line 315:
```suggestion
# these graphs using the `Bron-Kerbosch algorithm <https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm>`__.
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# .. math:: r_A = \frac{\langle H_C\rangle - \langle H_C\rangle_\text{min}}{|\langle H_C\rangle_\text{min}|}
+#
+# and (b) :math:`\phi`, the squared overlap with the (possibly degenerate) ground states for the cost Hamiltonian:
+#
+# .. math:: \phi = \sum_K |\langle \psi| \psi_K\rangle|^2
+#
+# where :math:`|\psi\rangle` is the prepared state, and each :math:`|\psi_K\rangle` is a ground state of the cost Hamiltonian.
+#
+# Running FALQON for all these graphs is
+# expensive, so we will only show the final results for these figures of merit (along with the values of :math:`\beta`) as a function of layer.
```
ixfoduap, line 327:
```suggestion
# expensive, so we only show the final results for these figures of merit along with the values of :math:`\beta`.
```
What do you mean by "as a function of layer?"

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# Note that, due to computational constraints, we have average over only :math:`3` random graphs per node size, for size :math:`n = 6, 7, 8`,
```
ixfoduap, line 328:
```suggestion
# Due to computational constraints, we have averaged over :math:`3` random graphs per node size, for size :math:`n = 6, 7, 8`,
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# with probability :math:`p = 0.1` of keeping an edge. We run FALQON for :math:`15` steps, with :math:`\Delta t = 0.01`:
+#
+# .. figure:: ../demonstrations/falqon/bench.png
+#     :align: center
+#     :width: 60%
+#
+# As expected, the relative error
+# decreases with layers, and also improves with graph size. Although :math:`r_A = 0.9` is a large relative error, we expect further improvement
```
ixfoduap, line 336: Should this be "decreases with the number of layers?"

ixfoduap: I would remove the sentence "Although r=0.9 is a large relative error..."

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# with layers and graph size. The ground state overlap :math:`\phi` increases with layer, indicating improved overlap with the true maximum clique(s).
+# Note that :math:`\phi` lies above :math:`1` due to large degeneracy in largest cliques for small, sparse (:math:`p=0.1`) graphs.
+
+######################################################################
+# Bonus: Seeding QAOA with FALQON (Bird Seed 🦅)
```
ixfoduap, line 341: I would remove the Bonus: I made it all the way here and I'm actually interested in reading! So why make it sound like it's something extra rather than something important?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# ---------------------------------------
+#
+# .. figure:: ../demonstrations/falqon/bird_seed.png
+#     :align: center
+#     :width: 90%
+#
+# Both FALQON and QAOA have unique benefits and drawbacks.
+# While FALQON requires no classical optimization and is guaranteed to decrease the cost function
+# with each iteration, its circuit depth grows linearly with the number of iterations. On the other hand, QAOA
```
ixfoduap, line 350: Like I asked you on slack, I think this is not the only catch. I think it's fine to leave as is, but maybe other experienced readers may wonder about additional drawbacks

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# has a fixed circuit depth, but does require classical optimization, and is therefore subject to all of the drawbacks that
+# come with probing a cost landscape for the ground state.
+#
+# Despite having unique issues, QAOA and FALQON have many similarities, most notably, their circuit structure. Both involve alternating layers
```
ixfoduap, line 354:
```suggestion
# Despite having unique featurs, QAOA and FALQON have many similarities, most notably their circuit structure. Both involve alternating layers
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# # of time-evolution operators corresponding to a cost and a mixer/driver Hamiltonian. This suggests to us that we may be able
+# to combine FALQON and QAOA to yield a new optimization algorithm that leverages the benefits of both algorithms!
```
ixfoduap, line 356:
```suggestion
# to combine FALQON and QAOA to yield a new optimization algorithm that leverages the benefits of both!
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+#
+# Suppose we want to run a QAOA circuit of depth :math:`p`. Our ansatz will be of the form:
+#
+# .. math:: U_{\text{QAOA}} = e^{-i \alpha_p H_m} e^{-i \gamma_p H_c} \cdots e^{-i \alpha_1 H_m} e^{-i \gamma_1 H_c}
+#
+# for sets of parameters :math:`\{\alpha_k\}` and :math:`\{\gamma_k\}`, which are optimized.
+# If we run FALQON for :math:`p` steps, setting :math:`H_d = H_m` and use the same cost Hamiltonian, we will end up with
```
ixfoduap, line 363:
```suggestion
# If we run FALQON for :math:`p` steps, setting :math:`H_d = H_m`, and use the same cost Hamiltonian, we will end up with
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# the following ansatz:
+#
+# .. math:: U_{\text{FALQON}} = e^{-i \Delta t \beta_p H_d} e^{-i \Delta t H_c} \cdots e^{-i \Delta t \beta_1 H_d} e^{-i \Delta t H_c}
+#
+# Thus, our strategy is to initialize our QAOA parameters using the :math:`\beta_k` values that FALQON yields. More specifically,
+# we set :math:`\alpha_k = \Delta t \beta_k` and :math:`\gamma_k = \Delta t`. We then optimize over these parameters. Our hope is that
```
ixfoduap, line 369:
```suggestion
# we set :math:`\alpha_k = \Delta t \beta_k` and :math:`\gamma_k = \Delta t`. We then optimize over these parameters. The goal is that
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+# these parameters provide QAOA a good place in the parameter space to begin its optimization.
+#
+# Using the code from earlier in the demonstration, we can easily prototype this process. To illustrate the power of this new
+# technique, we will attempt to solve MaxClique on a slightly more complicated graph:
```
ixfoduap, line 373:
```suggestion
# technique, we attempt to solve MaxClique on a slightly more complicated graph:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+
+new_edges = [(0, 1), (1, 2), (2, 0), (2, 3), (1, 4), (4, 5), (5, 2), (0, 6)]
+new_graph = nx.Graph(new_edges)
+nx.draw(new_graph, with_labels=True)
+
+######################################################################
+# We can now use the PennyLane QAOA module to create a QAOA circuit corresponding to the MaxClique problem. For this
+# demonstration, we will set the depth to :math:`5`:
```
ixfoduap, line 381:
```suggestion
# demonstration, we set the depth to :math:`5`:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+    for w in dev.wires:
+        qml.Hadamard(wires=w)
+    qml.layer(qaoa_layer, depth, params[0], params[1])
+
+# Creates a cost function with executes the QAOA circuit
+cost_fn = qml.ExpvalCost(qaoa_circuit, cost_h, dev)
+
+######################################################################
+# Now, all we have to do is run FALQON for :math:`5` steps to get our initial QAOA parameters.
+# We will set :math:`\Delta t = 0.02`:
```
ixfoduap, line 405:
```suggestion
# Now all we have to do is run FALQON for :math:`5` steps to get our initial QAOA parameters.
# We set :math:`\Delta t = 0.02`:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+    return qml.probs(wires=dev.wires)
+
+probs = prob_circuit()
+plt.bar(range(2**len(dev.wires)), probs)
+plt.xlabel("Bitstring")
+plt.ylabel("Measurement Probability")
+plt.show()
+
+######################################################################
+# Clearly, the state :math:`|112\rangle = |1110000\rangle` occurs with highest probability.
```
ixfoduap, line 444:
```suggestion
# The state :math:`|112\rangle = |1110000\rangle` occurs with highest probability.
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,452 @@
...
+convergence isn't guaranteed, and the optimization procedure can become "stuck" in local minima.
+
+.. figure:: ../demonstrations/falqon/global_min.png
+    :align: center
+    :width: 70%
+
+In this demo, we'll be implementing the FALQON algorithm: a feedback-based algorithm
+for quantum optimization, introduced by `Magann, Rudinger, Grace & Sarovar (2021) <https://arxiv.org/pdf/2103.08619.pdf>`__.
+It is similar in spirit to QAOA, but instead uses iterative feedback steps rather than a global optimization
+over parameters, therefore avoiding the issue of local minima. We will show how to implement FALQON in PennyLane
```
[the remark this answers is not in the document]
Lucaman99: This is true, although it avoids the issue to a large degree, at least empirically. The catch is massive circuit depth.

demonstrations/tutorial_falqon.py
Lucaman99: Ah, true!

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+To solve combinatorial optimization problems using a quantum computer, a typical strategy is to encode
+the solution to the problem as the ground state of *cost Hamiltonian* :math:`H_c`, and then use some procedure to drive
+the system from an initial state into the ground state of :math:`H_c`. FALQON falls under this broad scheme.
+
+Consider a quantum system governed by a Hamiltonian of the form :math:`H = H_c + \beta(t) H_d`. These kinds of
+Hamiltonians appear often in the theory of `quantum control <https://quantiki.org/wiki/quantum-control-theory>`__: a
+field of inquiry which studies how a quantum system can be driven from one state to another.
+The choice of :math:`\beta(t)` allows us to decide which state we
+want a system governed by such a Hamiltonian to evolve towards.
```
glassnotes, line 44: I found this sentence a bit confusing. It means to say here that tuning the choice of beta(t) affects the state that the system evolves towards? Won't the driver Hamiltonian also play a role?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+The time-dependent Schrodinger equation tells us that the dynamics of the system are given by:
+
+.. math:: i \frac{d}{dt} |\psi(t)\rangle = (H_c + \beta(t) H_d) |\psi(t)\rangle,
+
+where we set :math:`\hbar = 1`. Now suppose our objective is to drive the system
+to the state :math:`|\psi\rangle`: the ground state of :math:`H_c`. Phrased differently, we would like to minimize
```
glassnotes, line 51:
```suggestion
to the ground state of :math:`H_c`, which we denote by :math:`|\psi\rangle`. Phrased differently, we would like to minimize
```

Just to emphasize what the state conceptually is before giving it a label.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+Theory
+------
+
+To solve combinatorial optimization problems using a quantum computer, a typical strategy is to encode
+the solution to the problem as the ground state of *cost Hamiltonian* :math:`H_c`, and then use some procedure to drive
+the system from an initial state into the ground state of :math:`H_c`. FALQON falls under this broad scheme.
+
+Consider a quantum system governed by a Hamiltonian of the form :math:`H = H_c + \beta(t) H_d`. These kinds of
+Hamiltonians appear often in the theory of `quantum control <https://quantiki.org/wiki/quantum-control-theory>`__: a
```
glassnotes, line 41:
```suggestion
Hamiltonians appear often in the theory of `quantum control <https://quantiki.org/wiki/quantum-control-theory>`__, a
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+then :math:`\langle H_c \rangle` is guaranteed to strictly decrease, as desired!
+(Note that we bring the :math:`i` into the expectation to give a Hermitian operator.)
+
+Using `techniques from control theory <https://arxiv.org/pdf/1304.3997.pdf>`__, it is possible to rigorously show
+this choice of :math:`\beta(t)` will eventually drive the system into the ground state of :math:`H_c`. Thus, if we
+evolve some initial state :math:`|\psi_0\rangle` under the time-evolution operator corresponding to :math:`H`, given
+by:
+
+.. math:: U(T) = \mathcal{T} \exp \Big[ -i \displaystyle\int_{0}^{T} H(t) \ dt \Big]
+
```
glassnotes, line 72:
```suggestion
by

.. math:: U(T) = \mathcal{T} \exp \Big[ -i \displaystyle\int_{0}^{T} H(t) \ dt \Big],
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+we will arrive at the ground state of :math:`H_c`. This is exactly the procedure used by FALQON.
+
+In general, implementing a time-evolution unitary of the form :math:`U(T)` in a quantum circuit is
+difficult, so we use a
+`Trotter-Suzuki decomposition <https://en.wikipedia.org/wiki/Time-evolving_block_decimation#The_Suzuki%E2%80%93Trotter_expansion>`__
+to perform approximate time-evolution. We know that:
+
+.. math:: \displaystyle\int_{0}^{T} H(t) \ dt \approx \displaystyle\sum_{k = 0}^{T/\Delta t} H( k \Delta t) \Delta t
+
+for some small time-step :math:`\Delta t`. Thus, we will have:
```
glassnotes, line 83:
```suggestion
for some small time step :math:`\Delta t`. Thus, we will have:
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+.. math:: U(T) = \mathcal{T} \exp \Big[ -i \displaystyle\int_{0}^{T} H(t) \ dt \Big]
+
+where :math:`\mathcal{T}` is the `time-ordering operator <https://en.wikipedia.org/wiki/Path-ordering#Time_ordering>`__,
+we will arrive at the ground state of :math:`H_c`. This is exactly the procedure used by FALQON.
+
+In general, implementing a time-evolution unitary of the form :math:`U(T)` in a quantum circuit is
+difficult, so we use a
+`Trotter-Suzuki decomposition <https://en.wikipedia.org/wiki/Time-evolving_block_decimation#The_Suzuki%E2%80%93Trotter_expansion>`__
+to perform approximate time-evolution. We know that:
```
glassnotes, line 79:
```suggestion
to perform approximate time-evolution. We know that
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+.. math:: \displaystyle\int_{0}^{T} H(t) \ dt \approx \displaystyle\sum_{k = 0}^{T/\Delta t} H( k \Delta t) \Delta t
+
+for some small time-step :math:`\Delta t`. Thus, we will have:
+
+.. math:: U(T) \approx \mathcal{T} \exp \Big[ -i \displaystyle\sum_{k = 0}^{T/\Delta t} H( k \Delta t) \Delta t \Big] \approx
+            e^{-i\beta_n H_d \Delta t} e^{-iH_c \Delta t} \cdots e^{-i\beta_1 H_d \Delta t} e^{-iH_c \Delta t} = U_d(\beta_n) U_c \cdots U_d(\beta_1) U_c
+
+where :math:`n = T/\Delta t` and :math:`\beta_k = \beta(k\Delta t)`.
+
+For each layer of the time-evolution, we need to know the value :math:`\beta_k`. However,
```
glassnotes, line 90:
```suggestion
For each layer of the time evolution, we need to know the value :math:`\beta_k`. However,
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+.. math:: \beta_{k+1} = -A_k = -A(k\Delta t).
+
+This leads to the FALQON algorithm as a recursive process (in other words, it feeds back into itself).
+On step :math:`k`, we perform the following three substeps:
+
+1. Prepare the state :math:`|\psi_k\rangle = U_d(\beta_k) U_c \cdots U_d(\beta_1) U_c|\psi_0\rangle`.
+2. Measure the expectation value :math:`A_k = \langle i[H_c, H_d]\rangle_k`.
+3. Set :math:`\beta_{k+1} = -A_k`.
+
+We repeat for all :math:`k` from :math:`1` to :math:`n`. At the final step, we can evaluate :math:`\langle H_c \rangle`.
```
glassnotes, line 108: How is the number of steps, n, selected?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+import networkx as nx
+
+######################################################################
+# In this demonstration, we will be using FALQON to solve the
+# `maximum clique (MaxClique) problem <https://en.wikipedia.org/wiki/Clique_problem>`__: finding the
+# largest complete subgraph of some graph :math:`G`. For example, the following graph's maximum
+# clique is coloured in red:
+#
+# .. figure:: ../demonstrations/falqon/max_clique.png
+#     :align: center
```
glassnotes, line 134: For these and other networkx drawings, it would be helpful to have the nodes a lighter colour so that the labels are more clearly visible.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+#
+# We attempt to find the maximum clique of the following graph:
+#
+
+edges = [(0, 1), (1, 2), (2, 0), (2, 3), (1, 4)]
+graph = nx.Graph(edges)
+nx.draw(graph, with_labels=True)
+
+######################################################################
+# We must first encode this combinatorial problem into a cost Hamiltonian :math:`H_c`. This ends up being:
```
glassnotes, line 145:
```suggestion
# We must first encode this combinatorial problem into a cost Hamiltonian :math:`H_c`. This ends up being
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+#
+# .. math:: H_c = 3 \sum_{(i, j) \in E(\bar{G})} (Z_i Z_j - Z_i - Z_j) + \displaystyle\sum_{i \in V(G)} Z_i,
```
glassnotes, line 147: In the main text, define what \bar{G} is. This helps explain why it is computed in the Hamiltonian construction function below.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+    return H
+
+
+print("MaxClique Commutator")
+print(build_commutator(graph))
+
+######################################################################
+# We can now build the FALQON algorithm. Our goal is to evolve some initial state under the Hamiltonian :math:`H`,
+# with our chosen :math:`\beta(t)`. We first define one layer of our Trotterized time-evolution, which is of
```
glassnotes, line 218:
```suggestion
# with our chosen :math:`\beta(t)`. We first define one layer of our Trotterized time evolution, which is of
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+    return ansatz
+
+######################################################################
+# Finally, we implement the recursive process, where FALQON is able to determine the values
+# of :math:`\beta_k`, feeding back into itself as the number of layers increases. This is
+# straightforward using the methods defined above:
+
+def max_clique_falqon(graph, n, beta_1, delta_t, dev):
+    hamiltonian = build_commutator(graph) # Builds the commutator
```
glassnotes, line 252: Since `cost_h` and `driver_h` are defined, maybe keep with that convention and call this `commutator_h` or `comm_h` so it's clear which Hamiltonian it is.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+        # Creates a function which returns the expectation value of the cost Hamiltonian
+        cost_fn_energy = qml.ExpvalCost(ansatz, cost_h, dev)
+
+        # Adds a value of beta to the list and evaluates the cost function
+        beta.append(-1 * cost_fn(beta))
+        energy = cost_fn_energy(beta)
+        energies.append(energy)
+
+        print("Step {} Done, Cost = {}".format(i + 1, energy))
```
glassnotes, line 271: To save length, and since the results are going in a plot afterwards anyways, I recommend to remove the print statement.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+    return beta, energies
+
+######################################################################
+# Note that we return both the list of :math:`\beta_k` values, as well as the expectation value of the cost Hamiltonian
+# for each step.
+#
+# We can now run FALQON for our MaxClique problem! It is important that we choose :math:`\Delta t` small enough
+# such that the approximate time-evolution is close enough to the real time-evolution, otherwise we the expectation
+# value of :math:`H_c` may not strictly decrease. For this demonstration, we set :math:`\Delta t = 0.03`,
```
glassnotes, line 281: What is the motivation behind these choices?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+plt.bar(range(2**len(dev.wires)), probs)
+plt.xlabel("Bitstring")
+plt.ylabel("Measurement Probability")
+plt.show()
+
+######################################################################
+# The bitstring occurring with the highest probability is the state :math:`|28\rangle = |11100\rangle`.
+# This corresponds to nodes :math:`0`, :math:`1`, and :math:`2`, which is precisely the maximum clique.
+# FALQON has solved the MaxClique problem!
+#
```
glassnotes, line 324: Might be nice to show a picture of the same graph, to remind the reader what it looks like, but with the max clique nodes in a different colour.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+
+######################################################################
+# Benchmarking FALQON
+# -------------------
+#
+# Now that we have seen how FALQON works, it is worth noting how well FALQON performs according to a set of benchmarking
+# criteria. Let's see how FALQON performs on a batch of graphs! We generate graphs randomly using the
```
glassnotes, line 331:
```suggestion
# criteria on a batch of graphs! We generate graphs randomly using the
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+# and (b) :math:`\phi`, the squared overlap with the (possibly degenerate) ground states for the cost Hamiltonian:
+#
+# .. math:: \phi = \sum_K |\langle \psi| \psi_K\rangle|^2,
+#
+# where :math:`|\psi\rangle` is the prepared state, and each :math:`|\psi_K\rangle` is a ground state of the cost
+# Hamiltonian.
+#
+# Running FALQON for all these graphs is
+# expensive, so we will only show the final results for these figures of merit (along with the values of :math:`\beta`)
+# as a function of layer. Due to computational constraints, we have averaged over :math:`3` random graphs per node
```
glassnotes, line 349: I think there needs to be a lot more random graphs to make this section valuable (now that I see it in context, the version I ran had only 5 per step, which is still very low). Where is the bottleneck in these simulations?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+# size, for size :math:`n = 6, 7, 8`, with probability :math:`p = 0.1` of keeping an edge. We run FALQON for
+# :math:`15` steps, with :math:`\Delta t = 0.01`:
+#
+# .. figure:: ../demonstrations/falqon/bench.png
+#     :align: center
+#     :width: 60%
```
glassnotes, line 355: The metric \phi is not very intuitive because of the degeneracy; can it be normalized to between 0 and 1?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+#
+# As expected, the relative error
+# decreases with the number of layers, and also improves with graph size.
```
glassnotes, line 358: Why would the graph size have an effect on the results?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+#     :width: 90%
+#
+# Both FALQON and QAOA have unique benefits and drawbacks.
+# While FALQON requires no classical optimization and is guaranteed to decrease the cost function
+# with each iteration, its circuit depth grows linearly with the number of iterations. On the other hand, QAOA
+# has a fixed circuit depth, but does require classical optimization, and is therefore subject to all of the drawbacks
+# that come with probing a cost landscape for the ground state.
+#
+# Despite having unique issues, QAOA and FALQON have many similarities, most notably, their circuit structure. Both
+# involve alternating layers of time-evolution operators corresponding to a cost and a mixer/driver Hamiltonian.
```
glassnotes, line 378:
```suggestion
# involve alternating layers of time evolution operators corresponding to a cost and a mixer/driver Hamiltonian.
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+plt.bar(range(2**len(dev.wires)), probs)
+plt.xlabel("Bitstring")
+plt.ylabel("Measurement Probability")
+plt.show()
+
+######################################################################
+# The state :math:`|112\rangle = |1110000\rangle` occurs with highest probability.
+# This corresponds to nodes :math:`0`, :math:`1`, and :math:`2` of the graph, which is
+# the maximum clique! We have successfully combined FALQON and QAOA to solve a combinatorial
+# optimization problem 🎉.
```
glassnotes, line 472: This is an interesting application!

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+print("Running FALQON")
+res, res_energy = max_clique_falqon(new_graph, depth-1, 0.0, delta_t, dev)
+
+params = np.array([[delta_t for k in res], [delta_t * k for k in res]])
+
+######################################################################
+# Finally, we run our QAOA optimization procedure. We set the number of QAOA executions to :math:`40`:
+#
+
+print("Running QAOA")
```
glassnotes, line 443: I think this is out of scope for inclusion in the demo, but it could be interesting to explore and then discuss; can we quantify how much better it is to use initial parameters from FALQON rather than choosing them at random? Is the convergence time perhaps faster? If FALQON is resource intensive, what is the optimal number of FALQON layer that is likely to improve convergence? (Maybe just a few iterations is enough to show significant benefit?)

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+.. related::
+
+   tutorial_qaoa_intro Intro to QAOA
+   tutorial_qaoa_maxcut QAOA for MaxCut
+
+*Authors: David Wakeham and Jack Ceroni. Posted: XXX. Last updated: XXX.*
+
+-----------------------------
+
+While the
```
glassnotes, line 18: It might be worth adding a note somewhere in the introduction recommending that readers are familiar with QAOA before reading this demo (you can link to the Intro QAOA demo as a reference). Without that background, some of the theory especially might be tricky.

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,477 @@
...
+#     :width: 60%
+#
+# As expected, the relative error
+# decreases with the number of layers, and also improves with graph size.
+# We expect further improvement with layers and graph size. The ground state overlap :math:`\phi` increases with layer,
+# indicating improved overlap with the true maximum clique(s). Note that :math:`\phi` lies above :math:`1` due to large
+# degeneracy in largest cliques for small, sparse (:math:`p=0.1`) graphs.
+
+######################################################################
+# Seeding QAOA with FALQON (Bird Seed 🦅)
```
glassnotes, line 364: This appears to be something that is not explicitly in the paper, but just mentioned as a possible extension in the conclusion. I think it'd be worth briefly mentioning that here to differentiate this section from the demonstration of the original work.

glassnotes (changes requested):
@Lucaman99 @hapax This is a really neat demo! It's clear how much effort you've put into understanding the paper (and beyond!), so great work :100:

I've left a number of specific comments. As general feedback, I found especially the first section to be quite math heavy (though, well-explained!). I recommend to consider again each equation and remove any that aren't absolutely essential to explaining how things work. Another option to help break up the theory is to move the explanation and diagram of the FALQON algorithm earlier on, and use the algorithm itself to motivate the equations (i.e., show the diagram with the curvy arrows, and then explain that expectation values of these specific commutators are measured because it guarantees a strict decrease).

demonstrations/tutorial_falqon.py
hapax: For these small graphs, my computer, I would happy ask your computer for more :) For larger n, we need to tune Delta t more carefully to ensure the Trotterization works, which is a pain and will slow it down.

glassnotes: Feel free to shoot over another script and I can run it. Also, since the problems are independent, it may also be helpful to add some simple multiprocessing to parallelize it, either at the level of n, or the graph instances themselves (4-6 threads is reasonable and would go a long way towards speeding things up).

demonstrations/tutorial_falqon.py
Lucaman99: Well, yes, but it's best to think about the driver playing the role analogous to the mixer in QAOA. We don't really think of the mixer itself as determining which state QAOA evolves towards, that is the job of the parameters and the cost Hamiltonian. Nevertheless, we could make a bit more clear 👍

demonstrations/tutorial_falqon.py
Lucaman99: Sure!

demonstrations/tutorial_falqon.py
Lucaman99: Well, the choice of beta_1 is pretty arbitrary, so setting it to 0 just makes sense. As for n and delta_t, they just work 🤷 . It is possible to derive exactly what delta_t should be such that the expectation value strictly decreases, but that just seemed too complicated for this tutorial.

demonstrations/tutorial_falqon.py
Lucaman99: Yeah, this would be very cool, especially since it is not explored at all in the paper. However, I think you're right: it's outside the scope of this tutorial. Perhaps we could look into it in the future though!

demonstrations/tutorial_falqon.py
Lucaman99: I see, what do you think we should say?

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+    ansatz = build_maxclique_ansatz(cost_h, driver_h, delta_t) # Builds the FALQON ansatz
+
+    beta = [beta_1] # Records each value of beta_k
+    energies = [] # Records the value of the cost function at each step
+
+    for i in range(n):
+        # Creates a function which can evaluate the expectation value of the commutator
+        cost_fn = qml.ExpvalCost(ansatz, comm_h, dev)
+
+        # Creates a function which returns the expectation value of the cost Hamiltonian
```
ixfoduap, line 258:
```suggestion
        # Creates a function that returns the expectation value of the cost Hamiltonian
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+def max_clique_falqon(graph, n, beta_1, delta_t, dev):
+    comm_h = build_hamiltonian(graph) # Builds the commutator
+    cost_h, driver_h = qaoa.max_clique(graph, constrained=False) # Builds H_c and H_d
+    ansatz = build_maxclique_ansatz(cost_h, driver_h, delta_t) # Builds the FALQON ansatz
+
+    beta = [beta_1] # Records each value of beta_k
+    energies = [] # Records the value of the cost function at each step
+
+    for i in range(n):
+        # Creates a function which can evaluate the expectation value of the commutator
```
ixfoduap, line 255:
```suggestion
        # Creates a function that can evaluate the expectation value of the commutator
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+# iterations of the algorithm:
+#
+
+plt.plot(range(n+1)[1:], res_energies)
+plt.xlabel("Iteration")
+plt.ylabel("Cost Function Value")
+plt.show()
+
+######################################################################
+# As expected, the expectation value decreases!
```
ixfoduap, line 295:
```suggestion
# The expectation value decreases!
```
I don't like "as expected, the expectation value"

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+#
+# To get a better understanding of the performance of the FALQON algorithm,
+# we can create a graph showing the probability of measuring each possible bitstring. We define
```
ixfoduap, line 298:
```suggestion
# we can create a graph showing the probability of measuring each possible bitstring. We define the
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+probs = prob_circuit()
+plt.bar(range(2**len(dev.wires)), probs)
+plt.xlabel("Bitstring")
+plt.ylabel("Measurement Probability")
+plt.show()
+
+######################################################################
+# The bitstring occurring with the highest probability is the state :math:`|28\rangle = |11100\rangle`.
+# This corresponds to nodes :math:`0`, :math:`1`, and :math:`2`, which is precisely the maximum clique.
+# FALQON has solved the MaxClique problem!
```
ixfoduap, line 319: Good place for an emoji, if you want to add one :tada:

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+graph = nx.Graph(edges)
+cmap = ["#00b4d9"]*3 + ["#e377c2"]*2
+nx.draw(graph, with_labels=True, node_color=cmap)
+
+######################################################################
+# Benchmarking FALQON
+# -------------------
+#
+# After seeing how FALQON works, it is worth considering how well FALQON performs according to a set of benchmarking
+# criteria on a batch of graphs! We generate graphs randomly using the
```
ixfoduap, line 331:
```suggestion
# criteria on a batch of graphs. We generate graphs randomly using the
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+# size, for sizes :math:`n = 6, 7, 8, 9`, with probability :math:`p = 0.1` of keeping an edge. Running FALQON for
+# :math:`40` steps, with :math:`\Delta t = 0.01`, produces:
+#
+# .. figure:: ../demonstrations/falqon/bench.png
+#     :align: center
+#     :width: 60%
+#
+# The relative error decreases with the number of layers and graph size, except for :math:`n = 9` where the step size has become too large
+# and the Trotter-Suzuki decomposition breaks down.
+# The rate of decrease also slows. Even though the algorithm will converge to the ground state,
```
ixfoduap, line 353:
```suggestion
# The rate of decrease also slows down. Even though the algorithm will converge to the ground state,
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+# no one said anything about how quickly it gets there!
```
ixfoduap, line 354: Maybe you should have said something? :wink:

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+# ---------------------------------------
+#
+# .. figure:: ../demonstrations/falqon/bird_seed.png
+#     :align: center
+#     :width: 90%
+#
+# Both FALQON and QAOA have unique benefits and drawbacks.
+# While FALQON requires no classical optimization and is guaranteed to decrease the cost function
+# with each iteration, its circuit depth grows linearly with the number of iterations. The benchmarking data also shows
+# how the reduction in cost slows with layer, and the additional burden of correctly tuning the time step. On the other hand, QAOA
```
ixfoduap, line 367:
```suggestion
# how the reduction in cost slows with each layer, and the additional burden of correctly tuning the time step. On the other hand, QAOA
```
or maybe "slows with the number of layers"

ixfoduap, line 367:
```suggestion
# how the reduction in cost slows with each layer, and the additional burden of correctly tuning the time step. On the other hand, QAOA
```
or maybe "slows with the number of layers"

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+# has a fixed circuit depth, but does require classical optimization, and is therefore subject to all of the drawbacks
+# that come with probing a cost landscape for a set of optimal parameters.
+#
+# QAOA and FALQON also have many similarities, most notably, their circuit structure. Both
+# involve alternating layers of time evolution operators corresponding to a cost and a mixer/driver Hamiltonian.
+# The authors of the FALQON paper therefore suggest combining FALQON and QAOA to yield a new optimization algorithm that
+# leverages the benefits of both! In this final section of the tutorial, we will implement this procedure in PennyLane.
```
ixfoduap, line 374:
```suggestion
# leverages the benefits of both. In this final section of the tutorial, we will implement this procedure in PennyLane.
```
I feel there are too many exclamation marks in this section! AT LEAST YOU ARE NOT USING CAPS!

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,471 @@
...
+#
+# Suppose we want to run a QAOA circuit of depth :math:`p`. Our ansatz will be of the form:
+#
+# .. math:: U_{\text{QAOA}} = e^{-i \alpha_p H_m} e^{-i \gamma_p H_c} \cdots e^{-i \alpha_1 H_m} e^{-i \gamma_1 H_c}
```
ixfoduap, line 378:
```suggestion
# .. math:: U_{\text{QAOA}} = e^{-i \alpha_p H_m} e^{-i \gamma_p H_c} \cdots e^{-i \alpha_1 H_m} e^{-i \gamma_1 H_c},
```

ixfoduap (changes requested):
Fantastic tutorial! Congratulations to the authors :medal_sports: :medal_sports: 

I made a few suggestions as a final round of polish. Once they are addressed, I'm happy to approve :eagle:

demonstrations/tutorial_falqon.py
```diff
@@ -350,8 +350,8 @@ def prob_circuit():
 #
 # The relative error decreases with the number of layers and graph size, except for :math:`n = 9` where the step size has become too large
 # and the Trotter-Suzuki decomposition breaks down.
-# The rate of decrease also slows. Even though the algorithm will converge to the ground state,
-# no one said anything about how quickly it gets there!
+# The rate of decrease also slows down. Even though the algorithm will converge to the ground state, it won't always get
+# there in few time steps!
```
ixfoduap, line 354:
```suggestion
# there in a few time steps!
```

ixfoduap (approved):
:shipit:

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,473 @@
...
+# Both FALQON and QAOA have unique benefits and drawbacks.
+# While FALQON requires no classical optimization and is guaranteed to decrease the cost function
+# with each iteration, its circuit depth grows linearly with the number of iterations. The benchmarking data also shows
+# how the reduction in cost slows with each layer, and the additional burden of correctly tuning the time step. On the other hand, QAOA
+# has a fixed circuit depth, but does require classical optimization, and is therefore subject to all of the drawbacks
+# that come with probing a cost landscape for a set of optimal parameters.
+#
+# QAOA and FALQON also have many similarities, most notably, their circuit structure. Both
+# involve alternating layers of time evolution operators corresponding to a cost and a mixer/driver Hamiltonian.
+# The authors of the FALQON paper therefore suggest combining FALQON and QAOA to yield a new optimization algorithm that
```
glassnotes, line 375:
```suggestion
# The FALQON paper raises the idea of combining FALQON and QAOA to yield a new optimization algorithm that
```
(The original phrasing sounded like there was a discussion with the authors, but here it's just that the idea is  stated in the paper)

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,473 @@
...
+# leverages the benefits of both. In this final section of the tutorial, we will implement this procedure in PennyLane.
+#
+# Suppose we want to run a QAOA circuit of depth :math:`p`. Our ansatz will be of the form:
```
glassnotes, line 378:
```suggestion
# Suppose we want to run a QAOA circuit of depth :math:`p`. Our ansatz will be of the form
```

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,473 @@
...
+# Creates a cost function with executes the QAOA circuit
+cost_fn = qml.ExpvalCost(qaoa_circuit, cost_h, dev)
+
+######################################################################
+# Now all we have to do is run FALQON for :math:`5` steps to get our initial QAOA parameters.
+# We set :math:`\Delta t = 0.02`:
+
+delta_t = 0.02
+
+print("Running FALQON")
```
glassnotes, line 430:
```suggestion
```

glassnotes: Perhaps print the resulting params instead (or just leave with no output)

demonstrations/tutorial_falqon.py
```diff
@@ -0,0 +1,473 @@
...
+res, res_energy = max_clique_falqon(new_graph, depth-1, 0.0, delta_t, dev)
+
+params = np.array([[delta_t for k in res], [delta_t * k for k in res]])
+
+######################################################################
+# Finally, we run our QAOA optimization procedure. We set the number of QAOA executions to :math:`40`:
+#
+
+print("Running QAOA")
```
glassnotes, line 439:
```suggestion
```

glassnotes (approved):
@Lucaman99 @hapax this looks great! 

Just a couple typos and minor wording suggestions. Once addressed, remember to change the dates, and then we can merge it in :tada:

PR #266, opened 2021-05-12
Add demonstration for kernels module

johannesjmeyer:
**Title:**
Quantum Embedding Kernels with PennyLane's kernels module

**Summary:**
It is demonstrated how PennyLane's kernels module can be used to simplify the implementation of quantum embedding kernels.

**Relevant references:**
[Associated QHACK submission](https://github.com/thubregtsen/qhack)
[Associated preprint arXiv:2105.02276](https://arxiv.org/abs/2105.02276)


josh146:
@johannesjmeyer feel free to modify the `requirements.txt` file to install PennyLane from your branch so that the demo builds!

johannesjmeyer:
@josh146, if I run 
``` 
return kernel_circuit(x1, x2, params)[0]
```
where the `kernel_circuit` returns `qml.probs()`, I get a `TypeError: 'MeasurementProcess' object is not subscriptable`. How do I mitigate this?

josh146:
@johannesjmeyer, it sounds like `kernel_circuit` hasn't had the QNode decorator applied? As it is returning a 'bare' `qml.MeasurementProcess` rather than an array of numeric values

johannesjmeyer:
> @johannesjmeyer, it sounds like `kernel_circuit` hasn't had the QNode decorator applied? As it is returning a 'bare' `qml.MeasurementProcess` rather than an array of numeric values

Thanks, stupid mistake!

co9olguy:
Hi @johannesjmeyer, what is the status of this demo? We're super excited to have the kernels module in the next release (coming mid-June), so it would be awesome to have the demo come sometime near that if possible

johannesjmeyer:
@co9olguy I just realized that we have to add a fix to PL core to make the demo work, I'll do that. Otherwise the demo is more or less in shape I would say.

co9olguy:
@johannesjmeyer I think the ideal landing spot for a general-purpose non-quantum mathematical function is in `qml.math`. This will allow it to not only be automatically differentiable, but also make AD possible with every ML framework

@josh146 can comment more if he has more thoughts

johannesjmeyer:
Great, `qml.math` sounds like the right place! See https://github.com/PennyLaneAI/pennylane/pull/1388.

dwierichs:
I think this might be ready for review now. The example is not perfect by means of dataset complexity but previously the demo was running for quite a long time (still 7 minutes now for the updated reduced size). On the other hand, the benefit of alignment training is showcased and the functionalities are there.

josh146:
Thanks @dwierichs and @johannesjmeyer!

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+    tutorial_kernel_based_training Kernel-based training with scikit-learn
+    tutorial_data_reuploading_classifier Classification with data reuploading
+
+*Authors: Peter-Jan Derks, Paul Fährmann, Elies Gil-Fuster, Tom
+Hubregtsen, Johannes Jakob Meyer and David Wierichs. Posted: 24 June 2021*
+
+Kernel methods are one of the cornerstones of classical machine
+learning. To understand what a kernel method does, let’s first revisit
+one of the simplest methods to assign binary labels to datapoints:
+linear classification.
```
mariaschuld, line 19: I would tell a reader more what to expect here...the header may help advanced users to jump to the next section.

```suggestion
Kernel methods are one of the cornerstones of classical machine
learning. This tutorial will introduce the basic functionality of PennyLane's  
[kernels module](https://pennylane.readthedocs.io/en/latest/code/qml_kernels.html). 

What are kernel methods?
-----------------------------------

To understand what a kernel method does, let’s first revisit
one of the simplest methods to assign binary labels to datapoints:
linear classification.
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+   y(\boldsymbol{x}) = \operatorname{sgn}(\langle \boldsymbol{w}, \boldsymbol{x}\rangle + b).
+
+The vector :math:`\boldsymbol{w}` points perpendicular to the line and
+thus determine its slope. The independent term :math:`b` specificies the
+position on the plane. In this form, linear classification can also be
+extended to higher dimensional vectors :math:`\boldsymbol{x}`, where a
+line does not divide the entire space into two regions anymore. Instead
+one needs a *hyperplane*. It is immediately clear that this method is
+not very powerful, as datasets that are not separable by a hyperplane
+can’t be treated.
```
mariaschuld, line 43:
```suggestion
can’t be classified without error.
```
Maybe more accurate?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+   y(\boldsymbol{x}) = \operatorname{sgn}\left(\sum_i \alpha_i \langle \phi(\boldsymbol{x}_i), \phi(\boldsymbol{x})\rangle + b\right).
+
+This rewriting might not seem useful at first, but notice the above
+formula only contains inner products between vectors in the embedding
+space:
+
+.. math::
+   k(\boldsymbol{x}_i, \boldsymbol{x}_j) = \langle \phi(\boldsymbol{x}_i), \phi(\boldsymbol{x}_j)\rangle.
+
+We call this function the *kernel*. The clue now is that we can often
```
mariaschuld, line 77:
```suggestion
We call this function the *kernel*. The important point is that we can often
```

Lol, I _think_ this is an exclusively German formulation.

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+.. math::
+   \phi((x_1, x_2)) = (x_1^2, \sqrt{2} x_1 x_2, x_2^2) \qquad
+   k(\boldsymbol{x}, \boldsymbol{y}) = x_1^2 y_1^2 + 2 x_1 x_2 y_1 y_2 + x_2^2 y_2^2 = \langle \boldsymbol{x}, \boldsymbol{y} \rangle^2
+
+This means by just replacing the regular scalar product in our linear
+classification with the map :math:`k`, we can actually express much more
+intricate decision boundaries!
+
+This is very important, because in many interesting cases the embedding :math:`\phi`
+will be much more costlier to compute than the kernel :math:`k`.
```
mariaschuld, line 91:
```suggestion
will be much costlier to compute than the kernel :math:`k`.
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+   |\psi(\boldsymbol{x})\rangle = U(\boldsymbol{x}) |0 \rangle.
+
+The kernel value is then given by the *overlap* of the associated
+embedded quantum states
+
+.. math::
+   k(\boldsymbol{x}_i, \boldsymbol{x}_j) = | \langle\psi(\boldsymbol{x}_i)|\psi(\boldsymbol{x}_j)\rangle|^2.
+
+This demo is based on Ref. [#Training_QEKs]_ and we will show the functionalities
+of the PennyLane ``kernels`` module implemented alongside with it.
```
mariaschuld, line 110: Maybe remove that sentence and add the reference to the first sentence, so readers know this important paper from the start?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+"""
+
+##############################################################################
+# A toy problem
+# -------------
+# In this demo, we will treat a toy problem that showcases the
+# inner workings of our approach and the available functionalities
```
mariaschuld, line 117: Maybe not clear here what "our approach" is: the idea of a kernel trick, or something you want to do in the demo?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+
+np.random.seed(1359)
+
+##############################################################################
+# And we proceed right away to create a dataset to work with, the
+# ``DoubleCake`` dataset. We define it as a class with a convenience
+# plotting method and you may skip ahead if you are not interested in the
+# class itself.
```
mariaschuld, line 131:
```suggestion
# plotting method. The details of the class are not important 
# to understand the remainder of the demo, so do not worry if it confuses you.
```
Maybe this is less negative?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+dataset = DoubleCake(3)
+
+dataset.plot(plt.gca(), show_sectors=True)
+
+##############################################################################
+# Defining a Quantum Embedding Kernel
+# -----------------------------------
+# PennyLane’s ``kernels`` module allows for a particularly simple
+# implementation of Quantum Embedding Kernels. The first ingredient we
+# need for this is an *Ansatz* that represents the unitary
```
mariaschuld, line 222:
```suggestion
# need for this is an *ansatz* that represents the unitary
```
not sure, but feels better

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# :math:`U(\boldsymbol{x})` we use for embedding the data into a quantum
```
mariaschuld, line 223:
```suggestion
# :math:`U(\boldsymbol{x})` we use to embed the data into a quantum
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# state as well as its adjoint :math:`U(\boldsymbol{x})^\dagger`.
```
mariaschuld, line 224:
```suggestion
# state, as well as its adjoint :math:`U(\boldsymbol{x})^\dagger`.
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# 6 sectors:
+
+dataset = DoubleCake(3)
+
+dataset.plot(plt.gca(), show_sectors=True)
+
+##############################################################################
+# Defining a Quantum Embedding Kernel
+# -----------------------------------
+# PennyLane’s ``kernels`` module allows for a particularly simple
```
mariaschuld, line 220: Link to module? I immediately clicked the highlight, and there was this slightly disappointed feeling that comes with links not being links :)

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
+r"""Quantum Embedding Kernels with the kernels module
+=====================================================
```
josh146, line 2: I'm not too fussed either way, but we could shorten the title to simply "Quantum embedding kernels", and have use of the kernel module be implicit?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+    for j, wire in enumerate(wires):
+        qml.Hadamard(wires=[wire])
+        qml.RZ(x[i % len(x)], wires=[wire])
+        i += inc
+        qml.RY(params[0, j], wires=[wire])
+
+    qml.broadcast(unitary=qml.CRZ, pattern="ring", wires=wires, parameters=params[1])
+
+
+@qml.template
```
mariaschuld, line 242: I am wondering if one could simply delete this? We are kind of phasing this decorator out (unless it gets repurposed some time). `adjoint` should work without it!

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+adjoint_ansatz = qml.adjoint(ansatz)
+
+
+def random_params(num_wires, num_layers):
+    """Generate random variational parameters in the shape for the ansatz."""
+    return np.random.uniform(0, 2 * np.pi, (num_layers, 2, num_wires))
+
+
+##############################################################################
+# Together with the Ansatz we only need a device to run the quantum circuit on.
```
mariaschuld, line 258:
```suggestion
# Together with the ansatz we only need a device to run the quantum circuit on.
```

demonstrations/tutorial_kernel_module.py
josh146: Agree, was just about to suggest this 💯

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# For the purposes of this tutorial we will use PennyLane’s ``default.qubit``
```
mariaschuld, line 259:
```suggestion
# For the purpose of this tutorial we will use PennyLane’s ``default.qubit``
```

demonstrations/tutorial_kernel_module.py
josh146: In terms of the summary/intro, could add

- an additional sentence inbetween the two, making note of the *quantum* nature of the tutorial
- extend the final sentence, so that a reader knows what the demo will achieve
- the paper this is based on.

```markdown
Kernel methods are one of the cornerstones of classical machine
learning. Something something extends to quantum.
This tutorial will introduce the basic functionality of PennyLane's  
[kernels module](https://pennylane.readthedocs.io/en/latest/code/qml_kernels.html)
by exploring...
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# datapoint and then the adjoint of the embedding of the second datapoint. We
+# finally extract the probabilities of observing each basis state.
+
+
+@qml.qnode(dev)
+def kernel_circuit(x1, x2, params):
+    ansatz(x1, params, wires=wires)
+    adjoint_ansatz(x2, params, wires=wires)
+
+    return qml.probs(wires=wires)
```
mariaschuld, line 277: We now have a [projector observable](https://pennylane.readthedocs.io/en/latest/code/api/pennylane.Projector.html) :)  

Up to you if you want to use it, but could be cool.

Alternatively (or additionally) you could link the [other kernels](https://pennylane.ai/qml/demos/tutorial_kernel_based_training.html) demo here, where the logic of this adjoint circuit is explained with an equation, and the projector is used. Or add a note to clarify that there are different options?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+
+
+def kernel(x1, x2, params):
+    return kernel_circuit(x1, x2, params)[0]
+
+
+##############################################################################
+# Before focusing on the kernel values we have to provide values for the
+# variational parameters. At this point we fix the number of layers in the
+# Ansatz circuit to :math:`6`.
```
mariaschuld, line 293:
```suggestion
# ansatz circuit to :math:`6`.
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+linear classification.
+
+Imagine we want to discern two different classes of points that lie in
+different corners of the plane. A linear classifier corresponds to
+drawing a line and assigning different labels to the regions on opposing
+sides of the line:
+
+.. figure:: ../demonstrations/kernel_module/linear_classification.png
+   :align: center
+   :width: 30%
```
josh146, line 28: I love the drawings!

demonstrations/tutorial_kernel_module.py
josh146: (throughout this file): I recommend replacing unicode quotation marks with ASCII ones:
```suggestion
can't be treated.
```

There have been issues in the past with users on Windows attempting to run the demos, but experiencing invalid encoding errors.

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+from sklearn.svm import SVC
+
+##############################################################################
+# To construct the SVM, we need to supply ``sklearn.svm.SVC`` with a function
+# that takes two sets of datapoints and returns the associated kernel matrix.
+# We can make use of the function ``qml.kernels.kernel_matrix`` that provides
+# this functionality. It expects the kernel to not have additional parameters
+# besides the datapoints, which is why we again supply the variational
+# parameters via the ``lambda`` function from above.
+# Once we have this, we can let scikit adjust the SVM from our Quantum Embedding
```
mariaschuld, line 335:
```suggestion
# Once we have this, we can let scikit-learn adjust the SVM from our Quantum Embedding
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# Kernel.
+#
+# .. note::
+#     this step does *not* modify the variational parameters in our circuit
```
mariaschuld, line 339:
```suggestion
#     This step does *not* modify the variational parameters in our circuit
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+#     Ansatz. What it does is solving a different optimization task for the
+#     :math:`\alpha` and :math:`b` vectors we introduced in the beginning.
```
mariaschuld, line 341: Nice that you explain this, people get quite confused.

demonstrations/tutorial_kernel_module.py
mariaschuld: I think the link will be important because even advanced people sometimes get confused about this structure of the circuit.

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+   k(\boldsymbol{x}_i, \boldsymbol{x}_j) = \langle \phi(\boldsymbol{x}_i), \phi(\boldsymbol{x}_j)\rangle.
+
+We call this function the *kernel*. The clue now is that we can often
+find an explicit formula for the kernel :math:`k` that makes it
+superfluous to actually perform the embedding :math:`\phi`. Consider for
+example the following embedding and the associated kernel:
+
+.. math::
+   \phi((x_1, x_2)) = (x_1^2, \sqrt{2} x_1 x_2, x_2^2) \qquad
+   k(\boldsymbol{x}, \boldsymbol{y}) = x_1^2 y_1^2 + 2 x_1 x_2 y_1 y_2 + x_2^2 y_2^2 = \langle \boldsymbol{x}, \boldsymbol{y} \rangle^2
```
josh146, line 84:
```suggestion
.. math::
   \phi((x_1, x_2)) = (x_1^2, \sqrt{2} x_1 x_2, x_2^2), \qquad
   k(\boldsymbol{x}, \boldsymbol{y}) = x_1^2 y_1^2 + 2 x_1 x_2 y_1 y_2 + x_2^2 y_2^2 = \langle \boldsymbol{x}, \boldsymbol{y} \rangle^2.
```

josh146: Might be better to split this over two lines with `\\` rather than `\qquad`, for mobile users :)

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+
+##############################################################################
+# We see the outer points in the dataset can be correctly classified, but
+# we still struggle with the inner circle. But remember we have a circuit
+# with many free parameters! It is reasonable to believe we can give
+# values to those variational parameters which improve the overall accuracy
+# of our SVC.
+#
+# Training the Quantum Embedding Kernel
+# -------------------------------------
```
mariaschuld, line 401: Just a thought: Do you want to advertise this much higher up, maybe in the first sentence or even the title? This may increase clicks by a factor 10, people _love_ blindly training things, and kernels may be the next big thing to blindly train :)

demonstrations/tutorial_kernel_module.py
josh146: Agree

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+#
+# .. note::
+#     Seen from a more theoretical side, :math:`\operatorname{KA}`
+#     is nothing else than the cosine of the angle between the kernel
+#     matrices :math:`K_1` and :math:`K_2` if we see them as vectors
+#     in the space of matrices with the Hilbert-Schmidt (or
+#     Frobenius) scalar product
+#     :math:`\langle A, B \rangle = \operatorname{Tr}(A^T B)`. This
+#     reinforces the geometric picture of how this measure relates
+#     to objects, namely two kernels, being aligned in a vector space.
```
mariaschuld, line 427: Interesting!

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+#
+# The training data enters the picture by defining an *ideal* kernel
+# function that expresses the original labelling in the vector
+# :math:`\boldsymbol{y}` by assigning to two datapoints the product
+# of the corresponding labels:
+#
+# .. math::
+#    k_{\boldsymbol{y}}(\boldsymbol{x}_i, \boldsymbol{x}_j) = y_i y_j
```
mariaschuld, line 435:
```suggestion
#    k_{\boldsymbol{y}}(\boldsymbol{x}_i, \boldsymbol{x}_j) = y_i y_j.
```

mariaschuld: I am suddenly wondering if there should be punctuation in the equations? Ignore if not sure!

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# A toy problem
+# -------------
+# In this demo, we will treat a toy problem that showcases the
+# inner workings of our approach and the available functionalities
+# in PennyLane. We of course need to start with some imports:
+
+import pennylane as qml
+from pennylane import numpy as np
+import matplotlib as mpl
+import matplotlib.pyplot as plt
```
josh146, line 123: Minor point, but we have slowly started preferring placing the imports directly above the first code-blocks that require them. This serves two purposes:

- Pedagogically, readers can more easily associate the imports with the portions of the code that use them
- If a reader can't easily install or use a third party dependency (such as matplotlib), they are not limited from running earlier parts of the demo

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# where :math:`N` is the number of elements in :math:`\boldsymbol{y}`,
+# that is the number of datapoints in the dataset.
+#
+# In summary, the kernel-target alignment effectively captures how well
+# the kernel you chose reproduces the actual similarities of the data. It
+# does have one drawback, however: having a high kernel-target alignment
+# is only a necessary but not a sufficient condition for a good
+# performance of the kernel [#Alignment]_. This means having good alignment is
+# guaranteed for good performance, but optimal alignment will not always
+# bring optimal training accuracy with it.
```
mariaschuld, line 458: Great comment!

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# sample smaller subsets of the data at each step, we choose :math:`4`
+# datapoints at random. Remember that PennyLane’s inbuilt optimizer works
+# to *minimize* the cost function that is given to it, which is why we
+# have to multiply the kernel target alignment by :math:`-1` to actually
+# *maximize* it in the process.
+#
+# .. note::
+#     Currently, the function ``qml.kernels.target_alignment`` is not
+#     differentiable yet, making it unfit for gradient descent optimization.
+#     We therefore first define a differentiable version of this function.
```
mariaschuld, line 487: Just curious, could we simply rewrite it using the `qml.math` module?

demonstrations/tutorial_kernel_module.py
josh146, line 242: The `qml.template` decorator is no longer required/used internally, so you can safely remove it.

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# Defining a Quantum Embedding Kernel
+# -----------------------------------
+# PennyLane’s ``kernels`` module allows for a particularly simple
+# implementation of Quantum Embedding Kernels. The first ingredient we
+# need for this is an *Ansatz* that represents the unitary
+# :math:`U(\boldsymbol{x})` we use for embedding the data into a quantum
+# state as well as its adjoint :math:`U(\boldsymbol{x})^\dagger`.
+# We will use a structure where a single layer is repeated multiple times
+# and with variational parameters that determine :math:`U` in addition to
+# the datapoint :math:`\boldsymbol{x}`.
```
josh146, line 227: I would place this directly before the `ansatz` code definition, for two reasons:

- breaks up the large code block
- the description of the code is closer to the code :)

demonstrations/tutorial_kernel_module.py
josh146, line 277:
```suggestion
    adjoint_ansatz(x2, params, wires=wires)
    return qml.probs(wires=wires)
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+    ansatz(x1, params, wires=wires)
+    adjoint_ansatz(x2, params, wires=wires)
+
+    return qml.probs(wires=wires)
+
+
+##############################################################################
+# The kernel function itself is now obtained by looking at the probability
+# of observing the all-zero state at the end of the kernel circuit -- because
+# of the ordering in `qml.probs`, this is the first entry:
```
josh146, line 283:
```suggestion
# of the ordering in ``qml.probs``, this is the first entry:
```

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+# kernel. Thus, let’s build a second support vector classifier with the
+# trained kernel:
+
+svm_trained = SVC(
+    kernel=lambda X1, X2: qml.kernels.kernel_matrix(
+        X1,
+        X2,
+        lambda x1, x2: kernel(x1, x2, params),
+    )
+).fit(dataset.X, dataset.Y)
```
mariaschuld, line 557: I wonder if it's worth breaking this up. I know lambda functions are pretty for those who know them, but beginners usually find them confusing. And even I had to state hard at the nested structure...

(Optional)

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+
+kernel_value = kernel(dataset.X[0], dataset.X[1], init_params)
+print(f"The kernel value between the first and second datapoint is {kernel_value:.3f}")
+
+##############################################################################
+# The mutual kernel values between all elements of the dataset form the
+# *kernel matrix*. We can inspect it via the ``qml.kernels.square_kernel_matrix``
+# method. The option ``assume_normalized_kernel=True`` ensures that we do not
+# calculate the entries between the same datapoints, as we know them to be 1
+# for our noiseless simulation. To include the variational parameters, we
```
josh146, line 309: I assume it then only computes the k-1 upper-triangular portion of the matrix?

demonstrations/tutorial_kernel_module.py
josh146, line 339:
```suggestion
#     This step does *not* modify the variational parameters in our circuit
```

mariaschuld (changes requested):
Very cool, thanks so much for this great contribution. 

The demo is of a top quality already, and you'll find that all my suggestions are minor. The only ones that are important are to divide the first part into intro sentence and a new section, if possible to avoid using the template decorator, and to explain or reference an explanation for the kernel circuit.

Great work!

demonstrations/tutorial_kernel_module.py
josh146, line 487: Ah, I suppose we should open a github issue for this :)

josh146: Or is this covered by making `frobenius_inner_product` differentiable?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+
+    # Report the alignment on the full dataset every 50 steps.
+    if (i + 1) % 50 == 0:
+        current_alignment = target_alignment(
+            dataset.X,
+            dataset.Y,
+            lambda x1, x2: kernel(x1, x2, params),
+            assume_normalized_kernel=True,
+        )
+        print(f"Step {i+1} - Alignment = {current_alignment:.3f}")
```
josh146, line 544: I admit I am surprised - I expected an alignment close to 1 for an accuracy of 1. Or am I thinking about it incorrectly?

demonstrations/tutorial_kernel_module.py
```diff
@@ -0,0 +1,596 @@
...
+
+##############################################################################
+# We expect to see an accuracy improvement vs. the SVM with random
+# parameters:
+
+accuracy_trained = accuracy(svm_trained, dataset.X, dataset.Y)
+print(f"The accuracy of a kernel with trained parameters is {accuracy_trained:.3f}")
+
+##############################################################################
+# Very well! We now achieved perfect classification!
```
josh146, line 567:
```suggestion
# We have now achieved perfect classification! 🎆 
```

(Very well sounds a bit off to me. I tend to hear it more 'negatively'. E.g., 'Very well, I _won't_ do that 😠 ')

demos_research.rst
```diff
@@ -124,6 +124,12 @@ IBM, Google, Rigetti, and many more—or on real hardware devices.
     :description: :doc:`demos/learning2learn`
     :tags: tensorflow
 
+.. customgalleryitem::
+    :tooltip: Kernels and alignment training with PennyLane.
+    :figure: demonstrations/kernel_module/QEK_thumbnail.png
+    :description: :doc:`demos/tutorial_kernel_module`
+    :tags: kernels alignment classification
```
josh146, line 131: Thumbnail looks great!

josh146 (changes requested):
Agree with @mariaschuld, the content of the demo is very solid. My only main recommendation for improvement is structural; motivating what the demo will show/achieve early on, and breaking it into sections to help tell the story.

Other than that, my other suggestions are minor!

demonstrations/tutorial_kernel_module.py
josh146: Alternatively, it might make it easier to follow by splitting the class into three functions? I admit, I cannot see too much motivation for it being a class?

demonstrations/tutorial_kernel_module.py
josh146: 😆 I think @mariaschuld and I reviewed at the same time

demonstrations/tutorial_kernel_module.py
josh146: I agree with placing this in the title!

demonstrations/tutorial_kernel_module.py
josh146: yep! That is currently occurring in https://github.com/PennyLaneAI/pennylane/pull/1388

demonstrations/tutorial_kernel_module.py
mariaschuld: I would vote for "Training and evaluating quantum kernels" (since this will wake a lot of interest).

Also, consider dropping "Embedding" from the title. I am in full favour for the specification during the demo, but since it is not known in the literature, it adds no information to 99% of the readers, while potentially confusing them.

demonstrations/tutorial_kernel_module.py
dwierichs: Good catch :D

demonstrations/tutorial_kernel_module.py
dwierichs: I'd suggest not to change it in order to have variety in the demos, but the link works super nicely here I think :)
Or do we prefer to have the demos use the same circuit structure?

demonstrations/tutorial_kernel_module.py
dwierichs: I think there should :) Good catch!

demonstrations/tutorial_kernel_module.py
mariaschuld: No, I totally agree about the variety.

demonstrations/tutorial_kernel_module.py
dwierichs: Conceptually I find "Evaluating and training .. " better, but I "Training and evaluating.." sounds better to me. Any thought on that? :)

demonstrations/tutorial_kernel_module.py
dwierichs: Me too! @johannesjmeyer :)

demonstrations/tutorial_kernel_module.py
dwierichs: Added some details on this.

demonstrations/tutorial_kernel_module.py
dwierichs: Yes, it is :) This is where the fact that `frobenius_inner_product` isn't diffable became relevant in the first place :)

demonstrations/tutorial_kernel_module.py
dwierichs: It often does not optimize that far. IIRC we never saw alignments `>0.5` when working on qhack/the paper

dwierichs:
All proposed changes should be implemented now. I also moved the names to kernel**s**_module to match the module namespace.

demonstrations/tutorial_kernel_module.py
johannesjmeyer: An alignment of 1 essentially means that all training datapoints of the same class get mapped to the same state, so a value much lower than 1 is expected :)

johannesjmeyer:
Thanks again @dwierichs for finishing this off so well, I'm sorry that I didn't manage to pull through with this...

dwierichs:
All good @johannesjmeyer ! Given the diffability PR and the previous state of the demo, this was a :cake:

demonstrations/tutorial_kernel_module.py
josh146: Got it :)

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+
+Kernel methods are one of the cornerstones of classical machine learning.
+Here we are concerned with kernels that can be evaluated on quantum computers,
+*quantum kernels* for short.
+In this tutorial you will learn how to evaluate kernels, use them for classification
+and train them with gradient-based optimization, and all that using the
+functionality of PennyLane's
+`kernels module <https://pennylane.readthedocs.io/en/latest/code/qml_kernels.html>`__.
+The demo is based on Ref. [#Training_QEKs]_, a project from Xanadu's own
+`qhack <https://qhack.ai/>`__ hackathon.
```
mariaschuld, line 24: Nice!

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+This means by just replacing the regular scalar product in our linear
+classification with the map :math:`k`, we can actually express much more
+intricate decision boundaries!
+
+This is very important, because in many interesting cases the embedding :math:`\phi`
+will be much costlier to compute than the kernel :math:`k`.
+
+In this demo, we will explore one particular kind of kernel
+that can be realized on near-term quantum computers, namely *Quantum
+Embedding Kernels (QEKs)*. These are kernels that arise from embedding
```
mariaschuld, line 108: Nice introduction of the term!

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+##############################################################################
+# Let's now have a look at our dataset. In our example, we will work with
+# 3 sectors:
+
+import matplotlib.pyplot as plt
+
+num_sectors = 3
+X, Y = make_double_cake_data(num_sectors)
+
+plot_double_cake_data(X, Y, plt.gca(), num_sectors=num_sectors);
```
mariaschuld, line 227: Not blocking, but I wonder if `%inlineplot` would avoid the dummy output here:

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+
+##############################################################################
+#
+# .. note::
+#     An alternative way to set up the kernel circuit in PennyLane would be
+#     to use the observable type
+#     `Projector <https://pennylane.readthedocs.io/en/latest/code/api/pennylane.Projector.html>`__.
+#     This is shown in the
+#     `demo on kernel-based training of quantum models <https://pennylane.ai/qml/demos/tutorial_kernel_based_training.html>`__, where you will also find more
+#     background information on the kernel circuit structure itself.
```
mariaschuld, line 317: Very nice!

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+# Following on the results that SVM's have proven good generalisation
+# behavior, it will be interesting to inspect the decision boundaries of
+# our classifier:
+
+trained_plot_data = plot_decision_boundaries(svm_trained, plt.gca())
+
+##############################################################################
+# Indeed, we see that now not only every data instance falls within the
+# correct class, but also that there are no strong artifacts that would make us
+# distrust the model. In this sense, our approach benefits from both: On
```
mariaschuld, line 605:
```suggestion
# distrust the model. In this sense, our approach benefits from both: on
```
Took me years of code review to internalise that opposed to German, English spelling lower cases here... :)

mariaschuld (approved):
Very cool, the last changes make it even better! 

I only have non-blocking comments.

demonstrations/tutorial_kernels_module.py
josh146:
```suggestion
`QHack <https://qhack.ai/>`__ hackathon.
```

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+   :width: 30%
+
+We can mathematically formalize this by assigning the label :math:`y`
+via
+
+.. math::
+   y(\boldsymbol{x}) = \operatorname{sgn}(\langle \boldsymbol{w}, \boldsymbol{x}\rangle + b).
+
+The vector :math:`\boldsymbol{w}` points perpendicular to the line and
+thus determine its slope. The independent term :math:`b` specificies the
```
josh146, line 49:
```suggestion
thus determine its slope. The independent term :math:`b` specifies the
```

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+classification with the map :math:`k`, we can actually express much more
+intricate decision boundaries!
+
+This is very important, because in many interesting cases the embedding :math:`\phi`
+will be much costlier to compute than the kernel :math:`k`.
+
+In this demo, we will explore one particular kind of kernel
+that can be realized on near-term quantum computers, namely *Quantum
+Embedding Kernels (QEKs)*. These are kernels that arise from embedding
+data into the space of quantum states. We formalize this by considering
```
josh146, line 109: Optional: I tend to prefer bolding new terms, since it stands out more than italics.

```suggestion
that can be realized on near-term quantum computers, namely **Quantum
Embedding Kernels (QEKs)**. These are kernels that arise from embedding
data into the space of quantum states. We formalize this by considering
```

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+from pennylane import numpy as np
+import matplotlib as mpl
+
+np.random.seed(1359)
+
+##############################################################################
+# And we proceed right away to create a dataset to work with, the
+# ``DoubleCake`` dataset. We define a generating and a plotting function.
+# The details of these functions are not essential for understanding the demo,
+# so don't mind them if they are confusing.
```
josh146, line 141:
```suggestion
# ``DoubleCake`` dataset. Firstly, we define two functions to enable us to
# generate the data.
# The details of these functions are not essential for understanding the demo,
# so don't mind them if they are confusing.
```

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+
+    labels = np.hstack([labels1, -1 * labels2])
+
+    # Canonical form of labels
+    Y = labels.astype(int)
+
+    return X, Y
+
+
+def plot_double_cake_data(X, Y, ax, num_sectors=None):
```
josh146, line 175:
```suggestion

##############################################################################
# Next, we define a function to help plot the ``DoubleCake`` data
# and classification boundaries:

def plot_double_cake_data(X, Y, ax, num_sectors=None):
```

josh146: Optional, but breaks up the big code block :)

demonstrations/tutorial_kernels_module.py
josh146: I'm not sure if `%inlineplot` works in Sphinx gallery 🤔

demonstrations/tutorial_kernels_module.py
josh146: We should update the kernel based training demo to use the new `qml.Projector` - it would probably make it a bit faster as well!

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+# for our noiseless simulation. Overall this means that we compute
+# :math:`\frac{1}{2}(N^2-N)` kernel values for :math:`N` datapoints.
+# To include the variational parameters, we construct a ``lambda`` function that
+# fixes them to the values we sampled above.
+
+init_kernel = lambda x1, x2: kernel(x1, x2, init_params)
+K_init = qml.kernels.square_kernel_matrix(X, init_kernel, assume_normalized_kernel=True)
+
+with np.printoptions(precision=3, suppress=True):
+    print(K_init)
```
josh146, line 348: Didn't know this could be used as a context 🤯

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+print(f"The kernel-target alignment for our dataset and random parameters is {kta_init:.3f}")
+
+##############################################################################
+# Now let's code up an optimization loop and improve the kernel-target alignment!
+#
+# We will make use of regular gradient descent optimization. To speed up
+# the optimization we will not use the entire training set to compute
+# :math:`\operatorname{KTA}` but rather
+# sample smaller subsets of the data at each step, we choose :math:`4`
+# datapoints at random. Remember that PennyLane's inbuilt optimizer works
```
josh146, line 505:
```suggestion
# datapoints at random. Remember that PennyLane's built-in optimizer works
```

josh146: Weirdly, I think 'inbuilt' is a common Australia/British English term, but _not_ a common Canadian English term 😆 I myself have been guilty of using it in other demos and confusing our resident Canadians.

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+# is not expected to suffer from bad generalisation.
+#
+# References
+# ----------
+#
+# .. [#Training_QEKs]
+#
+#     Thomas Hubregtsen, David Wierichs, Elies Gil-Fuster, Peter-Jan H. S. Derks,
+#     Paul K. Faehrmann, and Johannes Jakob Meyer.
+#     “Training Quantum Embedding Kernels on Near-Term Quantum Computers.”
```
josh146, line 616:
```suggestion
#     "Training Quantum Embedding Kernels on Near-Term Quantum Computers."
```

demonstrations/tutorial_kernels_module.py
```diff
@@ -0,0 +1,623 @@
...
+#     `arXiv:2105.02276 <https://arxiv.org/abs/2105.02276>`__, 2021.
+#
+# .. [#Alignment]
+#
+#     Wang, Tinghua, Dongyan Zhao, and Shengfeng Tian.
+#     “An overview of kernel alignment and its applications.”
```
josh146, line 622:
```suggestion
#     "An overview of kernel alignment and its applications."
```

josh146 (approved):
Great work, I love the new title and slight structure modifications!

demonstrations/tutorial_kernels_module.py
dwierichs: Ah, does it not do so yet? :) I assumed this, so that there would be a demo using `qml.Projector` and one that doesn't.

demonstrations/tutorial_kernels_module.py
dwierichs: I'm definitely in the early stages of that process: still not liking it :D

demonstrations/tutorial_kernels_module.py
dwierichs: Very nice! :)

demonstrations/tutorial_kernels_module.py
josh146: It does it conceptually, but not in practice; it uses `qml.Hermitian` and a projector matrix 😆 Should be a relatively small update, we could even make a GitHub issue and mark it as 'good first issue'

demonstrations/tutorial_kernels_module.py
dwierichs: No, unfortunately the line magic functionality does not seem to be supported. [This](https://matplotlib.org/sampledoc/extensions.html) extensions summary might contain something for this, but I think that's not worth it at this point.

demonstrations/tutorial_kernels_module.py
dwierichs: Ah, I see :) Perfect! I could take care of this, if you want.

dwierichs:
That should be all, @mariaschuld @josh146  :-)

co9olguy:
Going live in 3...2...1 :rocket:

PR #271, opened 2021-05-21
Quantum Analytic Descent Demo

dwierichs:


dwierichs:
@mariaschuld and @hapax, we would be very happy if you could review this demo :)

mariaschuld (changes requested):
Challenging topic choice for the demo, and the parts are all there, but I think it needs another iteration simplifying the presentation:

* Make use of sections with headers, and in each section tell the reader what is done in the section (i.e., we first look at the equations, and want to convince you that we can build a model of the cost that requires fewer evaluations for optimisation than the original cost function; now we implement the steps one by one and compare model and true cost;...).
* Simplify the presentation of the equations. Readers do not mind long equations if they are warned of them, it is clear what is important about them, and they are well-motivated (and all symbols explained). But limit difficult content to small, contained spots.
* Simplify the plots - I suggest one subplot per surface. Explain what the points and the black lines from the points are.

Also, would you be open to renaming the "model" (since we use this a lot to describe a ML model)? For example, you could speak of the "approximate cost", which says it all (and explains immediately why you plot a "deviation", which you could even call "approximation error")?

Lastly, is this tutorial based on a paper or your own idea? I think this should be stated clearly in the introduction.

Good luck!

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,565 @@
...
+----------------------------------------------
+
+In a few words, what we have in mind when we talk about VQEs is a quantum circuit.
+Typically, this :math:`n`-qubit quantum circuit is initialized in the :math:`|0\rangle^{\otimes n}` state (in an abuse of notation, we name it simply :math:`|0\rangle` in the rest of the demo).
+The body of the circuit is populated with a *variational form* :math:`V(\boldsymbol{\theta})` -- a fixed architecture of quantum gates parametrized by an array of real-numbers :math:`\boldsymbol{\theta}\in\mathbb{R}^m`.
+After the variational form, the circuit ends with a measurement of an observable :math:`\mathcal{M}`, which is also fixed at the start.
+
+With these, the common choice of cost function :math:`E(\boldsymbol{\theta})` is
+
+.. math:: E(\boldsymbol{\theta}) = |\langle0|V^\dagger(\boldsymbol{\theta})\mathcal{M}V(\boldsymbol{\theta})|0\rangle|^2,
```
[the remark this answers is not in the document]
dwierichs: Good catch! :)

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,565 @@
...
+
+# Show the energy landscape on the grid.
+fig, ax = plt.subplots(1, 1, subplot_kw={"projection": "3d"}, figsize=(4, 4));
+surf = ax.plot_surface(X, Y, Z, label="$E(\\theta_1, \\theta_2)$", alpha=0.7, color='#209494');
+line1 = ax.plot([parameters[1]]*num_samples, theta_func, C1,
+        label="$E(\\theta_1, \\theta_2^{(0)})$", color='r', zorder=100);
+line2 = ax.plot(theta_func, [parameters[0]]*num_samples, C2,
+        label="$E(\\theta_1^{(0)}, \\theta_2)$", color='orange', zorder=100);
+
+###############################################################################
```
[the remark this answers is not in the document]
dwierichs: Whoops, very good catch!

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,565 @@
...
+# And hereby, the complete strategy:
+#
+# #. Set an initial reference point :math:`\boldsymbol{\theta}_0`.
+# #. Build the model :math:`\hat{E}(\boldsymbol{\theta})\approx E(\boldsymbol{\theta}_0+\boldsymbol{\theta})` at this point.
+# #. Find the global minimum :math:`\boldsymbol{\theta}_\text{min}` of the model.
+# #. Set :math:`\boldsymbol{\theta}_\text{min}` as the new reference point :math:`\boldsymbol{\theta}_0`, go back to step 2.
+# #. After convergence or a fixed number of iterations, output the last global minimum :math:`\boldsymbol{\theta}_\text{opt}=\boldsymbol{\theta}_\text{min}`.
+#
+# This provides an iterative strategy which will take us to a good enough solution much faster (in number of iterations) than for example regular stochastic gradient descent (SGD).
+# The procedure of Quantum Analytic Descent is shown in this flowchart:
```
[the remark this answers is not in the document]
dwierichs: Unfortunately, this is the title of the paper and I think thus not up for negotiations :D

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,565 @@
...
+    surf = ax[1].plot_surface(X, Y, Z_original-Z_model, label="Deviation",
+            color=green, alpha=alpha);
+    surf._facecolors2d = surf._facecolor3d
+    surf._edgecolors2d = surf._edgecolor3d
+    ax[0].legend();
+    ax[1].legend();
+
+# Define a mapped model that has the model coefficients fixed.
+mapped_model = lambda parameters: model_cost(parameters, *coeffs)
+plot_cost_and_model(circuit, mapped_model, parameters)
```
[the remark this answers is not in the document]
dwierichs: Good call! I will try plotting them separately but I am not sure how well one can resolve their similarity/differences then. I'll try :)

demonstrations/tutorial_quantum_analytic_descent.py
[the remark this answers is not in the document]
emgilfuster: This is a reply to an earlier message, but which doesn't have a reply box:

> Also, would you be open to renaming the "model" (since we use this a lot to describe a ML model)? For example, you could speak of the "approximate cost", which says it all (and explains immediately why you plot a "deviation", which you could even call "approximation error")?

In the paper they also refer to it as the model everywhere and in there it feels right reading it.
I think we should maybe spend a bit more time motivating the name "model", and maybe change it to "classical model"?

demonstrations/tutorial_quantum_analytic_descent.py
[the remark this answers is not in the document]
emgilfuster: As much as I like the double surface plot, it may be only because of how invested I am into it that I can read it.
Would it make sense to plot only one surface (either model or true), and the deviation next to it?

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,565 @@
...
+Next to the code snippets we provide with the theory required to understand
+what is going on at the different steps, too.
+So: sit down, relax, and enjoy your optimization!
+
+|
+
+.. figure:: ../demonstrations/quantum_analytic_descent/xkcd.png
+    :align: center
+    :width: 50%
+    :target: javascript:void(0)
```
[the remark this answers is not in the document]
emgilfuster: It shows it for me :O
Did David already solve this?

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+One of the main problems of many-body physics is that of finding the ground
+state and ground state energy of a given Hamiltonian.
+The Variational Quantum Eigensolver (VQE) combines a smart circuit
+design with a gradient-based optimization to solve this task
+(take a look at the `overview demo <https://pennylane.ai/qml/demos/tutorial_vqe.html>`_ for more details).
+Several practical demonstrations have pointed out how near-term quantum
+devices may be well-suited platforms for VQE and other variational quantum algorithms.
+One issue for such an approach is, though, that the optimization landscape is
+non-convex, so reaching a good enough local minimum quickly requires hundreds or
+thousands of update steps.
```
emgilfuster, line 31:
```suggestion
Using near-term devices for VQE becomes a non-convex optimization problem, which requires computing gradients at hundreds or thousands of update steps.
Computing gradients on a quantum computer carries the caveat of inefficiency when it comes to circuits with many parameters.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+
+At the same time, though, we do have a good understanding of the local shape
+of the cost landscape around any reference point.
+Cashing in on this, the Quantum Analytic Descent algorithm constructs a
```
emgilfuster, line 35:
```suggestion
Cashing in on this, the authors of the `Quantum Analytic Descent paper <https://arxiv.org/pdf/2008.13774.pdf>`_ propose and algorithm to construct a
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+local approximation to the landscape where gradient-based optimisation is much
```
emgilfuster, line 36:
```suggestion
classical model to approximate the landscape, so that the gradients can be calculated on a classical computer, which is much
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+cheaper.
+This approach avoids using the quantum computer to estimate the gradient
```
emgilfuster, line 38:
```suggestion
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+at every single optimization step.
```
emgilfuster, line 39:
```suggestion
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+Instead, we can use the quantum device to estimate the cost values at a few
```
emgilfuster, line 40:
```suggestion
In order to build the classical model, we need to use the quantum device to estimate the cost values on (a) a reference point :math:`\theta_0` and (b) a number of points in the vicinity of the reference one (more on that later).
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+points close to the reference one.
```
emgilfuster, line 41:
```suggestion
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+With the cost values at these points, we can build a classical model which
```
emgilfuster, line 42:
```suggestion
With the cost values at these points, we can build the classical model which
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+approximates the landscape locally, and then perform the optimization routine
```
emgilfuster, line 43:
```suggestion
approximates the landscape.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+
+VQEs give rise to trigonometric cost functions
+----------------------------------------------
+
+In a few words, what we have in mind when we talk about VQEs is a quantum circuit.
+Typically, this :math:`n`-qubit quantum circuit is initialized in the :math:`|0\rangle^{\otimes n}` state (in an abuse of notation, we name it simply :math:`|0\rangle` in the rest of the demo).
+The body of the circuit is populated with a *variational form* :math:`V(\boldsymbol{\theta})` -- a fixed architecture of quantum gates parametrized by an array of real-numbers :math:`\boldsymbol{\theta}\in\mathbb{R}^m`.
+After the variational form, the circuit ends with a measurement of an observable :math:`\mathcal{M}`, which is also fixed at the start.
+
+With these, the common choice of cost function :math:`E(\boldsymbol{\theta})` is
```
emgilfuster, line 69:
```suggestion
With these, our goal is to minimize the cost function :math:`E(\boldsymbol{\theta})`:
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+
+.. math:: E(\boldsymbol{\theta}) = \langle 0|V^\dagger(\boldsymbol{\theta})\mathcal{M}V(\boldsymbol{\theta})|0\rangle,
+
+or, in short
+
+.. math:: E(\boldsymbol{\theta}) = \operatorname{tr}[\rho(\boldsymbol{\theta})\mathcal{M}],
+
+where :math:`\rho(\boldsymbol{\theta})=V(\boldsymbol{\theta})|0\rangle\!\langle0|V^\dagger(\boldsymbol{\theta})` is the density matrix of the quantum state after applying the variational form on the initial state.
```
emgilfuster, line 77:
```suggestion
where :math:`\rho(\boldsymbol{\theta})=V(\boldsymbol{\theta})|0\rangle\!\langle0|V^\dagger(\boldsymbol{\theta})` is the density matrix of the quantum state after applying the variational form on :math:`|0\rangle`.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+
+# Create 1D sweeps through parameter space with the other parameter fixed.
+num_samples = 50
+theta_func = np.linspace(0, 2*np.pi, num_samples)
+C1 = [circuit(np.array([theta, .5])) for theta in theta_func]
+C2 = [circuit(np.array([3.3, theta])) for theta in theta_func]
+
+# Show the sweeps.
+fig, ax = plt.subplots(1, 1, figsize=(4, 3));
+ax.plot(theta_func, C1, label="$E(\\vartheta, \\theta_2^{(0)})$", color='r');
```
emgilfuster, line 129:
```suggestion
ax.plot(theta_func, C1, label="$E(\\theta, 0.5)$", color='r');
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+ax.plot(theta_func, C2, label="$E(\\theta_1^{(0)}, \\vartheta)$", color='orange');
```
emgilfuster, line 130:
```suggestion
ax.plot(theta_func, C2, label="$E(3.3
, \\vartheta)$", color='orange');
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+        label="$E(\\theta_1^{(0)}, \\theta_2)$", color='orange', zorder=100);
+plt.tight_layout();
+
+###############################################################################
+# Of course this is an overly simplified example, but the key take-home message so far is:
+# *if the variational parameters feed into Pauli rotations, the energy landscape is a 
+# multilinear combination of trigonometric functions*.
+# What is a good thing about trigonometric functions?
+# That's right!
+# We have studied them since high-school and know how to find their minima.
```
emgilfuster, line 164:
```suggestion
# We have studied them since high-school and know how their graphs look!
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+#
+# In our overly simplified example we had to query a quantum computer for every point on the surface.
```
emgilfuster, line 166:
```suggestion
# In our overly simplified example we had to query a quantum computer for every point on the surface.
# Querying the quantum computer in this example means calling the ``circuit`` function.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+# Could we have spared some computational resources?
+# Well, since we know the ultimate shape the landscape is supposed to have, we should in principle be able to construct it only from a few points (much the same way two points already uniquely specify a line, or three non-aligned points specify a circle, there should be a certain fixed number of points that completely specify the loss landscape).
+#
+# Quantum Analytic Descent
+# ------------------------
+#
+# Although in principle we should be able to reconstruct the cost function over the entire parameter space, in practice we are mostly interested in what is happening in the vicinity of a given reference point.
```
emgilfuster, line 173:
```suggestion
# One could aim at reconstructing the cost function over the entire parameter space.
# In practice, that would be *very* expensive in number of required quantum circuit evaluations, so the cure would be roughly as bad as the sickness.
# What can we do, then?
# Well, the authors of QAD claim that building a model that is only accurate in the region close to a given reference point is already good enough!
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+# This makes *all* the difference.
+# If we wanted to reconstruct the entire landscape, we would need to estimate around :math:`3^m` independent parameters, which would require the same number of function evaluations.
```
emgilfuster, line 175:
```suggestion
# If we wanted to reconstruct the entire landscape, we would need to estimate around :math:`3^m` independent parameters, which would require a comparable amount of circuit evaluations.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+# If, on the contrary, we are satisfied with an approximation that is cheaper to construct (polynomial in :math:`m` instead of exponential), we can borrow a page from Taylor's book.
+#
+# As explained in the paper, an approximation via trigonometric series up to second order is already a sound candidate.
```
emgilfuster, line 178:
```suggestion
# In the paper, the authors state that a trigonometric series up to second order is already a sound model candidate.
# Similar to Taylor's expansion, this requires we find the cost value at the reference point and compute the first and second order derivatives.
# Here, we will use parameter-shift rules (SHOULD WE CITE ANOTHER DEMO/PAPER?) for the derivatives.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+# In particular, we want to approximate
+#
+# .. math:: E(\boldsymbol{\theta}) := \operatorname{tr}\left[\mathcal{M}V(\boldsymbol{\theta})|0\rangle\!\langle 0| V(\boldsymbol{\theta})^\dagger\right]
+#
+# in the vicinity of a reference point :math:`\boldsymbol{\theta}_0`.
```
emgilfuster, line 183:
```suggestion
# START OF MAJOR RECONSTRUCTION, PLEASE KEEP OFF
# in the vicinity of a reference point :math:`\boldsymbol{\theta}_0`.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,586 @@
...
+#   B_k(\boldsymbol{\theta}) &= \tan\left(\frac{\theta_k}{2}\right)A(\boldsymbol{\theta})\\
+#   C_k(\boldsymbol{\theta}) &= \tan\left(\frac{\theta_k}{2}\right)^2 A(\boldsymbol{\theta})\\
+#   D_{kl}(\boldsymbol{\theta}) &= \tan\left(\frac{\theta_k}{2}\right)\tan\left(\frac{\theta_l}{2}\right)A(\boldsymbol{\theta})
+#
+# Using these, we can compute the classical surrogate model :math:`\tilde{E}(\boldsymbol{\theta})`:
+#
+# .. math::
+#
+#   \tilde{E}(\boldsymbol{\theta}) &= A(\boldsymbol{\theta}) E^{(A)} + \sum_k \left[B_k(\boldsymbol{\theta}) E^{(B)}_k + C_k(\boldsymbol{\theta}) E^{(C)}_k\right] + \sum_{l>k} D_{kl}(\boldsymbol{\theta}) E^{(D)}_{kl}\\
+#   \phantom{\tilde{E}(\boldsymbol{\theta})}&=A(\boldsymbol{\theta})\left[E^{(A)}+\sum_k \tan\left(\frac{\theta_k}{2}\right)E^{(B)}_k + \tan\left(\frac{\theta_k}{2}\right)^2 E^{(C)}_k + \sum_{l>k} \tan\left(\frac{\theta_k}{2}\right)\tan\left(\frac{\theta_l}{2}\right)E^{(D)}_{kl}\right]
```
emgilfuster, line 301:
```suggestion
#   \phantom{\tilde{E}(\boldsymbol{\theta})}&=A(\boldsymbol{\theta})\left[E^{(A)}+\sum_k \tan\left(\frac{\theta_k}{2}\right)E^{(B)}_k + \tan\left(\frac{\theta_k}{2}\right)^2 E^{(C)}_k + \sum_{l>k} \tan\left(\frac{\theta_k}{2}\right)\tan\left(\frac{\theta_l}{2}\right)E^{(D)}_{kl}\right]
# END OF THE MAJOR RECONSTRUCTION, YOU MAY PROCEED
```

mariaschuld:
Hey @dwierichs and @emgilfuster. Just request a re-review when you are ready, so I can have a second look :)

dwierichs:
We did some major restructuring and hope that the demo is more appealing to read now.
Sorry for taking so long. @mariaschuld and @glassnotes we would be happy about another review! Thank you!

demonstrations/tutorial_quantum_analytic_descent.py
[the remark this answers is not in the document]
dwierichs: Yes, this was resolved, sorry!

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#     In order to plot the surface for this example we had to query a quantum computer (call the ``circuit`` function) for every point inside the range.
+#     In real life, querying a quantum computer is very expensive, so could we have spared some computational resources?
+#     Well, since we know the ultimate shape the landscape is supposed to have, we should in principle be able to construct it from a fixed number of points:
+#     the same way two points specify a straight line, or three non-aligned points specify a circle.
+#     In particular, the number of points that completely determine the cost landscape should in depend mainly on the total number of parameters.
+#
+# The QAD strategy
+# ----------------
+#
+# Before jumping in on the specifics of the classical model they use, we are already in a good position to review the general strategy.
```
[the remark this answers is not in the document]
glassnotes: This paragraph would benefit from an additional sentence connecting the ideas here to the previous section. Relate things to the previous plot, or the idea of computational challenge discussed in the note, and explain how instead of computing this surface with a quantum computer, we are going to approximate it classically.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# However, you may have noticed all of those terms have large parts in common.
+# Indeed, we can rewrite the longer ones in a shorter way which is more decent to look at:
+#
+# .. math::
+#
+#   B_k(\boldsymbol{\theta}) &= 2\tan\left(\frac{\theta_k}{2}\right)A(\boldsymbol{\theta})\\
+#   C_k(\boldsymbol{\theta}) &= 2\tan\left(\frac{\theta_k}{2}\right)^2 A(\boldsymbol{\theta})\\
+#   D_{kl}(\boldsymbol{\theta}) &= 4\tan\left(\frac{\theta_k}{2}\right)\tan\left(\frac{\theta_l}{2}\right)A(\boldsymbol{\theta})
+#
+# And with that we know what type of terms we should expect to encounter in our local classical model:
```
glassnotes, line 306:
```suggestion
# With that. we know what type of terms we should expect to encounter in our local classical model:
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# The classical model we want to construct is a linear combination of the functions
+# :math:`\{A(\boldsymbol{\theta})`, :math:`B_k(\boldsymbol{\theta})` and :math:`C_k(\boldsymbol{\theta})`
```
glassnotes, line 308:
```suggestion
# :math:`A(\boldsymbol{\theta})`, :math:`B_k(\boldsymbol{\theta})` and :math:`C_k(\boldsymbol{\theta})`
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# Computing the expansion coefficients
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# As each of the terms in the linear combination has its own unique leading order it contributes to, we need to compute
+# the derivatives of the function we are approximating to obtain the coefficients of the linear combination -- just like
+# we would for a Taylor expansion.
+# As the terms we include in the expansion have leading orders :math:`0`, :math:`1` and :math:`2`, these derivatives are
+# :math:`E(\boldsymbol{\theta})`, :math:`\partial E(\boldsymbol{\theta})/\partial \theta_k`,
+# :math:`\partial^2 E(\boldsymbol{\theta})/\partial\theta_k^2`, and :math:`\partial^2 E(\boldsymbol{\theta})/\partial \theta_k\partial\theta_l`.
+# However, as the trigonometric polynomials contain multiple orders in :math:`\boldsymbol{\theta}`, and both
```
glassnotes, line 320:
```suggestion
# However, the trigonometric polynomials may contain multiple orders in :math:`\boldsymbol{\theta}`. For example, both
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# :math:`A(\boldsymbol{\theta})` and :math:`C_k(\boldsymbol{\theta})` contribute to the second order, we have to account
```
glassnotes, line 321:
```suggestion
# :math:`A(\boldsymbol{\theta})` and :math:`C_k(\boldsymbol{\theta})` contribute to the second order, so we have to account
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#
+#   E^{(A)} &= E(\boldsymbol{\theta})\Bigg|_{\boldsymbol{\theta}=0} \\
+#   E^{(B)}_k &= \frac{\partial E(\boldsymbol{\theta})}{\partial\theta_k}\Bigg|_{\boldsymbol{\theta}=0} \\
+#   E^{(C)}_k &= \frac{\partial^2 E(\boldsymbol{\theta})}{\partial\theta_k^2}\Bigg|_{\boldsymbol{\theta}=0} + \frac{1}{2}E(\boldsymbol{\theta})\Bigg|_{\boldsymbol{\theta}=0}\\
+#   E^{(D)}_{kl} &= \frac{\partial^2 E(\boldsymbol{\theta})}{\partial\theta_k\partial\theta_l}\Bigg|_{\boldsymbol{\theta}=0}
+#
+# .. note::
+#     In PennyLane, computing the gradient of a cost function with respect to an array of parameters can be easily done
+#     with the `parameter-shift rule <https://pennylane.ai/qml/glossary/parameter_shift.html>`_
+#     and by iterating the rule, we also obtain the second derivatives -- the Hessian (see for example [#higher_order_diff]_).
```
glassnotes, line 335:
```suggestion
#     with the `parameter-shift rule <https://pennylane.ai/qml/glossary/parameter_shift.html>`_.
#     By iterating the rule, we can obtain the second derivatives -- the Hessian (see for example [#higher_order_diff]_).
```

glassnotes: Have note as part of main text, because it leads into the code section.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#     Let us implement a function that does just that and prepares the coefficients :math:`E^{(A/B/C/D)}`:
+
+from pennylane import numpy as np
+
+
```
glassnotes, line 340:
```suggestion
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+    num_params = len(params)
+    E_A = fun(params)
+    # E_B contains the gradient.
+    E_B = qml.grad(fun)(params)
+    hessian = qml.jacobian(qml.grad(fun))(params)
+    # E_C contains the slightly adapted diagonal of the Hessian.
+    E_C = np.diag(hessian) + E_A / 2
+    # E_D contains the off-diagonal parts of the Hessian.
+    # We store each pair (k, l) only once, namely the upper triangle.
+    E_D = np.triu(hessian, 1)
```
glassnotes, line 352:
```suggestion
    E_A = fun(params)
    
    # E_B contains the gradient.
    E_B = qml.grad(fun)(params)
    
    hessian = qml.jacobian(qml.grad(fun))(params)
    
    # E_C contains the slightly adapted diagonal of the Hessian.
    E_C = np.diag(hessian) + E_A / 2
    
    # E_D contains the off-diagonal parts of the Hessian.
    # We store each pair (k, l) only once, namely the upper triangle.
    E_D = np.triu(hessian, 1)
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+print(f"Random parameters (params): {parameters}")
+coeffs = get_model_data(circuit, parameters)
+print(
+    f"Coefficients at params:",
+    f" E_A = {coeffs[0]}",
+    f" E_B = {coeffs[1]}",
+    f" E_C = {coeffs[2]}",
+    f" E_D = {coeffs[3]}",
+    sep="\n",
+)
```
glassnotes, line 370: Nice!

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+
+###############################################################################
+# The classical model (finally!)
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Bringing all of the above ingredients together, we have the following gorgeous trigonometric expansion up to second order:
+#
+# .. math:: \hat{E}(\boldsymbol{\theta}_0+\boldsymbol{\theta}) := A(\boldsymbol{\theta}) E^{(A)} + \sum_{k=1}^m\left[B_k(\boldsymbol{\theta})E_k^{(B)} + C_k(\boldsymbol{\theta}) E_k^{(C)}\right] + \sum_{k<l}^m\left[D_{kl}(\boldsymbol{\theta}) E_{kl}^{(D)}\right].
+#
+# Let us now take a few moments to breath deeply and admire the entirety of it.
```
glassnotes, line 380: Love the enthusiasm :star_struck:

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# On the one hand, we have the :math:`A`, :math:`B_k`, :math:`C_k`, and :math:`D_{kl}` functions, which we said would play the role of the monomials :math:`1`, :math:`x_k`, :math:`x_k^2`, and :math:`x_kx_l` in a regular Taylor expansion.
+# On the other hand we have the real-valued coefficients :math:`E^{(A/B/C/D)}` for the previous functions which are nothing but the derivatives in the corresponding input components.
+# Combining them yields the trigonometric expansion, which we implement with another function:
+
+
+def model_cost(params, E_A, E_B, E_C, E_D):
+    """Compute the model cost for relative parameters and given model data."""
+    A = np.prod(np.cos(0.5 * params) ** 2)
+    # For the other terms we only compute the prefactor relative to A
```
glassnotes, line 389:
```suggestion

    # For the other terms we only compute the prefactor relative to A
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+    B_over_A = 2 * np.tan(0.5 * params)
+    C_over_A = B_over_A ** 2 / 2
+    D_over_A = np.outer(B_over_A, B_over_A)
+    all_terms_over_A = [
+        E_A,
+        np.dot(E_B, B_over_A),
+        np.dot(E_C, C_over_A),
+        np.dot(B_over_A, E_D @ B_over_A),
+    ]
+    cost = A * np.sum(all_terms_over_A)
```
glassnotes, line 399:
```suggestion
    D_over_A = np.outer(B_over_A, B_over_A)
    
    all_terms_over_A = [
        E_A,
        np.dot(E_B, B_over_A),
        np.dot(E_C, C_over_A),
        np.dot(B_over_A, E_D @ B_over_A),
    ]
    
    cost = A * np.sum(all_terms_over_A)
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+
+    return cost
```
glassnotes, line 401: Could also just do `return A * np.sum...` instead of splitting over separate lines.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+    f"  Model:    {E_model}",
+    f"  Original: {E_original}",
+    sep="\n",
+)
+# Check that coeffs[0] indeed is the original energy and that the model is correct at 0.
+print(f"E_A and E_original are the same: {coeffs[0]==E_original}")
+print(f"E_model and E_original are the same: {E_model==E_original}")
+
+###############################################################################
+# As we can see, the model works fine and reproduces the correct energy at the parameter
```
glassnotes, line 419:
```suggestion
# As we can see, the model reproduces the correct energy at the parameter
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+    sep="\n",
+)
+print(f"E_model and E_original are the same: {E_model==E_original}")
+
+###############################################################################
+#
+# .. note::
+#     **Counting parameters and evaluations**
+#
+#     Now we can also summarize how many parameters our model has.
```
glassnotes, line 446:
```suggestion
#     How many parameters does our model have?
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#
+#     #. We have one :math:`E^{(A)}`.
+#     #. We have :math:`m` many :math:`E^{(B)}`'s, one for each component.
+#     #. We have :math:`m` many :math:`E^{(C)}`'s, also one for each component.
+#     #. We have :math:`(m-1)m/2` many :math:`E^{(D)}`'s, because we only need to check every pair once.
+#
+#     So, in total, there are :math:`m^2/2 + 3m/2 + 1` parameters, which is indeed polynomially many.
+#
+#     In practice, not every parameter needs the same amount of circuit evaluations, though!
+#     Without going into detail on the parameter-shift rule for the gradient and Hessian, we remark that we need :math:`1\times 1 + 2\times m + 1\times m + 4\times (m-1)m/2` many circuit evaluations, which amounts to a total of :math:`2m^2+m+1`.
```
glassnotes, line 457:
```suggestion
#     Without going into detail on the parameter-shift rule for the gradient and Hessian, we remark that we need a total of :math:`2m^2+m+1`.
```
Just show the result; maybe link to somewhere that does this computation though?

BTW this content works great as a note :+1:

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#     This is much cheaper than the :math:`3^m` we would need if we naively tried to construct the cost landscape exactly, without chopping after second order.
+#
+# Now this should be enough of theory, let's *take a look* at the model that results from our trigonometric expansion.
```
glassnotes, line 460:
```suggestion
# Now this should be enough theory, so let's visualize the model that results from our trigonometric expansion.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+    coords = np.linspace(-shift_radius, shift_radius, num_points)
+    X, Y = np.meshgrid(coords + params[0], coords + params[1])
+    # Compute the original cost function and the model on the grid.
+    Z_original = np.array([[fun(params + np.array([t1, t2])) for t2 in coords] for t1 in coords])
+    Z_model = np.array([[model(np.array([t1, t2])) for t2 in coords] for t1 in coords])
+    # Prepare sampled points for plotting.
+    shifts = [-np.pi / 2, 0, np.pi / 2]
+    samples = chain.from_iterable(
+        [
+            [[params[0] + s2, params[1] + s1, fun(params + np.array([s1, s2]))] for s2 in shifts]
```
glassnotes, line 479: This line is just a bit too long, and makes a horizontal scroll bar appear in the code. We try to avoid this in demos, so see if there is anywhere you can save some characters (e.g. variable names, `fun` ->`f`, something like this).

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+    ax1.plot_surface(X, Y, Z_model, color=orange, alpha=alpha)
+    ax1.set_title("Model energy")
+    ax2.plot_surface(X, Y, Z_original - Z_model, color=red, alpha=alpha)
+    ax2.set_title("Deviation")
+    plt.tight_layout(pad=2, w_pad=2.5)
+
+
+# Get some fresh random parameters
+parameters = np.random.random(2) * 2 * np.pi
+# Define a mapped model that has the model coefficients fixed.
```
glassnotes, line 503:
```suggestion

# Define a mapped model that has the model coefficients fixed.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+mapped_model = lambda params: model_cost(params, *coeffs)
+plot_cost_and_model(circuit, mapped_model, parameters)
+
+###############################################################################
+# In the first two plots, we see the true landscape in green, and the approximate model in orange.
+# The vertical rods indicate the points at which the cost function
+# was evaluated in order to obtain the model coefficients (we skip the additional
```
glassnotes, line 510: These plots look awesome!

I see now the reason for the earlier note about the number of samples. This might be a better place for it, like "notice how we evaluated the cost function at only a subset of points. this is because..." because it's directly required for the example.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+
+
+# Get some fresh random parameters
+parameters = np.random.random(2) * 2 * np.pi
+# Define a mapped model that has the model coefficients fixed.
+mapped_model = lambda params: model_cost(params, *coeffs)
+plot_cost_and_model(circuit, mapped_model, parameters)
+
+###############################################################################
+# In the first two plots, we see the true landscape in green, and the approximate model in orange.
```
glassnotes, line 508:
```suggestion
# In the first two plots, we see the true landscape, and the approximate model.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# The vertical rods indicate the points at which the cost function
+# was evaluated in order to obtain the model coefficients (we skip the additional
+# evaluations for :math:`E^{(C)}`, though, for clarity of the plot).
+# In the third plot, we see the difference between model and true landscape.
+# Around the reference point this difference is very small and changes very slowly,
```
glassnotes, line 513: It would be helpful to mark on the plots where the reference point is

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# only growing significantly for large simultaneous perturbations in both
+# parameters. This already hints at the value of the model for local optimization.
+#
+# Quantum Analytic Descent
+# ------------------------
+#
+# The underlying idea we are trying to exploit now for the optimization in VQEs is the following:
```
glassnotes, line 520:
```suggestion
# The underlying idea we will now try to exploit for optimization in VQEs is the following:
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# If we can model the cost around the reference point well enough, we will be able to find a rough
```
glassnotes, line 521:
```suggestion
# if we can model the cost around the reference point well enough, we will be able to find a rough
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# estimate of where the minimum of the landscape is.
+# Granted, our model represents the true landscape less accurately the further we go away from the
+# reference point :math:`\boldsymbol{\theta}_0`, but nonetheless the minimum *of the model*
+# will bring us much closer to the minimum *of the true cost* than a random walk.
```
glassnotes, line 525:
```suggestion
# will bring us much closer to the minimum *of the true cost* than a random choice.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# And hereby, we recap the complete strategy from above:
```
glassnotes, line 526:
```suggestion
# Recall the complete strategy from above:
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#
+# #. Set an initial reference point :math:`\boldsymbol{\theta}_0`.
+# #. Build the model :math:`\hat{E}(\boldsymbol{\theta})\approx E(\boldsymbol{\theta}_0+\boldsymbol{\theta})` at this point.
+# #. Find the minimum :math:`\boldsymbol{\theta}_\text{min}` of the model.
+# #. Set :math:`\boldsymbol{\theta}_0+\boldsymbol{\theta}_\text{min}` as the new reference point :math:`\boldsymbol{\theta}_0`, go back to Step 2.
+# #. After convergence or a fixed number of models built, output the last minimum :math:`\boldsymbol{\theta}_\text{opt}=\boldsymbol{\theta}_0+\boldsymbol{\theta}_\text{min}`.
+#
+# This provides an iterative strategy which will take us to a good enough solution much faster
+# (in number of iterations) than for example regular stochastic gradient descent (SGD).
+# The procedure of Quantum Analytic Descent is also shown in the following flowchart. Note that the minimization
```
glassnotes, line 536:
```suggestion
# This provides an iterative strategy which will take us to a good enough solution 
# in fewer iterations than, for example, regular stochastic gradient descent (SGD).
# The procedure of Quantum Analytic Descent is also shown in the following flowchart. Note that the minimization
```
Good place to link to the SGD demo.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# Set the number of iterations of Steps 2 to 4
+N_iter_outer = 3
+# Set the number of iterations for the model optimization in Step 3.
+N_iter_inner = 50
+# Prepare storage of the coefficients for the model and the parameter positions.
+past_coeffs = []
+past_parameters = []
+circuit_log = [circuit(parameters)]
+model_logs = []
+for iter_outer in range(N_iter_outer):
```
glassnotes, line 571: This is a very big code block, try to add a bit of spacing, and some lines could be saved by removing some comments that don't directly relate to the algorithm itself (e.g., `Logging`)

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+plot_cost_and_model(circuit, mapped_model, past_parameters[1])
+
+###############################################################################
+# **Iteration 2:** Now we observe the model to stay closer to the original landscape. In addition, the minimum of the model is within the displayed range -- we're getting closer.
+
+mapped_model = lambda params: model_cost(params, *past_coeffs[2])
+plot_cost_and_model(circuit, mapped_model, past_parameters[2])
+
+###############################################################################
+# **Iteration 3:** Both, the model and the original cost function now show a minimum close to our parameter position, Quantum Analytic Descent converged.
```
glassnotes, line 639: This is a really cool progression to see!

(Only suggestion is to mark somewhere the original / reference parameters so it can be seen how they correspond to the minimum)

glassnotes:
```suggestion
# **Iteration 3:** Both the model and the original cost function now show a minimum close to our parameter position--- Quantum Analytic Descent converged.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# Note how the larger deviations of the model close to the boundaries are not a problem at all because we only use the model in the central area,
+# in which both, the original energy and the model, form a convex bowl and the deviation plateaus at zero.
```
glassnotes, line 641:
```suggestion
# Note how the larger deviations of the model close to the boundaries are not a problem at all because we only use the model in the central area
# in which both the original energy and the model form a convex bowl and the deviation plateaus at zero.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+#
+# If we pay close attention to the values printed during the optimization, we can identify a curious phenomenon.
+# At the last epochs within some iterations, the *model cost* goes beyond :math:`-1`.
+# Could we visualize this behavior more clearly, please?
+
+fig, ax = plt.subplots(1, 1, figsize=(6, 4))
+ax.plot(range(N_iter_outer * N_iter_inner + 1), circuit_log, color="#209494", label="True")
+for i in range(N_iter_outer):
+    (line,) = ax.plot(
+        range(i * N_iter_inner, (i + 1) * N_iter_inner + 1), model_logs[i], ls="--", color="#ED7D31"
```
glassnotes, line 654: This line is also a bit too long, try to shorten it a bit or spread over more lines.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# Each of the orange lines corresponds to minimizing the model constructed at a
+# different reference point.
+# We can now more easily appreciate the phenomenon we just described:
+# towards the end of each "outer" optimization step, the model cost
+# can be seen to be significantly lower than the true cost.
+# Once the true cost itself approaches the absolute minimum, this means the
+# model cost can overstep the allowed range.
+# *Wasn't this forbidden? You guys told us the function could only take values in* :math:`[-1,1]` *(nostalgic* emoticon *time >:@)!*
+# Yes, but careful!
+# We said the *true cost* values are bounded, yet, that does not mean the ones of the *model* are!
```
glassnotes, line 673: Move to the beginning of this paragraph, after the current first sentence? Also could ju
```suggestion
# You might be thinking, *Wasn't this forbidden? The function can only take values in* :math:`[-1,1]`  >:@ *
# While the *true cost* values are bounded, that does not mean the ones of the *model* are!
```
Just a suggestion to rephrase and streamline.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+# We see how values exiting the allowed range of the true cost function does not have an
+# impact on the overall optimization.
+#
+# In this demo we've seen how to implement the Quantum Analytic Descent algorithm
+# using PennyLane for a toy-example of a Variational Quantum Eigensolver.
+# By making extensive use of 3D-plots we have also tried to illustrate exactly
+# what is going on in both the true cost landscape and the trigonometric expansion
+# approximation.
+# Recall we wanted to avoid working on the true landscape itself because we can
+# only access it via very costly quantum computations.
```
glassnotes, line 687: Great recap!

glassnotes (changes requested):
@dwierichs @emgilfuster great work, this is a very clear and detailed demonstration of the work! I'm a huge fan of the graphics, they look really nice and are well-chosen / placed throughout. 

My only major suggestion is to work on streamlining the content, as the demo is quite long. I've noted a few places where this can be done, and made some other wording/phrasing suggestions, but I suggest to go through yourselves as well, and ensure that all the contents that are there are essential to the structure of the explanation, and are as concise as possible (but without sacrificing all the personality!).

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+
+.. related::
+
+   tutorial_vqe A brief overview of VQE
+   tutorial_quantum_natural_gradient Quantum Natural Gradient (QNG)
+   tutorial_rotoselect Leveraging trigonometry with Rotoselect
+   tutorial_stochastic_parameter_shift Obtaining gradients stochastically
+
+
+*Authors: Elies Gil-Fuster, David Wierichs. Posted: ?? May 2021.*
```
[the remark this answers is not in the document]
mariaschuld: Whoops sorry, I think I suggested to take that out previously. My bad!

xanadu_theme/static/xanadu.css_t
```diff
@@ -1185,13 +1185,13 @@ div.topic.contents > ul > li {
 
 .admonition.note,
 .admonition.hint {
-  border-left-color: #31708f;
-  background-color: #dae1f0;
+  border-left-color: #19b37b;
+  background-color: #ebf2ee;
```
josh146, line 1189:
```suggestion
  background-color: #f2f9f5;
```

(Slightly less gray-ish 😆)

mariaschuld (approved):
@dwierichs @emgilfuster the demo is awesome now, well done. It explains a complex idea very clearly! Most of my comments coincide with those of @glassnotes, so if those are addressed it is ready to be merged from my side.

A tiny (optional) idea: do you want to show via vertical lines at epoch 50, 100 in the picture that there are different iterations of the model building algorithm? It is kind of clear from the text, but since the picture features at the start this could explain better what we see, i.e. why there are steps...

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+
+At the same time, we do have a good understanding of the *local* shape
+of the cost landscape around any reference point.
+Cashing in on this, the authors of the
+Quantum Analytic Descent paper [#QAD]_
+propose an algorithm that constructs a classical model which approximates the
+landscape, so that the gradients can be calculated on a classical computer
+for the next optimization steps, which is much cheaper.
+After this purely classical phase, a new model is built at the next reference
+point.
```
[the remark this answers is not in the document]
emgilfuster: Hi! I'm not sure I'm seeing it wrong, but if we also remove the "which" in line 41 then the sentence stops making sense grammatically, right?

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,715 @@
...
+with many parameters.
+
+At the same time, we do have a good understanding of the *local* shape
+of the cost landscape around any reference point.
+Cashing in on this, the authors of the
+Quantum Analytic Descent paper [#QAD]_
+propose an algorithm that constructs a classical model which approximates the
+landscape, so that the gradients can be calculated on a classical computer
+for the next optimization steps, which is much cheaper.
+After this purely classical phase, a new model is built at the next reference
```
emgilfuster, line 42:
```suggestion
After this purely classical phase, a new classical model is built at the next reference
```

demonstrations/tutorial_quantum_analytic_descent.py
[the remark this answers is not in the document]
glassnotes: Oops, that's my bad! Fixed it :)

demonstrations/tutorial_quantum_analytic_descent.py
emgilfuster: we wrote it out explicitly because it's wrong on the paper. Is it fine if it stays? Maybe phrased differently if all those \times are too ugly?

demonstrations/tutorial_quantum_analytic_descent.py
dwierichs: It seems like there is no scrollbar for us in this cell.
As we use black, shouldn't the line length limit of 100 avoid this happening? :-)

glassnotes: Oh weird, what browser are you using? (I've found when testing demos in the past that black with limit between 90-95 is safest)

dwierichs: Yes, sorry, actually there was a scrollbar in the end, I just fixed that.There must have been a change after I checked last...
I believe `black -l 94` should be around the limit :D

emgilfuster:
Hi! We implemented the suggested changes, we're ready for one more review iteration @glassnotes @mariaschuld

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+   tutorial_rotoselect Leveraging trigonometry with Rotoselect
+   tutorial_stochastic_parameter_shift Obtaining gradients stochastically
+
+
+*Authors: Elies Gil-Fuster, David Wierichs (Xanadu residents). Posted: ?? May 2021.*
+
+One of the main problems of many-body physics is that of finding the ground
+state and ground state energy of a given Hamiltonian.
+`The Variational Quantum Eigensolver (VQE) <https://pennylane.ai/qml/demos/tutorial_vqe.html>`_ combines smart circuit
+design with a gradient-based optimization to solve this task.
```
glassnotes, line 25:
```suggestion
design with gradient-based optimization to solve this task.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+with many parameters.
+
+At the same time, we have a good understanding of the *local* shape
+of the cost landscape around any reference point.
+Cashing in on this, the authors of the
+Quantum Analytic Descent paper [#QAD]_
+propose an algorithm that constructs a classical model which approximates the
+landscape, so that the gradients can be calculated on a classical computer, which is much cheaper.
+In order to build the classical model, we need to use the quantum device to
+evaluate the cost function on (a) a reference point :math:`\boldsymbol{\theta}_0`
```
glassnotes, line 41:
```suggestion
evaluate the cost function on (a) a reference point :math:`\boldsymbol{\theta}_0`,
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+
+
+VQEs give rise to trigonometric cost functions
+----------------------------------------------
+
+When we talk about VQEs we have a quantum circuit with :math:`n` qubits in mind, which are typically initialized in the base state :math:`|0\rangle`.
+The body of the circuit is a *variational form* :math:`V(\boldsymbol{\theta})` -- a fixed architecture of quantum gates parametrized by an array of real-valued parameters :math:`\boldsymbol{\theta}\in\mathbb{R}^m`.
+After the variational form, the circuit ends with the measurement of a chosen observable
+:math:`\mathcal{M}`, based on the problem
+we try to solve.
```
glassnotes, line 69:
```suggestion
we are trying to solve.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+
+We want to find the lowest possible energy the system can attain;
+this corresponds to running an optimization program to find the :math:`\boldsymbol{\theta}` that minimizes the function above.
+
+
+If the gates in the variational form are restricted to be Pauli rotations, then the cost function is a sum of *multilinear trigonometric terms* in each of the parameters.
+That's a scary sequence of words!
+What it means is that if we look at :math:`E(\boldsymbol{\theta})` but we focus only on one of the parameters, say :math:`\theta_i`, then we can write the functional dependence as a linear combination of three functions: :math:`1`, :math:`\sin(\theta_i)`, and :math:`\cos(\theta_i)`.
+
+That is, for each parameter :math:`\theta_i` there exist :math:`a_i`, :math:`b_i`, and :math:`c_i` such that the cost can be written as
```
glassnotes, line 83:
```suggestion
That is, for each parameter :math:`\theta_i` there exist :math:`a_i`, :math:`b_i`, and :math:`c_i` such that the cost can be written as
```

glassnotes: Fine to keep in the same paragraph since it is continuing the thought.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+    color="orange",
+    zorder=100,
+)
+
+###############################################################################
+# Of course this is an overly simplified example, but the key take-home message so far is:
+# *if the variational parameters feed into Pauli rotations, the energy landscape is a multilinear combination of trigonometric functions*.
+# What is a good thing about trigonometric functions?
+# That's right!
+# We have studied them since high school and know how their graphs look!
```
glassnotes, line 174:
```suggestion
# We have studied them since high school and know how their graphs look.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -215,6 +216,7 @@ def circuit(parameters):
 # In the QAD paper, the authors state that a *trigonometric expansion* up to second order is already a sound model candidate. Let's recap such expansions.
 #
 # .. admonition:: Taylor expansion vs. Trigonometric expansion
+#
 #     In spirit, a trigonometric expansion and a Taylor expansion are not that different: both are linear combinations of some basis functions, where the coefficients of the sum take very specific values usually related to the derivatives of the function we want to approximate.
 #     The difference between Taylor's and a trigonometric expansion is mainly what basis of functions we take.
 #     In Calculus I we learnt a Taylor series in one variable :math:`x` uses the integer powers of the variable namely :math:`\{1, x, x^2, x^3, \ldots\}`, in short :math:`\{x^n\}_{n\in\mathbb{N}}`:
```
glassnotes, line 222:
```suggestion
#     In Calculus I we learned that a Taylor series in one variable :math:`x` uses the integer powers of the variable namely :math:`\{1, x, x^2, x^3, \ldots\}`, in short :math:`\{x^n\}_{n\in\mathbb{N}}`:
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+#     The difference between Taylor's and a trigonometric expansion is mainly what basis of functions we take.
+#     In Calculus I we learnt a Taylor series in one variable :math:`x` uses the integer powers of the variable namely :math:`\{1, x, x^2, x^3, \ldots\}`, in short :math:`\{x^n\}_{n\in\mathbb{N}}`:
+#
+#     .. math:: f_\text{Taylor}(x) = \sum c_n(x-x_0)^n.
+#
+#     A trigonometric expansion instead uses a different basis, also for one variable: :math:`\{1, \sin(x), \cos(x), \sin(2x), \cos(2x), \ldots\}`, which we could call the set of trigonometric monomials with integer frequency, or in short :math:`\{\sin(nx),\cos(nx)\}_{n\in\mathbb{N}}`:
+#
+#     .. math:: f_\text{Trig}(x) = \sum a_n \cos(n(x-x_0))+ b_n \sin(n(x-x_0)).
+#
+#     For higher-dimensional variables we have to take products of the basis functions of each coordinate, i.e. of monomials or trigonometric monomials respectively.
```
glassnotes, line 230:
```suggestion
#     For higher-dimensional variables we have to take products of the basis functions of each coordinate, i.e., of monomials or trigonometric monomials respectively.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -215,6 +216,7 @@ def circuit(parameters):
 # In the QAD paper, the authors state that a *trigonometric expansion* up to second order is already a sound model candidate. Let's recap such expansions.
 #
 # .. admonition:: Taylor expansion vs. Trigonometric expansion
+#
```
glassnotes, line 219: This streamlined explanation looks great! I suggest actually to make this part of the main text, rather than an admonition, since it comprises most of the section.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -229,54 +231,35 @@ def circuit(parameters):
...
-# #. The :math:`1^{\text{st}}` order terms (would be :math:`x_k` for Taylor for each component :math:`k`) are :math:`B_k(\boldsymbol{\theta}) = 2\cos\left(\frac{\theta_k}{2}\right)\sin\left(\frac{\theta_k}{2}\right)\prod_{i\neq k} \cos\left(\frac{\theta_i}{2}\right)^2`.
-# #. The :math:`2^{\text{nd}}` order terms with respect to a single parameter (:math:`x_k^2` for Taylor for each component :math:`k`) are :math:`C_k(\boldsymbol{\theta}) = 2\sin\left(\frac{\theta_k}{2}\right)^2\prod_{i\neq k} \cos\left(\frac{\theta_i}{2}\right)^2`.
-# #. The :math:`2^{\text{nd}}` order terms with respect to mixed parameters (:math:`x_k x_l` for Taylor for each pair of *different* indices :math:`(k, l)`) are :math:`D_{kl}(\boldsymbol{\theta}) = 4\sin\left(\frac{\theta_k}{2}\right)\cos\left(\frac{\theta_k}{2}\right)\sin\left(\frac{\theta_l}{2}\right)\cos\left(\frac{\theta_l}{2}\right)\prod_{i\neq k,l} \cos\left(\frac{\theta_i}{2}\right)^2`.
+# One important aspect in which trigonometric series differ from regular
+# expansions is that there is not a clear separation between what terms
+# contribute to what order of th expansion (due to the fact that all order
```
glassnotes, line 239:
```suggestion
# contribute to each order of the expansion (due to the fact that all 
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -229,54 +231,35 @@ def circuit(parameters):
...
+#      - :math:`B_k(\boldsymbol{\theta}) = 2\cos\left(\frac{\theta_k}{2}\right)\sin\left(\frac{\theta_k}{2}\right)\prod_{i\neq k} \cos\left(\frac{\theta_i}{2}\right)^2`
+#      - :math:`x_k`
+#    * - 2
+#      - :math:`C_k(\boldsymbol{\theta}) = 2\sin\left(\frac{\theta_k}{2}\right)^2\prod_{i\neq k} \cos\left(\frac{\theta_i}{2}\right)^2`
+#      - :math:`x_k^2`
+#    * - 2
+#      - :math:`D_{kl}(\boldsymbol{\theta}) = 4\sin\left(\frac{\theta_k}{2}\right)\cos\left(\frac{\theta_k}{2}\right)\sin\left(\frac{\theta_l}{2}\right)\cos\left(\frac{\theta_l}{2}\right)\prod_{i\neq k,l} \cos\left(\frac{\theta_i}{2}\right)^2`
+#      - :math:`x_kx_l`
 #
 # Those are really large terms as compared to a Taylor series!
```
glassnotes, line 264:
```suggestion
# Those are really large terms compared to a Taylor series!
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -229,54 +231,35 @@ def circuit(parameters):
...
-# #. The :math:`2^{\text{nd}}` order terms with respect to a single parameter (:math:`x_k^2` for Taylor for each component :math:`k`) are :math:`C_k(\boldsymbol{\theta}) = 2\sin\left(\frac{\theta_k}{2}\right)^2\prod_{i\neq k} \cos\left(\frac{\theta_i}{2}\right)^2`.
-# #. The :math:`2^{\text{nd}}` order terms with respect to mixed parameters (:math:`x_k x_l` for Taylor for each pair of *different* indices :math:`(k, l)`) are :math:`D_{kl}(\boldsymbol{\theta}) = 4\sin\left(\frac{\theta_k}{2}\right)\cos\left(\frac{\theta_k}{2}\right)\sin\left(\frac{\theta_l}{2}\right)\cos\left(\frac{\theta_l}{2}\right)\prod_{i\neq k,l} \cos\left(\frac{\theta_i}{2}\right)^2`.
+# One important aspect in which trigonometric series differ from regular
+# expansions is that there is not a clear separation between what terms
+# contribute to what order of th expansion (due to the fact that all order
+# derivatives of sine and cosine are non-zero in general).
+# Because of this, we group the terms by their leading order contribution, and
+# in the following table write them next to their non-trigonometric analogues.
```
glassnotes, line 242: Add a single line here that just describes how these forms were obtained.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+# .. math::
+#
+#   B_k(\boldsymbol{\theta}) &= 2\tan\left(\frac{\theta_k}{2}\right)A(\boldsymbol{\theta})\\
+#   C_k(\boldsymbol{\theta}) &= 2\tan\left(\frac{\theta_k}{2}\right)^2 A(\boldsymbol{\theta})\\
+#   D_{kl}(\boldsymbol{\theta}) &= 4\tan\left(\frac{\theta_k}{2}\right)\tan\left(\frac{\theta_l}{2}\right)A(\boldsymbol{\theta})
+#
+# With that, we know what type of terms we should expect to encounter in our local classical model:
+# the model we want to construct is a linear combination of the functions
+# :math:`A(\boldsymbol{\theta})`, :math:`B_k(\boldsymbol{\theta})` and :math:`C_k(\boldsymbol{\theta})`
+# for each parameter, and :math:`D_{kl}(\boldsymbol{\theta})` for every pair of different variables :math:`(\theta_k,\theta_l)`.
```
glassnotes, line 277:
```suggestion
# for each parameter, and :math:`D_{kl}(\boldsymbol{\theta})` for every pair of different parameters :math:`(\theta_k,\theta_l)`.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -432,74 +418,98 @@ def model_cost(params, E_A, E_B, E_C, E_D):
...
-#     #. We have one :math:`E^{(A)}`.
-#     #. We have :math:`m` many :math:`E^{(B)}`'s, one for each component.
-#     #. We have :math:`m` many :math:`E^{(C)}`'s, also one for each component.
-#     #. We have :math:`(m-1)m/2` many :math:`E^{(D)}`'s, because we only need to check every pair once.
-#
-#     So, in total, there are :math:`m^2/2 + 3m/2 + 1` parameters, which is indeed polynomially many.
-#
-#     In practice, not every parameter needs the same amount of circuit evaluations, though!
-#     Without going into detail on the parameter-shift rule for the gradient and Hessian, we remark that we need :math:`1\times 1 + 2\times m + 1\times m + 4\times (m-1)m/2` many circuit evaluations, which amounts to a total of :math:`2m^2+m+1`.
+#     In the following table we count them for an :math:`m`-dimensional input
```
glassnotes, line 421: Table looks great! Consider even moving it into the main text, because the contents are brief. But it's fine as-is too.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -432,74 +418,98 @@ def model_cost(params, E_A, E_B, E_C, E_D):
...
 
 ###############################################################################
 # In the first two plots, we see the true landscape, and the approximate model.
-# The vertical rods indicate the points at which the cost function
+# The vertical rods indicate the points at which the original cost function
 # was evaluated in order to obtain the model coefficients (we skip the additional
 # evaluations for :math:`E^{(C)}`, though, for clarity of the plot).
-# In the third plot, we see the difference between model and true landscape.
-# Around the reference point this difference is very small and changes very slowly,
+# The rod with the bead on top indicates the reference point around which the model
```
glassnotes, line 508: Awesome, this really helps with clarity.

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -432,74 +418,98 @@ def model_cost(params, E_A, E_B, E_C, E_D):
...
+# is built and at which it coincides with the original cost function up to second
+# order. This is underlined in the third plot, where we see the difference between
+# model and true landscape.
```
glassnotes, line 511:
```suggestion
# the model and true landscapes.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+plot_cost_and_model(circuit, mapped_model, past_parameters[0])
+
+###############################################################################
+# **Iteration 1:** We see the cost function and the model around our starting point. This is the same as the plot before.
+
+mapped_model = lambda params: model_cost(params, *past_coeffs[1])
+plot_cost_and_model(circuit, mapped_model, past_parameters[1])
+
+###############################################################################
+# **Iteration 2:** Now we observe the model to stay closer to the original landscape. In addition, the minimum of the model is within the displayed range -- we're getting closer.
```
glassnotes, line 625:
```suggestion
# **Iteration 2:** Now we observe the model better resembles the original landscape. In addition, the minimum of the model is within the displayed range -- we're getting closer.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+ax.set_xlabel("epochs")
+ax.set_ylabel("cost")
+leg = ax.legend()
+
+###############################################################################
+# Each of the orange lines corresponds to minimizing the model constructed at a
+# different reference point.
+# We can now more easily appreciate the phenomenon we just described:
+# towards the end of each "outer" optimization step, the model cost
+# can be seen to be significantly lower than the true cost.
```
glassnotes, line 659:
```suggestion
# can potentially be significantly lower than the true cost.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+# While the *true cost* values are bounded, that does not mean the ones of the *model* are!
+# Notice also how this only happens at the first stages of analytic descent.
+#
+# Bringing together a few chords we have touched so far: once we fix a reference value, the further we go from it, the less accurate our model becomes.
+# Thus, if we start far off from the true minimum, it can happen that our model exaggerates how steep the landscape is and then the model minimum lies lower than that of the true cost.
+# We see how values exiting the allowed range of the true cost function does not have an
+# impact on the overall optimization.
+#
+# In this demo we've seen how to implement the Quantum Analytic Descent algorithm
+# using PennyLane for a toy-example of a Variational Quantum Eigensolver.
```
glassnotes, line 673:
```suggestion
# using PennyLane for a toy example of a Variational Quantum Eigensolver.
```

demonstrations/tutorial_quantum_analytic_descent.py
```diff
@@ -0,0 +1,706 @@
...
+# By making extensive use of 3D-plots we have also tried to illustrate exactly
```
glassnotes, line 674:
```suggestion
# By making extensive use of 3D plots we have also tried to illustrate exactly
```

glassnotes (approved):
@dwierichs @emgilfuster Great work! Left just a few small suggestions and caught some typos. Otherwise, this is merge-ready. :tada:

demonstrations/tutorial_quantum_analytic_descent.py
emgilfuster: At some point Maria proposed to somehow differentiate the scary/mathy stuff from the plain text. This used to be a note but you pointed out that might be too much of a separation given that it's critical information for the whole of the demo. It seemed to us this admonition thing could be a nice point inbetween, where you clearly see the margins are larger, so something is different, but it's not really that far away from the rest of the text.

demonstrations/tutorial_quantum_analytic_descent.py
dwierichs: Maybe we'd leave it as-is then, since it's just an unnecessary elaboration on the statement "the model requires polynomially many parameters", which is the big message that should (and already does) appear in the main text.

demonstrations/tutorial_quantum_analytic_descent.py
glassnotes: That's fair! :100:

demonstrations/tutorial_quantum_analytic_descent.py
glassnotes: Cool sounds good :+1:

PR #272, opened 2021-05-25
Use adjoint method in kernels demo

mariaschuld:
**Summary:**

Since we are moving to the `adjoint` method to invert an operation, this PR implements the change in the kernels demo, which may be an important reference for users of how to implement kernels in PennyLane.

The PR is made against master, since it uses functionality of the latest release.


mariaschuld:
Screenshot from 2021-05-25 11-41-25

The rendering and sanity check seem to work...

josh146 (approved):
image

My favourite type of PR to code review 💪

PR #273, opened 2021-05-25
Update `pyscf` version.

glassnotes:
Version `1.7.2` of `pyscf` causes installation issues when using Python 3.9 (as detailed [in this issue](https://github.com/PennyLaneAI/pennylane/issues/1145) and later resolved [here](https://github.com/PennyLaneAI/pennylane/pull/1254)
0. This PR updates `requirements.txt` file to use version `1.7.6` so that Python 3.9 can be used to build the demos.


josh146 (approved):
Thanks @glassnotes! Unfortunately, I don't think we can validate that this supports Python 3.9 yet, without also updating CircleCI to build the website using Python 3.9.

In particular, I think some of the other dependencies (e.g., TensorFlow <2.5) do not support Python 3.9.

glassnotes:
Ah, no worries! I see. Is there any disadvantage to bumping the `pyscf` version anyways? Or is better to wait until we can have everything support 3.9 (in which case, feel free to close this).

josh146:
No disadvantage if the build passes :)

PR #275, opened 2021-05-28
CH4398 | Fix Community Card Anchors

AndrewGardhouse:
**Title:**
Fix community card anchors

**Summary:**
On the Community page, if a URL has the `id` of a community card, the page will scroll to that card and highlight it. 

**Relevant references:**
- [CH4398](https://app.clubhouse.io/xanaduai/story/4398/fix-community-card-anchors)
- https://github.com/PennyLaneAI/qml/pull/194

### DIFF
--- a/custom_directives.py
+++ b/custom_directives.py
@@ -264,9 +264,6 @@ class CommunityCardDirective(Directive):
         if blog_url is not None:
             blog_footer = BLOG_FOOTER.format(blog=blog_url)
 
-        title_iterable = self.options["title"].split(" ")
-        title_first_letters = "".join(list(map(lambda word: word[0], title_iterable))).lower()
-
         card_rst = COMMUNITY_CARD_TEMPLATE.format(
             title=self.options["title"],
             author=self.options["author"],
@@ -276,7 +273,7 @@ class CommunityCardDirective(Directive):
             code_footer=code_footer,
             blog_footer=blog_footer,
             color=color,
-            id=title_first_letters + "-" + self.options["author"].split(" ")[-1].lower()
+            id=self.options["author"].split(" ")[0].lower() + self.options["date"].split("/")[-1] + self.options["title"].split(" ")[0].lower(),
         )
 
         thumbnail = StringList(card_rst.split('\n'))
--- a/custom_directives.py
+++ b/custom_directives.py
@@ -186,7 +186,7 @@ class YoutubeItemDirective(Directive):
 COMMUNITY_CARD_TEMPLATE = """
 .. raw:: html
 
-    <div class="card plugin-card">
+    <div class="card plugin-card" id={id}>
         <h4 class="card-header {color} lighten-4">{title}</h4>
         <div class="card-body">
             <h6>{author}</h6>
@@ -258,7 +258,7 @@ class CommunityCardDirective(Directive):
 
         if code_url is not None:
             code_footer = CODE_FOOTER.format(code=code_url)
-            
+
         blog_url = self.options.get("blog", None)
 
         if blog_url is not None:
@@ -272,7 +272,8 @@ class CommunityCardDirective(Directive):
             paper_footer=paper_footer,
             code_footer=code_footer,
             blog_footer=blog_footer,
-            color=color
+            color=color,
+            id=self.options["date"].replace("/", "-") + "-" + self.options["author"].split(" ")[-1].lower()
         )
 
         thumbnail = StringList(card_rst.split('\n'))
--- a/custom_directives.py
+++ b/custom_directives.py
@@ -264,6 +264,19 @@ class CommunityCardDirective(Directive):
         if blog_url is not None:
             blog_footer = BLOG_FOOTER.format(blog=blog_url)
 
+        def remove_accents(raw_text):
+            """Removes common accent characters."""
+
+            raw_text = re.sub(u"[àáâãäå]", 'a', raw_text)
+            raw_text = re.sub(u"[èéêë]", 'e', raw_text)
+            raw_text = re.sub(u"[ìíîï]", 'i', raw_text)
+            raw_text = re.sub(u"[òóôõö]", 'o', raw_text)
+            raw_text = re.sub(u"[ùúûü]", 'u', raw_text)
+            raw_text = re.sub(u"[ýÿ]", 'y', raw_text)
+            raw_text = re.sub(u"[ß]", 'ss', raw_text)
+            raw_text = re.sub(u"[ñ]", 'n', raw_text)
+            return raw_text
+
         card_rst = COMMUNITY_CARD_TEMPLATE.format(
             title=self.options["title"],
             author=self.options["author"],
@@ -273,7 +286,7 @@ class CommunityCardDirective(Directive):
             code_footer=code_footer,
             blog_footer=blog_footer,
             color=color,
-            id=self.options["author"].split(" ")[0].lower() + self.options["date"].split("/")[-1] + self.options["title"].split(" ")[0].lower(),
+            id=remove_accents(self.options["author"].split(" ")[-1].lower()) + self.options["date"].split("/")[-1] + self.options["title"].split(" ")[0].lower(),
         )
 
         thumbnail = StringList(card_rst.split('\n'))
--- a/custom_directives.py
+++ b/custom_directives.py
@@ -264,6 +264,9 @@ class CommunityCardDirective(Directive):
         if blog_url is not None:
             blog_footer = BLOG_FOOTER.format(blog=blog_url)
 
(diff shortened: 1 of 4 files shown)


custom_directives.py
```diff
@@ -258,12 +258,15 @@ def run(self):
...
             code_footer = CODE_FOOTER.format(code=code_url)
-            
+
         blog_url = self.options.get("blog", None)
 
         if blog_url is not None:
             blog_footer = BLOG_FOOTER.format(blog=blog_url)
 
+        title_iterable = self.options["title"].split(" ")
+        title_first_letters = "".join(list(map(lambda word: word[0], title_iterable))).lower()
```
josh146, line 268: Just trying to parse this one; from what I can tell, it's taking the first letter of each word in the title? So

> Linear regression using angle embedding and a single qubit by Hemant Gahankari

becomes

```
id = lruaeaasq-gahankari
```

josh146: I was trying to work out if there was a way to make it more readable, but I can't come up with anything else, because there are so many chances of clashes otherwise 😆

xanadu_theme/layout.html
```diff
@@ -250,10 +250,6 @@
     <!-- lightslider -->
     <script src="{{ pathto('_static/js/lightslider.min.js', 1) }}"></script>
 
-    <script type="text/javascript">
-      $(".document > .section").removeClass("section");
-      $(".nano").nanoScroller();
-    </script>
```
josh146, line 256: How come this was removed?

demos_community.rst.template
```diff
@@ -95,3 +95,46 @@ Community
...
+
+            // When a user hovers over a new card, we need to remove the active class from the previously active card
+            $('.plugin-card').on('mouseenter', function() {
+                $('.plugin-card').each(function(index, card) {
+                    if ($(card).hasClass('active')) { $(card).removeClass('active'); }
+                });
+            });
+
+            // When a URL is changed with a new hash, we want to scroll to the new card and add the active class to it
+            $(window).on('hashchange', function(e) {
```
josh146, line 124: This is great! My only nitpick (which is relatively minor) is that as soon as the mouse hovers over the card, the `active` class is lost. it would be great to keep this green glow more permanently while the URL hash is still present

josh146 (changes requested):
Thanks @AndrewGardhouse this is great! And kudos for puzzling out the community demos logical flow... it's very split between Jinja templates and Sphinx custom directives 😆 

It also works well in my testing when loading the URL from scratch 👍 The only thing I noticed though was that if you reload the URL, 

- The page scrolls just _below_ the card title
- The active class does not get re-added

I've added a small gif to show this:

ezgif com-gif-maker

custom_directives.py
glassnotes: What if we did similar to how arXiv generates BibTeX citation IDs, like `nameYEARfirstword`. So `gahankari2020using`, `huembeli2020characterizing`. There are still chances for things to clash (my bib file has an awful lot of `nameYEARquantum`...)  but could we work around that by storing the list of IDs and postfixing a number to avoid duplicates, `nameYEARquantum`, `nameYEARquantum-2`, etc.

glassnotes (commented):
Thanks @AndrewGardhouse !! :tada: 

This is maybe out of scope for this PR, but what if we also include a "copy link" button / text option on each card so that someone browsing the page can directly copy links to cards.

Like the same functionality as this:
image

Then the readability of the IDs is not as critical, as you can just extract them through copying rather than inferring through the title.

xanadu_theme/layout.html
AndrewGardhouse: I removed this because the same code is in the `<script>` tag above it

AndrewGardhouse:
> Thanks @AndrewGardhouse !! 🎉
> 
> This is maybe out of scope for this PR, but what if we also include a "copy link" button / text option on each card so that someone browsing the page can directly copy links to cards.
> 
> Like the same functionality as this:
> image
> 
> Then the readability of the IDs is not as critical, as you can just extract them through copying rather than inferring through the title.

This is a great idea, but I do feel like this is out of scope for this PR

custom_directives.py
AndrewGardhouse: This one was tricky to come up with a unique, yet readable ID 😅  I like @glassnotes' idea 👍

AndrewGardhouse: Screen Shot 2021-06-01 at 5 14 42 PM
I just found another issue with the current approach, if the name being used in the ID has an accent it will break the JS

glassnotes: Oh wow, good catch! Could we just replace the accented characters with the unaccented base character for the URL?

AndrewGardhouse:
@glassnotes @josh146 I addressed all of your comments 👍 Sorry for the delay, I didn't have notifications set up for GitHub PR mentions!

custom_directives.py
```diff
@@ -258,12 +258,25 @@ def run(self):
...
+
+            raw_text = re.sub(u"[àáâãäå]", 'a', raw_text)
+            raw_text = re.sub(u"[èéêë]", 'e', raw_text)
+            raw_text = re.sub(u"[ìíîï]", 'i', raw_text)
+            raw_text = re.sub(u"[òóôõö]", 'o', raw_text)
+            raw_text = re.sub(u"[ùúûü]", 'u', raw_text)
+            raw_text = re.sub(u"[ýÿ]", 'y', raw_text)
+            raw_text = re.sub(u"[ß]", 'ss', raw_text)
+            raw_text = re.sub(u"[ñ]", 'n', raw_text)
+            return raw_text
```
josh146, line 278: nice! This should be sufficient accent coverage :)

custom_directives.py
```diff
@@ -272,7 +285,8 @@ def run(self):
             paper_footer=paper_footer,
             code_footer=code_footer,
             blog_footer=blog_footer,
-            color=color
+            color=color,
+            id=remove_accents(self.options["author"].split(" ")[-1].lower()) + self.options["date"].split("/")[-1] + self.options["title"].split(" ")[0].lower(),
```
josh146, line 289: :+1:

demos_community.rst.template
```diff
@@ -95,3 +95,47 @@ Community
...
+            if (cardID && $(cardID).length) {
+                $(cardID).addClass('active');
+                $('html, body').animate({ scrollTop: $(cardID).offset().top - headerHeight }, scrollSpeed);
+            }
+
+            // When a user hovers over a new card, we need to remove the active class from the previously active card
+            $('.plugin-card').on('mouseenter', function(e) {
+                if ($(e.currentTarget).hasClass('active')) { return }
+                $('.plugin-card').each(function(index, card) {
+                    if ($(card).hasClass('active')) { $(card).removeClass('active'); }
```
josh146, line 120: I'm not sure exactly what is happening here, but it looks like this to me:

image

GitHub failing to syntax highlight?

josh146 (approved):
Thanks @AndrewGardhouse! Played around with it and the new ids work well :+1:

Only two comments:

- It seems that the `active` class disappearing on mouse hover still occurs, which means that the green shading can quickly disappear.
- The scrolling behaviour in the gif I posted before still seems to occur if reloading a URL with the same id 🤔 

However, I think this PR is currently a net benefit against master, so I'm happy to merge it in as-is and address those issues later if need be.

Oh, just one more thing - best not to force push into PRs, since it messes up GitHub's history, and it makes it hard to (a) view previous conversations (since they get marked as 'outdated') and (b) compare current changes to previous commits 🙂 

Even if the history within the PR becomes unwieldy, we only do merge commits into master, so there's an opportunity to re-phrase commit messages if need be!

custom_directives.py
glassnotes: Yep, plus the way it's coded makes it easy to add more in the future should we need to. Awesome :tada:

glassnotes (approved):
Thanks @AndrewGardhouse ! Looks good from my end :tada:

AndrewGardhouse:
@josh146 

> * The scrolling behaviour in the gif I posted before still seems to occur if reloading a URL with the same id 🤔
I wasn't able to reproduce this, could you provide the URL that is causing this issue, please?

> * It seems that the active class disappearing on mouse hover still occurs, which means that the green shading can quickly disappear.

Do we want the active class to remain on the card as long as the card's hash was navigated to through the URL, regardless of if a user is hovering over any of the other cards? Right now, the active class is only removed if the user hovers over another card.

josh146:
> Do we want the active class to remain on the card as long as the card's hash was navigated to through the URL, regardless of if a user is hovering over any of the other cards?

Yes, that was my intuition while browsing! So I found it a bit jarring when I saw the shadow change

PR #277, opened 2021-06-03
CH4568 | Fix Size Of Community Demo Cards

AndrewGardhouse:
**Summary:**
All the Community cards have the same height, and if the description text does not fit, there is a link to open a modal with the full description.

**Relevant references:**
[CH4568](https://app.clubhouse.io/xanaduai/story/4568/fix-size-of-community-demo-cards-and-add-a-read-more-dropdown-option
)

### DIFF
--- a/custom_directives.py
+++ b/custom_directives.py
@@ -192,8 +192,8 @@ COMMUNITY_CARD_TEMPLATE = """
                 <h4 class="card-header__text">{title}</h4>
             </div>
             <div class="card-body">
-                <div class="row h-100">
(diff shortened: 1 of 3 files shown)


josh146 (approved):
This is great @AndrewGardhouse! I tested it out on both my PC and mobile and it works well. I especially like how it only restricts the height/uses the models when there are two columns, and not when there is only one.

glassnotes (approved):
@AndrewGardhouse looks amazing!! :star2: Having a pop-up when clicking "read more" is a great idea. Thank you for implementing this :tada:

PR #281, opened 2021-06-09
Made sure to keep track of running best_cost

tvottra:
**Title:** Made sure to keep track of running best_cost (the current code forgets to update it)

**Summary:**

This ensures that our computed output_bloch_v is truly the best one encountered in gradient descent, instead of one computed by just any parameters that result in a cost less than our very first initial cost.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**


co9olguy:
Hi @tvottra, well spotted!

co9olguy (approved):

PR #285, opened 2021-06-15
Incorrect energy landscape in QNG demo

therooler:
**Title:**
Incorrect energy landscape in QNG demo

**Summary:**
The energy landscape in [the QNG demo](https://pennylane.ai/qml/demos/tutorial_vqe_qng.html#visualizing-the-results) is incorrect, as well as the description stating that red corresponds to the minimum.

We know that it is incorrect, because gradient descent should move orthogonal to the level curves of the function, which is not what we see here:
image

**Resolving the issue.**
I uploaded the correct `param_landscape.npy` generated by the code:
```python
parameter_landscape =[]
theta00, theta11 = np.meshgrid(theta0.flatten(), theta1.flatten())
for t0,t1 in zip(theta00.flatten(), theta11.flatten()):     
    parameter_landscape.append(cost_fn([t0,t1]))
parameter_landscape = np.array(parameter_landscape).reshape(100,100)
np.save("param_landscape",parameter_landscape)
```
which I inserted in line 185 of the tutorial file.
Loading these parameters gives the figure:
bug_fixed

Which is what we expect; gradient descent is orthogonal to the level curves of the function at each step of the optimization.


co9olguy:
@glassnotes:
@therooler and I found this bug, but would be good to get an independent sanity check

glassnotes (approved):
Good catch! Checked this myself, it looks like the underlying grid of cost values just needs to be the transpose of what it currently is, so that the x axis lines up with `theta0`, and the y with `theta1`:
```
for idx0, param0 in enumerate(theta0):
    for idx1, param1 in enumerate(theta1):
        grid[idx1, idx0] = cost_fn([param0, param1])
```

josh146:
Thanks for catching this 🤦 This was definitely a transpose issue, I remember (incorrectly) transposing it while writing the demo...

PR #286, opened 2021-06-15
Set `shots=1000` in device

thisac:
Minor update setting `shots=1000` in device so that results are noisy (as expected).

(also, quite a few whitespace changes, so I recommend ignoring them in the diff)


glassnotes (approved):
Thanks @thisac , nice catch! :tada:

antalszava (approved):
Thank you @thisac! :blush: Great catch!

Just double-checking: the output correct after the change as we'd expect, right? We can also check by incrementing the cache number in the circleci file if we'd like it to be built online.

PR #290, opened 2021-06-17
Add note about `qsimcirq` warning

thisac:
Adds a note related to the `qsimcirq` warning printed in accordance with the getting-started guide [here](https://quantumai.google/qsim/tutorials/qsimcirq#measurement_sampling).

Note that it's not possible to filter out this warning in any straight-forward way, since it's simply a print statement in the source code. It would technically be possibly to bypass any prints made when calling the `run` method, as in the following example by wrapping it with a `stopPrint` function (found [here](https://stackoverflow.com/questions/42952623/stop-python-module-from-printing)), although this would probably be a slightly hacky solution.

```python
import sys
import os
def stopPrint(func, *args, **kwargs):
    with open(os.devnull,"w") as devNull:
        original = sys.stdout
        sys.stdout = devNull
        res = func(*args, **kwargs)
        sys.stdout = original
    return res
```

### DIFF
--- a/demonstrations/qsim_beyond_classical.py
+++ b/demonstrations/qsim_beyond_classical.py
@@ -393,6 +393,13 @@ def fidelity_xeb(samples, probs):
 # their correponding bitstrings, since we need the samples to be in the
 # computational basis.
 #
+# .. note::
+#
+#     Every time the previously defined circuit is run without any intermediate
+#     measurements, using the ``qsim`` device, a warning will be printed. More
+#     information about this warning can be found in the `Measurement sampling
+#     section of the qsimcirq guide <https://quantumai.google/qsim/tutorials/qsimcirq#measurement_sampling>`__.
+#
 
 seed = np.random.randint(0, 42424242)
 probs = circuit(seed=seed, return_probs=True)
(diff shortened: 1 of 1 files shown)


demonstrations/qsim_beyond_classical.py
```diff
@@ -393,6 +393,13 @@ def fidelity_xeb(samples, probs):
 # their correponding bitstrings, since we need the samples to be in the
 # computational basis.
 #
+# .. note::
+#
+#     Every time the previously defined circuit is run without any intermediate
+#     measurements, using the ``qsim`` device, a warning will be printed. More
+#     information about this warning can be found in the `Measurement sampling
```
antalszava, line 400: (Minor): We cannot specify intermediate measurements, so the circuit is always run without intermediate measurements.
```suggestion
#     Every time the previously defined circuit is run using the ``qsim`` device, ``qsimcirq``
#     will print a warning message because the circuit has no intermediate measurements.
#     More information about this warning can be found in the `Measurement sampling
```

antalszava (approved):
Looks good to me! :+1:

demonstrations/qsim_beyond_classical.py
thisac: True, perhaps it might seem like we allow intermediate measurements just based on my wording. 😆

PR #291, opened 2021-06-17
Backprop more reps

antalszava:


demonstrations/tutorial_backprop.py
```diff
@@ -180,8 +180,8 @@ def circuit(params):
 
 import timeit
 
-reps = 3
-num = 10
+reps = 5
```
antalszava, line 183:
```suggestion
reps = 10
```

demonstrations/tutorial_backprop.py
```diff
@@ -276,8 +276,8 @@ def circuit(params):
...
```
antalszava, line 279:
```suggestion
reps = 10
```

josh146:
@antalszava curious the reason for this change?

antalszava:
@josh146 the two diagrams seem to be less accurate than before, quite likely due to the change to use `autograd` as per #256.

Nonetheless, the outputs are correct. This PR is just an attempt to make the outputs more accurate.

**stable**
image

image

**latest** (using `dev` without no changes)
image

image

**Increased reps and nums** (updating `dev` with this PR)
image

image

josh146:
@antalszava, thanks for investigating this!

I have to admit, however, that 

looks relatively okay to me --- it _is_ a bit 'messy', but the trend line is still clear.

With this change, the demo now takes 10 minutes to run rather than 4. Should we keep it as is to preserve the shorter running time?

antalszava:
@josh146 definitely! Happy to keep it as is :slightly_smiling_face: Will close this.

Issue #228, opened 2021-03-08
Search bar


ixfoduap:
The collection of PennyLane demos is amazing! Yet as it continues to grow, it's becoming difficult to find specific demos. We have a nice structure organizing them into categories, and useful filters for specific topics, both of which are very helpful. Yet it often still takes time to locate specific tutorials.

Adding a search bar to the page would make it much easier to navigate and to quickly identify the location of specific demos. For example, where do I find demos on barren plateaus? On kernel methods? On VQE in different spin sectors? 

![Screenshot from 2021-03-08 17-05-16](https://user-images.githubusercontent.com/40441298/110388566-425bd700-8031-11eb-9492-e275fbbe9cb2.png)

josh146:
This should be quite easy to do - we are actually hiding the search bar which is automatically added via sphinx!

In fact, it should be possible to get to the search page still: https://pennylane.ai/qml/search.html

Weirdly, though, the search isn't working for me 🤔

josh146:
Okay, I can't get the PL _nor_ the SF docs search working either:

https://pennylane.readthedocs.io/en/stable/search.html?q=backpropagation

Does it work for anyone else? Both were working for me during QHack, as I was using it to search for functions.

co9olguy:
It still works, but it seems to be picky about which queries it will return results for (e.g., try `rot`). I don't think it's a general-purpose search

josh146:
Weird, none of the search pages are working for me, in both chrome and firefox 🤔

glassnotes:
Closing this as the search bar is now live :tada:

PR #292, opened 2021-06-18
CH6501 | Add a search bar to the demo pages

AndrewGardhouse:
**Summary:**
This PR add a search field to the Filters sidebar on the `demos_research.html` and `demos_basics.html` pages.

**Relevant references:**
[CH6501](https://app.clubhouse.io/xanaduai/story/6501/add-a-search-bar-to-the-demo-pages)

**Related GitHub Issues:**
https://github.com/PennyLaneAI/qml/issues/228


xanadu_theme/layout.html
```diff
@@ -195,6 +196,38 @@
...
+
+                $("#tutorialDemoSearchInput").on("keyup", function(e) {
+                    e.preventDefault();
+                    var searchQuery = e.target.value.toLowerCase();
+                    var $galleryItems = $(".gallery-grid .sphx-glr-thumbcontainer");
+                    var filterVal = $(".tags input[type='radio']:checked").val();
+
+                    if (filterVal !== "all") {
+                        $(".filters label[for='all']").addClass("active");
+                        $(".filters label:not([for='all'])").removeClass("active");
```
glassnotes, line 208: These lines here, they are resetting the filter to "all" whenever something gets typed in the search box? Do we want to instead keep the chosen tags and further filter the tagged stuff? I'm actually not sure what makes more sense; but intuitively if I'd filtered by tags and then searched, I'd expect the filter to stay on.

xanadu_theme/layout.html
```diff
@@ -195,6 +196,38 @@
...
+                    }
+
+                    if (!searchQuery.length) {
+                        $galleryItems.each(function(index, item) { $(item).css("display", "block"); });
+                    } else {
+                        $galleryItems.each(function(index, item) {
+                            var $item = $(item);
+                            var tooltipText = $item.attr("tooltip").toLowerCase();
+                            var categoryText = $item.attr("data-category").toLowerCase();
+                            var descriptionText = $item.find("span.doc").text().toLowerCase();
```
glassnotes, line 218: So basically this searches based on tooltip, tags, and the little description of the demo that is in the file?

glassnotes (commented):
@AndrewGardhouse thank you so much for implementing this!! :tada: This is a much-needed improvement to the demo pages. It even auto-updates as you type :rocket:

Question: when I start typing, I see some suggestions coming up in a dropdown below, is that just my browser, or is that intentional? If it is, where do those values get pulled from? Some of the suggested values don't yield any results on the page.

xanadu_theme/layout.html
AndrewGardhouse: Exactly 👍

AndrewGardhouse:
@glassnotes 
> Question: when I start typing, I see some suggestions coming up in a dropdown below, is that just my browser, or is that intentional? If it is, where do those values get pulled from? Some of the suggested values don't yield any results on the page.

This is the browser, the same thing happens to me:
Screen Shot 2021-06-21 at 8 53 08 AM

xanadu_theme/layout.html
AndrewGardhouse: Since there was no indication that was the expected behaviour, I went with the simplest approach. But I can change this so if there is a filter selected, then only the filtered cards will be searched with the input field, and if there is search text and a filter is selected, then only the searched cards will be filtered 🙂

glassnotes (approved):
Thanks @AndrewGardhouse , this looks good to me!

josh146 (approved):
Amazing! 🙌  Looks great to me as well 💯 

This is out of scope, but reviewing this PR reminded me that the filters/search bar disappear if the window is not wide enough. I wonder if we could retain the search box (at the very least) for all window sizes?

Edit: Actually, it would be good to have the search bar on the top-level demos/index page, but this is even _further_ out of scope, since it would have to be an actual search bar, and not a filter.

glassnotes:
> Edit: Actually, it would be good to have the search bar on the top-level demos/index page, but this is even further out of scope, since it would have to be an actual search bar, and not a filter.

Yeah I was thinking about this as well, but wasn't sure what the interface would look like. Would it bring you to a new page with demos from both the basic *and* research section? How would you navigate to those things after the search? (In any case, out of scope and I'm a huge fan of the new bar :rocket: )

josh146:
@AndrewGardhouse definitely out of scope, but Sphinx actually has search functionality built into it (there is a remnant here: pennylane.ai/qml/search.html). So we could see if we could revive that?

PR #295, opened 2021-06-21
Hide designs demo

glassnotes:
Unless there is a reason to remove the files, just removing the entry from the table of contents is the easiest way to hide/unhide the demo.

### DIFF
--- a/demos_research.rst
+++ b/demos_research.rst
@@ -244,12 +244,6 @@ such as benchmarking and characterizing quantum processors.
    :tags: quantumcomputing
 
 .. customgalleryitem::
-   :tooltip: Explore the amazing applications of unitary t-designs.
-   :figure: demonstrations/unitary_designs/fano.png
-   :description: :doc:`demos/tutorial_unitary_designs`
-   :tags: quantumcomputing
-
-.. customgalleryitem::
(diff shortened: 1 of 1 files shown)


demos_research.rst
```diff
@@ -243,12 +243,6 @@ such as benchmarking and characterizing quantum processors.
    :description: :doc:`demos/tutorial_haar_measure`
    :tags: quantumcomputing
 
-.. customgalleryitem::
-   :tooltip: Explore the amazing applications of unitary t-designs.
-   :figure: demonstrations/unitary_designs/fano.png
-   :description: :doc:`demos/tutorial_unitary_designs`
-   :tags: quantumcomputing
-
```
co9olguy, line 251: :100:

demos_research.rst
```diff
@@ -292,5 +286,4 @@ such as benchmarking and characterizing quantum processors.
     demos/tutorial_haar_measure
     demos/tutorial_gbs
     demos/learning2learn
-    demos/tutorial_unitary_designs
```
co9olguy, line 295: yep, all we need to do to hide this

co9olguy (approved):

Issue #301, opened 2021-07-05
[BUG] Pennylane Tutorial Search Bar


sylviemonet:
-------------------------------------------------------------------------------------------------------------

#### Issue description

Description of the issue - Can not get the full search bar on safari.

<img width="1440" alt="searchBar" src="https://user-images.githubusercontent.com/16817699/124407940-b0ae6d00-dd77-11eb-8198-8537803c5b25.png">

* *Expected behavior:* Get the full search bar.

* *Actual behavior:* 75% search bar...

* *System information:* Nohting.

#### Additional information

Don not know if it is only an issue on macOS Big Sur (11.4 Ver.) or not.

josh146:
@AndrewGardhouse could this be related to screen/window width?

AndrewGardhouse:
@josh146 It looked like this issue came up from how Chrome and Safari calculated `width: auto` differently. I just opened https://github.com/PennyLaneAI/qml/pull/302 to fix this

AndrewGardhouse:
https://github.com/PennyLaneAI/qml/pull/302 has been merged to `master`

PR #302, opened 2021-07-05
Resolve Issue #301

AndrewGardhouse:
Fixes search bar width issue addressed in https://github.com/PennyLaneAI/qml/issues/301


josh146 (approved):
As I don't have safari I can't test this, but your gif was convincing!

PR #306, opened 2021-07-23
Update designs demo to use qml.apply and fix issues in dev

glassnotes:
Fixes issues that occur in `dev` as a result of [PR #1434](https://github.com/PennyLaneAI/pennylane/pull/1434). Also changes all instances of `op.queue()` to `qml.apply(op)`.


demos_research.rst
```diff
@@ -260,6 +260,12 @@ such as benchmarking and characterizing quantum processors.
    :description: :doc:`demos/tutorial_haar_measure`
    :tags: quantumcomputing
 
+.. customgalleryitem::
+   :tooltip: Explore the amazing applications of unitary t-designs.
+   :figure: demonstrations/unitary_designs/fano.png
+   :description: :doc:`demos/tutorial_unitary_designs`
+   :tags: quantumcomputing
+
```
glassnotes, line 268:
```suggestion
```

demos_research.rst
```diff
@@ -306,3 +312,4 @@ such as benchmarking and characterizing quantum processors.
     demos/tutorial_mol_geo_opt
     demos/tutorial_classical_shadows
     demos/tutorial_quantum_analytic_descent
+    demos/tutorial_unitary_designs
```
glassnotes, line 315:
```suggestion
```

demonstrations/tutorial_unitary_designs.py
```diff
@@ -503,24 +503,24 @@ def conjugate_with_unitary(tape, matrix):
...
-    # This particular expression requires state_1 to be a pure state
-    fid = np.trace(np.dot(state_1, state_2)) + 2 * np.sqrt(
-        np.linalg.det(state_1) * np.linalg.det(state_2)
-    )
+from scipy.linalg import sqrtm
+
+def fidelity(rho, sigma):
+    # Inputs rho and sigma are density matrices
+    sqrt_sigma = sqrtm(sigma)
+    fid = np.trace(sqrtm(sqrt_sigma @ rho @ sqrt_sigma))
```
josh146, line 523: How come this was changed?

josh146 (approved):
Approved, conditional on the demo building and rendering correctly!

demonstrations/tutorial_unitary_designs.py
glassnotes: When I was debugging, I originally thought the error might be in the trace function, so I reimplemented it to use the more standard expression rather than the special pure-state case. In the end this was not the problem, but I thought this version of it is much cleaner so I kept it :)

josh146: Nice 😆

PR #312, opened 2021-08-04
Update `tutorial_noisy_circuits.py`

antalszava:
Updates the `tutorial_noisy_circuits.py` demonstration to use 0-dimensional PennyLane tensors instead of 1-dimensional ones.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,9 +20,9 @@ jobs:
       - restore_cache:
           keys:
             # when lock file changes, use increasingly general patterns to restore cache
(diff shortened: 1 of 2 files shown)


josh146:
@antalszava can this only be merged in once https://github.com/PennyLaneAI/pennylane/pull/1495 is merged?

And I suppose that means we need to merge it into `dev`?

antalszava:
Oh right, thanks! Yes, both are required (changed to the dev branch).

josh146 (approved):

PR #319, opened 2021-08-13
Pin to `master` of PL and plugins in `dev`

antalszava:
#318 was merged into the `dev` branch, this PR reverts that change so that we're pinning to the `master` version of PL.


rmoyard (approved):
Thank you, it looks good to me!

PR #324, opened 2021-08-23
Pin `numba` and increase cache number

antalszava:
The check [when merging into `master` failed on the dependencies](https://github.com/PennyLaneAI/qml/commit/54722c5f22246454ca13d2a22cf47dba3cb21e4f). This PR increases the cache number such that dependencies are installed fresh.

The newest version of `numba` seems to cause import issues:
https://app.circleci.com/pipelines/github/PennyLaneAI/qml/3854/workflows/36affa59-2f4d-4ec3-9208-40dd64abe63d/jobs/4087/steps?invite=true#step-106-1478

Therefore, numba is pinned.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,9 +20,9 @@ jobs:
       - restore_cache:
           keys:
             # when lock file changes, use increasingly general patterns to restore cache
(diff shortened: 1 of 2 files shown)


josh146 (approved):
💯 Thanks Antal!

PR #329, opened 2021-09-07
Update the Torch version in `requirements.txt`

antalszava:
Bump the version of Torch due to `defaut.qubit.torch`:

https://github.com/PennyLaneAI/qml/runs/3515455887?check_suite_focus=true#step:8:2623


antalszava:
[ch8736]

albi3ro (approved):
Sounds good 👍

PR #330, opened 2021-09-10
Update demos to make use of the new qml.sample and qml.expval(H) functionality (story #: 8118) 

Jaybsoni:
### Before submitting

- [n] Ensure that your tutorial executes correctly, and conforms to the

- [y] Add a thumbnail link to your tutorial in `beginner.rst`, or if a

- [y] All QML tutorials conform to

**Update demos to use qml.sample and qml.expval(H):**

**Summary:**
Updating the tutorials to make use of the new functionality in qml.sample and removing the dependance on qml.ExpvalCost as it will be redacted

**Relevant references:**
story id: 8118

**Possible Drawbacks:**

**Related GitHub Issues:**


demonstrations/tutorial_measurement_optimize.py
```diff
@@ -141,14 +141,18 @@
...
+    return qml.expval(H)
 
 ##############################################################################
 # If we evaluate this cost function, we can see that it corresponds to 15 different
 # QNodes under the hood---one per expectation value:
 
 params = np.random.normal(0, np.pi, len(singles) + len(doubles))
-print("Cost function value:", cost(params))
+print("Cost function value:", cost_circuit(params))
 print("Number of quantum evaluations:", dev.num_executions)
```
Jaybsoni, line 156: Two things I would like to point out regarding the updates to this file: 
- This particular line of code (156), in the original tutorial produces 15 'executions'. The updated one only generates a single 'execution'. Not sure if this is a bug. If it isn't then I am not sure if we still want to keep this line of code. 
- Secondly, some of the results generated (in this entire tutorial) vary from the results of the original tutorial in the last 10^-13 ~ 10^-15 decimal places. I assume this is just noise on the python side. Not sure if that is worth looking into.

josh146: Good catches @Jaybsoni! I can provide some insights.

> This particular line of code (156), in the original tutorial produces 15 'executions'. The updated one only generates a single 'execution'. Not sure if this is a bug. If it isn't then I am not sure if we still want to keep this line of code.

This is because `return expval(H)` acts very differently to `ExpvalCost` internally. While `ExpvalCost` always creates a separate circuit per Hamiltonian commuting group (as you need to do on hardware), `return expval(H)` tries to be smarter:

- If `shots=None`, it computes the expectation value `<psi|H|psi>` _directly_, without needing to group Hamiltonian terms.
- If `shots>0`, it reverts to the existing behaviour of grouping.

In most case when running a simulation, we only care about the expectation value --- _how_ we computed the expectation value is not so important. But in this case, this particular demo actually _explores_ the grouping behaviour!

So we want to preserve the old behaviour here. We have several options:

1. One option is to _manually_ group the Hamiltonian, and print the commuting groups:

  ```pycon
  >>> H.compute_grouping()
  >>> len(H.grouping_indices)
  5
  ```

  (and ignore that the simulation is actually just doing one execution!). I don't really like this though, doesn't feel very clear.

2. Another option is to set `shots=1000` (or something similar) for the sections that use `ExpvalCost` (e.g., the first section, and the very last code-block). The rest can be left at `shots=None` for speed/accuracy.

3. A third option, which I think I like the most (but which doesn't exist yet!) is to extend PennyLane so that you can specify that you want grouping to occur even if `shots=None`. But this requires (a) thinking about how we want the UI to look, and (b) changes to PL master 🙂 

> Secondly, some of the results generated (in this entire tutorial) vary from the results of the original tutorial in the last 10^-13 ~ 10^-15 decimal places. I assume this is just noise on the python side. Not sure if that is worth looking into.

Yep, this would be related to the different logical pathway taken by `expval(H)` and `ExpvalCost` 🙂

josh146: Another tangential thought: I just realized that `dev.num_executions` is now outdated, and has been replaced with the device tracker: https://pennylane.readthedocs.io/en/stable/code/api/pennylane.Tracker.html

demonstrations/tutorial_vqe_qng.py
```diff
@@ -63,7 +63,11 @@ def circuit(params, wires=0):
 obs = [qml.PauliX(0), qml.PauliZ(0)]
 
 H = qml.Hamiltonian(coeffs, obs)
-cost_fn = qml.ExpvalCost(circuit, H, dev)
+
+@qml.qnode(dev)
```
Jaybsoni, line 67: The documentation paragraph associated with this bit of code (lines 58 - 60) describes some capabilities of qml.ExpvalCost, not sure if this is still true for qml.expval. Let me know how I should update that bit of documentation.

josh146: good catch @Jaybsoni, I would remove the "using the ``ExpvalCost`` class," part :)

demonstrations/tutorial_vqe_parallel.py
```diff
@@ -178,10 +187,9 @@ def circuit(param, wires):
...
 ##############################################################################
-# Finally, the energies as functions of rotation angle can be given using
-# :class:`~.pennylane.ExpvalCost`.
+# Finally, the energies as functions of rotation angle can be generated using:
 
-energies = [qml.ExpvalCost(circuit, h, devs) for h in hamiltonians]
+energies = [get_expval_qnode(h) for h in hamiltonians]
 
 ##############################################################################
 # Calculating the potential energy surface
```
Jaybsoni, line 195: The paragraph below also describes some functionality which is specific to ExpvalCost, any suggestions on how we can edit this one?

demonstrations/tutorial_vqe_qng.py
Jaybsoni: Ah awesome thanks !

demonstrations/learning2learn.py
```diff
@@ -239,8 +240,9 @@ def hamiltonian(params, **kwargs):
         # We set the default.qubit.tf device for seamless integration with TensorFlow
         dev = qml.device("default.qubit.tf", wires=len(graph.nodes))
 
-        # ExpvalCost evaluates the expectation value of an operator
-        cost = qml.ExpvalCost(circuit, cost_h, dev, interface="tf", diff_method="backprop")
+        # This qnode evaluates the expectation value of the cost hamiltonian operator
+        # cost = qml.ExpvalCost(circuit, cost_h, dev, interface="tf", diff_method="backprop")
```
Jaybsoni, line 244: It seems ExpvalCost has keyword arguments which (I assume) specify how derivatives are computed of the underlying expectation value with respect to the trainable parameters. How can I preserve these settings otherwise ?

antalszava: It should be possible to simply pass them to the QNode:

```python

@qml.qnode(dev, interface="tf", diff_method="backprop")
def circuit():
    # some circuit
    return expval(H)
```

demonstrations/tutorial_vqe_parallel.py
antalszava: Oh, that's a good question :thinking: Frankly, I'm not sure if using `parallel` option of QNodeCollection is still accessible with the `expval(H)` syntax.

@josh146 any thoughts?

josh146: I don't think it is! Perhaps best to leave this demo using `ExpvalCost`?

Jaybsoni: Cool, I will revert the changes for this demo then!

demonstrations/learning2learn.py
Jaybsoni: Ahh I see, thanks !

demonstrations/tutorial_measurement_optimize.py
Jaybsoni: Cool! I replaced `dev.num_executions` with `qml.Tracker`. I went with your 2nd suggestion as the 3rd one seems to be worth a separate PR. 

Also wanted to mention that I didn't set the shots for the very last code block because I don't think it is required, I think it explicitly computes the groupings.

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -141,15 +141,20 @@
...
 # If we evaluate this cost function, we can see that it corresponds to 15 different
 # QNodes under the hood---one per expectation value:
 
 params = np.random.normal(0, np.pi, len(singles) + len(doubles))
-print("Cost function value:", cost(params))
-print("Number of quantum evaluations:", dev.num_executions)
+with qml.Tracker(dev) as tracker:  # track the number of executions
+    print("Cost function value:", cost_circuit(params))
+
+print("Number of quantum evaluations:", tracker.totals['executions'])
```
antalszava, line 157: Nice!

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -176,15 +176,13 @@ def U_C(gamma):
 
 
 ##############################################################################
-# We will need a way to sample
-# a measurement of multiple qubits in the computational basis, so we define
-# a Hermitian operator to do this. The eigenvalues of the operator are
-# the qubit measurement values in integer form.
+# We will need a way to take our sampled measurement, which is a bitstring
+# representing the measured state in the computational basis, and convert it
+# to an int
```
antalszava, line 181:
```suggestion
# We will need a way to sample a measurement of multiple qubits in the computational basis
#  in integer form.
```

antalszava (commented):
Hi @Jaybsoni, this is looking nice! :tada: :slightly_smiling_face: Left some questions/suggestions.

1. There seem to be some occurrences of `ExpvalCost` in https://github.com/PennyLaneAI/qml/blob/master/demonstrations/tutorial_vqe_parallel.py too. Would they need to be changed?
2. Did the changed demos produce the expected outputs? (Including the ones that are not automatically run.)

demonstrations/tutorial_vqls.py
```diff
@@ -474,20 +474,23 @@ def prepare_and_sample(weights):
...
-samples = prepare_and_sample(w).astype(int)
-q_probs = np.bincount(samples) / n_shots
+raw_samples = prepare_and_sample(w)
+
+# Here we convert the raw samples (bit strings) into integers and count them
+samples = []
+for sam in raw_samples:
+    samples.append(int("".join(str(bs) for bs in sam), base=2))
 
+q_probs = np.bincount(samples, minlength=2 ** n_qubits) / n_shots
```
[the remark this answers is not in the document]
Jaybsoni: I think I specified it because I wanted to cover the case where we don't sample some of the bit strings at either end of the distributions. If we miss it, then the distribution might have fewer bins. But I guess its not necessary in this case as it wasn't included last time?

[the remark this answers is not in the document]
antalszava: Yep, maybe best left as is, if the change is not required. :)

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -125,7 +125,7 @@
 # on hardware. Let's generate the cost function to check this.
 
 # Create a 4 qubit simulator
-dev = qml.device("default.qubit", wires=num_qubits)
+dev = qml.device("default.qubit", wires=num_qubits, shots=1000)
```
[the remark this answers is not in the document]
Jaybsoni: Because when shots=None, the result is computed analytically the tracker only registers a single execution, but in this tutorial we want to track the number of executions if we were to explicitly run it.

demonstrations/tutorial_vqls.py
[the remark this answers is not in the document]
Jaybsoni: Got it !

demonstrations/tutorial_qaoa_intro.py
```diff
@@ -315,15 +315,18 @@ def circuit(params, **kwargs):
 # step is PennyLane's specialty: optimizing the circuit parameters.
 #
 # The cost function is the expectation value of :math:`H_C`, which we want to minimize. The
-# function :func:`~.pennylane.ExpvalCost` is designed for this purpose: it returns the
+# function :func:`~.pennylane.expval(H)` is designed for this purpose: it returns the
```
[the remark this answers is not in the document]
Jaybsoni: Got it,

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -745,12 +750,15 @@ def circuit(weights, group=None, **kwargs):
...
-# optimized:
+# problems), we can use the :func:`qml.expval` function
+# to generate our cost function:
 
 H = qml.Hamiltonian(coeffs=np.ones(len(terms)), observables=terms)
-cost_fn = qml.ExpvalCost(qml.templates.StronglyEntanglingLayers, H, dev, optimize=True)
+@qml.qnode(dev)
+def cost_fn(params):
+    qml.templates.StronglyEntanglingLayers(params, wires=dev.wires)
+    return qml.expval(H)
```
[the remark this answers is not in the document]
Jaybsoni: Ahh I see!

demonstrations/tutorial_falqon.py
```diff
@@ -234,6 +236,9 @@ def ansatz(beta, **kwargs):
             delta_t=delta_t
         )
 
+        if measurement_h is not None:
+            return qml.expval(measurement_h)
+
```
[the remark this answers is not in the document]
Jaybsoni: Got it!

demonstrations/tutorial_falqon.py
```diff
@@ -410,14 +411,14 @@ def qaoa_layer(gamma, beta):
     qaoa.cost_layer(gamma, cost_h)
     qaoa.mixer_layer(beta, mixer_h)
 
-# Creates the full QAOA circuit
+# Creates the full QAOA circuit as an executable cost function
+@qml.qnode(dev)
```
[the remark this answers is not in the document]
Jaybsoni: Ok, I think I get what you are suggesting, I implemented it in, let me know if this is what you had in mind.

Jaybsoni:
> Hi @Jaybsoni, this is looking nice! 🎉 🙂 Left some questions/suggestions.
> 
> 1. There seem to be some occurrences of `ExpvalCost` in https://github.com/PennyLaneAI/qml/blob/master/demonstrations/tutorial_vqe_parallel.py too. Would they need to be changed?
> 2. Did the changed demos produce the expected outputs? (Including the ones that are not automatically run.)

Hey Antal, I believe we discussed with Josh that we should leave the vqe_parallel tutorial as is. I have addressed all of your comments above. I confirmed that all of the automatically running demos are working as expected. I will go over the rest of them today and get back to you.

josh146:
> I believe we discussed with Josh that we should leave the vqe_parallel tutorial as is.

Yep, the reason was that the parallel VQE tutorial is designed to highlight the Dask parallelization of `ExpvalCost`, which no longer exists in the `expval(H)` pipeline, we may have to revisit this demo later

Jaybsoni:
Hi @antalszava and @josh146,

I finished checking most of the demos, there are just 2 left to check which I am having trouble running. 

1st: [qsim beyond classical demo](https://pennylane.ai/qml/demos/qsim_beyond_classical.html)
In this case I am running into a weird import error with qsimcirq: 
image
image
image

Have either of you come across this error before ? 

2nd: [braket parallel gradients](https://pennylane.ai/qml/demos/braket-parallel-gradients.html)
This one requires a Braket account to complete. 

If you could help me debug the first issue and provide me with at Braket account I can finish the reviews my self, otherwise it might be faster if someone who has a working setup can review them quickly so we can get this merged in? I am 90% sure it shouldn't cause any issues. 

Cheers,

josh146:
>  In this case I am running into a weird import error with qsimcirq:

I'm not 100% sure here, but I think @rmoyard or @thisac might know, I think they have run into similar issues!

>  This one requires a Braket account to complete.

Sure, that can be arranged!

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -176,15 +176,12 @@ def U_C(gamma):
 
 
 ##############################################################################
-# We will need a way to sample
-# a measurement of multiple qubits in the computational basis, so we define
-# a Hermitian operator to do this. The eigenvalues of the operator are
-# the qubit measurement values in integer form.
+# We will need a way to sample a measurement of multiple qubits in the computational basis
+#  in integer form.
```
mariaschuld, line 180:
```suggestion
# in integer form.
```

demonstrations/tutorial_vqe_qng.py
```diff
@@ -291,9 +295,12 @@ def ansatz(params, wires=[0, 1, 2, 3]):
 ##############################################################################
 # Note that the qubit register has been initialized to :math:`|1100\rangle`, which encodes for
 # the Hartree-Fock state of the hydrogen molecule described in the minimal basis.
-# Again, we define the cost function using the ``ExpvalCost`` class.
+# Again, we define the cost function using ``expval(H)``.
```
mariaschuld, line 298: I was wondering if here and in other places it would be clearer to say "we define the cost to be the expectation of the Hamiltonian" or "we define the cost to be the QNode itself"...It's not so important though.

demonstrations/tutorial_vqe_spin_sectors.py
```diff
@@ -185,17 +185,23 @@ def circuit(params, wires):
 dev = qml.device("default.qubit", wires=qubits)
 
 ##############################################################################
-# Next, we use the :class:`~.pennylane.ExpvalCost` class to define the cost function.
+# Next, we use the :func:`~.pennylane.expval` function to define the cost function.
```
mariaschuld, line 188: For example, we are not using the expval function, but the entire qnode to define the cost. Or in other words, the measurement outcome.

mariaschuld (commented):
Thanks for taking on this cumbersome task @Jaybsoni! 

@josh146  and I had a chat about how to deal with demos that use `ExpvalCost(..,parallel=True)` and we were wondering if one should just "unpack" the ExpvalCost and use dask explicitly. I.e., one could make one Qnode per Hamiltonian term, and then execute all Qnodes with dask. What do you think? 

This could be useful for users to get used to dask, and should have the same performance! It also shows an alternative, manual way to do VQE...

If you agree, just ping me if there are any problems! The important dask code is in the `QNodeCollection` class if I remember correctly.

demonstrations/tutorial_vqe_qng.py
Jaybsoni: That feels clearer to me as well,

demonstrations/tutorial_vqe_spin_sectors.py
Jaybsoni: Agreed!

antalszava:
@Jaybsoni just a guess, could there be something going on with the [`libomp` installation](https://formulae.brew.sh/formula/libomp)? That image seems to be missing. :thinking:

If it doesn't come around, let us know. :slightly_smiling_face:

thisac:
>  In this case I am running into a weird import error with qsimcirq:

I don't recognize the error, so I'm not sure what's causing it. Have you tried both installing with `pip` as well as directly from the GitHub master branch? I think I remember only getting one of them to work before.

@Jaybsoni I'm able to run the demo fine on my personal laptop, so I could help testing it out if you'd like.

Jaybsoni:
Hey all (@antalszava, @josh146), 

I can now confirm that all of the demos are running as expected. I was able to run the Braket parallel gradients demo today and @antalszava ran the Qsim beyond classical demo. Beyond that I have addressed any comments left on my changes. 


> Thanks for taking on this cumbersome task @Jaybsoni!
> 
> @josh146 and I had a chat about how to deal with demos that use `ExpvalCost(..,parallel=True)` and we were wondering if one should just "unpack" the ExpvalCost and use dask explicitly. I.e., one could make one Qnode per Hamiltonian term, and then execute all Qnodes with dask. What do you think?

@mariaschuld, thank you for the comments! That does seem to be a more informative solution, I think it might be better suited as its own separate story/ticket which I can take on after merging this one. 

I think the everything is good to go, let me know if you have any other suggestions and I can address them. 
Cheers,

antalszava:
[ch8118]

demonstrations/braket-parallel-gradients.py
```diff
@@ -96,7 +96,7 @@
 the address of the `S3 bucket <https://aws.amazon.com/s3/>`__ where results are to be stored:
 """
 
-my_bucket = "amazon-braket-Your-Bucket-Name"  # the name of the bucket
+my_bucket = "amazon-braket-Your-Bucket-Name"  # the name of the bucket, keep the 'amazon-braket-' and then include name
```
antalszava, line 99:
```suggestion
my_bucket = "amazon-braket-Your-Bucket-Name"  # the name of the bucket, keep the 'amazon-braket-' prefix and then include the bucket name
```

demonstrations/tutorial_falqon.py
```diff
@@ -410,14 +411,17 @@ def qaoa_layer(gamma, beta):
     qaoa.cost_layer(gamma, cost_h)
     qaoa.mixer_layer(beta, mixer_h)
 
-# Creates the full QAOA circuit
+# Creates the full QAOA circuit as an executable cost function
 def qaoa_circuit(params, **kwargs):
-    for w in dev.wires:
+    for w in new_graph.nodes:
```
antalszava, line 416: How come this was changed?

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -219,12 +220,14 @@ def circuit(params, wires):
...
-# Hamiltonian parameters.
+# Next, we define the `cost`` function :math:`g(\theta, x)` which depends on
+# both the circuit and the Hamiltonian parameters. Specifically we consider the
+# expectation values of the hamiltonian.
 
 def cost(params, x):
-    return qml.ExpvalCost(circuit, H(x), dev)(params)
+    hamiltonian = H(x)
+    run_circuit = qml.QNode(circuit, dev)
+    return run_circuit(params, obs=hamiltonian, wires=range(6))
```
antalszava, line 230: It might be worth creating an object for the number of wires e.g., on line 219 (`num_wires = 6`). Then this object can be used in the device definition and when passing the wires to `circuit`.

This would help with:
* Allowing that the number of wires needs to be changed only at one place (if applicable);
* Making the code slightly more descriptive by using a name such as `num_wires` or so.

antalszava (approved):
Looks good to me :champagne: :tada: :blush: Awesome addition @Jaybsoni, thank you! No major comments remain from my side.

Just double-checking: might it be that we're restoring the old cached outputs for the automatically executed demos? (ccing @josh146)

demonstrations/tutorial_falqon.py
Jaybsoni: Ah nice catch @antalszava, it must be a residual change left over when I was experimenting with the demo. I was sure I changed it back but I guess I must have missed it.

demonstrations/tutorial_mol_geo_opt.py
Jaybsoni: Ahh yes, it improves code readability!

Jaybsoni:
> Looks good to me 🍾 🎉 😊 Awesome addition @Jaybsoni, thank you! No major comments remain from my side.
> 
> Just double-checking: might it be that we're restoring the old cached outputs for the automatically executed demos? (ccing @josh146)

Thanks for the comments! I made sure to run each demo script locally as opposed to relying on the CI tests, so I can confidently say that each of the demos that have been changed here should run (at least locally on my setup). Let me know if there is anything else.

demonstrations/tutorial_coherent_vqls.py
```diff
@@ -496,18 +496,22 @@ def prepare_and_sample(weights):
     variational_block(weights)
 
     # We assume that the system is measured in the computational basis.
-    # If we label each basis state with a decimal integer j = 0, 1, ... 2 ** n_qubits - 1,
-    # this is equivalent to a measurement of the following diagonal observable.
-    basis_obs = qml.Hermitian(np.diag(range(2 ** n_qubits)), wires=range(n_qubits))
-
-    return qml.sample(basis_obs)
+    # Therefore, sampling the device will give us a value of 0 or 1 for each qubit (n_qubits)
```
mariaschuld, line 499:
```suggestion
    # Therefore, sampling from the device will give us a value of 0 or 1 for each qubit (n_qubits)
```

I think this is more correct :)

demonstrations/qsim_beyond_classical.py
```diff
@@ -297,8 +297,8 @@ def generate_single_qubit_gate_list():
 # From the QNode, we need both the probabilities of the measurement
 # results, as well as raw samples. To facilitate this, we add a keyword
 # argument to our circuit allowing us to switch between the two returns. We
-# sample from the Pauli-Z observable on all wires, which will give us the
-# eigenvalues :math:`\pm 1` of the observable, corresponding to the states
+# take samples from the computational basis from all wires, which will give us
```
mariaschuld, line 300:
```suggestion
# take samples from the computational basis state using all wires, which will return bitstrings of
```

demonstrations/tutorial_coherent_vqls.py
```diff
@@ -496,18 +496,22 @@ def prepare_and_sample(weights):
...
 
 
 ##############################################################################
 # To estimate the probability distribution over the basis states we first take ``n_shots``
 # samples and then compute the relative frequency of each outcome.
 
-samples = prepare_and_sample(w).astype(int)
+raw_samples = prepare_and_sample(w)
+
+# Here we convert the raw samples (bit strings) into integers and count them
```
mariaschuld, line 510:
```suggestion
# convert the raw samples (bit strings) into integers and count them
```

Not so important... :)

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -216,15 +217,18 @@ def circuit(params, wires):
...
+num_wires = 6
+dev = qml.device("default.qubit", wires=num_wires)
 
 ##############################################################################
-# Next, we use the PennyLane class :class:`~.pennylane.ExpvalCost` to define the
-# ``cost`` function :math:`g(\theta, x)` which depends on both the circuit and the
-# Hamiltonian parameters.
+# Next, we define the `cost`` function :math:`g(\theta, x)` which depends on
+# both the circuit and the Hamiltonian parameters. Specifically we consider the
+# expectation values of the hamiltonian.
```
mariaschuld, line 226: Nice!

demonstrations/tutorial_qaoa_intro.py
```diff
@@ -314,16 +314,19 @@ def circuit(params, **kwargs):
...
 # step is PennyLane's specialty: optimizing the circuit parameters.
 #
-# The cost function is the expectation value of :math:`H_C`, which we want to minimize. The
-# function :func:`~.pennylane.ExpvalCost` is designed for this purpose: it returns the
-# expectation value of an input Hamiltonian with respect to the circuit's output state.
-# We also define the device on which the simulation is
-# performed. We use the PennyLane-Qulacs plugin to
-# run the circuit on the Qulacs simulator:
+# The cost function is the expectation value of :math:`H_C`, which we want to minimize. We
+# use the function :func:`~.pennylane.expval` by passing it a Hamiltonian `H`: it returns the
```
mariaschuld, line 318:
```suggestion
# use the function :func:`~.pennylane.expval` which returns the
```

I am not sure if "passing it" is grammatically correct? Could just be simplified like this...

demonstrations/tutorial_vqe_qng.py
```diff
@@ -291,9 +295,12 @@ def ansatz(params, wires=[0, 1, 2, 3]):
 ##############################################################################
 # Note that the qubit register has been initialized to :math:`|1100\rangle`, which encodes for
 # the Hartree-Fock state of the hydrogen molecule described in the minimal basis.
-# Again, we define the cost function using the ``ExpvalCost`` class.
+# Again, we define the cost function to be the following QNode using the ``expval(H)`` function.
```
mariaschuld, line 298:
```suggestion
# Again, we define the cost function to be the following QNode that measures ``expval(H)``:
```

Only a suggestion...

demonstrations/tutorial_vqls.py
```diff
@@ -474,20 +474,23 @@ def prepare_and_sample(weights):
...
 
 ##############################################################################
 # To estimate the probability distribution over the basis states we first take ``n_shots``
 # samples and then compute the relative frequency of each outcome.
 
-samples = prepare_and_sample(w).astype(int)
-q_probs = np.bincount(samples) / n_shots
+raw_samples = prepare_and_sample(w)
 
+# Here we convert the raw samples (bit strings) into integers and count them
```
mariaschuld, line 488:
```suggestion
# convert the raw samples (bit strings) into integers and count them
```
just a suggestion

mariaschuld (approved):
Again, thanks so much for taking this on! I did not check if the demos return the correct results, and I added some suggestions to polish the comments, but they are not blocking. Other than that it looks good to me!

demonstrations/tutorial_chemical_reactions.py
```diff
@@ -297,7 +301,7 @@ def circuit(params, wires):
     prev_energy = 0.0
 
     for n in range(60):
-        params, energy = opt.step_and_cost(cost_fn, params)
+        params, energy = opt.step_and_cost(circuit, params)
         if np.abs(energy - prev_energy) < 1e-6:
             break
         prev_energy = energy
```
josh146, line 307: By switching to `expval(H)`, this demo went from 6 minutes to 1 minute execution time!

image

image

demonstrations/tutorial_falqon.py
```diff
@@ -447,11 +451,11 @@ def qaoa_circuit(params, **kwargs):
...
-def prob_circuit():
+def prob_circuit(params):
     qaoa_circuit(params)
     return qml.probs(wires=dev.wires)
 
-probs = prob_circuit()
+probs = prob_circuit(params)
 plt.bar(range(2**len(dev.wires)), probs)
 plt.xlabel("Bit string")
 plt.ylabel("Measurement Probability")
```
josh146, line 461: This one is another amazing speed improvement 😍 

image

image

josh146: @Jaybsoni oddly enough, after your changes here, the output is also _better_; the state |112> has a higher probability!

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -141,15 +141,20 @@
...
 params = np.random.normal(0, np.pi, len(singles) + len(doubles))
-print("Cost function value:", cost(params))
-print("Number of quantum evaluations:", dev.num_executions)
+with qml.Tracker(dev) as tracker:  # track the number of executions
+    print("Cost function value:", cost_circuit(params))
+
+print("Number of quantum evaluations:", tracker.totals['executions'])
 
 ##############################################################################
 # How about a larger molecule? Let's try the water molecule :download:`h2o.xyz </demonstrations/h2o.xyz>`:
```
josh146, line 160: @Jaybsoni is there a reason for leaving this at the bottom?

image

@mariaschuld, I suppose in order to change this, we need to modify `qml.expval(H)` in order to optionally support grouping when requested with `shots=None`?

demonstrations/tutorial_mol_geo_opt.py
josh146:
```suggestion
# expectation values of the Hamiltonian.
```

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -188,11 +188,12 @@ def H(x):
 # The ``hf`` array is used by the :class:`~.pennylane.BasisState` operation to initialize
```
josh146, line 188: image

image

😻

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -216,15 +217,18 @@ def circuit(params, wires):
...
 
-dev = qml.device("default.qubit", wires=6)
+num_wires = 6
+dev = qml.device("default.qubit", wires=num_wires)
 
 ##############################################################################
-# Next, we use the PennyLane class :class:`~.pennylane.ExpvalCost` to define the
-# ``cost`` function :math:`g(\theta, x)` which depends on both the circuit and the
-# Hamiltonian parameters.
+# Next, we define the `cost`` function :math:`g(\theta, x)` which depends on
```
josh146, line 224:
```suggestion
# Next, we define the ``cost`` function :math:`g(\theta, x)` which depends on
```

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -216,15 +217,18 @@ def circuit(params, wires):
...
+# both the circuit and the Hamiltonian parameters. Specifically we consider the
+# expectation values of the hamiltonian.
 
 def cost(params, x):
-    return qml.ExpvalCost(circuit, H(x), dev)(params)
+    hamiltonian = H(x)
+    run_circuit = qml.QNode(circuit, dev)
+    return run_circuit(params, obs=hamiltonian, wires=range(num_wires))
```
josh146, line 231: @Jaybsoni since it appears that the QNode is only really needed to be created once, should we instead use the decorator above,

```python
num_wires = 6
dev = qml.device("default.qubit", wires=num_wires)

@qml.qnode(dev)
def circuit(params, obs, wires):
    qml.BasisState(hf, wires=wires)
    qml.DoubleExcitation(params[0], wires=[0, 1, 2, 3])
    qml.DoubleExcitation(params[1], wires=[0, 1, 4, 5])
    return qml.expval(obs)
```
and then here, we can simply have

```python
def cost(params, x):
    hamiltonian = H(x)
    return circuit(params, obs=hamiltonian, wires=range(num_wires)
```

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -240,14 +244,13 @@ def cost(params, x):
...
-# :class:`~.pennylane.ExpvalCost` is used to evaluate the expectation value of
-# the gradient components :math:`\frac{\partial H(x)}{\partial x_i}`. This is implemented by
+# the Hamiltonian using a central-difference approximation. Then, we evaluate the expectation
+# value of the gradient components :math:`\frac{\partial H(x)}{\partial x_i}`. This is implemented by
 # the function ``grad_x``:
 
 def grad_x(x, params):
     grad_h = qml.finite_diff(H)(x)
-    grad = [qml.ExpvalCost(circuit, obs, dev)(params) for obs in grad_h]
+    grad = [qml.QNode(circuit, dev)(params, obs=obs, wires=range(num_wires)) for obs in grad_h]
```
josh146, line 253: The above suggestion would also make this line more efficient:

```python
grad = [circuit(params, obs=obs, wires=range(num_wires)) for obs in grad_h]
```

demonstrations/tutorial_qaoa_intro.py
```diff
@@ -314,16 +314,19 @@ def circuit(params, **kwargs):
 # ``params[0]`` and ``params[1]`` into each layer of the circuit. That's it! The last
```
josh146, line 314: @mariaschuld this demo is actually a good example of the slowdown incurred for QAOA using the new sparse default when `shots=None`:

Before:
image

After:
image

Makes me think we should definitely consider how to avoid the sparse pipeline when all terms commute

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -176,15 +176,12 @@ def U_C(gamma):
```
josh146, line 176: Interestingly, since we are using `shots=1` here, this demo has no speed increase/decrease :)

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -176,15 +176,12 @@ def U_C(gamma):
...
-# a measurement of multiple qubits in the computational basis, so we define
-# a Hermitian operator to do this. The eigenvalues of the operator are
-# the qubit measurement values in integer form.
+# We will need a way to sample a measurement of multiple qubits in the computational basis
+# in integer form.
 
-
-def comp_basis_measurement(wires):
-    n_wires = len(wires)
-    return qml.Hermitian(np.diag(range(2 ** n_wires)), wires=wires)
```
josh146, line 187: So glad this is finally gone!

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -176,15 +176,12 @@ def U_C(gamma):
 
 
 ##############################################################################
-# We will need a way to sample
-# a measurement of multiple qubits in the computational basis, so we define
-# a Hermitian operator to do this. The eigenvalues of the operator are
-# the qubit measurement values in integer form.
+# We will need a way to sample a measurement of multiple qubits in the computational basis
+# in integer form.
```
josh146, line 180:
```suggestion
# We will need a way to convert a bitstring, representing a sample of multiple qubits
# in the computational basis, to integer or base-10 form.
```
Perhaps this is slightly clearer?

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -218,7 +215,7 @@ def circuit(gammas, betas, edge=None, n_layers=1):
         U_B(betas[i])
     if edge is None:
         # measurement phase
-        return qml.sample(comp_basis_measurement(range(n_wires)))
+        return qml.sample()
```
josh146, line 218: 🔥

demonstrations/tutorial_vqe_qng.py
```diff
@@ -55,15 +55,19 @@ def circuit(params, wires=0):
```
josh146, line 55: image

image

💪

josh146 (changes requested):
Awesome work @Jaybsoni! I've left some very minor suggestions, once these have been considered, happy to approve 💪

demonstrations/tutorial_chemical_reactions.py
Jaybsoni: Thats amazing!

demonstrations/tutorial_falqon.py
Jaybsoni: Interesting!

demonstrations/tutorial_measurement_optimize.py
Jaybsoni: I think we decide to leave it in because the `optimize=True` keyword argument wasn't accessible through `expval(H)`? 

> Likely we'd have to keep this as is as if I recall correctly `optimize=False` is the default under the hood when using a QNode & `qml.expval(H)`.

josh146: Thanks for the reminder!

demonstrations/tutorial_mol_geo_opt.py
Jaybsoni: Ahh yes this is much cleaner. I didn't think of moving the device definition up to the circuit definition because in a few of the demos the device was example specific, but in this case its just a standard 6 wire device.

demonstrations/tutorial_mol_geo_opt.py
Jaybsoni: Got it

antalszava (approved):
:100:

josh146 (approved):
Looks great @Jaybsoni 💪

PR #331, opened 2021-09-15
Add more `plt.show` calls in the l2l demo

antalszava:
**Context**

There seems to be something going wrong with the plots that are produced by the learning 2 learn demonstration. The following are plots that are being produced when running it locally:

image

image

Multiple plots seemed to be mixed together on the outputs.

**Changes**

Adds further `plt.show` calls in the demo. This helps produce the outputs that are originally in the demo.


demonstrations/learning2learn.py
```diff
@@ -1069,7 +1072,7 @@ def call(self, inputs):
 #
 #  .. code-block:: none
 #
-#        Model: "model"
+#        Model: "functional_1"
```
antalszava, line 1075: The name seems to have been off.

albi3ro (approved):
Does seem like the demos tend to use `plt.show()`. Thanks for the update.

PR #336, opened 2021-09-16
Update spsa

rmoyard:
**Summary:**
Update the spsa tutorial, to be independent form external file and also explain how to connect to different IBM hardware backends.

### DIFF
--- a/demonstrations/spsa.py
+++ b/demonstrations/spsa.py
@@ -429,10 +429,9 @@ print(f"Device execution ratio: {grad_desc_exec_min/spsa_exec_min}.")
 
 from pennylane import qchem
 
-geometry = "h2.xyz"
-
-symbols, coordinates = qchem.read_structure(geometry)
-h2_ham, num_qubits = qchem.molecular_hamiltonian(symbols, coordinates)
+symbols = ["H", "H"]
+coordinates = np.array([0.0, 0.0, -0.6614, 0.0, 0.0, 0.6614])
+h2_ham, num_qubits = qml.qchem.molecular_hamiltonian(symbols, coordinates)
 
 # Variational ansatz for H_2 - see Intro VQE demo for more details
 def circuit(params, wires):
@@ -459,6 +458,10 @@ from qiskit.providers.aer import noise
 
 # Note: you will need to be authenticated to IBMQ to run the following code.
 # Do not run the simulation on this device, as it will send it to real hardware
+# IBMQ.save_account(TOKEN)
+# IBMQ.load_account() # Load account from disk
+# IBMQ.providers()    # List all available providers
+
 dev_melbourne = qml.device(
     "qiskit.ibmq", wires=num_qubits, backend="ibmq_16_melbourne"
 )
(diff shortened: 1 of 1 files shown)


demonstrations/spsa.py
```diff
@@ -459,6 +458,10 @@ def circuit(params, wires):
 
 # Note: you will need to be authenticated to IBMQ to run the following code.
 # Do not run the simulation on this device, as it will send it to real hardware
+# IBMQ.save_account(TOKEN)
+# IBMQ.load_account() # Load account from disk
```
antalszava, line 462: Would we need both of these statements?

demonstrations/spsa.py
```diff
@@ -459,6 +458,10 @@ def circuit(params, wires):
...
+# IBMQ.providers()    # List all available providers
```
antalszava, line 463: Would this list only the providers or the backends too?

demonstrations/spsa.py
```diff
@@ -429,9 +429,8 @@ def cost_gd(params):
 
 from pennylane import qchem
 
-geometry = "h2.xyz"
-
-symbols, coordinates = qchem.read_structure(geometry)
+symbols = ["H", "H"]
+coordinates = np.array([0.0, 0.0, -0.6614, 0.0, 0.0, 0.6614])
```
antalszava, line 433: How come this was changed?

rmoyard: Loading the file did not work on my end and I asked Soran about how to do it.

demonstrations/spsa.py
rmoyard: "Provides access to quantum devices and simulators, collectively called ‘backends’, and additional services tailored to a specific backend instance."

demonstrations/spsa.py
rmoyard: I think it is necessary to save and load, in order to run it locally.

demonstrations/spsa.py
```diff
@@ -459,6 +458,10 @@ def circuit(params, wires):
 
 # Note: you will need to be authenticated to IBMQ to run the following code.
 # Do not run the simulation on this device, as it will send it to real hardware
+# IBMQ.save_account(TOKEN)
```
antalszava, line 461:
```suggestion
# For access to IBMQ, the following statements will be useful:
# IBMQ.save_account(TOKEN)
```

demonstrations/spsa.py
antalszava:
```suggestion
# List the providers to pick an available backend:
# IBMQ.providers()    # List all available providers
```

antalszava (approved):
Looks good to me! :100:

demonstrations/spsa.py
josh146: For future reference, it might be good to instead link directly to the IBMQ documentation (https://quantum-computing.ibm.com/lab/docs/iql/manage/account/ibmq), and avoiding having code snippets here.

E.g.,

```rest
In order to set up your system for access to IBMQ, please refer to
the `IBMQ documentation <https://quantum-computing.ibm.com/lab/docs/iql/manage/account/ibmq>`__.
```

That way we don't need to worry about keeping this comment up to date 🙂

PR #337, opened 2021-09-17
Add dev version of lightning as dependency to dev branch

trbromley:
Adds the development branch of PL lightning as a dependency to the QML repo.

PL-lightning will be a dependency of the 0.18 version of PL, so we mirror the dependency here to confirm that no issues are caused in the QML repo demos.

See also https://github.com/PennyLaneAI/qml/pull/337 for the `master` branch version.


trbromley:
Closing since none of the demos currently depend on development features of lightning.

PR #360, opened 2021-10-26
Update templates access and remove template decorator

rmoyard:
**Summary:**

Replacement of how to access the templates in order to encourage the user to use `qml.Template` instead of `qml.templates.Template`.

Also remove the template decorator in the demo `tutorial_quantum_metrology.py`.


antalszava (approved):
Looks good to me! :100: Just to make sure: the functions for which `@template` was removed continue to function correctly without the decorator, correct?

rmoyard:
@antalszava I'll double check it :)

PR #363, opened 2021-10-28
Update demo checker path

antalszava:
Updates the path being used in the GitHub actions script for the demo output checker added in #356.

This update is required after a [commit in the linked PR](https://github.com/PennyLaneAI/qml/pull/356/commits/a0cf7ee2a694dfd366b0624bbb5310261b85fd51) that has changed the location of the output file. The reason for the location change was a [failure when running the check](https://github.com/PennyLaneAI/qml/runs/4027361936?check_suite_focus=true#step:12:25):

```
refusing to allow a GitHub App to create or update workflow `.github/workflows/demo_output_results/demo_diffs.md` without `workflows` permission
```


requirements.txt
```diff
@@ -1,7 +1,7 @@
 --find-links https://download.pytorch.org/whl/torch_stable.html
 appdirs==1.4.4
 autograd==1.3
-numpy==1.18.5
+numpy==1.19.2
```
antalszava, line 4: These requirements were changed to pin to the latest versions of TF and Torch. TensorFlow requires `numpy~=1.19.2`, hence that was also included here.

antalszava:
The [checks ran well](https://github.com/PennyLaneAI/qml/actions/runs/1406369094) on the latest major commit https://github.com/PennyLaneAI/qml/pull/363/commits/778d8d54ca49c40da9e8b0231ab652dc21f62c00.

The generated file resides on the `demo_output_comparison` branch [here](https://github.com/PennyLaneAI/qml/blob/demo_output_comparison/demo_checker/demo_diffs.md).

Jaybsoni (approved):
Awesome! Were we able to pin down why it takes so long to run ?

antalszava:
@Jaybsoni thanks! It runs both `build-dev` and `build-master` anew, so it's expected to run so long. An alternative solution would be to gather the latest artifacts of those two CI checks, but gathering those files would require another push (and the use of GitHub API).

PR #364, opened 2021-11-01
Requires grad updates

antalszava:
**Changes**

Updates demonstrations that emitted the warning related to:
* the `requires_grad=True` deprecation with the Autograd interface;
* `qml.metric_tensor` deprecation.

Furthermore, removed a line styling option from `demonstrations/tutorial_vqe.py` where the following warnings were originally emitted:
```
demonstrations/tutorial_vqe.py:211: UserWarning: linestyle is redundantly defined by the 'linestyle' keyword argument and the fmt string "go-" (-> linestyle='-'). The keyword argument will take precedence.
  ax1.plot(range(n + 2), energy, "go-", ls="dashed")
```
```
demonstrations/tutorial_vqe.py:222: UserWarning: linestyle is redundantly defined by the 'linestyle' keyword argument and the fmt string "go-" (-> linestyle='-'). The keyword argument will take precedence.
  ax2.plot(range(n + 2), angle, "go-", ls="dashed")
```
*Note:* the `tutorial_quantum_natural_gradient.py` and `demonstrations/tutorial_vqe_qng.py` continue to raise warnings due to how the `QNGOptimizer` works at the moment, see [related PennyLane PR](https://github.com/PennyLaneAI/pennylane/pull/1834). No further changes are needed in the demos themselves.

TODO:

* [x] tutorial_variational_classifier.py

### DIFF
--- a/demonstrations/tutorial_adaptive_circuits.py
+++ b/demonstrations/tutorial_adaptive_circuits.py
@@ -148,7 +148,7 @@ doubles_select
 
 opt = qml.GradientDescentOptimizer(stepsize=0.5)
 
-params_doubles = [0.0] * len(doubles_select)
+params_doubles = np.zeros(len(doubles_select), requires_grad=True)
 
 for n in range(20):
     params_doubles = opt.step(cost_fn, params_doubles, excitations=doubles_select)
@@ -213,7 +213,7 @@ singles_select
 
 cost_fn = qml.ExpvalCost(circuit_1, H, dev, optimize=True)
 
-params = [0.0] * len(doubles_select + singles_select)
+params = np.zeros(len(doubles_select + singles_select), requires_grad=True)
 
 gates_select = doubles_select + singles_select
 
@@ -259,8 +259,8 @@ H_sparse
 opt = qml.GradientDescentOptimizer(stepsize=0.5)
 
 excitations = doubles_select + singles_select
-params = [0.0] * len(excitations)
 
+params = np.zeros(len(excitations), requires_grad=True)
 
 @qml.qnode(dev, diff_method="parameter-shift")
 def circuit(params):
--- a/demonstrations/tutorial_chemical_reactions.py
+++ b/demonstrations/tutorial_chemical_reactions.py
@@ -106,7 +106,7 @@ hf = qml.qchem.hf_state(electrons=2, orbitals=4)
 # the equilibrium bond length, and the point where the bond is broken, which occurs when the atoms
 # are far away from each other.
 
-import numpy as np
+from pennylane import numpy as np
 
 # atomic symbols defining the molecule
 symbols = ['H', 'H']
@@ -146,7 +146,7 @@ for r in r_range:
         return qml.expval(H)  # we are interested in minimizing this expectation value
 
     # initialize the gate parameters
-    params = np.zeros(3)
+    params = np.zeros(3, requires_grad=True)
 
     # initialize with converged parameters from previous point
     if pes_point > 0:
@@ -293,7 +293,7 @@ for r in r_range:
         AllSinglesDoubles(parameters, range(qubits), hf, singles, doubles)
         return qml.expval(H)  # we are interested in minimizing this expectation value
 
-    params = np.zeros(len(singles) + len(doubles))
+    params = np.zeros(len(singles) + len(doubles), requires_grad=True)
 
     if pes_point > 0:
         params = params_old
--- a/demonstrations/tutorial_local_cost_functions.py
+++ b/demonstrations/tutorial_local_cost_functions.py
@@ -437,13 +437,7 @@ for runs in range(samples):
     print("--- New run! ---")
     has_been_trained = False
 
-    params_global = np.array(
-        [
-            [np.random.uniform(-np.pi, np.pi) for i in range(wires)],
-            [np.random.uniform(-np.pi, np.pi) for i in range(wires)],
-        ],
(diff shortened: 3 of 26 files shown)


demonstrations/tutorial_chemical_reactions.py
```diff
@@ -146,7 +146,7 @@ def circuit(parameters):
         return qml.expval(H)  # we are interested in minimizing this expectation value
 
     # initialize the gate parameters
-    params = np.zeros(3)
+    params = np.zeros(3, requires_grad=True)
```
antalszava, line 149: `requires_grad=True` was just added so that the change is more apparent: this is the default anyways when using NumPy from PennyLane.

demonstrations/tutorial_variational_classifier.py
```diff
@@ -137,9 +137,7 @@ def circuit(weights, x):
 # variables for easy use in the quantum node.
 
 
-def variational_classifier(var, x):
-    weights = var[0]
-    bias = var[1]
+def variational_classifier(weights, bias, x):
```
antalszava, line 140: This tutorial used ragged arrays internally. Now the `weights` and `bias` objects have been separated into two different trainable parameters.

demonstrations/tutorial_adaptive_circuits.py
```diff
@@ -148,7 +148,7 @@ def circuit_1(params, wires, excitations):
 
 opt = qml.GradientDescentOptimizer(stepsize=0.5)
 
-params_doubles = [0.0] * len(doubles_select)
+params_doubles = np.zeros(len(doubles_select), requires_grad=True)
```
josh146, line 151: 🙌

demonstrations/tutorial_chemical_reactions.py
josh146: @antalszava good call. We should make `requires_grad=True` explicit everywhere, since soon we will be defaulting to `requires_grad=False` by default!

demonstrations/tutorial_variational_classifier.py
josh146: 😍

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -436,10 +436,15 @@ def cost_tunable(rotations):
...
-        [np.random.uniform(-np.pi, np.pi) for i in range(wires)],
-    ]
+
+    params_global = np.array(
+        [
+            [np.random.uniform(-np.pi, np.pi) for i in range(wires)],
+            [np.random.uniform(-np.pi, np.pi) for i in range(wires)],
+        ],
+        requires_grad=True,
+    )
```
albi3ro, line 446:
```suggestion
    np.random.uniform(-np.pi, np.pi, (2, wires), requires_grad=True)
```

demonstrations/tutorial_chemical_reactions.py
antalszava: :+1: For this PR, I'd rather not increase the scope, as it mostly targetted those cases where the change was necessary as per the dev version of PennyLane.

demonstrations/tutorial_variational_classifier.py
```diff
@@ -234,22 +233,23 @@ def cost(var, X, Y):
...
+weights = weights_init
+bias = bias_init
 for it in range(25):
 
     # Update the weights by one optimizer step
     batch_index = np.random.randint(0, len(X), (batch_size,))
     X_batch = X[batch_index]
     Y_batch = Y[batch_index]
-    var = opt.step(lambda v: cost(v, X_batch, Y_batch), var)
+    weights, bias = opt.step(lambda w, b: cost(w, b, X_batch, Y_batch), weights, bias)
```
albi3ro, line 244:
```suggestion
    weights, bias, _ = opt.step(cost, weights, bias, X_batch, Y_batch)
```

Don't need to use a lambda function anymore either.

demonstrations/tutorial_variational_classifier.py
```diff
@@ -488,26 +488,27 @@ def cost(weights, features, labels):
...
+weights = weights_init
+bias = bias_init
 for it in range(60):
 
     # Update the weights by one optimizer step
     batch_index = np.random.randint(0, num_train, (batch_size,))
     feats_train_batch = feats_train[batch_index]
     Y_train_batch = Y_train[batch_index]
-    var = opt.step(lambda v: cost(v, feats_train_batch, Y_train_batch), var)
+    weights, bias = opt.step(lambda w, b: cost(w, b, feats_train_batch, Y_train_batch), weights, bias)
```
albi3ro, line 499:
```suggestion
    weights, bias, _ = opt.step(cost, weights, bias, feats_train_batch, Y_train_batch)
```

albi3ro (commented):
Just a suggestion to get rid of the `lambda` function in the optimizer while we're making changes there anyway.

demonstrations/tutorial_chemical_reactions.py
josh146: Definitely we can do this in a new PR :+1: I think it will still be somewhat important to add explicit `requires_grad=True` everywhere it is needed sooner rather than later, however, since we will soon be making NumPy arrays default to `False`.

So good to get readers into the habit of being explicit 🙂

albi3ro (approved):
Thanks for this!

antalszava:
[sc-10710]

PR #365, opened 2021-11-02
Use `approx='block-diag'` instead of `diag_approx=False`

antalszava:
Updates the `demonstrations/tutorial_vqe_qng.py` as per the [`QNGOptimizer` deprecation in the development version of PennyLane](https://github.com/PennyLaneAI/pennylane/pull/1834).


josh146 (approved):
💪

PR #367, opened 2021-11-03
Updating demos to be consistent with the new breaking changes.

Jaybsoni:
- Updated link in pytorch noise
- Removed .template when calling templates
- Need to update IBMQ decice

### Before submitting

- [x] Add a thumbnail link to your tutorial in `beginner.rst`, or if a

- [ ] All QML tutorials conform to

**Title:**
Update Demos to be up-to-date with current breaking changes

**Summary:**
Making changes to the demos in the demonstrations folder to ensure they are still useful to users today


demonstrations/pytorch_noise.py
```diff
@@ -25,7 +25,7 @@
 To follow along with this tutorial on your own computer, you will require the
 following dependencies:
 
-* The `Forest SDK <https://rigetti.com/forest>`_, which contains the quantum virtual
+* The `Forest SDK <https://qcs.rigetti.com/sdk-downloads>`_, which contains the quantum virtual
```
antalszava, line 28: The demo looks good on GPU :ok_hand:

demonstrations/quantum_volume.py
```diff
@@ -382,11 +385,20 @@ def qv_circuit_layer(num_qubits):
...
+#     qml.layer(qv_circuit_layer, m, num_qubits=m)
+#
+# print(tape.expand().draw(wire_order=dev_ideal.wires, show_all_wires=True))
+
+@qml.qnode(dev_ideal)
+def our_circuit():
+    for i in range(m):
+        qv_circuit_layer(m)
+
+    return qml.sample()
```
josh146, line 398: I might double check with @glassnotes here, I feel like there was a reason for using the low level tape in this tutorial 🤔

josh146: if it can easily be converted to QNodes, though, without any issues, I would be in favour 👍

Jaybsoni: I spoke with @glassnotes yesterday and she was able to find an elegant solution to the issue we were running into! might not have to convert to qnodes at all.

glassnotes: The reason it was done using tapes is because we need to create a random circuit, and have that exact same circuit be run on two different devices. Just using QNodes I think it wasn't possible to do this because the construction would happen within the QNode and we couldn't guarantee they are the same.

As for what Jay and I found yesterday, it is very strange - there was a problem with the `qiskit.aer` device, it was complaining about not being able to implement the `Rot` gate. This is solvable by doing `tape.expand(depth=2).execute(dev)`, but this previously wasn't an issue. @josh146 was the default expansion depth changed in the new version of PL, or could it be something with how the `aer` device performs the expansion?

josh146: That's odd, I don't think anything in the logic has changed in this release 🤔 

Having said that, I suddenly realized that `tape.execute(device)` is due to be removed, and replaced with

```python
qml.execute(list_of_tapes, device, gradient_fn)
```

In this demo, you could try making the change

```python
tape.execute(dev) -> qml.execute([tape], dev, None)
```

and see if that avoids the need for that weird expansion?

josh146: Here is the docstring from `qml.execute`: https://pennylane.readthedocs.io/en/latest/code/api/pennylane.execute.html

Jaybsoni: Awesome, I'll try that

Jaybsoni: > The reason it was done using tapes is because we need to create a random circuit, and have that exact same circuit be run on two different devices. Just using QNodes I think it wasn't possible to do this because the construction would happen within the QNode and we couldn't guarantee they are the same.
> 
> As for what Jay and I found yesterday, it is very strange - there was a problem with the `qiskit.aer` device, it was complaining about not being able to implement the `Rot` gate. This is solvable by doing `tape.expand(depth=2).execute(dev)`, but this previously wasn't an issue. @josh146 was the default expansion depth changed in the new version of PL, or could it be something with how the `aer` device performs the expansion?

This might be due to my particular machine, but even with the solution you recommended (@glassnotes), I still get a segmentation fault. It never gets through the whole doc

image

antalszava (approved):
Looks good to me! :muscle:

antalszava:
[sc-10709]

PR #369, opened 2021-11-03
Pin keras 2.6

antalszava:
Keras 2.7 was just released, however, it appears to be incompatible with TF 2.6.

This PR is meant to replicate https://github.com/PennyLaneAI/pennylane/pull/1846. TF 2.6 would by default pull in Keras 2.7, but there's a compatibility issue between the two.

See https://github.com/PennyLaneAI/qml/runs/4097045949?check_suite_focus=true#step:8:1430 for a log.


Jaybsoni (approved):
Looks good !

PR #370, opened 2021-11-03
Further updates pre-v0.19.0

antalszava:
A couple of previously missed cases where an explicit `requires_grad=True` marking is required to not raise a warning:

* Rosalin https://github.com/PennyLaneAI/qml/blob/demo_output_comparison/demo_checker/demo_diffs.md#demo7
* Local cost functions: https://github.com/PennyLaneAI/qml/blob/demo_output_comparison/demo_checker/demo_diffs.md#demo15
```
/opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/site-packages/pennylane/_grad.py:96: UserWarning: Starting with PennyLane v0.20.0, when using Autograd, inputs have to explicitly specify requires_grad=True (or the argnum argument must be passed) in order for trainable parameters to be identified.
```


albi3ro (approved):
Sorry for missing these while checking ...

antalszava:
I missed them too, no worries :slightly_smiling_face: thanks for the fast review!

PR #376, opened 2021-11-10
Add demo of error mitigation

trbromley:
**Title:** Error mitigation with Mitiq and PennyLane

**Summary:** Showcases how to do error mitigation in PL by using Mitiq as a backend.

**Relevant references:** See [here](https://4546-214003948-gh.circle-artifacts.com/0/_build/html/demos/tutorial_error_mitigation.html#references)

**Possible Drawbacks:** Some discussion to be had for [this](https://github.com/PennyLaneAI/qml/pull/376/files#r752234911) and [this](https://github.com/PennyLaneAI/qml/pull/376/files#r752236099) comment.


demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,544 @@
...
+def executor(circuit):
+    with qml.tape.QuantumTape() as circuit_with_meas:
+        for o in circuit.operations:
+            qml.apply(o)
+        qml.expval(qml.PauliZ(0))
+
+    return qml.execute([circuit_with_meas], dev_noisy, gradient_fn=None)
+
+
+execute_with_zne(circuit, executor, factory=factory, num_to_average=10)
```
trbromley, line 390: Once when running this I encountered the error:
```
mitiq.zne.inference.ExtrapolationError: The extrapolation fit failed to converge. The problem may be solved by switching to a more stable extrapolation model such as `LinearFactory`.
```
Since this only happened once, it may be to do with the user of random folding. I might come back to this and consider using a deterministic folder.

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,544 @@
...
+#
+# .. code-block:: python
+#
+#     from qiskit import IBMQ
+#     import qiskit.providers.aer.noise as noise
+#
+#     provider = IBMQ.load_account()
+#     backend = provider.get_backend('ibmq_lima')
+#
+#     noise_model = noise.NoiseModel.from_backend(backend)
```
trbromley, line 427: Just to note that all code blocks from here on are not actually executed, since we are not authenticated to IBMQ.

How do we ensure that these blocks keep working as new versions of PL and IBMQ get released?

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,545 @@
...
+
+##############################################################################
+# We need to do this step as part of the Mitiq integration with the low-level PennyLane
+# :class:`QuantumTape <pennylane.tape.QuantumTape>`. You will not have to worry about these details
+# when using the main :func:`mitigate_with_zne <pennylane.transforms.mitigate_with_zne>` function we
+# encountered earlier.
+#
+# Now, let's execute these circuits:
+
+qml.execute(folded_circuits_with_meas, dev_ideal, gradient_fn=None)
```
andreamari, line 253: This code seems equal to the body of the `executor` function defined later:

```
def executor(circuit):
    with qml.tape.QuantumTape() as circuit_with_meas:
        for o in circuit.operations:
            qml.apply(o)
        qml.expval(qml.PauliZ(0))

    return qml.execute([circuit_with_meas], dev_noisy, gradient_fn=None)
```
I wonder if we could replace this block of code by directly defining an `executor` function here.

demonstrations/tutorial_error_mitigation.py
andreamari: Yes, this is probably the case.  `fold_global` is deterministic and should fix the problem.

Otherwise switching to simpler fit methods  (`LinearFactory` or `Poly_Factory` ) could also fix the problem.

andreamari (commented):
Thanks @trbromley !  This is a great demo!

All sections are good for me. If I must find a slightly negative aspect is that
the demo is quite long and the reader may be intimidated. On the other hand it has a smooth learning curve so this is probably not big problem.

I left some comments and (optional) suggestions.

demonstrations/tutorial_error_mitigation.py
andreamari: If you are interested in simulating this without authenticating into IBMQ a trick that could be used is this:

```
from qiskit.test.mock import FakeLima
from qiskit.providers.aer.noise import NoiseModel

backend = FakeLima()
noise_model = NoiseModel.from_backend(backend)
```

I just tested this code cell and it works. But I don't know if the rest of the example will work.

trbromley: Thanks @andreamari, this is really cool! I've now updated it so that the code runs live: fd8fd6871cfa9e9118daf4acecb4f4f8baf376b6.

The only thing I'm less sure on is how realistic these mocked devices are. I had a look at `FakeLima` and `FakeBelem` which are similar devices, and we can at least establish that the fake noise models differ between the two :+1:.

trbromley (commented):
Thanks @andreamari, very useful feedback!

> If I must find a slightly negative aspect is that the demo is quite long and the reader may be intimidated

Yes that's a fair comment. I'm taking the philosophy that it's ok to have a longer demo and a shorter blog post.

andreamari (approved):
Thanks for applying all the suggestions! 
I think is a great overview of the PL-Mitiq integration and, at the same time, a nice introduction to the general idea of unitary folding and zero-noise extrapolation.

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+
+.. related::
+
+   tutorial_chemical_reactions Modelling chemical reactions
+   tutorial_noisy_circuits Explore NISQ devices
+
+*Author: Mitiq and PennyLane dev teams. Last updated: 10 November 2021*
+
+Have you ever run a circuit on quantum hardware and not quite got the result you were expecting?
+If so, welcome to the world of noisy intermediate scale quantum (NISQ) devices! These devices must
```
alvaro-at-xanadu, line 18:
```suggestion
If so, welcome to the world of noisy intermediate-scale quantum (NISQ) devices! These devices must
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+
+noise_gate = qml.PhaseDamping
+noise_strength = 0.1
+n_wires = 4
+
+dev_ideal = qml.device("default.mixed", wires=n_wires)
+dev_noisy = qml.transforms.insert(noise_gate, noise_strength)(dev_ideal)
+
+###############################################################################
+# In the above, we load a noise-free device ``dev_ideal`` and a noisy device ``dev_noisy`` which
```
alvaro-at-xanadu, line 67:
```suggestion
# In the above, we load a noise-free device ``dev_ideal`` and a noisy device ``dev_noisy``, which
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+
+##############################################################################
+# Amazing! Using PennyLane's :func:`mitigate_with_zne <pennylane.transforms.mitigate_with_zne>`
+# transform, we can create a new ``mitigated_qnode`` whose result is closer to the ideal noise-free
+# value of ``1``. How does this work?
+#
+# Understanding error mitigation
+# ------------------------------
+#
+# Error mitigation can be realized through a number of techniques and the Mitiq
```
alvaro-at-xanadu, line 154:
```suggestion
# Error mitigation can be realized through a number of techniques, and the Mitiq
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+# `documentation <https://mitiq.readthedocs.io/en/stable/>`__ is a great resource to start learning
+# more. In this demo we focus upon the zero-noise extrapolation (ZNE) method originally
```
alvaro-at-xanadu, line 156:
```suggestion
# more. In this demo, we focus upon the zero-noise extrapolation (ZNE) method originally
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+
+scale_factors = [1, 2, 3]
+folded_circuits = [fold_global(circuit, scale_factor=s) for s in scale_factors]
+
+for s, c in zip(scale_factors, folded_circuits):
+    print(f"Globally-folded circuit with a scale factor of {s}:")
+    print(c.draw())
+
+##############################################################################
+# Although these circuits are a bit deep, if you look carefully you might be able to convince
```
alvaro-at-xanadu, line 203:
```suggestion
# Although these circuits are a bit deep, if you look carefully, you might be able to convince
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+
+##############################################################################
+# Although this degradation may seem undesirable, it is part of the standard recipe for ZNE error
+# mitigation: we have a family of equivalent circuits that experience a varying amount of noise
+# when executed on hardware, and we are able to control the amount of noise by varying the folding
+# scale factor :math:`s` which determines the circuit depth. The final step is to extrapolate our
+# results back to :math:`s=0`, providing us with an estimate of the noise-free result of the
+# circuit.
+#
+# Performing extrapolation is a well-studied numeric method in mathematics and Mitiq provides
```
alvaro-at-xanadu, line 281:
```suggestion
# Performing extrapolation is a well-studied numeric method in mathematics, and Mitiq provides
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+##############################################################################
+# Since we are using the Richardson extrapolation method, the fitted function (dashed line)
+# corresponds to a polynomial interpolation of the measured data (blue points).
+#
+# The zero-noise limit corresponds to the value of the extrapolation function evaluated at `x=0`.
+#
+# Error mitigation in PennyLane
+# -----------------------------
+#
+# Now that we understand the ZNE method for error mitigation we can provide a few more details on
```
alvaro-at-xanadu, line 317:
```suggestion
# Now that we understand the ZNE method for error mitigation, we can provide a few more details on
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+# how it can be performed using PennyLane. As we have seen, the
+# :func:`mitigate_with_zne <pennylane.transforms.mitigate_with_zne>` function provides the main
+# entrypoint. This function is an example of a :doc:`circuit transform </code/qml_transforms>` and
```
alvaro-at-xanadu, line 320:
```suggestion
# entry point. This function is an example of a :doc:`circuit transform </code/qml_transforms>` and
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+# it can be applied to pre-constructed QNodes as well as being used as a decorator when constructing
+# new QNodes. For example, suppose we have a ``qnode`` already defined. A mitigated QNode can be
+# created using:
+#
+# .. code-block:: python
+#
+#     mitigated_qnode = mitigate_with_zne(scale_factors, folding, extrapolate)(qnode)
+#
+# When using ``mitigate_with_zne``, we must specify the target scale factors as well as providing
```
alvaro-at-xanadu, line 329:
```suggestion
# When using ``mitigate_with_zne``, we must specify the target scale factors as well as provide
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+for _ in range(3):
+    print(folding(circuit, scale_factor=1.1).draw())
+
+##############################################################################
+# To accommodate for this randomness, we can perform multiple repetitions of random folding for a
+# fixed :math:`s` and average over the execution results to generate the value :math:`f(s)` used
+# for extrapolation. As shown above, the number of repetitions is controlled by setting the optional
+# ``reps_per_factor`` argument.
+#
+# We conclude this section by highlighting the possibility to instead work directly with the core
```
alvaro-at-xanadu, line 370:
```suggestion
# We conclude this section by highlighting the possibility of working directly with the core
```

alvaro-at-xanadu: Or move 'instead' to the end, just doesn't flow well in my opinion :)

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+#
+# We're now ready to apply our knowledge to a more practical problem in quantum chemistry:
+# calculating the potential energy surface of molecular hydrogen. This is achieved by finding the
+# ground state energy of :math:`H_{2}` as we increase the bond length between the hydrogen atoms. As
+# shown in :doc:`this <tutorial_chemical_reactions>` tutorial, one approach to finding the ground
+# state energy is to calculate the corresponding qubit Hamiltonian and to fix an ansatz variational
+# quantum circuit that returns its expectation value. We can then vary the parameters of the
+# circuit to minimize the energy.
+#
+# To find the potential energy surface of :math:`H_{2}`, we must choose a range of inter-atomic
```
alvaro-at-xanadu, line 403:
```suggestion
# To find the potential energy surface of :math:`H_{2}`, we must choose a range of interatomic
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+# distances and calculate the qubit Hamiltonian corresponding to each distance. We then optimize the
+# variational circuit with a new set of parameters for each Hamiltonian and plot the resulting
+# energies for each distance. In this demo we compare the potential energy surface reconstructed
```
alvaro-at-xanadu, line 406:
```suggestion
# energies for each distance. In this demo, we compare the potential energy surface reconstructed
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+    :property="og:description": Learn how to mitigate quantum circuits using Mitiq and PennyLane.
+
+    :property="og:image": https://pennylane.ai/qml/_images/laptop.png
+
+.. related::
+
+   tutorial_chemical_reactions Modelling chemical reactions
+   tutorial_noisy_circuits Explore NISQ devices
+
+*Author: Mitiq and PennyLane dev teams. Last updated: 10 November 2021*
```
trbromley, line 15:
```suggestion
*Author: Mitiq and PennyLane dev teams. Last updated: 29 November 2021*
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+# ------------------------------
+#
+# Error mitigation can be realized through a number of techniques and the Mitiq
+# `documentation <https://mitiq.readthedocs.io/en/stable/>`__ is a great resource to start learning
+# more. In this demo we focus upon the zero-noise extrapolation (ZNE) method originally
+# introduced by Temme et al. [#temme2017error]_ and Li et al. [#li2017efficient]_.
+#
+# The ZNE method works by assuming that the amount of noise present when a circuit is run on a
+# noisy device is enumerated by a parameter :math:`\gamma`. Suppose we have an input circuit
+# that experiences an amount of noise equal to :math:`\gamma = \gamma_{0}` when executed.
```
alvaro-at-xanadu, line 161:
```suggestion
# that experiences an amount of noise :math:`\gamma = \gamma_{0}` when executed.
```

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,532 @@
...
+# Ideally, we would like to evaluate the result of the circuit in the :math:`\gamma = 0`
+# noise-free setting.
+#
+# To do this, we create a family of equivalent circuits whose ideal noise-free value is the
+# same as our input circuit. However, when run on a noisy device, each circuit experiences
+# an amount of noise equal to :math:`\gamma = s \gamma_{0}` for some scale factor :math:`s \ge 1`. By
```
alvaro-at-xanadu, line 167:
```suggestion
# an amount of noise :math:`\gamma = s \gamma_{0}` for some scale factor :math:`s \ge 1`. By
```

alvaro-at-xanadu (commented):
This is an excellent demo and it is very well-written.  I feel I have quickly learned a topic I knew nothing about! I would not worry about the length too much. A few suggestions to improve it would be
1. Comment code a little bit more. It can be hard to parse at times. Just one-line comments that explain the more important blocks should be enough.  More so when doing quantum chemistry, which not all readers are familiar with.
2. It ends a bit abruptly. A short paragraph summarizing what we learned and motivating the reader to learn more could work well as a closing.

trbromley:
Thanks @alvaro-at-xanadu for the great feedback!

- I have added some code comments here: feea2aa8644eb361437abce7cf7879211e76114e
- I have added a summary paragraph here: 6918fb42ab8b32b48cb409fdd00869d22211a0ec

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,570 @@
...
+from mitiq.zne.inference import RichardsonFactory
+from pennylane.transforms import mitigate_with_zne
+
+extrapolate = RichardsonFactory.extrapolate
+scale_factors = [1, 2, 3]
+
+mitigated_qnode = mitigate_with_zne(scale_factors, fold_global, extrapolate)(
+    noisy_qnode
+)
+mitigated_qnode(w1, w2)
```
CatalinaAlbornoz, line 148: When I run this I get "ModuleNotFoundError: No module named 'pennylane_qiskit'"
We clearly need the pennylane-qiskit plugin so we should say it here somewhere and show them how to install it (pip install pennylane-qiskit)

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,570 @@
...
+# To simplify this demo, we will load pre-trained parameters for our variational circuit.
+
+params = np.load("error_mitigation/params.npy")
+
+##############################################################################
+# These parameters can be downloaded by clicking
+# :download:`here <../demonstrations/error_mitigation/params.npy>`. We are now ready to set up the
+# variational circuit and run on the ideal and noisy devices:
+
+from pennylane import qchem
```
CatalinaAlbornoz, line 454: If we need to use qchem we should ask the users to install it.

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,570 @@
...
+    shots=10000,
+)
+
+##############################################################################
+# Note the use of the ``optimization_level=0`` argument when loading the noisy device. This prevents
+# the ``qiskit.aer`` transpiler from performing a pre-execution circuit optimization.
+#
+# To simplify this demo, we will load pre-trained parameters for our variational circuit.
+
+params = np.load("error_mitigation/params.npy")
```
CatalinaAlbornoz, line 447: Is there any particular reason why you added the error_mitigation folder? I feel that it sometimes confuses users because they don't realize that they have to create an additional folder.

CatalinaAlbornoz (commented):
Great tutorial. Congratulations.
I added a few comments, most of them regarding libraries that the users need to install.

alvaro-at-xanadu (commented):
Nice closing paragraph! I do agree with Catalina's comments. Also remember that you can write many line comments using """ (yes, even in restructured text, download the py file for the [https://pennylane.ai/qml/demos/tutorial_classical_shadows.html](classical shadows tutorial) too see how!)

demonstrations/tutorial_error_mitigation.py
trbromley: Good spot! I've added the details here: 27c39e335a8880d3e9344c136fb632d760d1a4c7

demonstrations/tutorial_error_mitigation.py
trbromley: > Is there any particular reason why you added the error_mitigation folder?

This was to follow the [precedent](https://github.com/PennyLaneAI/qml/tree/master/demonstrations) of other demos, which have the main `tutorial_*.py` in the `demonstrations` folder, with any additional files like images and parameters contained in a subfolder. I'd say the main reason is to keep the `demonstrations` folder less polluted, though there are so many `tutorial_*.py` files already that it might make sense for everything to go in a subfolder.

> I feel that it sometimes confuses users because they don't realize that they have to create an additional folder.

Yes I completely agree. @antalszava do you have any thoughts on how to best approach this going forward?

demonstrations/tutorial_error_mitigation.py
trbromley: Good point, I've added here: af8d7f643be7aa5850f838152767f575243fcc0a

trbromley (commented):
Thanks @CatalinaAlbornoz!

demonstrations/tutorial_error_mitigation.py
```diff
@@ -0,0 +1,592 @@
...
+# variational circuit with a new set of parameters for each Hamiltonian and plot the resulting
+# energies for each distance. In this demo, we compare the potential energy surface reconstructed
+# when the optimized variational circuits are run on ideal, noisy, and noise-mitigated devices.
+#
+# Instead of modifying the :mod:`default.mixed <pennylane.devices.default_mixed>` device to add
+# simple noise as we do above, let's choose a noise model that is a little closer to physical
+# hardware. Suppose we want to simulate the ``ibmq_lima`` hardware device available on IBMQ. We
+# can load a noise model that represents this device using:
+
+from qiskit.test.mock import FakeLima
```
CatalinaAlbornoz, line 433: We haven't asked them to install Qiskit. I would add the note.

CatalinaAlbornoz (commented):
Great improvements Tom! My only comment would be to add a note asking the users to install qiskit if they haven't already. Great demo!

demonstrations/tutorial_error_mitigation.py
trbromley: Users should already have `qiskit` installed from the warning box to install the `pennylane-qiskit` plugin (where `qiskit` is a dependency), so I think this is addressed in this commit: 27c39e335a8880d3e9344c136fb632d760d1a4c7

CatalinaAlbornoz: Oh perfect.

demonstrations/tutorial_error_mitigation.py
antalszava: I think a simple alternative would be to place the source into the same folder as the corresponding files. That should help users. Not completely sure about the exact changes required, but I think it should be doable with CircleCI too.

PR #379, opened 2021-11-15
Explicit `requires_grad` definitions

antalszava:
When creating an array using the NumPy version of PennyLane, currently `requires_grad=True` is set as a default.

Multiple demonstrations make use of this implicit setting. However, using the default implicitly comes with its drawbacks:

* It may hide that trainable parameters need to be marked using the Autograd interface (either by setting the `requires_grad` attribute or by passing the `argnum` argument to `qml.grad`/`qml.jacobian`);
* The `requires_grad=True` default is planned to be changed to `requires_grad=False` in a future release.

This PR updates demonstrations that implicitly depend on marking parameters as trainable.

### DIFF
--- a/demonstrations/tutorial_data_reuploading_classifier.py
+++ b/demonstrations/tutorial_data_reuploading_classifier.py
@@ -251,7 +251,7 @@ def density_matrix(state):
 
 label_0 = [[1], [0]]
 label_1 = [[0], [1]]
-state_labels = [label_0, label_1]
+state_labels = np.array([label_0, label_1], requires_grad=False)
 
 
 ##############################################################################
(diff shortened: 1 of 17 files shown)


antalszava:
[sc-11205]

demonstrations/tutorial_data_reuploading_classifier.py
```diff
@@ -251,7 +251,7 @@ def density_matrix(state):
 
 label_0 = [[1], [0]]
 label_1 = [[0], [1]]
-state_labels = [label_0, label_1]
+state_labels = np.array([label_0, label_1], requires_grad=False)
```
dime10, line 254: Is there a particular reason here (and other instances below) for explicitly setting `requires_grad` to False if that is the default?

dime10 (approved):
Hi @antalszava, this looks good to me!
One instance you potentially may have missed: https://github.com/PennyLaneAI/qml/blob/06d7f970dc9a97af635e393d8a318c397d41ba9a/demonstrations/tutorial_plugins_hybrid.py#L334

demonstrations/tutorial_data_reuploading_classifier.py
antalszava: Yep, just to make it explicit for users such that they get accustomed to seeing this option being used. :slightly_smiling_face:

PR #380, opened 2021-11-18
Update the general par shift demo to use `qnode_spectrum`

antalszava:


demonstrations/tutorial_general_parshift.py
```diff
@@ -232,19 +232,19 @@ def cost(x):
...
+#     The analysis tool :func:`~.pennylane.fourier.circuit_spectrum` makes use of the internal
 #     structure of the :class:`~.pennylane.QNode` that encodes the cost function.
 #     As we used the ``jax.jit`` decorator when defining the cost function above, we
-#     here need to pass the wrapped function to ``spectrum``, which is stored in
+#     here need to pass the wrapped function to ``circuit_spectrum``, which is stored in
 #     ``cost_function.__wrapped__``.
 
 
-from pennylane.fourier import spectrum
+from pennylane.fourier import circuit_spectrum
```
josh146, line 242: just pinging @dwierichs, it could be the case that the gen param shift demo can now use `qnode_spectrum` :)

josh146: or even `qml.fourier.reconstruct` 🤔

dwierichs: Yes, it could use `qnode_spectrum` but it is the same for the examples in the demo, as no classical preprocessing is used in the circuits. It comes down to whether we want to promote `circuit_spectrum` or `qnode_spectrum`.

Regarding the reconstruction I agree, we should make use of `reconstruct` now that it is (freshly) merged in :)
I can take care of this, @antalszava , if you want.

antalszava: @dwierichs sure that would be great! In this PR, I've made the change because a warning was emitted for the demo. If any further logic changes would be required, I'd appreciate further help. :slightly_smiling_face:

josh146: > Whether we use circuit_spectrum or qnode_spectrum maybe is more a matter of taste, @josh146 what would you suggest?

I have no preference 😆 whichever one leads to code that is easier to parse for pedagogical purposes.

dwierichs: In that case, maybe `qnode_spectrum` is nicer because it does not require users to dive into the gate `id` concept? (although that is a very simply concept...)
I personally also find `qnode_spectrum` a bit more intuitive in this context but that is certainly a subjective opinion.

antalszava: Hi @dwierichs, I've opened a separate PR that targets the `dev` branch and changes the use of `spectrum`: https://github.com/PennyLaneAI/qml/pull/397. Would you be able to leave a review?

antalszava:
Closing this PR for now because:
* Bumping the JAX number seems to require more exploration (issue with Torch + sphinx backreferences);
* The changes to the demo have been separated into https://github.com/PennyLaneAI/qml/pull/397.

PR #385, opened 2021-11-19
Increase cache (`dev`)

antalszava:
* Increases the cache number such that tests are re-run instead of restoring the cache.


demonstrations/tutorial_general_parshift.py
```diff
@@ -135,6 +135,10 @@ def random_observable(N, seed):
 from jax import numpy as np
 import pennylane as qml
 
+
+# NOTE: dummy change to trigger running this demo
+a = 1
+
```
antalszava, line 141:
```suggestion
```

Jaybsoni (approved):
Looks good, we are just updating the version number here correct?

PR #386, opened 2021-11-20
[WIP] - Maximum Independent Set with DQVA ansatz

pangara:
### Before submitting

- [ ] Add a thumbnail link to your tutorial in `beginner.rst`, or if a

- [ ] All QML tutorials conform to

**Title:** Maximum Independent Set with DQVA ansatz

**Summary:**

**Relevant references:** [Approaches to Constrained Quantum Approximate Optimization](https://arxiv.org/abs/2010.06660)

**Possible Drawbacks:**

**Related GitHub Issues:**


demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,793 @@
...
+# In this demo, we looked into solving a constrained combinatorial
+# optimization problem using the QAO-Ansatz and the DQVA. The QAO-Ansatz
+# defines mixer unitaries with partial mixers in such a way that we never
+# leave the set of feasible states. However, this comes at a high cost of
+# using multi-controlled Toffoli gates. To overcome this, we looked at the
+# DQVA which turned partial mixers on and off, and randomized the partial
+# mixer ordering. Furthermore, Saleem et al., recommend starting with a
+# set of initial states that are independent sets found by any classical
+# polynomial time approximation algorithm to "warm start" the
+# initialization. Furthermore, they propose the first useful application
```
anthayes92, line 777:
```suggestion
# initialization. They also propose the first useful application
```

anthayes92 (changes requested):
Amazing work @pangara! :sparkles: I certainly felt like I understand DQVA better after reading this demo!

I've left some comments and suggestions throughout and admittedly will need to come back to go through the code in more detail

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,793 @@
...
+#
+#
+#    C_{obj} = H = \sum_{i \in V} b_i,
+#
+# where
+#
+# .. math::
+#
+#
+#    b_i = \frac{1}{2} (1 - Z_i).
```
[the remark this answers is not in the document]
pangara: This is what is in the paper> The way I understood it was that Z_i is PauliZ applied to qubit i

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,793 @@
...
+#
+#    \tilde{B} = \prod_{j=1}^{l} \tilde{b}_{v_j}.
+#
+# Here, where :math:`v_j` are the neighbors and :math:`l` is the number of
+# neighbors for the :math:`i`-th node and
+#
+# .. math::
+#
+#
+#    \tilde{b}_{v_j} = \frac{1+Z_{v_j}}{2}.
```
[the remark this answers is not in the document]
pangara: Same here> The way I understood it was that Z__v_j is PauliZ applied to qubit v_j

demonstrations/tutorial_dqva_mis.py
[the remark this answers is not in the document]
angusjlowe: I think Ant is referring to the fact that Z_i is an operator, whereas the notation you've chosen for the identity operation "1" could be confused for the integer 1. I know the paper does this, but I would second Ant's suggestion to change this to "I", perhaps.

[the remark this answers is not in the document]
pangara: Ah okay, I was being cautious, but yes.. its more unambiguous this way 👍

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,793 @@
...
+        # Uncompute the ancilla
+        qml.MultiControlledX(
+            control_wires=ctrl_qubits, wires=[ancilla], control_values="0" * len(neighbors)
+        )
+
+
+######################################################################
+# We are now ready to build the circuit consisting of alternating cost and
+# mixer layers. A list of ``params`` includes parameters for the mixer
+# layer (even elements) and cost layer (odd elements).
```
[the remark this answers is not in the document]
pangara: Followed the structure of the authors here, but I can update it to two separate lists if that's easier.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,793 @@
...
+# near-term quantum computers.
+#
+# This brings us to the **Dynamic Quantum Variational Ansatz (DQVA)** that
+# maximizes the performance of the QAO-Ansatz by:
+#
+# 1. *Warm starting the optimization* by starting with an initial state that is a feasible state
+# or a superposition of feasible states. A feasible state can be found by
+# using a classical approximate polynomial-time algorithm.
+# 2. *Updating the ansatz dynamically* by turning partial mixers *on* and *off*
+# 3. *Randomizing the ordering of the partial mixers* in the mixer unitaries
```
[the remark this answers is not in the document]
pangara: I could be wrong but I think it doesn't evolve outside the set of feasible states (when I was running the algorithm, I didn't see probabilities for the infeasible states). The paper doesn't mention anything in this regard as far as I know.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,793 @@
...
+# :math:`H_C`, and a *mixer Hamiltonian*, :math:`H_M`, to evaluate the
+# objective function and a classical optimization loop updates the ansatz
+# parameters. You can learn more about this in `PennyLane's tutorial on
+# QAOA <https://pennylane.ai/qml/demos/tutorial_qaoa_intro.html>`__.
+#
+# For solving constrained combinatorial optimization problems such as MIS
+# using QAOA, constraints can be imposed in two ways:
+#
+# 1. By adding a penalty term to the problem's objective function (the Lagrange multiplier approach)
+# 2. By constructing the variational ansatz in a way such that constraints are satisfied at all times
```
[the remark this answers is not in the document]
pangara: Would it be the other way around? "unconstrained" add penalty terms to the cost Hamiltonian, and "constrained" is QAO-Ansatz style where complexity is added to the mixer Hamiltonian?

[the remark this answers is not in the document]
angusjlowe: ^What you have described is correct

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+# For solving constrained combinatorial optimization problems such as maximal independent set (MIS) 
+# using QAOA, constraints can be imposed in two ways:
+#
+# 1. By adding a penalty term to the problem's objective function (the Lagrange multiplier approach)
+# 2. By constructing the variational ansatz in a way such that constraints are satisfied at all times
+#
+# Hadfield et al., extend the QAOA framework and introduce the **Quantum
+# Alternating Operator Ansatz (QAO-Ansatz)** which alternates between more
+# general families of operators thereby supporting potentially more useful
+# sets of states by focusing on unitaries rather than Hamiltonians. For
```
angusjlowe, line 43:
```suggestion
# Hadfield et al., extend QAOA by introducing the **Quantum
# Alternating Operator Ansatz (QAO-Ansatz)** which alternates between more
# general families of unitary operators. This has the potential to narrow the focus 
# of the algorithm on a more useful set of states. For
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+# example, a mixing operator :math:`U_M (\beta)` is decomposed into a
+# sequence of partial mixers :math:`U_{M, x}` which may not correspond to
+# time evolution under a fixed mixer Hamiltonian :math:`H_M`.
```
angusjlowe, line 46:
```suggestion
# example, a mixing operator :math:`U_M (\beta)` may not correspond to
# time evolution under a fixed mixer Hamiltonian :math:`H_M`.
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+#    :width: 90%
+#
+#    ..
+#
+#    Structure of a partial mixer
+#
+# The QAO-Ansatz guarantees that the output never leaves the set of
+# feasible states (in contrast, the penalty term optimization requires an
+# additional pruning step since the output contains both feasible and
+# infeasible solutions). However, the QAO-Ansatz requires more complex
```
angusjlowe, line 59:
```suggestion
# The QAO-Ansatz can be used to guarantee that the state of the circuit never leaves the set of
# feasible states. In contrast, the penalty term approach requires an
# additional pruning step since the output can correspond to an
# infeasible solution. However, the QAO-Ansatz requires more complex
```

angusjlowe (changes requested):
Awesome work @pangara :) I just finished a pass through the first half of the tutorial and will leave additional comments soon.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+#
+#    ..
+#
+#    Formulation of the mixer unitary
+#
+# In the figure above, a PauliX rotation is applied to qubit :math:`\vert i \rangle` between two multi-controlled Toffoli gates
+# controlled by the neighbors of the :math:`i`'th vertex.
+
+def mixer_layer(beta, ancilla, mixer_order):
+
```
anthayes92, line 275:
```suggestion
def mixer_layer(beta: List, ancilla: int, mixer_order: Optional[List]=None):
      """
      Builds the QAO-Ansatz mixer layer for max independent set problem using PennyLane operations
      
      Args:
          beta (List): A list of values for the mixing parameter
          ancilla (int): The qubit to be used as the ancilla
          mixer_order (Optional[List]=None): The desired permutation of the partial mixers
          
      Returns: 
          None
      """
```

anthayes92: Note that this will require adding
```python
from typing import List, Optional
```
before calling these type hints

demonstrations/tutorial_dqva_mis.py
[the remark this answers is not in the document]
anthayes92: I think it's fine to leave as it is, was just wondering if there was an advantage I missed!

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+    nq = len(graph.nodes)
+    # Initialize
+    if init_state is None:
+        init_state = "0" * nq
+
+    else:
+        for qb, bit in enumerate(reversed(init_state)):
+            if bit == "1":
+                qml.PauliX(wires=qb)
+    assert len(params) == 2 * P, "Incorrect number of parameters!"
```
anthayes92, line 314:
```suggestion
    if len(params) == 2 * P:
      raise Exception("Incorrect number of parameters!")
```

anthayes92: Seems a bit more natural to raise an exception here

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+            x = [int(bit) for bit in list(np.binary_repr(sample, len(graph.nodes)))]
+            # Cost function is Hamming weight
+            avg_cost += probs[sample] * sum(x)
+
+        return -avg_cost
+
+    # Begin outer optimization loop
+    best_indset = init_state
+    best_init_state = init_state
+    cur_init_state = init_state
```
anthayes92, line 377:
```suggestion
    best_indset = best_init_state = cur_init_state = init_state
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+
+
+######################################################################
+# Let's run this to find the MIS!
+#
+
+base_str = "0" * len(graph.nodes)
+
+out = solve_mis_qaoa(base_str, P=1, m=4, threshold=1e-5, cutoff=1)
+print("Init string: {}, Best MIS: {}".format("".join(base_str), out[0]))
```
anthayes92, line 469:
```suggestion
print(f"Init string: {base_str}, Best MIS: {out[0]}")
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+# indicates the number of bitstrings (sorted by the highest probabilities)
+# that we consider that may improve the Hamming weight.
+#
+
+base_str = "0" * len(graph.nodes)
+for i in range(len(graph.nodes)):
+    init_str = list(base_str)
+    init_str[i] = "1"
+    out = solve_mis_qaoa("".join(init_str), P=1, m=4, threshold=1e-5, cutoff=2)
+    print("Init string: {}, Best MIS: {}".format("".join(init_str), out[0]))
```
anthayes92, line 486:
```suggestion
    print(f"Init string: {init_str}, Best MIS: {out[0]}")
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+
+        new_hamming_weight = hamming_weight(cur_init_state)
+        inner_round = 1
+
+        while inner_round < 2:
+            print(
+                "Start round {}.{}, Initial state = {}".format(
+                    mixer_round, inner_round, cur_init_state
+                )
+            )
```
anthayes92, line 392:
```suggestion
            print(
                f"Start round {mixer_round}.{inner_round}, Initial state = {cur_init_state}"
            )
```

anthayes92: Just an aesthetic preference for [f-strings](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/)

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+            best_indset, new_hamming_weight = better_strs[0]
+            best_init_state = cur_init_state
+            best_params = opt_params.copy()
+            best_perm = copy.copy(cur_permutation)
+            cur_init_state = best_indset
+            print(
+                "\tFound new independent set: {}, Hamming weight = {}".format(
+                    best_indset, new_hamming_weight
+                )
+            )
```
anthayes92, line 452:
```suggestion
            print(
                f"\tFound new independent set: {best_indset}, Hamming weight = {new_hamming_weight}"
            )
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+            # Obtain probabilites
+            probs = probability_circuit(
+                P, params=opt_params, init_state=cur_init_state, mixer_order=cur_permutation
+            )
+
+            # Sort bitstrings by decreasing probability
+            top_counts = list(
+                map(lambda x: np.binary_repr(x, len(graph.nodes)), np.argsort(probs))
+            )[::-1]
+            # print (top_counts)
```
anthayes92, line 421:
```suggestion
```

anthayes92 (changes requested):
Code is looking really well structured @pangara! Easy to follow and learn from :mortar_board: 

I've added some suggestions, mainly tidying up. One of the bigger suggestion would be to add dosctrings for some of the bigger functions. This will help users if they wanted to re-use some of your code (which I'm sure they will!). I've left and example to follow in the code.

There's a couple more function toward the end of the demo I need to come back to review

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+# may have different outputs. Therefore, the variational ansatz is defined
+# up to a permutation:
+#
+# .. math::
+#
+#
+#    U_M(\beta) \simeq \mathcal{P}(V_1(\beta)V_2(\beta)\cdots V_N(\beta))
+#
+# where P is the permutation's function of labels from :math:`1` to
+# :math:`N`.
```
angusjlowe, line 253: I know this is inherited from the paper as well, but this notation doesn't really make sense to me. (I can infer the meaning, but not ideal to have to do that.)

angusjlowe (changes requested):
Second pass, still more comments to come!

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+                    best_indset, new_hamming_weight
+                )
+            )
+            inner_round += 1
+
+        # Choose a new permutation of the mixer unitaries
+        cur_permutation = np.random.permutation(list(graph.nodes)).tolist()
+
+    print("\tRETURNING, best hamming weight:", new_hamming_weight)
+    return best_indset, best_params, best_init_state, best_perm
```
anthayes92, line 744: This function uses a lot of the same code that `solve_mis_qaoa` uses on line `363`. Again we could think abut creating smaller functions to be reused here. The above examples of helper function suggestions can be used as a guide here :smiley:  

This may also have the bonus of breaking up these mis solver functions into smaller digestible pieces

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+# trivial - any bitstring with a single "1" is a valid independent set)
+#
+
+base_str = "0" * len(graph.nodes)
+all_init_strs = []
+for i in range(len(graph.nodes)):
+    init_str = list(base_str)
+    init_str[i] = "1"
+    out = solve_mis_dqva("".join(init_str), P=1, m=4, threshold=1e-5, cutoff=1)
+    print("Init string: {}, Best MIS: {}".format("".join(init_str), out[0]))
```
anthayes92, line 760:
```suggestion
    print(f"Init string: {init_str}, Best MIS: {out[0]}")
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+
+######################################################################
+# We also run the algorithm for several initial states (instead of
+# just one) to check what is the best Maximum Independent Set we obtain.
+# The initial states are a set of independent sets of size one (this is
+# trivial - any bitstring with a single "1" is a valid independent set)
+#
+
+base_str = "0" * len(graph.nodes)
+all_init_strs = []
```
anthayes92, line 755:
```suggestion
```

anthayes92: unused variable

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+# that we consider that may improve the Hamming weight.
+#
+
+base_str = "0" * len(graph.nodes)
+for i in range(len(graph.nodes)):
+    init_str = list(base_str)
+    init_str[i] = "1"
+    out = solve_mis_qaoa("".join(init_str), P=1, m=4, threshold=1e-5, cutoff=2)
+    print(f"Init string: {init_str}, Best MIS: {out[0]}")
+    print()
```
anthayes92, line 491: This could be made into a function since it is used again after using DQVA to solve MIS on line `754`

anthayes92 (changes requested):
Final few suggestion on the code have been added, the main suggestion being to break down some of the code that gets repeated into smaller helper functions that can be re-used. 

The logic within the code is looking fantastic, amazing work @pangara :raised_hands:

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+#    :width: 80%
+#
+#    ..
+#
+#    Formulation of the mixer unitary
+#
+# In the figure above, a PauliX rotation is applied to qubit :math:`\vert i \rangle` between two multi-controlled Toffoli gates
+# controlled by the neighbors of the :math:`i`'th vertex.
+
+def mixer_layer(beta: List, ancilla: int, mixer_order: Optional[List]=None):
```
pangara, line 274:
```suggestion
def mixer_layer(beta, ancilla, mixer_order):
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+    return qml.probs(wires=range(wires - 1))
+
+
+######################################################################
+# Finally, we run the quantum-classical loop for ``m`` different mixer
+# permutations. The expectation value is also calculated in the cost
+# function ``f`` by running the probability circuit and calculating the
+# Hamming weight (which is defined as the cost). In each iteration, an
+# ansatz is constructed based on the mixer order, with optimal parameters
+# from classical minimization of the cost function using gradient descent:
```
angusjlowe, line 352:
```suggestion
# permutations. To evaluate the cost function ``f``, the expectation value
# :math:`\langle C_{obj}\rangle is obtained by running the probability circuit 
# and calculating the average Hamming weight. In each iteration, an
# ansatz is constructed based on the mixer order and the optimal parameters
# are determined by a classical minimization of the cost function using gradient descent.
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+#
+# .. math::
+#
+#
+#    \langle \beta, \gamma \vert C_{obj} \vert \beta, \gamma \rangle.
+#
```
angusjlowe, line 358: Would suggest removing this, or mentioning that the function f(\beta,\gamma) is defined to be this quantity earlier.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+    init_str[i] = "1"
+    out = solve_mis_qaoa("".join(init_str), P=1, m=4, threshold=1e-5, cutoff=2)
+    print(f"Init string: {init_str}, Best MIS: {out[0]}")
+    print()
+
+
+######################################################################
+# Dynamic Quantum Variational Ansatz
+# ----------------------------------
+# We will now formulate the MIS using the DQVA Ansatz [#Saleem2020]. The cost function
```
angusjlowe, line 495: The reference "[#Saleem2020]" doesn't seem to be working on my end.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+#
+# In the DQVA, the way mixers are defined is slightly different from the
+# QAO-Ansatz and are allowed to be independent.
+#
+# .. math::
+#
+#
+#    U_M^k(\alpha_k) = \mathcal{P}(V_1^k(\alpha_k^1)V_2^k(\alpha_k^2)\cdots V_N^k(\alpha_k^N))
+#
+# where :math:`k = 1, 2 \cdots p`.
```
angusjlowe, line 506:
```suggestion
# where :math:`k = 1, 2, \dots, p`.
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+#    \langle \alpha, \gamma \vert C_{obj} \vert \alpha, \gamma \rangle.
+#
+# The dynamic ansatz update includes:
+#
+# 1. Optimization of parameters using Gradient Descent and finding the Hamming weight with new parameters
+# 2. If the Hamming weight of the new state is larger than the initial state, the initial state gets updated to this new state
+# 3. Based on this new state, partial mixers are updated (i.e., turned off for ones and turned on for zeros)
+# 4. Steps 2 and 3 are repeated until the Hamming weight can no longer be improved
+# 5. If no new Hamming weight is obtained, the partial mixers are randomized and steps 2 and 3 are repeated to check if
+# a better Hamming weight is found. The number of randomizations is controlled via a hyperparameter.
```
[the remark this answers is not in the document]
angusjlowe: Seems so on my end as well

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,799 @@
...
+        cost_layer(gamma)
+
+
+######################################################################
+# Now, let's build the circuit that finds the probabilities of measuring
+# each bitstring.
+#
+
+
+@qml.qnode(dev)
```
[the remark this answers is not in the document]
anthayes92: Oh didn't notice! In what way is this not a real circuit?

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,805 @@
...
+#    \langle \alpha, \gamma \vert C_{obj} \vert \alpha, \gamma \rangle.
+#
+# The dynamic ansatz update includes:
+#
+# 1. Optimization of parameters using Gradient Descent and finding the Hamming weight with new parameters
+# 2. If the Hamming weight of the new state is larger than the initial state, the initial state gets updated to this new state
+# 3. Based on this new state, partial mixers are updated (i.e., turned off for ones and turned on for zeros)
+# 4. Steps 2 and 3 are repeated until the Hamming weight can no longer be improved
+# 5. If no new Hamming weight is obtained, the partial mixers are randomized and steps 2 and 3 are repeated to check if
+# a better Hamming weight is found. The number of randomizations is controlled via a hyperparameter.
```
anthayes92, line 638:
```suggestion
# 5. If no new Hamming weight is obtained, the partial mixers are randomized and steps 2 and 3 are repeated to check if
#     a better Hamming weight is found. The number of randomizations is controlled via a hyperparameter.
```

demonstrations/tutorial_dqva_mis.py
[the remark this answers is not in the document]
pangara: Do you mean that because it is run on a simulator?

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,795 @@
...
+    ind_set = []
+    for idx, bit in enumerate(bitstr):
+        if bit == "1":
+            cur_neighbors = list(G.neighbors(idx))
+            for node in ind_set:
+                if node in cur_neighbors:
+                    return False
+            else:
+                ind_set.append(idx)
+    return True
```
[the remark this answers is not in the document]
pangara: This makes a lot of sense and is much cleaner! I will use this.

demonstrations/tutorial_dqva_mis.py
pangara: I have made two helpers for this - hopefully it's better.  This is a long function!

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# parameters. You can learn more about this in `PennyLane's tutorial on
+# QAOA <https://pennylane.ai/qml/demos/tutorial_qaoa_intro.html>`__.
+#
+# For solving constrained combinatorial optimization problems such as maximal independent set (MIS) 
+# using QAOA, constraints can be imposed in two ways:
+#
+# 1. By adding a penalty term to the problem's objective function (the Lagrange multiplier approach)
+# 2. By constructing the variational ansatz in a way such that constraints are satisfied at all times
+#
+# Hadfield et al., extend QAOA by introducing the **Quantum
```
glassnotes, line 38:
```suggestion
# Hadfield et al. extend QAOA by introducing the **Quantum
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#
+# The **Quantum Approximate Optimization Algorithm (QAOA)**, a hybrid
+# quantum-classical technique due to `Farhi et
+# al. <https://arxiv.org/abs/1411.4028>`__ [#Farhi2014]_  is an approach to solving combinatorial optimization problems using quantum computers. The
+# quantum part evaluates the objective function and involves alternating between unitaries corresponding to a *cost Hamiltonian*,
+# :math:`H_C`, and a *mixer Hamiltonian*, :math:`H_M`. A classical optimization loop updates the ansatz
+# parameters. You can learn more about this in `PennyLane's tutorial on
+# QAOA <https://pennylane.ai/qml/demos/tutorial_qaoa_intro.html>`__.
+#
+# For solving constrained combinatorial optimization problems such as maximal independent set (MIS)
```
glassnotes, line 32:
```suggestion
# For solving constrained combinatorial optimization problems such as maximum independent set (MIS) 
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#
+# 1. By adding a penalty term to the problem's objective function (the Lagrange multiplier approach)
+# 2. By constructing the variational ansatz in a way such that constraints are satisfied at all times
+#
+# Hadfield et al., extend QAOA by introducing the **Quantum
+# Alternating Operator Ansatz (QAO-Ansatz)** which alternates between more
+# general families of unitary operators. This has the potential to narrow the focus 
+# of the algorithm on a more useful set of states. For
+# example, the mixing operator :math:`U_M (\beta)` can be a product
+# of partial mixers :math:`U_{M, x}(\beta)` which may not commute.
```
glassnotes, line 43: Here it would be helpful to describe, in a few words in parens, what a partial mixer is.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#    ..
+#
+#    Structure of a partial mixer
+#
+# The QAO-Ansatz can be used to guarantee that the state of the circuit never leaves the set of
+# feasible states. In contrast, the penalty term approach requires an
+# additional pruning step since the output can correspond to an
+# infeasible solution. However, the QAO-Ansatz requires more complicated
+# quantum circuits. In the case of MIS, one must apply
+# Multi-Controlled Toffoli gates that require high connectivity between
```
glassnotes, line 60:
```suggestion
# multi-controlled Toffoli gates that require high connectivity between
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# qubits, limiting the practicality of the QAO-Ansatz for large graphs on 
+# near-term quantum computers.
+#
+# This brings us to the **Dynamic Quantum Variational Ansatz (DQVA)** that
+# maximizes the performance of the QAO-Ansatz by:
+#
+# 1. *Warm starting the optimization* by starting with an initial state that is a feasible state
```
glassnotes, line 67: The rendering of this firs bullet point looks odd:

image

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+
+*Author: Priya Angara*
+
+This demo discusses the `Dynamic Quantum Variational Ansatz (DQVA) <https://arxiv.org/abs/2010.06660>`__ [#Saleem2020]_ and the `Quantum Alternating Operator Ansatz (QAO-Ansatz) <https://arxiv.org/abs/1709.03489)>`__ [#Hadfield2019]_ for constrained quantum approximate optimization in the context of solving the Maximum Independent Set problem. 
+
+"""
+
+######################################################################
+#QAO-Ansatz and DQVA
+# --------------------------
```
glassnotes, line 22:
```suggestion
# QAO-Ansatz and DQVA
# --------------------------
```
Make sure the --- line matches the line length of the above. This is not properly rendering as a section header

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#
+# Now that we have the basics in place, let's implement the QAO-Ansatz and
+# the DQVA!
+
+
+######################################################################
+# Formulating the Maximum Independent Set (MIS) problem
+# -----------------------------------------------------
+#
+# Given a graph, :math:`G = (V, E)` where :math:`V` is the set of vertices
```
glassnotes, line 81:
```suggestion
# Given a graph :math:`G = (V, E)` where :math:`V` is the set of vertices
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# .. figure:: ../demonstrations/dqva_mis/independent-sets.png
+#    :align: center
+#    :width: 100%
+#
+#    ..
+#
+#    Independent sets of a graph with the MIS highlighted.
+#
+# Here, independent sets can be represented as bitstrings
+# :math:`b = \{b_1, b_2, \cdots b_n\} \in \{0, 1\}^n`, where the i'th bit being one represents inclusion of the ith vertex in the set. This is convenient as the outputs of
```
glassnotes, line 98:
```suggestion
# :math:`b = \{b_1, b_2, \cdots b_n\} \in \{0, 1\}^n`, where the i'th bit being 1 represents inclusion of the ith vertex in the set. This is convenient as the outputs of 
```
Check throughout and make usage of i'th consistent (have seen ith and $i$th other places).

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#    C_{obj} = H = \sum_{i \in V} b_i,
+#
+# where
+#
+# .. math::
+#
+#
+#    b_i = \frac{1}{2} (I - Z_i).
+#
+# Here :math:`Z_i` is the Pauli-Z operator acting on the :math:`i`-th
```
glassnotes, line 155: Check consistency between Pauli-Z and PauliX, etc. throughout.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# qubit.
+#
+
+
+######################################################################
+# We need a couple of helper functions here: ``hamming_weight``, which
+# calculates the Hamming weight of a given bitstring and ``is_indset``,
```
glassnotes, line 162:
```suggestion
# calculates the Hamming weight of a given bitstring, and ``is_indset``,
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# which verifies whether a given bitstring is a valid independent set.
+#
+
+
+def hamming_weight(bitstr: str) -> int:
+    """Helper function to calculate the hamming weight of a bitstring"""
+    return sum([1 for bit in bitstr if bit == "1"])
+
+
+def is_indset(bitstr:str, G: nx.Graph) -> bool:
```
glassnotes, line 172: Would be good to run the code through black to ensure that the formatting is consistent.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+
+    for edge in list(G.edges):
+        if bitstr[edge[0]] == "1" and bitstr[edge[1]] == "1":
+            return False
+    return True
+
+
+######################################################################
+# Next, we define a cost unitary that incorporates the Hamming weight operator,
+# :math:`H` and is parameterized by :math:`\gamma`:
```
glassnotes, line 183:
```suggestion
# :math:`H`, and is parameterized by :math:`\gamma`:
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+
+def cost_layer(gamma: List) -> None:
+    """
+    Builds the QAO-Ansatz cost layer for max independent set problem using PennyLane operations
+    
+    Args:
+        gamma (List): A list of values for the cost parameter
+        
+    Returns: 
+        None
```
glassnotes, line 200:
```suggestion
```

glassnotes: Some of the docstrings include explicit parameter descriptions and some don't, recommend to make uniform.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# As mentioned earlier, the partial mixers may not commute with each other, i.e.,
+# :math:`[V_i, V_j] \neq 0`. Therefore, the variational ansatz is defined
+# up to a permutation:
+#
+# .. math::
+#
+#
+#    U_M(\beta) \simeq \mathcal{P}(V_1(\beta)V_2(\beta)\cdots V_n(\beta))
+#
+# where P is the permutation's function of labels from :math:`1` to
```
glassnotes, line 253:
```suggestion
# where P is the permutation's function of labels from :math:`1` to
```
Make p mathcal?

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#
+# .. figure:: ../demonstrations/dqva_mis/mixer-unitary.png
+#    :align: center
+#    :width: 60%
+#
+#    ..
+#
+#    Formulation of the mixer unitary
+#
+# In the figure above, a PauliX rotation is applied to qubit :math:`\vert i \rangle` between two multi-controlled Toffoli gates
```
glassnotes, line 272:
```suggestion
# In the figure above, an :math:`X` rotation is applied to qubit :math:`\vert i \rangle` between two multi-controlled Toffoli gates
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+    )
+
+    qml.CRX(2 * beta, wires=[ancilla, qubit])
+
+    # Uncompute the ancilla
+    qml.MultiControlledX(
+        control_wires=ctrl_qubits, wires=[ancilla], control_values="0" * len(neighbors)
+    )
+
+# We are now ready to build the mixer layer for the QAO-Ansatz by applying the
```
glassnotes, line 296:
```suggestion
######################################################################
# We are now ready to build the mixer layer for the QAO-Ansatz by applying the
```
(presumably want this to be parto f the main text?)

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+    """
+    # Permute the order of mixing unitaries
+    if mixer_order is None:
+        mixer_order = list(graph.nodes)
+
+    for qubit in list(mixer_order):
+        apply_multi_controlled_toffoli(ancilla, beta, qubit)
+
+######################################################################
+# We are now ready to build the circuit consisting of alternating cost and
```
glassnotes, line 320: Rephrase to avoid repetition with previous sentences of text

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+def initialize_ansatz(init_state: Optional[str]=None) -> None:
+    """Helper function to initialize circuit ansatz"""
+    nq = len(graph.nodes)
+
+    if init_state is None:
+        init_state = "0" * nq
+
+    else:
+        for qb, bit in enumerate(reversed(init_state)):
+            if bit == "1":
```
glassnotes, line 337: Could use `qml.BasisStatePreparation` here in lieu of the loop

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+                qml.PauliX(wires=qb)
+
+
+def qaoa_ansatz(P: int, params: Optional[List]=[], init_state: Optional[str]=None, mixer_order: Optional[List]=None) -> None:
```
glassnotes, line 341: Line is too long; reformat (running through `black` with line length of about 90 should take care of it)

demonstrations/tutorial_dqva_mis.py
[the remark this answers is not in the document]
glassnotes: I don't quite understand the comment ehre :confused:

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+    """Obtains probabilities for the QAO-Ansatz given a set of parameters, initial state, and the mixer order"""
+
+    qaoa_ansatz(P, params, init_state, mixer_order)
+    return qml.probs(wires=range(wires - 1))
+
+
+######################################################################
+# Finally, we run the quantum-classical loop for ``m`` different mixer
+# permutations. To evaluate the cost function ``f``, the expectation value
+# :math:`\langle C_{obj}\rangle is obtained by running the probability circuit
```
glassnotes, line 388:
```suggestion
# :math:`\langle C_{obj}\rangle` is obtained by running the probability circuit 
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+    optimizer = qml.GradientDescentOptimizer(stepsize=0.5)
+    cur_params = init_params.copy()
+
+    for i in range(70):
+        cur_params, opt_cost = optimizer.step_and_cost(f, cur_params)
+
+    return cur_params, opt_cost
+
+
+# We will also define a helper function that compares the independent sets
```
glassnotes, line 409:
```suggestion

######################################################################
# We will also define a helper function that compares the independent sets 
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+
+    for i in range(70):
+        cur_params, opt_cost = optimizer.step_and_cost(f, cur_params)
+
+    return cur_params, opt_cost
+
+
+# We will also define a helper function that compares the independent sets 
+# generated by the solver with a currently-known best independent set.
+# This is done by comparing the hamming weights of the bitstrings
```
glassnotes, line 411:
```suggestion
# This is done by comparing the Hamming weights of the bitstrings 
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+
+        this_hamming = hamming_weight(bitstr)
+        if is_indset(bitstr, graph) and this_hamming > best_hamming_weight:
+            better_strs.append((bitstr, this_hamming))
+
+    better_strs = sorted(better_strs, key=lambda t: t[1], reverse=True)
+    return better_strs  
+
+
+def solve_mis_qaoa(init_state: str, P: Optional[int]=1, m: Optional[int]=1, mixer_order: Optional[List]=None,
```
glassnotes, line 435: Format header and docstring here.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+            print(
+                f"\tFound new independent set: {best_indset}, Hamming weight = {new_hamming_weight}"
+            )
+
+            inner_round = inner_round + 1
+        # Choose a new permutation of the mixer unitaries
+        cur_permutation = np.random.permutation(list(graph.nodes)).tolist()
+
+    print("\tRETURNING, best hamming weight:", new_hamming_weight)
+    return best_indset, best_params, best_init_state, best_perm
```
glassnotes, line 532: This is a large amount of code and I lost track of what we were trying to do here; recommend to split into parts and have some more explanation throughout

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+        cutoff (Optional[int]=1): number of "better" independent sets to consider
+    Returns: 
+        str: Best independent set
+        np.ndarray: Best parameters
+        str: Best initial state
+        List: Best mixer order
+    """
+    # Select an ordering for the partial mixers
+    if mixer_order == None:
+        cur_permutation = np.random.permutation(list(graph.nodes)).tolist()
```
glassnotes, line 455: Set a seed for the random number generator to ensure reproducibility

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#
+
+base_str = "0" * len(graph.nodes)
+
+out = solve_mis_qaoa(base_str, P=1, m=4, threshold=1e-5, cutoff=1)
+print(f"Init string: {base_str}, Best MIS: {out}")
+print()
+
+######################################################################
+# Starting with an all-zero initial string will give us an independent
```
glassnotes, line 547: Depending on how the randomness works out this may be confusing; in my local build, the all-zero start found an MIS of size 3, but the subsequent code block, not all of them did, and I had to scroll through a lot of output. Maybe for this case it would be better to streamline the output so that it only returns the final result of the loop  (like only indicate which string was the starting string and the best MIS it found)

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+    print()
+
+######################################################################
+# Dynamic Quantum Variational Ansatz
+# ----------------------------------
+# We will now formulate the MIS using the DQVA Ansatz [#Saleem2020]_. The cost function
+# is the same as the QAO-Ansatz (the Hamming weight operator).
+#
+# In the DQVA, the way mixers are defined is slightly different from the
+# QAO-Ansatz and are allowed to be independent.
```
glassnotes, line 570:
```suggestion
# QAO-Ansatz and are allowed to be independent:
```
Independent of what?

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+#
+# .. math::
+#
+#
+#    U_M^k(\alpha_k) = \mathcal{P}(V_1^k(\alpha_k^1)V_2^k(\alpha_k^2)\cdots V_N^k(\alpha_k^N))
+#
+# where :math:`k = 1, 2, \dots, p`.
+#
+# In the DQVA, whenever the :math:`j`-th bit of the initial state is one,
```
glassnotes, line 579:
```suggestion
# In the DQVA, whenever the :math:`j`-th bit of the initial state is 1,
```

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# 1. Optimization of parameters using Gradient Descent and finding the Hamming weight with new parameters
+# 2. If the Hamming weight of the new state is larger than the initial state, the initial state gets updated to this new state
+# 3. Based on this new state, partial mixers are updated (i.e., turned off for ones and turned on for zeros)
+# 4. Steps 2 and 3 are repeated until the Hamming weight can no longer be improved
+# 5. If no new Hamming weight is obtained, the partial mixers are randomized and steps 2 and 3 are repeated to check if
+#    a better Hamming weight is found. The number of randomizations is controlled via a hyperparameter.
+#
+
+
+def solve_mis_dqva(init_state: Optional[str], P: Optional[int]=1, m: Optional[int]=1, mixer_order: Optional[List]=None, threshold: Optional[float]=1e-5, cutoff: Optional[int]=1) -> Tuple[str, np.ndarray, str, List]:
```
glassnotes, line 718: As a general comment, I had a hard time following the code blocks and outputs because there was just a lot of "stuff" in them (type hints, full docstrings, print statements, etc.) For a demo, there isn't usually full documentation unless there is something particularly unclear (e.g., parameters that weren't discussed in the text), since in a sense the prose is documenting what is being done.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,878 @@
...
+# DQVA which turned partial mixers on and off, and randomized the partial
+# mixer ordering. Furthermore, Saleem et al., recommend starting with a
+# set of initial states that are independent sets found by any classical
+# polynomial time approximation algorithm to "warm start" the
+# initialization. They also propose the first useful application
+# of `circuit-cutting techniques <https://arxiv.org/abs/2107.07532>`__ [#Saleem2021]_ to
+# solve MIS, which involves classical partitioning of a large graph into
+# sub-graphs, finding a solution using DQVA on a subgraph, and then
+# preparing a set of states to be passed as an input, essentially
+# stitching together solutions.
```
glassnotes, line 859: Nice conclusion :+1:

glassnotes (commented):
@pangara great to see this come together :tada: I've just done a quick pass, I didn't go thoroughly through all the math or code examples.

demonstrations/tutorial_dqva_mis.py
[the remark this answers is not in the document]
angusjlowe: Sorry for the unclear wording -- the unitary operation is of course fine. What I mean to say is that returning the probabilities of obtaining the various outcomes by measuring in the computational basis corresponds to estimating the (noiseless) expectations of 2^n observables. I wanted to potentially flag this up for readers so as not to give the impression that this subroutine can be executed any more efficiently than a brute force search over every possible solution to the optimization problem, even with an ideal quantum computer... Note that this magic "compute distribution" subroutine is required to execute the proposed algorithm on real hardware -- it is not an artifact of the classical simulation being performed for the demo. @glassnotes @anthayes92 did you have any thoughts about what the appropriate disclaimer might look like here?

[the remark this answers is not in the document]
angusjlowe: ^Just a follow up -- contrary to my initial comment it seems like the original paper does indeed give the pseudocode in a way that does not manipulate the probabilities directly. Perhaps a simple comment like "this circuit can only be executed on a simulator, but it is also possible to run the algorithm on real hardware" will suffice.

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,881 @@
...
+
+edges = [(0, 1), (0, 3), (1, 2), (1, 3), (3, 4)]
+graph = nx.Graph(edges)
+nx.draw(graph, with_labels=True)
+plt.show()
+
+
+######################################################################
+# We will also instantiate a device with the number of wires being one
+# more than the number of nodes (the extra wire is an ancilla).
```
anthayes92, line 130: It would be useful to introduce why we need an ancilla wire here

demonstrations/tutorial_dqva_mis.py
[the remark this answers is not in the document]
anthayes92: Thanks for clarifying @angusjlowe! This sounds like a good idea to highlight that the probabilities are being obtained in using a different method to the original paper and doing this on hardware would be expensive!

demonstrations/tutorial_dqva_mis.py
anthayes92: Ah so I actually advocated for docstrings and typehints but perhaps I was too quick to assume that this is makes things clearer for everyone! 

IMO typehints could be dropped but docstrings could be useful for re-usability (though not sure if this is the intention of demos). Happy to leave this to your best judgement.

josh146: As @glassnotes notes, we generally avoid docstrings and typehints in demos 🙂 They can be a distraction from the content and code for readers that are not familiar with Python software development, and can in some cases hinder from the flow of the text

demonstrations/tutorial_dqva_mis.py
```diff
@@ -0,0 +1,881 @@
...
+    qml.MultiControlledX(
+        control_wires=ctrl_qubits, wires=[ancilla], control_values="0" * len(neighbors)
+    )
+
+    qml.CRX(2 * beta, wires=[ancilla, qubit])
+
+    # Uncompute the ancilla
+    qml.MultiControlledX(
+        control_wires=ctrl_qubits, wires=[ancilla], control_values="0" * len(neighbors)
+    )
```
JerryChen97, line 294: It would be much better if we can update the call of `qml.MultiControlledX` to be compatible with latest interface here. See https://docs.pennylane.ai/en/stable/code/api/pennylane.MultiControlledX.html for more details.

Specifically, the following two features have been deprecated and will be soon removed:
 - string values of `control_values`
 - `control_wires`. Instead, we should specify `wires` to be control wire(s) followed by a single target wire where the operation acts on.

JerryChen97 (changes requested):

Alan-eMartin:
Hey @pangara,

As we transition our demo pipeline, we're closing PRs that haven't been updated in a while. Don't worry, the history is saved, and you can always reopen it if it's still relevant. We can help you transfer any changes to the new pipeline, or you can do it yourself once it's ready.

Feel free to reach out to the @PennyLaneAI/cloud-software team if you have any concerns.

PR #391, opened 2021-11-30
Update the backprop demo to work with the latest PL master

josh146:
Changes made include:

- removing the `templates` namespace from templates that are used
- removing the `mutable` keyword argument (doesn't exist with the new QNode)


albi3ro (approved):
Seems good to me :)

PR #393, opened 2021-11-30
Fix jax demo to work with the latest version of PennyLane

josh146:
The JAX demo uses `diff_method=None` with the JAX JIT, which is not supported on the latest version of PennyLane (setting `diff_method=None` will 'break' the JIT, as the QNode is not expecting backpropagation).

This is fixed by removing this keyword argument.


antalszava (approved):
Looks good! :tada: Thank you for fixing this. :slightly_smiling_face:

PR #396, opened 2021-12-02
`dev` branch fixes

antalszava:
**Context**

The latest [`build-dev` CI check fails due to 3 failing demonstrations](https://github.com/PennyLaneAI/qml/runs/4389087107?check_suite_focus=true#step:8:2401).

This PR addresses two of the failures, the 3rd one being an issue with the Qiskit plugin (fix upcoming).

**Changes:**

* Updates the `demonstrations/tutorial_classical_shadows.py` demo not to use a keyword syntax for passing the device;
* Pulls in the latest version of Lightning.


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: circleci/python:3.7.2
+      - image: circleci/python:3.7.12
```
antalszava, line 9: Need a higher version of `gcc` to compile Lightning. This image pulls in GCC 10.

.circleci/config.yml
```diff
@@ -30,7 +30,7 @@ jobs:
             sudo apt install pandoc -qq
             python3 -m venv venv
             . venv/bin/activate
-            pip install pip setuptools --upgrade
+            pip install pip setuptools cmake --upgrade
```
antalszava, line 33: Need a high version of `cmake`, pip installing pulls in a high enough version (as opposed to using `apt-get install`).

demonstrations/tutorial_classical_shadows.py
```diff
@@ -200,7 +200,7 @@ def calculate_classical_shadow(circuit_template, params, shadow_size, num_qubits
 dev = qml.device("default.qubit", wires=num_qubits, shots=1)
 
 # simple circuit to prepare rho
-@qml.qnode(device=dev)
```
antalszava, line 203: The new decorator was taking a `dev` argument instead of `device`. This has now been changed in PennyLane too, https://github.com/PennyLaneAI/pennylane/pull/1969, but left the change here too just to be on the safe side.

requirements.txt
```diff
@@ -11,6 +11,7 @@ sympy==1.6.2
 pillow==7.1.0
 pip==21.1
 pyscf==1.7.6
+git+https://github.com/PennyLaneAI/pennylane-lightning.git
```
antalszava, line 14: By pulling the latest version of Lightning, the related demo should pass.

Jaybsoni: So the demo was working with an older version of lightning, then due to recent changes in lightning it started failing, and now the most recent version of lightning fixes this ? Is this correct?

Jaybsoni: This should be changed to the latest version lightning once we release it correct? (after we release PL and lightning)

antalszava: > So the demo was working with an older version of lightning, then due to recent changes in lightning it started failing, and now the most recent version of lightning fixes this ? Is this correct?

The demo that's failing is new, though it was working at the time of merging it. The reason for failure was (I think) the change in the QNodes, it broke Lightning. Therefore, when pulling the latest of PennyLane, but the released version of Lightning, we get an error.

The issues were resolved in the latest version of Lightning, hence we pull it from GitHub.

> This should be changed to the latest version lightning once we release it correct? (after we release PL and lightning)

Yes. That will be relevant when merging the `dev` branch into `master` (after the v0.20.0 release of PennyLane and the plugins).

Jaybsoni (approved):
Looks good!

PR #397, opened 2021-12-03
Use `qnode_spectrum` in the general parshift demo

antalszava:
The `spectrum` function is being removed from PennyLane, so we'll change to using `qnode_spectrum`.

See more details here: https://github.com/PennyLaneAI/qml/pull/380.


rmoyard (approved):
Thank you @antalszava for changing that, it looks good to me 💯

PR #400, opened 2021-12-08
Use `plt.show()` general parshift

antalszava:
The general parameter shift tutorial includes plots generated by running the code in the tutorial.

When running the script locally, no plots appear, however. The matplotlib figures are not shown, hence they do not appear.

This PR adds in `plt.show()` calls to have the plots appear during the execution.

This addition makes the execution be "blocked" each time a plot is shown. Yet, it seems more user-friendly to show such plots by default as done with other demonstrations.

@dwierichs any thoughts?

### DIFF
--- a/demonstrations/tutorial_general_parshift.py
+++ b/demonstrations/tutorial_general_parshift.py
@@ -219,6 +219,7 @@ for ax, N, E in zip(axs, Ns, evaluated_cost):
     ax.set_xlabel("$x$")
 
 _ = axs[0].set_ylabel("$E$")
+plt.show()
 
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


Jaybsoni (approved):
Looks good, but why was this not included in the tutorial before hand?

antalszava:
@Jaybsoni good question :thinking: Maybe David will have the answer. Let me merge this though, we can revisit if there was a distinct reason.

dwierichs:
Just to make this thread complete: I had not thought too much about the local, downloadable version but mostly about the online experience, that's why I did not add the `plt.show()`s :)

PR #402, opened 2021-12-08
Fix demo diffs with PennyLane release v0.20.0

dime10:
Fixes:
- new warning emitted by [Quantum natural gradient](https://pennylane.ai/qml/demos/tutorial_quantum_natural_gradient.html)
  ```
  UserWarning: The device does not have a wire that is not used by the tape. Reverting to the block-diagonal approximation. It will often be much more efficient to request the block-diagonal approximation directly!
  ```
- use `expansion_strategy='device'` to decompose templates before drawing (as before)
- correct Hessian calculation now requires `max_diff=2` parameter to QNode

### DIFF
--- a/demonstrations/tutorial_general_parshift.py
+++ b/demonstrations/tutorial_general_parshift.py
@@ -232,19 +232,19 @@ _ = axs[0].set_ylabel("$E$")
 #
 # .. note ::
 #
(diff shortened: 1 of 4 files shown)


antalszava (approved):
LGTM, thanks! :tada:

dime10:
[sc-9645]

PR #418, opened 2022-02-01
Specify `diff_method=None` in `tutorial_jax_transformations.py`

antalszava:


requirements.txt
```diff
@@ -12,7 +12,7 @@ pillow==7.1.0
 pip==21.1
 wheel==0.37.0
 pyscf==1.7.6
-git+https://github.com/PennyLaneAI/pennylane-lightning.git
+pennylane-lightning
```
antalszava, line 15: This is temporary pinning to get this PR and the `dev` action running. Currently, 2 automatically run demos use Lightning. If all else fails, in the short term we can check those locally for correctness.

The Lightning compilation process has become heavier in terms of resource requirements and results in timeouts/failed builds. We're investigating the best approach forward but would also require the demos to be run in the short term. Hence this pinning.

albi3ro (approved):
Thanks for figuring out a workaround for the CI!

requirements.txt
josh146: Would the easiest approach be to simply have release candidate wheels available for Lightning?

PR #420, opened 2022-02-02
Fixed arxiv links in the Reference Section

ChaakuDaaku:

**Title:**
Fixed arxiv links in the Reference section
**Summary:**
Swapped the arxiv links for the two paper references in Reference section.


josh146 (approved):
Thanks @ChaakuDaaku! Good catch.

PR #421, opened 2022-02-02
added plt.show() after each plot 

Jaybsoni:
## Context: 
Add `plt.show()` after each plot to ensure that the graphs are displayed when running the script locally.


antalszava (approved):
Thank you! :tada:

PR #425, opened 2022-02-16
Add Tensor Network Quantum Circuits Demo

DSGuala:
**Title:**
Tensor-Network Quantum Circuits
**Summary:**
Explains how to simulate tensor-network quantum circuits with PennyLane using qml.MPS and qml.TTN.
**Relevant references:**
Towards Quantum Machine Learning with Tensor Networks: https://arxiv.org/abs/1803.11537
A Practical Introduction to Tensor Networks: Matrix Product States and Projected Entangled Pair States: https://arxiv.org/abs/1306.2164


demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+
+.. math::
+    d=2^{\frac{n}{2}}
+
+
+PennyLane Design
+----------------
+
+
+Import PennyLane and Numpy
```
ixfoduap, line 97: Here and throughout the demo: do not use headers such as

Build the circuit with PennyLane
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Leave this only for major section titles. Instead use regular text to guide the reader through what you are doing

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+"""
+
+import pennylane as qml
+from pennylane import numpy as np
+
+##############################################################################
+# Build the circuit with PennyLane
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ixfoduap, line 106: You could preface this better. Maybe something like saying that the first step is to define the circuit that will be used in each block

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+qml.draw_mpl(circuit)(weights)
+
+##############################################################################
+# Easy circuit editing
+# ^^^^^^^^^^^^^^^^^^^^ 
+#
+# Using the `qml.MPS` template we can easily change the block type, depth, and size. For example:
+#
+# A different block:
+# ^^^^^^^^^^^^^^^^^^
```
ixfoduap, line 139: remove

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+PennyLane Design
+----------------
+
+
+Import PennyLane and Numpy
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+"""
+
```
ixfoduap, line 100: **IMPORTANT:** The entire rest of this demo needs text to clearly explain what each part of the code is doing. Right now you are basically just adding text as if they were comments in a python file. Not good

I will henceforth review only the code itself

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+def circuit(template_weights):
+    qml.MPS(wires=range(4),n_block_wires=2,block=block,n_params_block=2,template_weights=template_weights)
+    return qml.expval(qml.PauliZ(wires=3))
+
+
+weights = np.random.random(size=[3,2])
+print(f'The circuit result is {circuit(weights):.02f}')
+print('The circuit looks like:')
+qml.drawer.use_style('black_white')
+qml.draw_mpl(circuit)(weights)
```
ixfoduap, line 130: Not sure if there's anything we can do, but the circuit image also looks really big

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+
+##############################################################################
+# Easy circuit editing
+# ^^^^^^^^^^^^^^^^^^^^ 
+#
+# Using the `qml.MPS` template we can easily change the block type, depth, and size. For example:
```
ixfoduap, line 136:
```suggestion
# Using the :class:`~.pennylane.MPS` template we can easily change the block type, depth, and size. For example:
```
Maybe it should be class:`~.pennylane.templates.MPS`. Either way, you can check other demos to see how we reference functions in the code so that in the demo, there is a built hyperlink to the function in the docs

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+#
+# A different block:
+# ^^^^^^^^^^^^^^^^^^
+
+
+def different_block(weights, wires):
```
ixfoduap, line 142: You may have too many examples. Here, the circuit is at first glance very similar to the one before. Consider skipping right to the next one

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+    :property="og:description": This demonstration explins how to simulate tensor network quantum circuits.
+    :property="og:image": https://pennylane.ai/qml/_images/thumbnail_tn_circuits.png
+
+.. related::
+
+   tutorial_plugins_hybrid Plugins and Hybrid computation
+   tutorial_gaussian_transformation Gaussian transformation
+   tutorial_state_preparation Training a quantum circuit with PyTorch
+
+*Author: PennyLane dev team. Last updated: 2 March 2022.*
```
ixfoduap, line 17: It may be good to add your name explicitly, as well as potentially Esther and Shaoming. It's a nice way to credit them and place their BMW affiliation. We would of course ask them to review the demo

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+print(f'The circuit result is {circuit(weights):.02f}')
+print('The circuit looks like:')
+qml.draw_mpl(circuit)(weights)
+
+##############################################################################
+# A deeper block
+# ^^^^^^^^^^^^^^
+
+
+def deep_block(weights, wires):
```
ixfoduap, line 165: Nice!

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+
+##############################################################################
+# Build the circuit with PennyLane
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# The following cell builds a quantum circuit in the shape of a matrix product state tensor network and prints the result.
+# 
+# Block defines a variational quantum circuit that takes the position of tensors in the network.
+
+
+def block(weights,wires):
```
ixfoduap, line 112: For this and all following circuits: **the CNOTs should not be placed at the beginning** since they do nothing when acting on the initial state |00..0>

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+    qml.MPS(wires=range(4),n_block_wires=2,block=deep_block,n_params_block=5,template_weights=template_weights)
+    return qml.expval(qml.PauliZ(wires=3))
+
+
+weights = np.random.random(size=[3,5])
+print(f'The circuit result is {circuit(weights):.02f}')
+print('The circuit looks like:')
+qml.draw_mpl(circuit)(weights)
+
+##############################################################################
```
ixfoduap, line 188: I like that you are changing different parameters to show how we can build different types of circuits. For at least on of these examples, maybe actually two, please define the circuit directly in terms of a template. Besides strongly entangling layers, I think that the SimplifiedTwoDesign could make a nice example

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+
+BAS =[[1,1,0,0],[0,0,1,1],[1,0,1,0],[0,1,0,1]]
+j=1
+for i in BAS:
+    plt.subplot(1,4,j)
+    j+=1
+    plt.imshow(np.reshape(i,[2,2]))
+
+##############################################################################
+# Then we define a tree tensor quantum circuit that will learn to label these images
```
ixfoduap, line 261:
```suggestion
# Then we define a tree tensor network quantum circuit that will learn to label these images
```

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,310 @@
...
+weights = np.random.random(size=[3,2])
+print(f'The circuit result is {circuit(BAS[0],weights):.02f}')
+print('The circuit looks like:')
+qml.draw_mpl(circuit)(BAS[0],weights)
+
+##############################################################################
+# We define a cost function and a gradient descent optimizer to train the circuit.
+
+optimizer = qml.GradientDescentOptimizer(stepsize=0.1)
+def costfunc(params):
```
ixfoduap, line 286: There has got to be a more elegant way to write this cost function. How would you do it if the dataset had millions of images?

ixfoduap (changes requested):

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+   tutorial_gaussian_transformation Gaussian transformation
+   tutorial_state_preparation Training a quantum circuit with PyTorch
+
+*Authors: PennyLane Dev Team, Diego Guala* :superscript:`1` *, Esther Cruz-Rico* :superscript:`2` *,
+Shaoming Zhang* :superscript:`2` *, Juan Miguel Arrazola* :superscript:`1` *Last updated: 9 March 2022.*
+
+.. raw:: html
+
+    <center> <sup>1</sup> Xanadu, Toronto, ON, M5G 2C8, Canada <br> <sup> 2 </sup> BMW AG, Munich, Germany</center> <br>
+
```
DSGuala, line 23: This method follows how affiliations are written in papers. But I'm not sure it looks very good here. Let me know if there are any ideas.

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+        n_params_block=2,
+        template_weights=template_weights,
+    )
+    return qml.expval(qml.PauliZ(wires=7))
+
+shapes = qml.SimplifiedTwoDesign.shape(n_layers=1, n_wires=4)
+weights = [np.random.random(size=shape) for shape in shapes]
+template_weights = [weights] * 3
+print(f"The circuit result is {circuit(template_weights):.02f}")
+fig, ax = qml.draw_mpl(circuit)(template_weights)
```
ixfoduap, line 220: This doesn't look like an MPS to me... Is there a mistake? Or am I missing something? :thinking: Either way, it is currently confusing

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+weights = np.random.random(size=[3, 2])
+print(f"The circuit result is {circuit(weights):.02f}")
+fig, ax = qml.draw_mpl(circuit)(weights)
+fig.set_size_inches((6, 3.8))
+##############################################################################
+# Classifying the bars and stripes data set
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# Next, we use a tensor-network quantum circuit to tackle a toy machine learning problem.
+# For this, we use the bars and stripes data set and optimize a parametrized circuit to label the images as
+# either *bars* or *stripes*.
```
ixfoduap, line 256:
```suggestion
# either bars or stripes.
```

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+# The data set is composed of binary black and white images of size :math:`n \times n` pixels.
+# In images that should receive the bars label, all pixels in any given column have the same color.
+# In images with the stripes label, all pixels in any given row have the same color.
+# The full data set for :math:`4\times 4` images is shown in the image below:
+#
+# .. figure:: ../demonstrations/tn_circuits/BAS.png
+#   :align: center
+#   :height: 300
+#
+# A successful quantum circuit should accept any image from the data set as input, process
```
ixfoduap, line 266: It's weird to say "a successful quantum circuit". Consider rephrasing

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+# the input and, after training, output the correct label for that input.
+# We will therefore choose a data encoding strategy that can record the input image in
+# a qubit register, a processing circuit that can analyze the data, and a final measurement
+# that can serve as a label of either "stripes" or "bars".
+#
+# The first step is to generate the bars and stripes data set.
+# For :math:`2\times 2` images, we can manually define the full data set, giving white pixels a value of 1
+# and black pixels a value of 0:
```
ixfoduap, line 274: There is a lot of white space in the image below. Try to see if there is something you can do about it

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+
+def block(weights, wires):
+    qml.RY(weights[0], wires=wires[0])
+    qml.RY(weights[1], wires=wires[1])
+    qml.CNOT(wires=wires)
+
+
+##############################################################################
+# As for the tensor-network architecture, since :class:`~pennylane.MPS` is better suited
+# for one-dimensional data [#Huggins]_,
```
ixfoduap, line 303: Don't justify the use of TTN, just state that we use it

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+optimizer = qml.GradientDescentOptimizer(stepsize=0.1)
+
+for k in range(100):
+    if k % 20 == 0:
+        print(f"Step {k}, cost: {costfunc(params)}")
+    params = optimizer.step(costfunc, params)
+
+##############################################################################
+# With the circuit trained and the parameters stored in ``params``,
+# we can now show the full circuits and the resulting output for each image input.
```
ixfoduap, line 367:
```suggestion
# we can now show the full circuits and the resulting output for each image.
```

ixfoduap (changes requested):
Looking great! After you've addressed this round of suggestions, please open for further review. I expect only minor additional suggestions

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,395 @@
...
+@qml.qnode(dev, expansion_strategy="device")
+def circuit(template_weights):
+    qml.MPS(
+        wires=range(4),
+        n_block_wires=2,
+        block=deep_block,
+        n_params_block=3,
+        template_weights=template_weights,
+    )
+    return qml.expval(qml.PauliZ(wires=3))
```
chaeyeunpark, line 171: It may be worth mentioning that the resulting circuit can be represented by MPS with bond dimension 2.

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,395 @@
...
+        template_weights=template_weights,
+    )
+    return qml.expval(qml.PauliZ(wires=7))
+
+shapes = qml.SimplifiedTwoDesign.shape(n_layers=1, n_wires=4)
+weights = [np.random.random(size=shape) for shape in shapes]
+template_weights = [weights] * 3
+fig, ax = qml.draw_mpl(circuit)(template_weights)
+
+##############################################################################
```
chaeyeunpark, line 223: The generated MPS here is slightly more complicated. As just cutting between [0,1] and [2,3] gives 4 CNOTs which generate 16 indices, one naively expect that we have bond dim=16. However, as each side of the circuit has only 2 qubits, it is possible to compress this to an MPS with bond dim=4. Mentioning all these details must confuse readers, but I would suggest saying (at least), something like "we expect higher bond dimension for the corresponding MPS of this circuit (than the above case)".

chaeyeunpark:
I also have a general comment. A large subclass of MPS is a unitary MPS and can be generated by applying a unitary gate sequentially on a qubit and a big ancilla system (whose dimension is equal to the bond dimension) (https://arxiv.org/pdf/quant-ph/0612101.pdf). However, it seems like we cannot simulate this scheme directly in Pennylane right now as we cannot mix a device with a qubit and a general `d` dimensional system (qudit). Still, as it is possible that we use `log_2(d)` qubits instead of `d` dimensional ancilla, it would be also interesting to see this workaround in any future.

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,401 @@
...
+systems in classical simulations [#orus]_.
+We can reverse this, instead designing quantum systems, e.g., quantum circuits, that follow the structure and connectivity of tensor networks.
+We call these circuits *tensor-network quantum circuits*.
+
+In tensor-network quantum circuits, the tensor network architecture acts as a
+guideline for the shape of the quantum circuit.
+More specifically, the tensors in the tensor networks above are replaced with
+unitary operations to obtain quantum circuits of the form in the following image.
+
+.. image:: ../demonstrations/tn_circuits/MPS_TTN_Circuit_Color.PNG
```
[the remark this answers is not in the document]
esthercruz: I agree. We should maybe say that every quantum circuit can be represented as a tensor network, with the bond dimension related to the depth of the circuit. Moreover, one can design the architecture of the circuit to have the same connectivity as some well known subclasses of tensor networks. Note that the connectivity of these subclasses is related to how entanglement is distributed and how correlations spread in your system. So one can try to leverage this fact by "shaping" the quantum circuit in this way, with the hope that it captures best the information you want to extract.  (a bit informal explanation but may give intuition).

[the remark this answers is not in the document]
esthercruz: Of course usually the main motivation when representing quantum circuits as tensor networks is to actually simulate them more efficiently. In that case the connectivity is chosen accordingly to how the bond dimension increases the least when applying the gates. But that's not the (main) motivation here.

Maybe one could say something like: "The motivation to represent quantum circuits as TN is two-folded. First, if the depth of the circuit is not too large, tensor networks provide a way of simulating the circuits in a much more efficient way than usual state-based simulators. Second, we can leverage the fact that tensor networks are a very successful Ansatz for quantum many-body systems where interactions are local and correlations are short range. If the states that our quantum circuit is preparing exhibit some locality in the way information is distributed, then tensor networks architectures might be advantageous with respect to more general variational architectures." (feel free to ignore this proposal or modify it to your taste).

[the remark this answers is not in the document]
esthercruz: > We should talk more about the correspondence between the TNs and the corresponding circuits. I'm not sure we're interpreting this correctly

I agree. We should maybe say that every quantum circuit can be represented as a tensor network, with the bond dimension related to the depth of the circuit. Moreover, one can design the architecture of the circuit to have the same connectivity as some well known subclasses of tensor networks. Note that the connectivity of these subclasses is related to how entanglement is distributed and how correlations spread in your system. So one can try to leverage this fact by "shaping" the quantum circuit in this way, with the hope that it captures best the information you want to extract. (a bit informal explanation but may give intuition).



Of course usually the main motivation when representing quantum circuits as tensor networks is to actually simulate them more efficiently. In that case the connectivity is chosen accordingly to how the bond dimension increases the least when applying the gates. But that's not the (main) motivation here.

Maybe one could say something like: "The motivation to represent quantum circuits as TN is two-folded. First, if the depth of the circuit is not too large, tensor networks provide a way of simulating the circuits in a much more efficient way than usual state-based simulators. Second, we can leverage the fact that tensor networks are a very successful Ansatz for quantum many-body systems where interactions are local and correlations are short range. If the states that our quantum circuit is preparing exhibit some locality in the way information is distributed, then tensor networks architectures might be advantageous with respect to more general variational architectures." (feel free to ignore this proposal or modify it to your taste).

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,395 @@
...
+can be seen in the figure below. An MPS is shown on the left and a TTN on the right.
+
+.. image:: ../demonstrations/tn_circuits/MPS_TTN_Color.PNG
+    :align: center
+    :width: 50 %
+
+These tensor networks are commonly used to efficiently represent many-body quantum
+states [#orus]_.
+We can also design quantum circuits that follow the structure and connectivity of tensor networks.
+We call these **tensor-network quantum circuits**.
```
esthercruz, line 80: > We should talk more about the correspondence between the TNs and the corresponding circuits. I'm not sure we're interpreting this correctly

I agree. We should maybe say that every quantum circuit can be represented as a tensor network, with the bond dimension related to the depth of the circuit. Moreover, one can design the architecture of the circuit to have the same connectivity as some well known subclasses of tensor networks. Note that the connectivity of these subclasses is related to how entanglement is distributed and how correlations spread in your system. So one can try to leverage this fact by "shaping" the quantum circuit in this way, with the hope that it captures best the information you want to extract. (a bit informal explanation but may give intuition).



Of course usually the main motivation when representing quantum circuits as tensor networks is to actually simulate them more efficiently. In that case the connectivity is chosen accordingly to how the bond dimension increases the least when applying the gates. But that's not the (main) motivation here.

Maybe one could say something like: "The motivation to represent quantum circuits as TN is two-folded. First, if the depth of the circuit is not too large, tensor networks provide a way of simulating the circuits in a much more efficient way than usual state-based simulators. Second, we can leverage the fact that tensor networks are a very successful Ansatz for quantum many-body systems where interactions are local and correlations are short range. If the states that our quantum circuit is preparing exhibit some locality in the way information is distributed, then tensor networks architectures might be advantageous with respect to more general variational architectures." (feel free to ignore this proposal or modify it to your taste).

demonstrations/tutorial_tn_circuits.py
DSGuala: Thanks Chae-Yeun. I'm puzzling over this.
Grouping them as in the picture I attached should lead to bond dim=2.
Temp

chaeyeunpark: bond_dim
Hi Diego, this is the way we need to cut to make it MPS. Your cut makes it a proper tensor network, but a tensor has 4 legs (instead of a maximum 3 for MPS cases).

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,395 @@
...
+specific patterns of connections between tensors and can be extended to have
+many or few indices. Examples of these architectures with only a few tensors 
+can be seen in the figure below. An MPS is shown on the left and a TTN on the right.
+
+.. image:: ../demonstrations/tn_circuits/MPS_TTN_Color.PNG
+    :align: center
+    :width: 50 %
+
+These tensor networks are commonly used to efficiently represent many-body quantum
+states [#orus]_.
```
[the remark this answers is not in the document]
DSGuala: Thanks! I went with "certain many-body quantum states".

demonstrations/tutorial_tn_circuits.py
DSGuala: I ended up adding the first paragraph of this comment. It's a great explanation of tensor-network quantum circuits and why we would use them/ what we try to do with them.

I think the explanation of why we would want to represent quantum circuits as tensor networks is not necessary in this demo since we don't discuss how to simulate quantum circuits with tensor networks.

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+
+
+###############################################################################
+# To implement this wider block, we can use the :class:`~pennylane.MPS` template
+# as before. To account for the extra wires per block, we simply set the ``n_block_wires``
+# argument to a higher number. The figure below shows the resulting circuit. Notice
+# that, in the circuit diagram, gates are left-justified. Therefore parts of later blocks
+# appear near the beginning of the circuit. Furthermore, we note that according to Ref. [#Huggins]_,
+# this circuit has a higher bond dimension than the previous ones and
+# would correspond to an MPS with a bond dimension of four.
```
chaeyeunpark, line 211: I like this wider circuit example.

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+    qml.SimplifiedTwoDesign(initial_layer_weights=weights[0], weights=weights[1], wires=wires)
+
+
+###############################################################################
+# To implement this wider block, we can use the :class:`~pennylane.MPS` template
+# as before. To account for the extra wires per block, we simply set the ``n_block_wires``
+# argument to a higher number. The figure below shows the resulting circuit. Notice
+# that, in the circuit diagram, gates are left-justified. Therefore parts of later blocks
+# appear near the beginning of the circuit. Furthermore, we note that according to Ref. [#Huggins]_,
+# this circuit has a higher bond dimension than the previous ones and
```
ixfoduap, line 210:
```suggestion
# appear near the beginning of the circuit. Furthermore,
# this circuit has a higher bond dimension than the previous ones and
```
check for spacing since I'm removing text, don't just accept the suggestion

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+    qml.RY(weights[1], wires=wires[1])
+    qml.CNOT(wires=wires)
+
+
+##############################################################################
+# As for the tensor-network architecture, we use the tree tensor-network quantum circuit.
+# We use :class:`~pennylane.BasisStatePreparation` to encode the input images.
+# The following code implements the :class:`~pennylane.BasisStatePreparation` encoding,
+# followed by a :class:`~pennylane.TTN` circuit using the above ``block``. Finally, we compute the expectation
+# value of a :class:`~pennylane.PauliZ` measurement as the output
```
ixfoduap, line 313:
```suggestion
# value of a :class:`~pennylane.PauliZ` measurement as the output.
```
Attention to details!

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+    )
+    return qml.expval(qml.PauliZ(wires=3))
+
+weights = np.random.random(size=[3, 2])
+fig, ax = qml.draw_mpl(circuit)(BAS[0], weights)
+fig.set_size_inches((6, 3.5))
+
+##############################################################################
+# When the output of the above circuit is less than zero, we label the image stripes,
+# otherwise we label it bars. Based on these labels, we define a cost function to train the circuit.
```
ixfoduap, line 337:
```suggestion
# When the output of the above circuit is less than zero, we label the image "stripes",
# otherwise we label it "bars". Based on these labels, we define a cost function to train the circuit.
```

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+for k in range(100):
+    if k % 20 == 0:
+        print(f"Step {k}, cost: {costfunc(params)}")
+    params = optimizer.step(costfunc, params)
+
+##############################################################################
+# With the circuit trained and the parameters stored in ``params``,
+# we can now show the full circuits and the resulting output for each image.
+
+for image in BAS:
```
ixfoduap, line 371: Could you mage the BAS images a bit smaller? If it's too much trouble, don't worry about it

DSGuala: :mage:

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,403 @@
...
+   tutorial_variational_classifier Variational classifier
+
+*Authors: Diego Guala* :superscript:`1` *, Esther Cruz-Rico* :superscript:`2` *,
+Shaoming Zhang* :superscript:`2` *, Juan Miguel Arrazola* :superscript:`1` *Last updated: 9 March 2022.*
+
+| :sup:`1` Xanadu, Toronto, ON, M5G 2C8, Canada
+| :sup:`2` BMW AG, Munich, Germany
+
+This demonstration explains how to use PennyLane templates to design and implement tensor-network quantum circuits
+as in Ref. [#Huggins]_. Tensor-network quantum circuits emulate the shape and connectivity of tensor networks such as matrix product states
```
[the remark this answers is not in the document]
josh146: Ah, I think this is because Sphinx < 3.0 inserts the extra square brackets I think, whereas Sphinx > 3 doesn't 🤔 or something along those lines (not entirely sure I got the version numbers correct)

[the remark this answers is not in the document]
DSGuala: Great. I added the brackets manually. Thanks Josh!

antalszava (commented):
Hi @DSGuala, a note on references: the plan would be to add a CSS rule for how references are handled in the repository (see #452) and hence a ref like `\[[#orus]_\]` could become `[#orus]_` once the relevant PR is merged.

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,408 @@
...
+   tutorial_variational_classifier Variational classifier
+
+*Authors: Diego Guala* :superscript:`1` *, Esther Cruz-Rico* :superscript:`2` *,
+Shaoming Zhang* :superscript:`2` *, Juan Miguel Arrazola* :superscript:`1` *Last updated: 17 March 2022.*
+
+| :sup:`1` Xanadu, Toronto, ON, M5G 2C8, Canada
+| :sup:`2` BMW Group, Munich, Germany
+
+This demonstration explains how to use PennyLane templates to design and implement tensor-network quantum circuits
+as in Ref. \[[#Huggins]_\]. Tensor-network quantum circuits emulate the shape and connectivity of tensor networks such as matrix product states
```
co9olguy, line 22:
```suggestion
as in Ref. [#Huggins]_. Tensor-network quantum circuits emulate the shape and connectivity of tensor networks such as matrix product states 
```

demonstrations/tutorial_tn_circuits.py
```diff
@@ -0,0 +1,408 @@
...
+These follow specific patterns of connections between tensors and can be extended to have
+many or few indices. Examples of these architectures with only a few tensors 
+can be seen in the figure below. An MPS is shown on the left and a TTN on the right.
+
+.. image:: ../demonstrations/tn_circuits/MPS_TTN_Color.PNG
+    :align: center
+    :width: 50 %
+
+These tensor networks are commonly used to efficiently represent certain many-body quantum
+states \[[#orus]_\]. Every quantum circuit can be represented as a tensor network, with the bond
```
co9olguy, line 78:
```suggestion
states [#orus]_. Every quantum circuit can be represented as a tensor network, with the bond
```

DSGuala: Thanks Nathan!

PR #429, opened 2022-02-24
remove jinja2 from requirements

albi3ro:
Test fix


josh146:
I think Jinja2 is a dependency of Sphinx!

It looks like `markupsafe` [released a new version](https://pypi.org/project/MarkupSafe/#history) this week that removes the `soft_unicode` function.

A likely fix could be adding `markupsafe==2.0.1` to the `requirements.txt` file

albi3ro:
hence why the test failed.

PR #431, opened 2022-02-26
Updated demo to avoid the soon to be deprecated qml.finite_diff func

Jaybsoni:
**Summary:**
Updated the molecular geometry optimization demo to remove dependance on the soon to be deprecated `qml.finite_diff` function


demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -243,13 +243,25 @@ def cost(params, x):
 #
 #     \nabla_x g(\theta, x) = \langle \Psi(\theta) \vert \nabla_x H(x) \vert \Psi(\theta) \rangle.
 #
-# We use the :func:`~.pennylane.finite_diff` function to compute the gradient of
+# We use the :func:`finite_diff` function to compute the gradient of
 # the Hamiltonian using a central-difference approximation. Then, we evaluate the expectation
 # value of the gradient components :math:`\frac{\partial H(x)}{\partial x_i}`. This is implemented by
 # the function ``grad_x``:
 
+def finite_diff(f, x, delta=0.01):
```
antalszava, line 251: Hi @Jaybsoni, just double-checking: could the functionality from `pennylane.gradients.finite_diff` be used here?

Jaybsoni: Hey @antalszava, the function `pennylane.gradients.finite_diff` is used to compute the gradient of a Qnode or Tape, but this finite_diff function is being used to compute the gradient of an observable which we then pass into the Qnode to compute the expectation value. Since it is static in the Qnode, we can't use this functionality to compute the gradient.

antalszava: Thanks! Maybe we could chat a bit about this. Another thing: let's target the `dev` branch with such changes, where the deprecation warning will arise.

antalszava (approved):
Looks good! :100: Thank you :)

PR #434, opened 2022-03-02
Small fix "posted on DD MMMM" demo on classical kernels

emgilfuster:
The demo on "How to approximate a classical kernel with a quantum computer" has the wrong publication date "Posted on DD MMMM 2021", I changed it to "01 Mar 2022".


josh146 (approved):
thanks @emgilfuster!

PR #437, opened 2022-03-07
Change querying matrices as per the new API

antalszava:
Addresses failures with the `dev` CI action:
https://github.com/PennyLaneAI/qml/runs/5435754352?check_suite_focus=true#step:8:2293


josh146 (approved):
Thanks Antal!

PR #438, opened 2022-03-08
Update circuit drawing output for quantum optical neural network

albi3ro:
With the new text circuit drawer, spacing is slightly different and the text wraps at a line length of 100 by default.


antalszava (approved):
:100:

PR #440, opened 2022-03-09
Collect changes to demos for the v0.22.0 release

dime10:
List of changes:

- `tutorial_vqt.py`: show matrix when drawing qnode (as before)


antalszava (approved):
:100:

PR #445, opened 2022-03-10
Update quantum volume Demo with instructions to configure IBM Q account 

Jaybsoni:
**Summary:**
Updated the demo by adding a note with instructions on how to setup an ibmq account using API token.


demonstrations/quantum_volume.py
```diff
@@ -538,6 +538,19 @@ def heavy_output_set(m, probs):
...
 # .. note::
 #
+#    In order to access the IBM Q backend, users must have an IBM Q account
+#    configured. This can be done by running:
+#
+#         .. code-block:: python3
+#
+#             from qiskit import IBMQ
+#             IBMQ.save_account('MY_API_TOKEN')
+#
```
dime10, line 548:
```suggestion
#
#    Alternatively, the API token can be passed directly to the device function:
#
#         .. code-block:: python3
#
#             qml.device("qiskit.ibmq", wires=5, backend="imbq_bogota", ibmqx_token="MY_API_TOKEN")
#
```

dime10 (approved):
Thanks @Jaybsoni, this should very helpful :)

demonstrations/quantum_volume.py
```diff
@@ -538,6 +538,25 @@ def heavy_output_set(m, probs):
...
+#    Alternatively, the API token can be passed directly to the device function:
+#
+#         .. code-block:: python3
+#
+#             qml.device("qiskit.ibmq", wires=5, backend="imbq_bogota", ibmqx_token="MY_API_TOKEN")
+#
```
Jaybsoni, line 554:
```suggestion
```

Jaybsoni: The note looks abnormal in the docs, and we don't really need to provide all this information.

PR #452, opened 2022-03-22
Add CSS rule to enclose references with extra pair of square brackets

antalszava:
**Context**

With https://github.com/PennyLaneAI/qml/pull/427, the QML repository has started using a Sphinx version higher than 2.0.0.

With 2.0.0, however, Sphinx has changed how in-line references are rendered.

An inline reference that used to render as `[1]` renders `1` with Sphinx v0.2.0 or higher.

To make up for the difference, the authors of demonstrations would have to add an extra pair of square brackets around references. Instead, a CSS rule is added in this PR to automatically add such a pair of square brackets.

#### Inline
image (14)

#### Footnote
Screenshot from 2022-03-22 16-03-34

## Previous rendering (desired)
Screenshot from 2022-03-22 14-16-22

#### Footnote
Screenshot from 2022-03-22 16-04-15

**Changes**

Adds a CSS rule to enclose the affected bracket classes with another pair of square brackets.

### DIFF
--- a/demonstrations/tutorial_sc_qubits.py
+++ b/demonstrations/tutorial_sc_qubits.py
@@ -12,28 +12,28 @@ Quantum computing with superconducting qubits
 
 *Author: PennyLane dev team. Posted: 22 March 2022. Last updated: 22 March 2022.*
 
-**Superconducting qubits** are among the most promising approaches to building quantum computers. 
-It is no surprise that this technology is being used by well-known tech companies in their quest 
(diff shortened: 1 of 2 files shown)


AndrewGardhouse (approved):
Brackets are rendering 🎉 
Screen Shot 2022-03-22 at 4 51 27 PM

antalszava:
[sc-16444]

PR #321, opened 2021-08-17
Demo on how to emulate classical kernels

emgilfuster:
This demo introduces a re-uploading in depth quantum embedding kernel and demonstrates how it can approximate any stationary classical kernel up to arbitrary precision.

PR #453, opened 2022-03-22
Update square brackets in the superconducting demo

antalszava:
Amends #452 such that the superconducting demo uses the old reference syntax along with the CSS rule introduced.

Note: another demonstration https://github.com/PennyLaneAI/qml/pull/321/ was added after the change in Sphinx version (https://github.com/PennyLaneAI/qml/pull/427), though that demonstration doesn't require changes.


rmoyard (approved):
Thanks @antalszava , it looks good to me 💯

PR #465, opened 2022-04-19
Remove `get_unitary_matrix` from  sc qubits demo

rmoyard:
**Summary:**
`get_unitary_matrix` was deprecated, therefore we replace by `qml.matrix`in this demo.


antalszava (approved):
:tada:

PR #467, opened 2022-04-20
Fix QChem related demo errors, get latest Lightning & pin Cirq

antalszava:
Fixes issues that block the `build-dev` action from passing.

1. QChem related changes:
* `demonstrations/tutorial_measurement_optimize.py`: updates the number of terms mentioned as per using the DHF solver;
* Adds `openfermionpyscf==0.5` and `h5py<=3.2.1` to the `requirements.txt` file such that `tutorial_quantum_chemistry` and `tutorial_adaptive_circuits` continue working.

Note the related PR in PennyLane: https://github.com/PennyLaneAI/pennylane/pull/2465

2. Lightning requirement change: recent changes of PennyLane-Lightning are required to run the checks, so pulls Lightning from Test PyPi.
3. Cirq: requirements of Cirq are pinned such that a working version is guaranteed ([failures arose otherwise](https://app.circleci.com/pipelines/github/PennyLaneAI/qml/4879/workflows/32e6e93c-cfbe-418b-b5d5-7c7bc766dd28/jobs/5223)).

**More details**
Updates the measurement optimization tutorial. The switch from using `pyscf` to using the differentiable HF pipeline results in using a different cutoff value and more Hamiltonian terms are being considered. This change is captured by the updates in this PR. Closes https://github.com/PennyLaneAI/qml/issues/468.


.circleci/config.yml
```diff
@@ -91,6 +91,7 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             pip install pip setuptools cmake --upgrade
+            pip install -i https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade
```
antalszava, line 94: This should help make `tutorial_quantum_gans.py` pass (no import error).

demonstrations/tutorial_measurement_optimize.py
```diff
@@ -776,7 +776,7 @@ def circuit(weights, group=None, **kwargs):
 print("Number of required measurements after optimization:", len(groups))
 
 ##############################################################################
-# We went from 2050 required measurements/circuit evaluations to 523 (just over *two thousand*
+# We went from 2110 required measurements/circuit evaluations to 556 (just over *two thousand*
```
josh146, line 779: woah @antalszava any reason why this is the case? The number of executions shouldn't have increased 🤔

antalszava: After a chat with @soranjh, this seems to be due to a different cutoff value used by `molecular_hamiltonian` when using `method="dhf"` (new default) vs. `method="pyscf"` (previously default).

josh146: Thanks for checking this @antalszava!

.circleci/config.yml
```diff
@@ -14,9 +14,9 @@ jobs:
       - restore_cache:
           keys:
             # when lock file changes, use increasingly general patterns to restore cache
-            - pip-v34c-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements_no_deps.txt" }}
-            - pip-v34c-{{ .Branch }}-
-            - pip-v34c-
+            - pip-v38c-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "requirements_no_deps.txt" }}
```
antalszava, line 17: This is meant to increase the cache of the CircleCI build such that it is built from scratch.

demonstrations/tutorial_adaptive_circuits.py
```diff
@@ -80,7 +80,8 @@
     symbols,
     geometry,
     active_electrons=2,
-    active_orbitals=5
+    active_orbitals=5,
+    method='pyscf'
```
soranjh, line 84: I think this tutorial should run with the default method ('dhf') after merging [2465](https://github.com/PennyLaneAI/pennylane/pull/2465).

requirements.txt
```diff
@@ -11,9 +11,12 @@ sympy==1.6.2
 pillow==7.1.0
 pip==21.1
 wheel==0.37.0
+h5py<=3.2.1
 pyscf==1.7.6
-pennylane-lightning
-git+https://github.com/PennyLaneAI/pennylane.git
+cirq-core==0.14.1
+cirq-pasqal==0.14.1
+openfermionpyscf==0.5   # required by 2 tutorials: tutorial_quantum_chemistry and tutorial_adaptive_circuits
```
soranjh, line 18: Minor point: the latest version should be fine.

requirements.txt
```diff
@@ -11,9 +11,12 @@ sympy==1.6.2
 pillow==7.1.0
 pip==21.1
 wheel==0.37.0
+h5py<=3.2.1
 pyscf==1.7.6
```
soranjh, line 15: Minor point: `openfermionpyscf` installs these automatically, latest versions should be fine.

requirements.txt
antalszava: This seems to be the latest version :+1: It is pinned strictly (i.e., by having `openfermionpyscf==0.5` instead of `openfermionpyscf`) such that we hit no unexpected behaviour for example if `openfermionpyscf` 0.6 was released.

requirements.txt
antalszava: Pinned `h5py` here because an error seems to have come up:
https://app.circleci.com/pipelines/github/PennyLaneAI/qml/4877/workflows/356b013d-efc9-40e0-bfe8-14c2195d4912/jobs/5221?invite=true#step-106-460

demonstrations/tutorial_adaptive_circuits.py
antalszava: Can confirm that this is the case after a local run. :+1:

soranjh (approved):

antalszava:
[sc-17245]

Issue #470, opened 2022-04-20
[FEATURE] Revisit the Quantum volume demo


antalszava:
The quantum volume demonstration and its results are device-specific. It could be revisit it and see how it could be remade to make it independent of the hardware device being used.

![image](https://user-images.githubusercontent.com/24476053/164323910-365cdb27-af86-46b1-a4d7-8f969f51350c.png)

PR #471, opened 2022-04-20
Update quantum volume demonstration as per drawing updates

antalszava:
`qml.draw` takes a `show_matrices` argument has been updated to be set to `False` by default.

**Changes**
Explicitly passes `show_matrices=True` to the `qml.draw` call and updates the matrix labels. Note, that these changes are akin to a possible output, rather than the real output of the demonstration.

The output of the demo is device-specific (see the related issue here https://github.com/PennyLaneAI/qml/issues/470).

### DIFF
--- a/demonstrations/quantum_volume.py
+++ b/demonstrations/quantum_volume.py
@@ -385,7 +385,7 @@ m = 3  # number of qubits
 with qml.tape.QuantumTape() as tape:
     qml.layer(qv_circuit_layer, m, num_qubits=m)
 
-print(tape.expand(stop_at=lambda op: isinstance(op, qml.QubitUnitary)).draw(wire_order=dev_ideal.wires, show_all_wires=True))
+print(tape.expand(stop_at=lambda op: isinstance(op, qml.QubitUnitary)).draw(wire_order=dev_ideal.wires, show_all_wires=True, show_matrices=True))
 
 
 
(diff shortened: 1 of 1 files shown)


demonstrations/quantum_volume.py
```diff
@@ -385,7 +385,7 @@ def qv_circuit_layer(num_qubits):
 with qml.tape.QuantumTape() as tape:
     qml.layer(qv_circuit_layer, m, num_qubits=m)
 
-print(tape.expand(stop_at=lambda op: isinstance(op, qml.QubitUnitary)).draw(wire_order=dev_ideal.wires, show_all_wires=True))
+print(tape.expand(stop_at=lambda op: isinstance(op, qml.QubitUnitary)).draw(wire_order=dev_ideal.wires, show_all_wires=True, show_matrices=True))
```
albi3ro, line 388:
```suggestion
expanded_tape = tape.expand(stop_at=lambda op: isinstance(op, qml.QubitUnitary))
print(qml.drawer.tape_text(expanded_tape, wire_order=dev_ideal.wires, show_all_wires=True, show_matrices=True))
```

I personally think we should favour the use of `qml.drawer.tape_text` over `tape.draw`.  This also splits the process into two lines, which may be more readable.

antalszava:
Thank you for the suggestion! :heart_eyes: @albi3ro Indeed looks better.

albi3ro (approved):

PR #473, opened 2022-04-21
Make hamiltonian coeffs non-differentiable in VQE QNG

soranjh:
The PR changes the Hamiltonian coefficients to non-differentiable parameters in `tutorial_vqe_qng` to prevent an error with the QNG optimizer used in VQE optimizations.


antalszava (approved):
:tada:

PR #481, opened 2022-04-25
Update build to large resource class in config.yml

josh146:


.circleci/config.yml
```diff
@@ -77,6 +79,8 @@ jobs:
 
       - store_artifacts:
           path: _build/test-results
+    # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass
+    resource_class: large
```
josh146, line 83: @smokingroosters

PR #485, opened 2022-05-04
Demo on measurement-based quantum computation

soosub:
### Before submitting

- [x] Ensure that your tutorial executes correctly, and conforms to the

- [x] Add a thumbnail link to your tutorial in `beginner.rst`, or if a

- [x] All QML tutorials conform to

**Title:** Measurement-based quantum computation

**Summary:** A demo on measurement-based quantum computation

**Relevant references:**
- https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.86.5188
- https://quantum-journal.org/papers/q-2021-02-04-392/

**Possible Drawbacks:**
None

**Related GitHub Issues:**
None


soosub:
Resources:
- https://phas.ubc.ca/~raussen/Supplements/VideoLectures.html
- https://arxiv.org/pdf/1711.04921.pdf

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,277 @@
...
+
+code_distance = 3
+RHG = SurfaceCode(code_distance)
+# RHG.draw(backend="plotly") #TODO: uncomment this line after merging #103
+
+##############################################################################
+#
+# .. raw:: html
+#    :file: ../demonstrations/mbqc/rhg-graph.html
+#
```
soosub, line 230: Ideally I automatically generate the html instead of uploading it separately

soosub:
Can someone add @isschoch and @mauriceweber as reviewers here?

isschoch:
Cannot wait to learn some cool stuff about MBQC!

demos_quantum-computing.rst
```diff
@@ -83,6 +83,13 @@ such as benchmarking and characterizing quantum processors.
     :figure: demonstrations/photonics/photonics_tn.png
     :description: :doc:`demos/tutorial_photonics`
     :tags: quantumcomputing
+
+.. customgalleryitem::
+    :tooltip: Measurement-based quantum computation
+    :figure: demonstrations/mbqc/mbqc_blueprint.png
+    :description: :doc:`demos/tutorial_mbqc`
+    :tags: quantumcomputing
```
soosub, line 91: Somehow the gallery looks weird here. I tried to resolve it but have not found a solution yet.

image

requirements.txt
```diff
@@ -46,3 +46,7 @@ pyyaml==5.4.1
 jax==0.2.10
 jaxlib==0.1.64
 neural-tangents==0.3.6
+flamingpy==0.9.0b0
+plotly==5.9.0
+ipython==7.4.0
+nbformat>=4.2.0
```
soosub, line 52: Some of these might be removed when the demo is finalised

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,450 @@
...
+def one_bit_teleportation(input_state):
+    # Prepare input state
+    qml.QubitStateVector(input_state, wires=0)
+    qml.Hadamard(wires=1)
+
+    qml.CNOT(wires=[1, 0])
+    
+    # Measure the first qubit and apply an X-gate conditioned on the outcome
+    m = qml.measure(wires = [0])
+    qml.cond(m == 1, qml.PauliX)(wires=1)
```
soosub, line 147: Just a note that the circuit is not the same as the figure @radras

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,450 @@
...
+        qml.Hadamard(wires=[node])
+
+    for edge in G.edges:
+        i, j = edge
+        qml.CZ(wires=[str(i), str(j)])
+
+    return qml.state()
+
+
+print(qml.draw(cluster_state)())
```
soosub, line 96: To make the visualisation of the circuit more aesthetic, we can have a look at the comments in [this thread on slack](https://xanaduhq.slack.com/archives/C036VDB901W/p1653595712897369)

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,461 @@
...
+#
+#    ..
+#
+
+##############################################################################
+# Arbitrary quantum circuits
+# ```````````````````````````
+# Once we have established the ability to implement arbitrary single-qubit rotations and a two-qubit
+# gate, the CNOT, the final step is to show that we can implement arbitrary quantum circuits. To do
+# so, we simply have to note that we have a *universal gate set* [#DiVincenzo]_. However, you might wonder - how
```
soosub, line 293: I want to reference the DiVincenzo criteria paper here, still need to do that

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,461 @@
...
+#
+# Conclusion and further reading
+# -------------------------------
+#
+# Xanadu's approach toward a universal quantum computer involves continuous-variable cluster states 
+# [#CV-MBQC]_. If you would like to learn more about the architecture, you can read our blueprint 
+# papers [#XanaduBlueprint]_ and [#XanaduPassiveArchitecture]_. In the meantime on the hardware side, 
+# efforts are made to develop the necessary technology. This includes the recent `Borealis 
+# experiment <https://xanadu.ai/blog/beating-classical-computers-with-Borealis>`_ [#Borealis]_ where a 
+# 3-dimensional photonic graph state was created that was used to demonstrate quantum advantage.
```
soosub, line 386: We can show the Borealis graph state here with FlamingPy and mention that it is not a cluster state.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,461 @@
...
+# side of the cluster to the other. In this section, we will tackle the remaining parts concerning
+# logical operations.
+#
+
+##############################################################################
+#
+# .. _single-qubit-rotations:
+#
+# Single-qubit rotations
+# ```````````````````````
```
soosub, line 228: In this section, we still want to add PennyLane snippets that show the equivalence between MBQC and the gate-based approach

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,461 @@
...
+# means that the entanglement is made independently from the computation, like how a blank sheet of
+# paper is made separately from the text of a book. 
+#
+
+##############################################################################
+# Universality of MBQC
+# ----------------------
+# How do we know if this measurement-based scheme is just as powerful as its gate-based brother? We
+# have to prove it! In particular, we want to show that a measurement-based quantum computer is a
+# called quantum Turing machine (WTM) also known as a universal quantum computer. To do this, we
```
soosub, line 207:
```suggestion
# called quantum Turing machine (QTM) also known as a universal quantum computer. To do this, we
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,461 @@
...
+# .. math::    |\psi\rangle=\Pi_{(i,j)\in E}CZ_{ij}|+⟩^{\otimes n}.
+#
+# where :math:`CZ_{ij}` is the controlled-:math:`Z`` gate between qubits :math:`i` and :math:`j`, 
+# and :math:`|+\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle + |1\rangle\big)` is the :math:`+1`` 
+# eigenstate of the Pauli-:math:`X` operator.
+#
+# The distinction between a graph state and a cluster state is that the latter has the additional 
+# condition that the underlying graph has to be a lattice. Physically, this means that ...
+# ... require Pauli measurements ... fully disentangle bla bla.
+#
```
soosub, line 66: We need to explain the distinction between graph states and cluster states here, and why it is physically and QIT-ically different

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,461 @@
...
+# We can also describe the creation of a cluster state in the gate-based description. Let us first 
+# define a graph we want to look at, and then construct a circuit in PennyLane to create the 
+# corresponding graph state.
+#
+
+import networkx as nx
+
+a, b = 5, 1 # dimensions of the graph (lattice)
+G = nx.grid_graph(dim=[a, b]) # there are a * b qubits
+nx.draw(G, node_color='black')
```
soosub, line 76: Preferably, I want to have the figure a bit smaller

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,463 @@
...
+#
+# .. math::
+#   \mathcal{B}_j(\theta_j) \equiv \left\{\frac{|0\rangle + e^{i\theta_j}|1\rangle}{\sqrt{2}}, 
+#   \frac{|0\rangle - e^{i\theta_j}|1\rangle}{\sqrt{2}}\right\},
+#
+# where the angles :math:`\theta_j`. Explicitly, these angles are given as
+#
+# .. :math:: \theta_{\mathrm{in}} = 0,  \theta_{1} = (-1)^{m_{\mathrm{in}} + 1} \alpha, 
+#   \theta_{2} = (-1)^{m_1} \beta, \theta_{3} = (-1)^{m_{\mathrm{in}} + m_2} \gamma
+#
```
soosub, line 260: Somehow this does not render properly on the website

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,463 @@
...
+# :math:`t_\mathrm{in}`, :math:`a_1`, :math:`a_2`, and :math:`a_3`, respectively. Note that the 
+# measurement basis is adaptive - the measurement on :math:`a_3` for example depends on the outcome 
+# of earlier measurements on the chain. After these operations, the state of qubit 
+# :math:`t_\mathrm{out}` is given by
+#
+# .. math:: \hat{U}(\alpha, \beta, \gamma) = X^{m_1 + m_3}Z^{m_{\mathrm{in}} + m_2}U(\alpha, \beta, \gamma)
+#    |\psi_{\mathrm{in}}\rangle.
+#
+# Now note that this unitary :math:`\hat{U}` is related to our desired unitary :math:`U` up to 
+# possible the first two Pauli terms. However, we can correct for these additional Pauli gates by
```
soosub, line 271:
```suggestion
# Now note that this unitary :math:`\hat{U}` is related to our desired unitary :math:`U` up to 
# the first two Pauli terms. However, we can correct for these additional Pauli gates by 
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,463 @@
...
+# Propagation of logical information
+# ````````````````````````
+# Essentially, we keep logical information in one end of our cluster state which we progagate to the
+# other end using the teleportation protocol. By choosing adaptive measurements, we can "write" our
+# circuit onto the cluster state. Later, we will see how we can actually do this.
+#
+# It's good to emphasize that the entanglement of the cluster state is created *off-line*. This
+# means that the entanglement is made independently from the computation, like how a blank sheet of
+# paper is made separately from the text of a book. 
+#
```
soosub, line 202: This section is still incomplete

demonstrations/tutorial_mbqc.py
soosub: and the linear chain should be straight

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,463 @@
...
+# RHG.draw(backend="plotly") #TODO: uncomment this line after merging FP#103
+
+##############################################################################
+#
+# .. raw:: html
+#    :file: ../demonstrations/mbqc/rhg-graph.html
+#
+
+##############################################################################
+#
```
soosub, line 361: I'd still like to explain dual and primal sheets, how it relates to the gate-based surface code etc.

demonstrations/tutorial_mbqc.py
soosub: Here are some relevant references:

1. [Persistent Entanglement in Arrays of Interacting Particles](https://journals.aps.org/prl/pdf/10.1103/PhysRevLett.86.910)
2. [Entanglement in Graph States and its Applications](https://arxiv.org/abs/quant-ph/0602096)
3. Section on Cluster States by Hans Briegel in the _Compendium of Quantum Physics_

demonstrations/tutorial_mbqc.py
soosub: There are some intricate differences between CV/DV lands that would make it too complicated for this demo so let's not do it.

demonstrations/tutorial_mbqc.py
soosub: @KetpuntoG do you have any suggestions for this? Is there a better way than using the `draw` method here?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,512 @@
...
+# .. math:: |\psi_{\mathrm{out}}\rangle = \hat{U}(\alpha, \beta, \gamma)|\psi_{\mathrm{in}}\rangle = X^{m_1 + m_3}Z^{m_{\mathrm{in}} + m_2}U(\alpha, \beta, \gamma)
+#    |\psi_{\mathrm{in}}\rangle.
+#
+# Now note that this unitary :math:`\hat{U}` is related to our desired unitary :math:`U` up to
+# the first two Pauli terms. Luckily, we can correct for these additional Pauli gates by
+# choosing the final the measurement basis appropriately or correcting for them classically after 
+# the quantum computation.
+#
+# Let's now see how this pans out in PennyLane. [WORK IN PROGRESS]
```
soosub, line 280: @radras can you take care of this?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,512 @@
...
+# the figure below, and measuring qubits :math:`t_\mathrm{in}` and :math:`a` in the Hadamard basis,
+# we implement the CNOT gate between qubits :math:`c` and :math:`t_\mathrm{out}` up to Pauli corrections.
+#
+# .. figure:: ../demonstrations/mbqc/cnot.png
+#    :align: center
+#    :width: 50%
+#
+#    ..
+#
+# Let's now see how this pans out in PennyLane. [WORK IN PROGRESS]
```
soosub, line 298: and this too, please @radras?

mauriceweber (commented):
Very nice demo!! I learned quite a bit about MBQC 🎉 

I think the main point from my side is to make a concrete example for translating an arbitrary quantum circuit to a measurement based quantum computation (it's still a bit for me to imagine how exactly this is achieved). 

The quantum error correction section is a nice intro, although probably quite hard for people who don't now anything about QEC. So making a simple example with a small quantum circuit would illustrate MBQC for a wider audience.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,512 @@
...
+#    ..
+#
+#    We can also consume the cluster state while we grow it.
+#
+# This feature makes it particularly interesting for photonic-based quantum computers: we can use
+# expendable qubits that don't have to stick around for the full calculation. If we can find a
+# reliable way to produce qubits and stitch them together through entanglement, we can use it to
+# produce our cluster state resource! Essentially, we need some kind of qubit factory and a
+# stitching mechanism that puts it all together.
+#
```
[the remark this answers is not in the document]
soosub: Cool, I'll mention that this can be done with an Ising interaction or beamsplitters in case of photonics ⚡

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,512 @@
...
+# .. math::    |\psi\rangle=\Pi_{(i,j)\in E}CZ_{ij}|+⟩^{\otimes n}.
+#
+# where :math:`CZ_{ij}` is the controlled-:math:`Z`` gate between qubits :math:`i` and :math:`j`,
+# and :math:`|+\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle + |1\rangle\big)` is the :math:`+1``
+# eigenstate of the Pauli-:math:`X` operator.
+#
+# The distinction between a graph state and a cluster state is that the latter has the additional
+# condition that the underlying graph has to be a lattice. Physically, this means that ...
+# ... require Pauli measurements ... fully disentangle bla bla.
```
soosub, line 67: This is still a TODO

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+# In this demo, we have learned that a one-way quantum computer capable of cluster state 
+# entanglement together with adaptive arbitrary single-qubit measurements is universal. 
+# The MBQC framework is a powerful quantum computing approach and offers several advantages over
+# the gate-based model and is particularly useful in platforms that allow for many expendable
+# qubits. For example, it circumvents the need for applying in-line entangling gates which are 
+# often the most noisy operations in gate-based quantum computers with trapped-ions or 
+# superconducting circuits.
+#
+#
+#   Cluster state entanglement + adaptive arbitrary single-qubit measurements = universal quantum computer
```
soosub, line 592: This box quote looks a bit weird rn, we can change this to make it look better

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+#
+# with :math:`m_{\mathrm{in}}, m_1, m_2` and :math:`m_3` being the measurement outcomes on nodes
+# :math:`t_\mathrm{in}`, :math:`a_1`, :math:`a_2`, and :math:`a_3`, respectively. Note that the
+# measurement basis is adaptive - the measurement on :math:`a_3` for example depends on the outcome
+# of earlier measurements on the chain. After these operations, the state of qubit
+# :math:`t_\mathrm{out}` is given by
+#
+# .. math:: |\psi_{\mathrm{out}}\rangle = \hat{U}(\alpha, \beta, \gamma)|\psi_{\mathrm{in}}\rangle 
+#    = X^{m_1 + m_3}Z^{m_{\mathrm{in}} + m_2}U(\alpha, \beta, \gamma)
+#    |\psi_{\mathrm{in}}\rangle.
```
KetpuntoG, line 280: I would have liked to follow the calculations, do you know any reference to put here to see it?

KetpuntoG (commented):
I leave you a few comments from the first part. On Monday I will finish the second half. I'm really enjoying it! Congratulations 😄
I think I understand the problem you commented @soosub . When talking about rotations you talk about 5 qubits but then you can't do more than one measurement, right? You could explain that those examples, the coefficients we know are zero and it is simplified somehow. But if not, it looks strange to see small circuits when you were expecting something with 5 qubits

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+
+    # Measure the first qubit in the Pauli-X basis
+    # and apply an X-gate conditioned on the outcome
+    qml.Hadamard(wires=0)
+    m = qml.measure(wires=[0])
+    qml.cond(m == 1, qml.PauliX)(wires=1)
+    qml.Hadamard(wires=1)
+
+    # Return the density matrix of the output state
+    return qml.density_matrix(wires=[1])
```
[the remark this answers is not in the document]
soosub: Hmm in principle this should be a pure state, it's just that PennyLane does not support outputting the state vector if there was a mid-circuit measurement (which in general makes sense). Do you think we should still comment on this @KetpuntoG?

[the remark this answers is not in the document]
soosub: Maybe we can show that the purity of the state / density matrix to get the point across

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+#    :width: 75%
+#
+#    ..
+#
+#    The flow of information in a measurement-based quantum computation [#OneWay2001]_
+#
+# One-bit Teleportation
+# `````````````````````
+# Let us have a deeper look at the principles behind the protocol using a simple example of one-bit
+# teleportation. We start with one qubit in the state :math:`|\psi\rangle` that we want to transfer
```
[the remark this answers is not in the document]
soosub: Good call, we'll change that

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
KetpuntoG: by putting a link to wikipedia of what the density matrix is the first time it appears, it will be enough :)

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+#
+# Now note that this unitary :math:`\hat{U}` is related to our desired unitary :math:`U` up to
+# the first two Pauli terms. Luckily, we can correct for these additional Pauli gates by
+# choosing the measurement basis of qubit :math:`t_\mathrm{out}` appropriately or correcting for them classically after
+# the quantum computation.
+#
+# To demonstrate that this actually works, we will use PennyLane. For simplicity, we will just 
+# show the ability will to perform single-axis rotations :math:`R_z(\theta)` and 
+# :math:`R_x(\theta)`. Note these two plus the CNOT also constitute also a universal gate set. 
+# To start of, we define the :math:`R_z(\theta)` gate using two qubits with the gate-based approach.
```
KetpuntoG, line 290: remember to comment that being a case with 2 coefficients equal to 0, it is simplified and we do not need 5 qubits but 2. (same RX)

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+# show how to perform a CNOT operation in the measurement-based framework. The input state is given on two qubits,
+# control qubit :math:`c` and target qubit :math:`t_\mathrm{in}`. Preparing the cluster state shown in
+# the figure below, and measuring qubits :math:`t_\mathrm{in}` and :math:`a` in the Hadamard basis,
+# we implement the CNOT gate between qubits :math:`c` and :math:`t_\mathrm{out}` up to Pauli corrections.
+#
+# .. figure:: ../demonstrations/mbqc/cnot.png
+#    :align: center
+#    :width: 50%
+#
+#    ..
```
KetpuntoG, line 414: I missed the mathematical expression here. As you had done before in which expresses |phi_out> as a function of U and |phi_in>.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+# teleportation is very real and it's one of the driving concepts behind MBQC. Moreover, it has related applications
+# in safe communication protocols that are impossible with classical communication so it's certainly
+# worth learning about. In this protocol, we do not transport matter but *information* between systems. Admittedly, it has a
+# rather delusive name because it is not instantaneous but requires communication of additional classical information,
+# which is naturally limited by the speed of light.
+#
+# .. figure:: ../demonstrations/mbqc/mbqc_info_flow.png
+#    :align: center
+#    :width: 75%
+#
```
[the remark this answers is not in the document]
radras: If you look at the examples with a single-qubit rotation and a CNOT gate you can see how the state "flows" from left to right and ends up at the rightmost qubits. Looking at this figure you can relate any T shape with a 2-qubit gate and any straight path with a 1-qubit gate. Hence, in a sense, the layout of the MBQC protocol resembles the graphical representation of a circuit in a gate-based form where information flows through time while here information flows through space as well.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+# teleportation. We start with one qubit in the state :math:`|\psi\rangle` that we want to transfer
+# to the second qubit initially in the state :math:`|0\rangle`. The figure below represents the 
+# one-bit teleportation protocol. The green box represents the creation of a cluster state while
+# the red box represents the measurement of a qubit with the appropriate correction applied to 
+# the second qubit based on the single bit acquired through the measurement.
+#
+# .. figure:: ../demonstrations/mbqc/one-bit-teleportation.png
+#    :align: center
+#    :width: 75%
+#
```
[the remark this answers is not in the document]
radras: Well, this is just a convention we chose, maybe we should briefly mention the other possibility of applying a Hadamard. I think that this is more natural since this is what is actually happening. Take for example a case where you want to first perform a single-qubit unitary and then a CNOT. You can initialize all of the required qubits into the cluster state in the beginning. Then after a single qubit gate, you will have a qubit that will hold the rotated state |psi>. This state can then be considered as an input into a CNOT protocol and there is no need for another Hadamard since it has been performed at the beginning of the entire calculation to generate the cluster state. So basically I did this so it resembles the latter examples but we can change it if you think this might be confusing.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,687 @@
...
+# .. raw:: html
+#    :file: ../demonstrations/mbqc/rhg-graph.html
+#
+#
+# For the sake of intuition, you can think of the state as having two spatial dimensions (:math:`x` 
+# and :math:`y`) and one temporal dimension (:math:`z`). The cluster state alternates between *primal* and *dual sheets*, shown in the
+# figure above on the xy-plane. In principle, any quantum error correction code can be foliated into 
+# a graph state for measurement-based QEC. However, the foliations are particularly nice for CSS 
+# codes, named after Calderbank, Shor and Steane. CSS codes have stabilizers that exclusively contain
```
KetpuntoG, line 557: It would be great a link to deep into the CSS codes

KetpuntoG (commented):
I have become a fan of MBQC! 🚀
I leave a couple more comments but I see it as very consistent

demonstrations/tutorial_mbqc.py
radras: I added [this reference that I used](https://arxiv.org/pdf/2112.11601.pdf) although it doesn't have detailed derivations, I had to do some by hand to adapt the formulas for Rx and Rz.

.github/workflows/build-dev.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
```
KetpuntoG, line 31: by the way @soosub @radras , what are all these changes you are making in these files. It seems like it's something you shouldn't be modifying

soosub: I checked with @josh146 and if it doesn't break anything it's fine. We did it because FlamingPy requires Python 3.8.

demonstrations/tutorial_mbqc.py
soosub: I don't understand what you mean @KetpuntoG

KetpuntoG: Sorry, I mean to put this [link](https://en.wikipedia.org/wiki/CSS_code), or another one for CSS codes

soosub: Ah I see, I will find a good reference. The wiki on CSS codes is shit 🙈

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
soosub: We can put your comment above in the demo @radras 😏

demonstrations/tutorial_mbqc.py
KetpuntoG: if it's not too complicated it would be great! 😄

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
soosub: I will put this in text and adjust the figure so that's clear later this week 🖌️

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,732 @@
...
+# reads as follows
+#
+# .. math::    |\Phi\rangle=\Pi_{(i,j)\in E}CZ_{ij}|+⟩^{\otimes n}.
+#
+# where :math:`n` is the number of qubits, :math:`CZ_{ij}` is the controlled-:math:`Z`` gate between
+# qubits :math:`i` and :math:`j`, and :math:`|+\rangle = \frac{1}{\sqrt{2}}\big(|0\rangle + |1\rangle\big)` 
+# is the :math:`+1`` eigenstate of the Pauli-:math:`X` operator.
+#
+# The distinction between a graph state and a cluster state is that the latter has the additional
+# condition that the underlying graph has to be a lattice. Physically, this means that TODO ...
```
KetpuntoG, line 70: I leave the comment so that this will not be forgotten

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,732 @@
...
+#
+
+##############################################################################
+# Information propagation and teleportation
+# ------------------------------------------
+#
+# Measurement-based quantum computation heavily relies on the idea of information propagation. In
+# particular, we make use of a protocol called *quantum teleportation*, one of the driving concepts behind MBQC. Despite its esoteric name, quantum
+# teleportation is very real and experimentally demonstrated multiple times in the last few decades 
+# [#Hermans2022]_, [#Furusawa1998]_, [#Riebe2004], [#Nielsen1998]. Moreover, it has related applications
```
KetpuntoG, line 133: Rendering failure, maybe the "_" has something to do with it?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,732 @@
...
+    qml.cond(m == 1, qml.PauliX)(wires=1)
+    qml.Hadamard(wires=1)
+
+    # Return the density matrix of the output state
+    return qml.density_matrix(wires=[1])
+
+##############################################################################
+#
+# Seems good! As we can see, the resulting states are practically the same as we wanted to show. 
+# Note that we return `*density matrices* <https://en.wikipedia.org/wiki/Density_matrix>`_ in the
```
KetpuntoG, line 343: I think you wanted to put the bold print here. On the other hand, the clarification of the density matrix I would put it in the teleportation part since it is the first time where it appears

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,732 @@
...
+#   C. Becher, F. Schmidt-Kaler, D. F. V. James & R. Blatt. (2002) *Deterministic quantum 
+#   teleportation with atoms* `Nature 429, 734-737 <https://www.nature.com/articles/nature02570>`_.
+#
+
+##############################################################################
+#
+# About the authors
+# ----------------
+#
+# .. bio:: Joost Bus
```
KetpuntoG, line 724: The biographies have been decided to put in third person (I know that in your how-to it is in first person @soosub, but it will be changed)

KetpuntoG (commented):
Completing the missing TODO's, for my part it's reviewed!
 As soon as it is ready, a second person will review and give the go-ahead 💪
 A very good demo topic, congratulations  🎉

demonstrations/tutorial_mbqc.py
soosub:
```suggestion
# [#Hermans2022]_, [#Furusawa1998]_, [#Riebe2004]_, [#Nielsen1998]_. Moreover, it has related applications
```

soosub: It would be even nicer if we can group the references between one pair of brackets, do you know if that's possible @KetpuntoG ?

demonstrations/tutorial_mbqc.py
soosub: doesn't that sound terribly impersonal?

soosub: I prefer first person

soosub: *He prefers to speak in first person

demonstrations/tutorial_mbqc.py
soosub: what line do you mean?

soosub: @KetpuntoG do you know how to make the link text cursive?

KetpuntoG: I've been looking at a few demos and haven't seen any in cursive. I don't know how to do it 🤔

josh146: Unfortunately, this is a limitation of reStructuredText. While it is generally much more powerful than markdown, unlike markdown it does not allow nested formatting!

There are [workarounds](https://stackoverflow.com/questions/4743845/format-text-in-a-link-in-restructuredtext) however, but they do feel a bit ugly.

soosub: Thanks @josh146!

demonstrations/tutorial_mbqc.py
KetpuntoG: Ok, you can leave the bio in 1 person 👌

KetpuntoG:
Hey! I cannot pass on the review to the next person until you complete the TODO's
Can you take a look at them? 🙂

soosub:
We're working on it @KetpuntoG 😌

requirements.txt
```diff
@@ -46,3 +46,4 @@ pyyaml==5.4.1
 jax==0.3.15
 jaxlib==0.3.15
 neural-tangents==0.6.0
+flamingpy==0.9.0b0
```
soosub, line 49: this needs to be updated once FP PR#103 is in for the plotly plot

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,772 @@
...
+#    |\psi_{\mathrm{in}}\rangle.
+#
+# Now note that this unitary :math:`\hat{U}` is related to our desired unitary :math:`U` up to
+# the first two Pauli terms. Luckily, we can correct for these additional Pauli gates by
+# choosing the measurement basis of qubit :math:`t_\mathrm{out}` appropriately or correcting for them classically after
+# the quantum computation.
+#
+# To demonstrate that this actually works, we will use PennyLane. For simplicity, we will just 
+# show the ability will to perform single-axis rotations :math:`R_z(\theta)` and 
+# :math:`R_x(\theta)` for arbitrary :math:`\theta \in [0, 2 \pi)`. Note that these two operations
```
soosub, line 302: Hey @radras, could you add some more explanation of why we use less qubits here? Maybe also include the citation of the paper that you used

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
soosub: Can you include something that makes this clear in the text @radras?

demonstrations/tutorial_mbqc.py
soosub: Could you include this reference @radras?

soosub:
Hey it's ready for another review. I still have two questions though:
- How do I put the floor operation in here? It seems like there are some problems in the surface code figure caption. Is this because it's a caption?
- Can I combine references into one pair of brackets? I have a couple of places where we reference multiple papers but it comes out as [1] [2] [3]. It would be nice to have something like [1, 2, 3] as per usual.

KetpuntoG:
> Hey it's ready for another review. I have one questions about formatting though:
> 
> * Can I combine references into one pair of brackets? I have a couple of places where we reference multiple papers but it comes out as [1] [2] [3]. It would be nice to have something like [1, 2, 3] as per usual.

Thanks! 
Not to my knowledge. It is something that appears in other demos.
 If I find out, I'll let you know 👌

soosub:
Ready for review @KetpuntoG

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# show the ability will to perform single-axis rotations :math:`R_z(\theta)` and
+# :math:`R_x(\theta)` for arbitrary :math:`\theta \in [0, 2 \pi)`. Note that these two operations
+# plus the CNOT also constitute a universal gate set.
+#
+# To start of, we define the :math:`R_z(\theta)` gate using two qubits with the gate-based approach
+# so we can later compare our MBQC approach to it.
+
+dev = qml.device("default.qubit", wires=1)
+
+
```
isaacdevlugt, line 309:
```suggestion
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+##############################################################################
+#
+# Now he have a base case, let's implement an :math:`R_z` gate on an arbitrary state
+# in the MBQC formalism.
+#
+
+mbqc_dev = qml.device("default.qubit", wires=2)
+
+
```
isaacdevlugt, line 330:
```suggestion
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+##############################################################################
+#
+# Next, we will prepare a random input state and compare the two approaches.
+#
+
+# Generate a random input state
+input_state = generate_random_state()
+theta = 2 * np.pi * np.random.random()
+
+np.allclose(RZ(theta, input_state), RZ_MBQC(theta, input_state))
```
isaacdevlugt, line 361: 🤯

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+    qml.cond(m2 == 1, qml.PauliX)(wires=2)
+    qml.cond(m1 == 1, qml.PauliZ)(wires=2)
+
+    # Return the density matrix of the output state
+    return qml.density_matrix(wires=[2])
+
+
+##############################################################################
+#
+# Finally, we again compare the two implementations with random state as an input.
```
isaacdevlugt, line 416:
```suggestion
# Finally, we again compare the two implementations with a random state as an input.
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+    qml.CNOT(wires=[0, 1])
+
+    return qml.density_matrix(wires=[0, 1])
+
+
+# Let's now implement a CNOT in MBQC formalism!
+# We will associate qubits c, t_in, a, and t_out in the figure with qubits 0, 1, 2, 3, respectively.
+mbqc_dev = qml.device("default.qubit", wires=4)
+
+
```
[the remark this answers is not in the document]
isaacdevlugt: Just tidying up the code to reduce scroll time!

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# gate, the final step is to show that we can implement arbitrary quantum circuits. To do so,
+# we simply have to note that we have a *universal gate set* [#DiVincenzo]_. However, you might
+# wonder - is it even feasible to construct the large cluster states that one-way quantum
+# computation requires?
+#
+# The number of qubits needed to construct a circuit can grow to be very large, as it not only
+# depends on the number of logical qubits, but also on the depth of the circuit. At this point,
+# it's good to reiterate that the entanglement of the cluster
+# state is created *off-line*. This means that the entanglement is made independently from the
+# computation, like how a blank sheet of paper is made separately from the text of a book.
```
isaacdevlugt, line 514: Interesting analogy. I dig it 👍

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# Interestingly enough, we do not have to prepare all the entanglement at once. Just like we can
```
isaacdevlugt, line 515:
```suggestion
# Interestingly enough, we do not have to prepare all of the entanglement at once. Just like we can
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# already start printing text upon the first few pages, we can apply measurements to one end of the
+# cluster, while growing it at the same time, as shown in the figure below. That is, we can start
```
isaacdevlugt, line 517:
```suggestion
# cluster while growing it at the same time, as shown in the figure below. That is, we can start
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# printing the text on the first few pages while at the same time reloading the printer's paper
+# tray!
```
isaacdevlugt, line 519: You really went heavy into the printer paper analogy! 😆 I love it

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+#    :width: 75%
+#
+#    ..
+#
+#    We can also consume the cluster state while we grow it. The blue qubits are in a cluster state,
+#    where the bonds between them represent entanglement. The gray qubits have been measured,
+#    destroying the entanglement and removing them from the cluster. At the same time, the green
+#    qubits are being added to the cluster by entangling them with it. Prior measurement outcomes
+#    determine the basis for future measurements [#OpticalQuantumComputing]_.
+#
```
isaacdevlugt, line 532: This is rendering as the figure caption. Is that on purpose?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# This feature makes it particularly interesting for photonic-based quantum computers: we can use
+# expendable qubits that don't have to stick around for the full calculation. If we can find a
+# reliable way to produce qubits and stitch them together through entanglement, we can use it to
+# produce our cluster state resource! Essentially, we need some kind of qubit factory and a
+# stitching mechanism that puts it all together. The stitching mechanism depends on the physical
+# platform, and can for example be implemented with an Ising interaction [#OneWay2001]_ or in the
+# context of photonics by interfering two modes with a beamsplitter.
```
isaacdevlugt, line 539:
```suggestion
# platform, and can, for example, be implemented with an Ising interaction [#OneWay2001]_ or, in the
# context of photonics, by interfering two modes with a beamsplitter.
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+#
+
+##############################################################################
+# Quantum error correction
+# -------------------------
+#
+# To mitigate the physical errors that can (and will) happen during a quantum computation, we
+# require some kind of error correction. Error correction is a technique for detecting errors and
+# reconstructing the logical data without losing any information. It is not exclusive to quantum computing;
+# it is also used in "classical" computing, memories and communication where one also has to deal `with
```
isaacdevlugt, line 549:
```suggestion
# reconstructing the logical data without losing any information. It is not exclusive to quantum computing;
# it is also used in "classical" computing, memories, and communication where one also has to deal `with
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# being flipped to a 0 or vice versa.
+#
+# A whole research field is devoted
+# to combating these challenges since Peter Shor published the seminal paper in 1995 [#ShorQEC1995]_.
+# Full coverage of this topic is beyond the scope of this tutorial, but a good place to start is
+# `Daniel Gottesman's thesis <https://arxiv.org/abs/quant-ph/9705052>`_ or `this blog post by
+# Arthur Pesah <https://arthurpesah.me/blog/2022-01-25-intro-qec-1/>`_ for a more compact
+# introduction. Instead, what we will do here is show how to implement error correction in the
+# MBQC framework by using the surface code [#FowlerSurfaceCode]_ [#FowlerPolyestimate]_ [#GoogleQEC2022]_ as
+# an example.
```
isaacdevlugt, line 568: Excellent! :chefskiss:

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+#
+# .. figure:: ../demonstrations/mbqc/surface_code_d3.png
```
isaacdevlugt, line 570: Nice!

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+from flamingpy.codes import SurfaceCode
+
+code_distance = 3
+RHG = SurfaceCode(code_distance)
+# RHG.draw(backend="plotly", label=["type"], showbackground=True) #TODO: uncomment this line after merging FP#103
+
+##############################################################################
+#
+# .. raw:: html
+#    :file: ../demonstrations/mbqc/rhg-graph.html
```
isaacdevlugt, line 599: Oooooo! Pretty! 

My only concern is that the interactive picture here is _very_ large... Can you make it smaller by default?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+#
+#
+# For the sake of intuition, you can think of the graph shown above as having two spatial dimensions (:math:`x`
+# and :math:`y`) and one temporal dimension (:math:`z`). The cluster state alternates between *primal* and *dual sheets*, shown in the
+# figure above on the :math:`xy`-plane. In principle, any quantum error correction stabilizer code can be `foliated <https://arxiv.org/abs/1607.02579>`_ into
+# a graph state for measurement-based QEC [#FoliatedQuantumCodes]_, [#UniversalFTMBQC]_. However, the foliations are particularly nice for `CSS
+# codes <https://errorcorrectionzoo.org/c/css>`_, named after Calderbank, Shor and Steane. CSS codes have stabilizers that exclusively contain
```
isaacdevlugt, line 606:
```suggestion
# codes <https://errorcorrectionzoo.org/c/css>`_, named after Calderbank, Shor, and Steane. CSS codes have stabilizers that exclusively contain
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# :math:`X`-stabilizers *or* :math:`Z`-stabilizers, and include the *surface code* and *colour code* families.
+# For these CSS codes, you can view the primal and dual sheets as measuring the
+# :math:`Z`-stabilizers and :math:`X`-stabilizers, respectively.
+#
+# The computation and error correction are again performed with single-qubit measurements, as illustrated below.
+# At each timestep, we measure all the qubits on one sheet of the lattice. The binary
+# outcomes of these measurements determine the measurement bases for future measurements, and the
+# last sheet of the lattice contains the encoded result of the computation which can be read out by yet another measurement.
+#
+# .. figure:: ../demonstrations/mbqc/gif_measuring.gif
```
isaacdevlugt, line 616: 🤩

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# expendable qubits that don't have to stick around for the full calculation. If we can find a
+# reliable way to produce qubits and stitch them together through entanglement, we can use it to
+# produce our cluster state resource! Essentially, we need some kind of qubit factory and a
+# stitching mechanism that puts it all together. The stitching mechanism depends on the physical
+# platform, and can for example be implemented with an Ising interaction [#OneWay2001]_ or in the
+# context of photonics by interfering two modes with a beamsplitter.
+#
+
+##############################################################################
+# Quantum error correction
```
isaacdevlugt, line 543: I think it needs to be made clear somewhere in this section that you won't be showing any code for error correction in the MBQC formalism. I got to the end of this section and was a little surprised. Obviously, it's totally fine that there's no code in this section! It's already a very long and amazing demo!

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# useful in platforms that allow for many expendable qubits, and offers
+# several advantages over the gate-based model. For example, it circumvents the need for applying
+# in-line entangling gates which are often the most noisy operations in gate-based quantum computers
+# with trapped-ions or superconducting circuits. Instead, the required entanglement is created
+# off-line which is often simpler to implement.
+#
+# .. container:: alert alert-block alert-info
+#
+#   **A system that is capable of creating an entangled cluster state and performing adaptive
+#   arbitrary single-qubit measurements on it is a universal quantum computer!**
```
isaacdevlugt, line 643: Feels out of place 🤔

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# beyond the scope of this tutorial, but a good place to start is `this
+# paper <https://arxiv.org/abs/quant-ph/0403025>`_ [#MagicStates]_.
+#
+# Xanadu's approach toward a universal quantum computer involves *continuous-variable* cluster states
+# [#CV-MBQC]_. If you would like to learn more about the architecture, you can read our blueprint
+# papers [#XanaduBlueprint]_ and [#XanaduPassiveArchitecture]_. We highly recommend watching `this
+# video <https://www.youtube.com/watch?v=SD6TH7GZ1rM&feature=emb_title>`_ outlining the main ideas.
+# On the hardware side, efforts are made to develop the necessary technology. This includes the recent `Borealis
+# experiment <https://xanadu.ai/blog/beating-classical-computers-with-Borealis>`_ [#Borealis]_ where
+# a 3-dimensional photonic graph state was created that was used to demonstrate quantum advantage.
```
isaacdevlugt, line 657:
```suggestion
# a 3-dimensional photonic graph state was created that was used to demonstrate quantum computational advantage.
```

isaacdevlugt (changes requested):
Really great demo guys! Comments are all very minor. Nice work!

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# ------------------------------------------
+#
+# Measurement-based quantum computation heavily relies on the idea of information propagation. In
+# particular, we make use of a protocol called *quantum teleportation*, one of the driving concepts behind MBQC. Despite its esoteric name, quantum
+# teleportation is very real and experimentally demonstrated multiple times in the last few decades
+# [#Hermans2022]_, [#Furusawa1998]_, [#Riebe2004]_, [#Nielsen1998]_. Moreover, it has related applications
+# in safe communication protocols that are impossible with classical communication so it's certainly
+# worth learning about. In this protocol, we do not transport matter but *information* between systems. Admittedly, it has a
+# rather delusive name because it is not instantaneous but requires communication of additional classical information,
+# which is naturally limited by the speed of light.
```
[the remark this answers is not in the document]
soosub: Hmm what I tried to convey is that it's we don't have faster than light communication. Should we rephrase?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+
+##############################################################################
+#
+# Note that we return a `density matrix <https://en.wikipedia.org/wiki/Density_matrix>`_ in the
+# function above because we are dealing with a *quantum channel* that includes a measurement on a
+# subset of the qubits. These matrices allow for description of mixed quantum states and are an
+# extension of pure state vectors. Using this formalism, we can describe operations beyond unitaries
+# such as the teleportation protocol.
+#
+# Now, let's prepare a random qubit state and see if the teleportation protocol is working as
```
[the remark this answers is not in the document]
soosub: Do you want to omit the reason why we use density matrices?

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+##############################################################################
+#
+# Finally, we again compare the two implementations with random state as an input.
+#
+
+# Generate a random input state
+input_state = generate_random_state()
+theta = 2 * np.pi * np.random.random()
+
+np.allclose(RX(theta, input_state), RX_MBQC(theta, input_state))
```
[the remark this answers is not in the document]
soosub: Shout out to @radras 🙌🏼

demonstrations/tutorial_mbqc.py
soosub: Yes, do you think we should have a smaller caption and put the rest in the main text @isaacdevlugt?

demonstrations/tutorial_mbqc.py
soosub: I can make it smaller 👍🏼

demonstrations/tutorial_mbqc.py
soosub: Alright, what do you think should we put something like "giving the gist of QEC in the MBQC framework" in the section introduction? @isaacdevlugt

demonstrations/tutorial_mbqc.py
soosub: Leave it out completely? Place it somewhere else? @isaacdevlugt

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+
+a, b = 1, 5  # dimensions of the graph (lattice)
+G = nx.grid_graph(dim=[a, b])  # there are a * b qubits
+
+plt.figure(figsize=(5, 1))
+nx.draw(G, pos={node: node for node in G}, node_size=500, node_color="black")
+
+##############################################################################
+#
+# This is a fairly simple cluster state, but we `will later see <#single-qubit-rotations>`_ how even
```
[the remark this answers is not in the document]
soosub:
```suggestion
# This is a fairly simple cluster state, but we will later see how even
```

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
soosub: That seems to leave away a lot of info, don't you think? @isaacdevlugt

demonstrations/tutorial_mbqc.py
soosub: I'll leave it out

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# extension of pure state vectors. Using this formalism, we can describe operations beyond unitaries
+# such as the teleportation protocol.
+#
+# Now, let's prepare a random qubit state and see if the teleportation protocol is working as
+# expected. To do so, we generate a random complex vector and normalize it to create a valid
+# quantum state :math:`|\psi\rangle = \alpha |0\rangle + \beta |1\rangle`.
+# We then apply the teleportation protocol and see if the resulting density matrix of the output
+# state of the second qubit is the same as the input state of the first qubit.
+
+# Define helper function for random input state on n qubits
```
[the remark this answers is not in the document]
soosub: Hmm I prefer it this way

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# :math:`t_\mathrm{in}` is prepared in some input state :math:`|\psi_\mathrm{in}\rangle`,
+# and we are interested in the final state of the output qubit :math:`t_\mathrm{out}`.
+#
+# .. figure:: ../demonstrations/mbqc/single-qubit-rotation.png
+#    :align: center
+#    :width: 75%
+#
+#    ..
+#
+# The input qubit :math:`t_\mathrm{in}`, together with the intermediate qubits :math:`a_1`,
```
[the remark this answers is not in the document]
soosub: That will get a bit messy 🤔 You would have to keep track of density matrices and properly do the projective measurements. I don't think showing it here will make thinks clearer.

soosub:
```
from flamingpy.codes import SurfaceCode
from plotly.subplots import make_subplots
import plotly.graph_objects as go

# RHG lattice
RHG = SurfaceCode(3)

# Layout
size = 450
fig = make_subplots(rows=1, cols=2, 
                    specs=[[{'type':'scene'}, {'type':'scene'}]], 
                    column_titles=["Primal", "Dual"], 
                    horizontal_spacing = 0.02)
fig.update_layout(width=2*size, height=size)

# Primal
x = 2
fig1 = RHG.draw(backend="plotly", label=["type"], showbackground=True, width=size, height=size)
fig.update_layout(scene = dict(xaxis = {'nticks': 2, 'range': [x+0.5, x-0.5], 'dtick':1}), showlegend=False)
fig.update_layout(scene = dict(yaxis = {'dtick':1}))
fig.update_layout(scene = dict(zaxis = {'dtick':1}))
fig.add_trace(fig1.data[0], row=1, col=1)
fig.add_trace(fig1.data[1], row=1, col=1)

# Dual
x = 1
fig2 = RHG.draw(backend="plotly", label=["type"], showbackground=True, width=size, height=size)
fig.get_subplot(1,2)['xaxis'].update({'nticks': 2, 'range': [x+0.5, x-0.5]})
fig.get_subplot(1,2)['yaxis'].update({'dtick':1})
fig.get_subplot(1,2)['zaxis'].update({'dtick':1})
fig.add_trace(fig2.data[0], row=1, col=2)
fig.add_trace(fig2.data[1], row=1, col=2)

# Showing figure
fig.show(width=size, height=size)

# Saving figure
fig.write_image("primal_dual.jpeg")
fig.write_html("primal_dual.html")
```
primal_dual

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
isaacdevlugt: Hmm yea we can find some middle ground here. How about this?

```suggestion
# Note that we return a `density matrix <https://en.wikipedia.org/wiki/Density_matrix>`_ in the
# function above.
# This allows for us to describe operations beyond unitaries
# such as the teleportation protocol.
#
# Now, let's prepare a random qubit state and see if the teleportation protocol is working as
```

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
isaacdevlugt: okay! I would maybe leave comments in the code block then. E.g., 

```python
def generate_random_state(n=1):
    input_state = np.random.random(2 ** n) + 1j * np.random.random(2 ** n) # psi = alpha |0> + beta |1>`
    return input_state / np.linalg.norm(input_state)
```

or something like that

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
isaacdevlugt: Okay! Good to know. Was just wondering if it would be feasible.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,809 @@
...
+# .. math:: \theta_{\mathrm{in}} = 0, \qquad \theta_{1} = (-1)^{m_{\mathrm{in}} + 1} \alpha, \qquad
+#   \theta_{2} = (-1)^{m_1} \beta, \quad \text{and} \quad \theta_{3} = (-1)^{m_{\mathrm{in}} + m_2} \gamma
+#
+# with :math:`m_{\mathrm{in}}, m_1, m_2, m_3 \in \{0, 1\}` being the measurement outcomes on nodes
+# :math:`t_\mathrm{in}`, :math:`a_1`, :math:`a_2`, and :math:`a_3`, respectively. Note that the
+# measurement basis is adaptive - the measurement on :math:`a_3` for example depends on the outcome
+# of earlier measurements on the chain. After these operations, the state of qubit
+# :math:`t_\mathrm{out}` is given by
+#
+# .. math:: |\psi_{\mathrm{out}}\rangle = \hat{U}(\alpha, \beta, \gamma)|\psi_{\mathrm{in}}\rangle
```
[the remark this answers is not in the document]
isaacdevlugt: It's a small pet peeve of mine that we don't put hats on anything to begin with, but it's a settled-upon convention in the quantum community so it is what it is. But, either everything should have a hat on it, or no hats at all in my opinion. 

I'd opt for no hats _anywhere_, and then renaming `\hat{U}` to something else, like `V` for example.

demonstrations/tutorial_mbqc.py
isaacdevlugt: No I think it's fine, but just make it a bit more explicit that it is indeed a figure caption. E.g.

```suggestion
#    A diagrammatic perspective on simultaneously consuming and growing a cluster state.
#    The blue qubits are in a cluster state,
#    where the bonds between them represent entanglement. The gray qubits have been measured,
#    destroying the entanglement and removing them from the cluster. At the same time, the green
#    qubits are being added to the cluster by entangling them with it. Prior measurement outcomes
#    determine the basis for future measurements [#OpticalQuantumComputing]_.
#
```

demonstrations/tutorial_mbqc.py
isaacdevlugt: Yeah something like that! Maybe slightly less conversational. I wouldn't use "gist"

isaacdevlugt (changes requested):
Just addressing comments

demonstrations/tutorial_mbqc.py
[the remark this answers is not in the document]
soosub: Ah I see the confusion. I am using the hat just to make clear that U and \hat{U} are different, not to designate that \hat{U} is a operator. I could use a tilde or apostrophe instead @isaacdevlugt

demonstrations/tutorial_mbqc.py
soosub: 👍🏼

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,812 @@
...
+# dictates the structure of the cluster state. The cluster state that is associated with the surface code is known as the RHG lattice,
+# named after its architects Raussendorf, Harrington, and Goyal. We can visualize this cluster
+# state with FlamingPy.
+#
+
+from flamingpy.codes import SurfaceCode
+
+code_distance = 3
+RHG = SurfaceCode(code_distance)
+# RHG.draw(backend="plotly", label=["type"], showbackground=True) #TODO: uncomment this line after merging FP#103
```
soosub, line 596: TODO ⚠️

soosub (commented):
Just adding a note here that the demo should not be posted before these changes are added to FlamingPy

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+# codes <https://errorcorrectionzoo.org/c/css>`_, named after Calderbank, Shor, and Steane. CSS codes have stabilizers that exclusively contain
+# :math:`X`-stabilizers *or* :math:`Z`-stabilizers, and include the *surface code* and *colour code* families.
+# For these CSS codes, you can view the primal and dual sheets as measuring the
+# :math:`Z`-stabilizers and :math:`X`-stabilizers, respectively. We encourage you to have another 
+# look at the :ref:`figure <fig-surfacecode>` with the distance-3 surface code and try to link it 
+# with the dual and primal sheets shown here!
+#
+# .. raw:: html
+#    :file: ../demonstrations/mbqc/primal_dual.html
+#
```
soosub, line 623: @ilantz I think I might have interchanged primal/dual. Am I correct?

requirements.txt
```diff
@@ -46,3 +46,4 @@ pyyaml==5.4.1
 jax==0.3.15
 jaxlib==0.3.15
 neural-tangents==0.6.0
+git+https://github.com/XanaduAI/flamingpy@main
```
josh146, line 49:
```suggestion
git+https://github.com/XanaduAI/flamingpy.git
```

demonstrations/tutorial_mbqc.py
soosub: Installing it with git causes errors for some reason. We might need to wait for this month's release of FlamingPy before publishing.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#    A distance-3 surface code. Circles represent qubits and bubbles represent operators, called
+#    stabilizers, used to detect errors. The stabilizers are tensor products of Pauli-:math:`X` or Pauli-:math:`Z`
+#    operators and each is associated with its own ancilla qubit. The combined system encodes one
+#    logical qubit and can correct any combination of :math:`\lfloor (d-1)/2 \rfloor` errors.
+#    [#FowlerPolyestimate]_
+#
+# In the measurement-based picture, quantum error correction requires cluster states that are at
+# least 3-dimensional [#XanaduBlueprint]_, contrary to the 2-dimensional cluster states required for
+# universal quantum computation discussed in the previous section. The error correcting code that you want to implement
```
ilan-tz, line 592:
```suggestion
# In the measurement-based picture, successive rounds of quantum error correction requires generating cluster states
# that are at least 3-dimensional [#XanaduBlueprint]_, contrary to the 2-dimensional cluster states required for
# universal quantum computation discussed in the previous section. The error correcting code that you want to implement
```

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#    :file: ../demonstrations/mbqc/rhg-graph.html
+#
+#
+# For the sake of intuition, you can think of the graph shown above as having two spatial dimensions (:math:`x`
+# and :math:`y`) and one temporal dimension (:math:`z`). The cluster state alternates between *primal* and *dual sheets*, shown below in more detail. 
+# In principle, any quantum error correction stabilizer code can be `foliated <https://arxiv.org/abs/1607.02579>`_ into
+# a graph state for measurement-based QEC [#FoliatedQuantumCodes]_, [#UniversalFTMBQC]_. However, the foliations are particularly nice for `CSS
+# codes <https://errorcorrectionzoo.org/c/css>`_, named after Calderbank, Shor, and Steane. CSS codes have stabilizers that exclusively contain
+# :math:`X`-stabilizers *or* :math:`Z`-stabilizers, and include the *surface code* and *colour code* families.
+# For these CSS codes, you can view the primal and dual sheets as measuring the
```
ilan-tz, line 616:
```suggestion
# For these CSS codes, you can roughly view the primal and dual sheets as measuring the
```

demonstrations/tutorial_mbqc.py
ilan-tz: Primal sheets are at even Z slices, so they are flipped in the figure

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#
+
+
+##############################################################################
+#
+# Conclusion
+# -------------------------------
+#
+# We have learned that a one-way quantum computer capable of cluster state
+# entanglement together with adaptive arbitrary single-qubit measurements allows for universal
```
ilan-tz, line 646:
```suggestion
# entanglement together with adaptive arbitrary single-qubit measurements allows for universal
```
Consider modifying in light of my comments above

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#
+#    Schematic showing how we can also consume the cluster state while we grow it. The blue qubits are in a cluster state,
+#    where the bonds between them represent entanglement. The gray qubits have been measured,
+#    destroying the entanglement and removing them from the cluster. At the same time, the green
+#    qubits are being added to the cluster by entangling them with it. Prior measurement outcomes
+#    determine the basis for future measurements [#OpticalQuantumComputing]_.
+#
+# This feature makes it particularly interesting for photonic-based quantum computers: we can use
+# expendable qubits that don't have to stick around for the full calculation. If we can find a
+# reliable way to produce qubits and stitch them together through entanglement, we can use it to
```
ilan-tz, line 535:
```suggestion
# This feature makes it particularly attractive for photonic quantum computers: we can use
# expendable qubits that don't have to stick around for the full calculation. If we can find a
# reliable way to produce qubits and stitch them together through entanglement, we can use it to
```
The don't stick around, but they also can't, because they are likely to be lost or corrupted by loss

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+
+##############################################################################
+#
+# Conclusion
+# -------------------------------
+#
+# We have learned that a one-way quantum computer capable of cluster state
+# entanglement together with adaptive arbitrary single-qubit measurements allows for universal
+# quantum computation. The MBQC framework is a powerful quantum computing approach, particularly
+# useful in platforms that allow for many expendable qubits, and offers
```
ilan-tz, line 648: Consider: platforms with flying qubits and easy physical entangling gates

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+# several advantages over the gate-based model. For example, it circumvents the need for applying
+# in-line entangling gates which are often the most noisy operations in gate-based quantum computers
+# with trapped-ions or superconducting circuits. Instead, the required entanglement is created
+# off-line which is often simpler to implement.
```
ilan-tz, line 652: For photonics: the depth of the optical circuit should not increase with the depth of the logical circuit because (a) the losses become intolerable, and (b) for encodings with probabilistic gates, e.g. dual rail entangling gates, the success probability of the computation will exponentially decay

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#
+# In this demo, we assumed that the system is capable of performing arbitrary
+# single-qubit measurements. This is not a strict requirement, as one can acquire the same
+# capabilities by sprinkling *magic states* into the cluster state. A discussion of this topic is
+# beyond the scope of this tutorial, but a good place to start is `this
+# paper <https://arxiv.org/abs/quant-ph/0403025>`_ [#MagicStates]_.
```
ilan-tz, line 658: Ah, I see you address this. I think it may be worthwhile to insinuate this earlier on

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#
+# Xanadu's approach toward a universal quantum computer involves *continuous-variable* cluster states
+# [#CV-MBQC]_. If you would like to learn more about the architecture, you can read our blueprint
+# papers [#XanaduBlueprint]_ and [#XanaduPassiveArchitecture]_. We highly recommend watching `this
+# video <https://www.youtube.com/watch?v=SD6TH7GZ1rM&feature=emb_title>`_ outlining the main ideas.
+# On the hardware side, efforts are made to develop the necessary technology. This includes the recent `Borealis
+# experiment <https://xanadu.ai/blog/beating-classical-computers-with-Borealis>`_ [#Borealis]_ where
+# a 3-dimensional photonic graph state was created that was used to demonstrate quantum computational advantage.
```
ilan-tz, line 666: image

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#
+
+import networkx as nx
+import matplotlib.pyplot as plt
+
+a, b = 1, 5  # dimensions of the graph (lattice)
+G = nx.grid_graph(dim=[a, b])  # there are a * b qubits
+
+plt.figure(figsize=(5, 1))
+nx.draw(G, pos={node: node for node in G}, node_size=500, node_color="black")
```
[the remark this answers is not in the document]
soosub: I think that might complicate things too much.

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+
+##############################################################################
+#
+# Observe that the structure of the circuit is fairly simple. It only requires Hadamard
+# gates on each qubit and then a controlled-:math:`Z` gate between connected qubits. This part of
+# the computation is not actually computing anything. In fact, aside from the width and depth of
+# the desired quantum circuit, the cluster state generation is essentially independent of the
+# calculation. If you have a reliable way of applying these two operations (Hadamard and
+# controlled-:math:`Z`), you are ready for the next step: worrying about conditional single-qubit
+# measurements.
```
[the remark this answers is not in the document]
soosub: first worry, then perform 😌

demonstrations/tutorial_mbqc.py
soosub: What do you mean? Don't you think graph state is the right term? @ilan-tz

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+#
+# .. _fig-surfacecode:
+#
+# .. figure:: ../demonstrations/mbqc/surface_code_d3.png
+#    :align: center
+#    :width: 50%
+#
+#    ..
+#
+#    A distance-3 surface code. Circles represent qubits and bubbles represent operators, called
```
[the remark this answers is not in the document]
soosub:
```suggestion
#    A distance :math:`d=3` surface code. Circles represent qubits and bubbles represent operators, called
```

demonstrations/tutorial_mbqc.py
soosub: @josh146 do you happen to know if this is possible? It seems like the plotly output is not shown 😢

soosub: I guess this might be related: https://github.com/readthedocs/sphinx_rtd_theme/issues/788

requirements.txt
```diff
@@ -46,3 +46,5 @@ pyyaml==5.4.1
 jax==0.3.15
 jaxlib==0.3.15
 neural-tangents==0.6.0
+git+https://github.com/XanaduAI/flamingpy@mbqc-demo
```
soosub, line 49: > **Warning**: this will fail once the branch is deleted. I opened up a [PR here](https://github.com/XanaduAI/flamingpy/pull/118) to get the changes from the `mbqc-demo` branch in flamingpy, but eventually we'd want to use the `pip` version anyway.

demonstrations/tutorial_mbqc.py
soosub: I think it makes sense to assume that ability here, and then later state that this is not a strict requirement. What do you think?

soosub: https://github.com/PennyLaneAI/qml/pull/485/files#r955431398 nevermind 😌

soosub: I mention the assumption now in the section on universality 👍🏼

demonstrations/tutorial_mbqc.py
```diff
@@ -0,0 +1,818 @@
...
+# section, we will tackle the remaining parts concerning logical operations.
+#
+
+##############################################################################
+#
+# .. _single-qubit-rotations:
+#
+# Single-qubit rotations
+# ```````````````````````
+# Arbitrary single-qubit rotations are an essential operation for a universal quantum computer. In
```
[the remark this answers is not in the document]
soosub: imo that's just a different interpretation - they are just implemented in a different way (and still essential)

requirements.txt
soosub: Fixed https://github.com/XanaduAI/flamingpy/pull/118

requirements.txt
```diff
@@ -46,3 +46,5 @@ pyyaml==5.4.1
 jax==0.3.15
 jaxlib==0.3.15
 neural-tangents==0.6.0
+git+https://github.com/XanaduAI/flamingpy
+plotly>=4.5.0
```
soosub, line 50: Plotly can be uninstalled when FlamingPy is installed with `pip`. However, this should wait until this month's (August 2022) release of FlamingPy.

demonstrations/tutorial_mbqc.py
soosub: Just had a talk with Raf: he agreed and I will just drop the connection with the Borealis experiment

demonstrations/tutorial_mbqc.py
soosub: @radras

soosub:
@KetpuntoG @isaacdevlugt I addressed all the comments. There are little things that @radras might want to address still, but I don't think they are strictly necessary. We could always update the demo in the future. I will leave the finalisation in your capable hands 👍🏼

KetpuntoG:
Don't worry about updating the branch @soosub , the demo is perfect 😄 when we upload it we will update everything

soosub:
@KetpuntoG I was planning to address [this ](https://github.com/PennyLaneAI/qml/pull/485#discussion_r956315514) now the FlamingPy August release is out but there [seem to be some issues](https://xanadu-quantum.slack.com/archives/C03MEBR3T5E/p1662741868767199?thread_ts=1661910934.558379&cid=C03MEBR3T5E) 🥵

KetpuntoG:
> @KetpuntoG I was planning to address [this ](https://github.com/PennyLaneAI/qml/pull/485#discussion_r956315514) now the FlamingPy August release is out but there [seem to be some issues](https://xanadu-quantum.slack.com/archives/C03MEBR3T5E/p1662741868767199?thread_ts=1661910934.558379&cid=C03MEBR3T5E) 🥵

Ah okey! Let's see if Narinam answer you :)

isaacdevlugt (approved):
My comments were addressed or resolved!

KetpuntoG:
@soosub , @radras thanks for this great demo! 
Tomorrow we will announce the release of this demo on Pennylane twitter 🚀

PR #491, opened 2022-05-18
Added required_context to deployment creation to stop it from failing if unrelated workflows fail

rashidnhm:
**Title:** Added required_context to deployment creation to stop it from failing if unrelated workflows fail

**Summary:**
This pull request resolves the error in deploy-pr where workflows extraneous to build-pr failing caused the deployment creation to fail. This is due to the reason that by default if the `required_context` is empty, GitHub will check to ensure all status checks pass before creating the deployment.

Previous failure:
- [deploy-pr](https://github.com/PennyLaneAI/qml/actions/runs/2346363675)

**Relevant references:**
- [Octokit docs](https://octokit.github.io/rest.js/v18#repos-create-deployment)
- [Commit statuses](https://docs.github.com/en/rest/commits/statuses)

**Possible Drawbacks:**
If any build related to the the deploy (in other words, any builds in `build-pr` workflow) fails, then the docs will not be deployed. This is as expected since we need sphinx-build to pass successfully before the docs can be fully deployed.

**Related GitHub Issues:**
N/A


josh146 (approved):
Thanks @rashidnhm!

PR #502, opened 2022-06-12
Fix tutorial for error mitigation

rmoyard:
**Summary:**

Remove `.numpy()` from qnode results.


albi3ro (approved):
Just wondering why it was there in the first place and what it broke, but this seems fine.

rmoyard:
> Just wondering why it was there in the first place and what it broke, but this seems fine.
My guess is the refractor of mixed device for autodiff

PR #503, opened 2022-06-13
Update the drawn circuit in quantum volume demo

antalszava:
* Updates the drawn circuit;
* Doesn't update any other output as the existing outputs seem like outputs one might still be getting.


albi3ro (approved):
👍 
Looks like there is some randomness without a controlled seed, but that should be fine.

Issue #508, opened 2022-06-14
Functionality for parallel `qml.expval(H)` execution needed


Qottmann:
In the [VQE with parallel QPUs on Rigetti Forest](https://pennylane.ai/qml/demos/vqe_parallel.html) demo, deprecated `qml.ExpvalCost` functionality is used that makes use of internally parallelizing the executions with `dask`. Currently, something similar is not available with `qml.expval(H)`, and, more importantly, not possible manually with user-facing functions and dask.

@antalszava and I concluded that there is no point to re-write the `ExpvalCost` logic with non-user-facing functions now, but we should instead find possibilities to provide parallelization possibilities with `qml.expval(H)` in the near future.

For now, we leave the demo as is with a warning about the deprecation, see https://github.com/PennyLaneAI/qml/pull/506

josh146:
@Qottmann I imagine the closest approach would be to do something like:

```python
@qml.qnode(dev)
def circuit(x, h):
    qml.RX(x, wires=0)
    qml.RY(x * 2, wires=0)
    return qml.expval(h)

H = qml.PauliZ(0) + qml.PauliX(1)

results = [dask.delayed(circuit)(0.2, h) for h in H.ops]
results = dask.compute(*results, scheduler="threads") @ H.coeffs
```

Would this be sufficient in the tutorial? It avoids `ExpvalCost`, while making the Dask usage explicit.

Qottmann:
In principle this would work, but do you see a way to incorporate the measurement optimization as well? Already in this example it is executing 2 expvals whereas only 1 is necessary. In many Hamiltonians you have big commuting Pauli groups, so executing _all individual expvals in parallel_ is most likely slower (and wastes unnecessary hardware resources) than just default execution: 
```python
Zs = [qml.PauliZ(i) for i in range(10)]
Xs = [qml.PauliZ(i) for i in range(10)]
H = qml.Hamiltonian(coeffs = np.arange(20), observables = Zs + Xs, grouping_type="qwc")

results = [dask.delayed(circuit)(0.2, h) for h in H.ops]

>>> %timeit result = H.coeffs @ dask.compute(*results, scheduler="threads")
73.9 ms ± 5.17 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

>>> %timeit circuit(0.2, H)
11.2 ms ± 289 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

[...]

PR #510, opened 2022-06-15
Updating the VQE parallel demo with `qml.expval` functionalities and explicit dask

Qottmann:
Since `qml.ExpvalCost` is deprecated, needed to update this demo, see https://github.com/PennyLaneAI/qml/issues/508.

The current version performs the following:

1. "naive" execution using expval
2. parallelized execution using dask
3. parallelized execution using dask and perfoming optimized measurements

Curiously, the "naive" execution is actually very good and takes only 30 sec, whereas if I run the old demo using `ExpvalCost` I get around 200 sec. This creates the paradoxical situation that the parallelized execution is actually slower. Am I unintentionally cheating here and unknowingly have some optimizations in the "naive" version that creates this speedup? Or vice versa, am I doing something wrong in the dasking that slows it down?


demonstrations/vqe_parallel.py
```diff
@@ -251,37 +297,51 @@ def calculate_surface(parallel=True):
...
+#    6 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 1.3 Å
+#    7 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 1.5 Å
+#    8 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 1.7 Å
+#    9 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 1.9 Å
+#    10 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 2.1 Å
+#    Evaluation time: 26.51 s
+
 
 ##############################################################################
 # We have seen how a :class:`~.pennylane.QNodeCollection` can be evaluated in parallel. This results
```
rmoyard: Maybe mention here that it is parallel and optimization with grouping

demonstrations/vqe_parallel.py
```diff
@@ -251,37 +297,51 @@ def calculate_surface(parallel=True):
...
+#    4 / 10: Sequential execution; Running for inter-atomic distance 0.9 Å
+#    5 / 10: Sequential execution; Running for inter-atomic distance 1.1 Å
+#    6 / 10: Sequential execution; Running for inter-atomic distance 1.3 Å
+#    7 / 10: Sequential execution; Running for inter-atomic distance 1.5 Å
+#    8 / 10: Sequential execution; Running for inter-atomic distance 1.7 Å
+#    9 / 10: Sequential execution; Running for inter-atomic distance 1.9 Å
+#    10 / 10: Sequential execution; Running for inter-atomic distance 2.1 Å
+#    Evaluation time: 39.33 s
 #
 #    Evaluating the potential energy surface in parallel
```
rmoyard, line 313: I wonder why it is doing so poorly in parallel... Do you think it is still relevant ro keep it, or maybe we should remove this part?

rmoyard (commented):
Thank you for fixing the demo @Qottmann! Some questions about the parallel performance, otherwise I am waiting for someone else to review it as well before approving.

demonstrations/vqe_parallel.py
Qottmann: Yes this was a relic from the old version, I changed that :)

demonstrations/vqe_parallel.py
Qottmann: Good question! I think in combination with measurement optimization it is still relevant. And also maybe those examples are not ideal to showcase the power since "naive" expval is already very fast. It might be worth to generally overhaul this demo and rewrite it into a general paralellization / dasking demo /how-to. The quantum chemistry part here seems irrelevant to me.

Qottmann:
I noticed I can get further speed-ups when using `grouping_type="commuting"`, which results in fewer groups. Is there any restriction for real devices that I cant see at the moment for using "commuting" instead of qubit-wise-commuting "qwc"?

demonstrations/vqe_parallel.py
rmoyard: The combination is relevant indeed! I think we should explain why it is slower if we want to keep parallel 👍

Qottmann:
Conclusions:
For now, it is fine as is as it shows how to dask which can be valuable. The example is sub-optimal because actually the speed-up from just dasking is negligible. However, as seen in the example below, also the opposite can be true so it is worth to keep both examples. In the future, we it would be worth it to write a dedicated demo on parallelization.

Here is an example where the opposite is true. The Hamiltonian is actually the same, but the circuit is different. The functions for `parallel` and `parallel_optimized` execution are the same as in the demo.

```python
import time
import dask
from pennylane import numpy as np
import pennylane as qml

devs = [qml.device("forest.qvm", device="4q-qvm") for _ in range(20)]

symbols, coordinates = (['H', 'H'], np.array([0., 0., -0.66140414, 0., 0., 0.66140414]))
H, n_wires = qml.qchem.molecular_hamiltonian(symbols, coordinates)
def circuit(params, H):
    qml.StronglyEntanglingLayers(params, wires=range(n_wires))
    return qml.expval(H)

params = np.random.rand(2,n_wires,3)

# Vanilla execution
dev = devs[0]

t0 = time.time()
E_seq = qml.QNode(circuit, dev)(params, H)
dt_seq = time.time() - t0

print(f"Evaluation time with standard expval: {dt_seq:.2f} s")


# dasked execution
def compute_energy_parallel(H, devs, param):
    results = []

    for i in range(len(H.ops)):
        qnode = qml.QNode(circuit, devs[i%len(devs)])
        results.append(dask.delayed(qnode)(param, H.ops[i]))

    result = H.coeffs @ dask.compute(*results, scheduler="threads")
    return result


t0 = time.time()
E_par = compute_energy_parallel(H, devs, params)
dt_par = time.time() - t0

print(f"Evaluation time with dask: {dt_par:.2f} s")


# dasked and optimized execution
devs = [qml.device("forest.qvm", device="10q-qvm") for _ in range(5)]
def compute_energy_parallel_optimized(H, devs, param):
    results = []

    obs_groupings, coeffs_groupings = qml.grouping.group_observables(H.ops, H.coeffs, "qwc")

    for i, (obs, coeffs) in enumerate(zip(obs_groupings, coeffs_groupings)):
        H_part = qml.Hamiltonian(coeffs, obs)
        qnode = qml.QNode(circuit, devs[i%len(devs)])
        results.append(dask.delayed(qnode)(param, H_part))

    result = qml.math.sum(dask.compute(*results, scheduler="threads"))
    return result

t0 = time.time()
E_par_opt = compute_energy_parallel_optimized(H, devs, params)
dt_par_opt = time.time() - t0

print(f"Evaluation time with dask and measurement optimization: {dt_par_opt:.2f} s")
```

output:
```pycon
Evaluation time with standard expval: 9.14 s
Evaluation time with dask: 2.18 s
Evaluation time with dask and measurement optimization: 12.10 s
```

(no significant difference between using "commuting" instead of "qwc", also, no significant difference between using vanilla expval and doing the qwc grouping with vanilla expval.

rmoyard (approved):
Thank you for updating the demo and the PR, it looks good to me 💯 I think we could come back to it for a more impactful example, but update is still valuable in itself for showcasing dask and qvm. Wait @antalszava or @josh146 approval to merge 👍

PR #511, opened 2022-06-15
Match plot legend in benchmarking code to demo [adjoint differentiation]

dime10:
Match order of plot labels in the code to those currently displayed in the demo.


albi3ro (approved):
Wonder how that difference slipped in, but good catch anyway.

PR #512, opened 2022-06-15
Fix randomly generated graph to be inline with seed [learning 2 learn]

dime10:
Fix randomly generated graph to be the one obtained when running the entire demo once with the initial seed.


rmoyard (approved):
Thanks @dime10, great catch 💯

PR #527, opened 2022-06-27
Add author bios to tensor network circuits tutorial

DSGuala:
**Title:**
Add short biographical information to the tensor network quantum circuits demo.


soranjh (approved):
Thanks @DSGuala, looks good to me. Might consider:

1. Updating "Last updated: 17 March 2022."
2. Removing the space before the superscript, e.g. `Diego Guala* :superscript:` to `Diego Guala*:superscript:` etc. Not sure if it works properly though.

PR #532, opened 2022-07-26
Removed search page from theme

rashidnhm:
**Title:** Fix soft 404 error for QML site

**Summary:**
Currently, Google Search Console raises 404 errors for certain pages on the QML site. If you look at the discovery source of both the 404 pages, it is the sitemap.xml file. 

Since we are building the docs with sphinx-build, any file that sphinx processes and renders, it will put it in the sitemap that is generated at the end. However, if those pages don't exist or don't respond properly, then Google marks them with the soft 404 flag.

This pull request introduces a new step in the build pipeline that removes these files from all nodes that build the docs. In addition to deleting the html files, the sitemap that is generated by sphinx is also updated in place to remove the references to these files. I could not find any good way to tell sphinx-build to not generate these files, so this was the second best option.

**Relevant references:**
Image references from Google Console:
image
image

**Possible Drawbacks:**
search.html and sg_execution_times.html pages will no longer exist in the QML site. But we were already not supporting these pages so it should not be of any concern.

**Related GitHub Issues:**
N/A


rashidnhm:
As a reference for reviewers: If you take a look at the sitemap for this PRs deployment, the pages in question are not listed: https://qml-build-previews.pennylane.ai/pull_request_build_preview/532/sitemap.xml

josh146 (approved):
Thanks @rashidnhm! Looks great to me, no major comments or suggestions

smokingroosters (approved):

PR #534, opened 2022-08-02
Update Jax to `0.3.15` for fixing general_param_shift demo

rmoyard:
**Changes:**

Update `jax`, `jax-lib` and `neural-tangents` to the latest version.

PR #537, opened 2022-08-04
Update build pipeline invalidate cache

rashidnhm:
**Title:** Update the QML Build pipeline to ignore the GitHub Actions cache under certain scenarios

**Summary:**
This PR introduces two scenarios where the Python and Sphinx Cache is ignored and all demos are re-run for a given pull request:

A) A pull request updates `requirements*.txt`
In this case, it will invalidate the Python cache and sphinx cache, but caching itself will not be disabled. The very first build of the the pull request will occur without the use of cache, once the build is successful, the cache for that branch will be updated with the new build items. Subsequent pushes to that branch will use the new cache. If any of the requirements files are updated, the cache is invalidated and the cycle restart.

The reason for this behaviour is because the hash of the requirement files are used as [part of the key for the Python package caches](https://github.com/PennyLaneAI/qml/blob/48eb4d3c529b9737a61c7cb2cdcd2f0de8374bb3/.github/workflows/build-pr.yml#L106). If the requirement files change, so do the hash and there is a cache-miss. Which is then updated upon a successful (This is default caching behaviour in CI/CD)

B) A pull request is assigned the new label `ignore-qml-cache`
This is a special label to completely turn off the use of caching for a pull request. It may be important to re-run all demos due to changes in a pull request even if it does not change the requirements files. 

Once this label is attached to a PR, all subsequent builds of that PR will ignore the GitHub cache completely.

**Relevant references:**
- Need to ignore cache in #534 
- sc-24106

**Possible Drawbacks:**
Previous functionality still preserved.

**Related GitHub Issues:**
None.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -45,6 +45,8 @@ jobs:
     strategy:
       matrix:
         offset: ${{ fromJson(needs.compute-build-strategy-matrix.outputs.strategy-matrix) }}
+    env:
+      ignore_cache: ${{ contains(github.event.pull_request.labels.*.name, 'ignore-qml-cache') }}
     steps:
       - name: Checkout PR
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


rashidnhm:
Notes for reviewers:
After adding the `ignore-qml-cache`, the following [build](https://github.com/PennyLaneAI/qml/actions/runs/2798895749) did not use the GitHub cache and re-ran all demos.

josh146 (approved):
Thanks Rashid, both workflows look really good to me! Especially the fact that the requirements one rebuilds the cache once only.

antalszava (approved):
Looks good :+1:

PR #543, opened 2022-08-09
Fix QML pipeline bug to push changes to appropriate environment site

rashidnhm:
**Title:** Update QML Build pipeline to fix bug where PR merge to either master or dev branch would always push changes to Prod S3 bucket

**Summary and RCA of bug**
The current QML pipeline attempts to push the changes from a sphinx build to either the QML Prod or Dev sites based on branch. Determination of that logic was done at [this step](https://github.com/PennyLaneAI/qml/blob/23977d59e29f1095fc264405231548014d395a86/.github/workflows/deploy-pr.yml#L90-L96). The `GITHUB_REF` variable generally contains the branch that the update was pushed for, however, in this case the `deploy-pr` workflow is not triggered directly from a push; `build-pr` is triggered from a push and `deploy-pr` runs from a [workflow_run call](https://github.com/PennyLaneAI/qml/blob/23977d59e29f1095fc264405231548014d395a86/.github/workflows/deploy-pr.yml#L2-L7).

In other words, the push context (information regarding which branch was updated) is not available to `deploy-pr`, it runs from the default branch. This is why the step referring to `GITHUB_REF` always resolved to master, and all changes were pushed to Prod.

This pull request resolves this issue by adding a new step to the `build-pr` workflow, it creates an artifact with the required information from the push context. That artifact is then downloaded by `deploy-pr` and parsed. 

**Relevant references:**
- sc-24324

**Possible Drawbacks:**
The master branch may need to be merged to dev afterwards.

**Related GitHub Issues:**
None.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -18,7 +18,7 @@ jobs:
   compute-build-strategy-matrix:
     runs-on: ubuntu-20.04
     steps:
-      - name: Checkout PR
+      - name: Checkout
         uses: actions/checkout@v3
         with:
           fetch-depth: 1
@@ -48,7 +48,7 @@ jobs:
     env:
       ignore_cache: ${{ contains(github.event.pull_request.labels.*.name, 'ignore-qml-cache') }}
     steps:
-      - name: Checkout PR
+      - name: Checkout
         uses: actions/checkout@v3
         with:
           fetch-depth: 1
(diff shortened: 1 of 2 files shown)


.github/workflows/build-pr.yml
```diff
@@ -228,23 +228,41 @@ jobs:
...
+      - name: Save Pull Request Event Context
         if: github.event_name == 'pull_request' && matrix.offset == 0
         run: |
           mkdir -p /tmp/pr
-          cat >/tmp/pr/pr_info.json <<EOL
+          cat >/tmp/pr/info.json <<EOL
           {
             "id": "${{ github.event.pull_request.number }}",
-            "title": "${{ github.event.pull_request.title }}",
-            "author": "${{ github.event.pull_request.user.login }}",
```
antalszava, line 239: Why were these removed?

.github/workflows/build-pr.yml
```diff
@@ -228,23 +228,41 @@ jobs:
...
         if: github.event_name == 'pull_request' && matrix.offset == 0
         uses: actions/upload-artifact@v3
         with:
           name: pr_info.zip
           path: /tmp/pr
           retention-days: 30
+
+      # Will run to create an artifact containing key push event information
+      - name: Save Push Event Context
+        if: github.event_name == 'push' && matrix.offset == 0
```
antalszava, line 253: Just to make sure: does `matrix.offset == 0` mean that the main process is being run (i.e., not any of the workers)?

antalszava (approved):
Looks good! :clap: Thank you for resolving this so quick! :1st_place_medal: :fireworks:

.github/workflows/build-pr.yml
rashidnhm: They are not required by the `deploy-pr` workflow. And rendering the title like that also left us open to unexpected code injection if someone were to craft a malicious title to a PR. Overall thought it was better to remove since we don't use it for anything.

.github/workflows/build-pr.yml
rashidnhm: `matrix.offset == 0` means that the first worker only would execute that step. Reason for having it this way is because there are certain artifacts that only need to be created once instead of being created by each worker.

josh146 (approved):
Thanks Rashid for fixing this so quickly!

PR #549, opened 2022-08-15
Replaced Hermitian call with PauliZ in expval

quantshah:
**Title:**Replaced Hermitian call with PauliZ in expval

**Summary:** The Hamiltonian construction in the circuit uses the old Hermitian() call which makes it difficult to JIT the QAOA circuit in JAX. Additionally, the new Hamiltonian/PauliZ class could possibly be more efficient than the Hamiltonian construction from a matrix. It is a simple update.

@josh146


josh146 (approved):
Perfect, thanks @quantshah!

PR #556, opened 2022-08-22
Update docs URL

josh146:


rashidnhm (approved):
I see the nodes passing so this should be the fix we needed. I've also made the `latest` build version of pennylane public so other references to it will work as well.

PR #563, opened 2022-09-07
fix typo

nickdgardner:


josh146 (approved):
Thanks for catching this @nickdgardner!

PR #570, opened 2022-09-12
print actual final energies for VQE with SPSA

timmysilv:
**Summary:**
The VQE with SPSA final output being shown was actually the final output from the GradientDescentOptimizer. Also, only the second-to-last energy was being shown. I'm updating it to show the last energy in both cases.

**Relevant references:**
N/A

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A


rmoyard:
Hi @timmysilv, thanks for the PR! Could you merge the latest changes of dev and also target dev for this PR? (dev is using master version of Pennylane)

rmoyard:
Don't worry about that, we will remove it separately. Thank you!

demonstrations/spsa.py
```diff
@@ -615,10 +620,12 @@ def callback_fn(xk):
     callback=callback_fn,
 )
 
+final_energy = h2_spsa_energies_ibm[-1]
```
rmoyard, line 623: Very good catch here!

rmoyard (approved):
Thanks for the update and the good catch for the last final energy 💯

PR #575, opened 2022-09-14
Gradient method changed PyTorch_noise demo

Jaybsoni:
**Summary:**
The method for computing the gradient has changed, updated the demo to match this.

PR #576, opened 2022-09-14
Update the print output with new Gradient func in PyTorch Noise Demo

Jaybsoni:
Update the print output with new Gradient func in PyTorch Noise Demo. 

Same changes as [this PR](https://github.com/PennyLaneAI/qml/pull/575)


rmoyard (approved):
💯

PR #583, opened 2022-09-23
QSP: remove comment about reverse order

timmysilv:
~It doesn't truly make a difference, but we should apply the angles in reverse order.~
EDIT: don't wanna update values (below comment aged poorly) so I'm just removing the comment.

My Mac couldn't run the whole demo, but it only froze many lines after the change so we good. Also TIL `torch.tensor` can't have a negative step (I tried `phi[:0:-1]` and it failed 😢)

### DIFF
--- a/demonstrations/function_fitting_qsp.py
+++ b/demonstrations/function_fitting_qsp.py
@@ -191,11 +191,11 @@ def QSP_circ(phi, W):
     representation of the final unitary are polynomials!
     """
     qml.Hadamard(wires=0)  # set initial state |+>
-    for i in range(len(phi) - 1):  # iterate through rotations in reverse order
-        qml.RZ(phi[i], wires=0)
+    for angle in phi[1:].flip(0):  # iterate through rotations in reverse order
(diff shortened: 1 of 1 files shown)


Jaybsoni (approved):
Looks good! Thanks for catching this 💯 !

demonstrations/function_fitting_qsp.py
```diff
@@ -191,8 +191,8 @@ def QSP_circ(phi, W):
     representation of the final unitary are polynomials!
     """
     qml.Hadamard(wires=0)  # set initial state |+>
-    for i in range(len(phi) - 1):  # iterate through rotations in reverse order
-        qml.RZ(phi[i], wires=0)
+    for angle in phi[:-1]:
```
Jaybsoni, line 194: Just to confirm, I always forget, but this does not include the final element correct?

timmysilv: correct!
```python
>>> [3,4,5][:-1]
[3, 4]
```

PR #585, opened 2022-09-23
Add more youtube videos

isaacdevlugt:


isaacdevlugt:
General question for @josh146. How do you want to handle the QHack 2021 & 2022 videos? Should QHack have its own separate URL just organize things a bit better?

josh146:
@isaacdevlugt how do you mean? There are several options I can think of:

- You could add sections to the existing video page using standard ReST syntax

  ```rest
  QHack 2021
  ----------

  .. video:: dsfkjhadfjklh
      :author: ahsadfjknvka
  ```

- You could create a new page (a new `rst` file), and include it in the `videos.rst` table of contents.

isaacdevlugt:
@josh146 I'm good with both of those options! I think I prefer the second?

alvaro-at-xanadu (approved):

PR #605, opened 2022-10-12
[sc-26860]: Update QML cache strategy to allow for refreshing of cache

rashidnhm:
**Title:** Add ability to refresh cache for QML CI/CD

**Summary:**
This pull request aims to add a new feature to the QML CI/CD pipeline by allowing the refreshing of cache. Previously, you either had to use existing cache, or completely ignore cache. There was no way to update/refresh the cache without using workarounds.

After this pull request is merged, users will be able to use the existing `ignore-qml-cache` label which now build fresh each time, but also refresh the cache with the new results each time. Once the label is removed, the most recent cache will be used for subsequent build.

Please note that this is vastly different from prior behavior. Before this PR, if the `ignore-qml-cache` was attached to a PR, it would ignore the cache but never update it. So removing the label was never an option to use the cache again, which slowed down a lot of PRs.

This pull request brings the following key changes:
- The Python venv will no longer be cached
- The cache key for items has been updated

Pros:
- Builds will always fetch the latest libraries
- Initial build of PRs will fetch the latest libraries at the time the PR is opened.
- No further changes are required in PRs to support a cache refresh. The existing `qml-ignore-cache` label can be used.

Cons:
- Build times will increase by 3-7 minutes

**Relevant references:**
- sc-26860

**Possible Drawbacks:**
Please see Cons list above.

Since this PR updates most cache keys, once it is merged, all existing caches will be invalidated. This should not impact anything but expect a longer build time on the very first build .. similar situation to when this workflow was first launched.

**Related GitHub Issues:**
None.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -16,7 +16,7 @@ concurrency:
 jobs:
   # Generates a JSON list that is used by the strategy.matrix of the build job to spawn multiple workers
   compute-build-strategy-matrix:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -24,22 +24,22 @@ jobs:
           fetch-depth: 1
 
       - name: Setup Python
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
       - name: Generate Build Matrix
         id: compute-strategy-matrix
         run: |
-          echo "::set-output name=strategy-matrix::$(python3 .github/workflows/github_job_scheduler.py \
+          echo "strategy-matrix=$(python3 .github/workflows/github_job_scheduler.py \
             build-strategy-matrix \
             ${{ github.workspace }} \
-            --num-workers=${{ env.NUM_WORKERS }})"
+            --num-workers=${{ env.NUM_WORKERS }})" >> $GITHUB_OUTPUT
     outputs:
       strategy-matrix: ${{ steps.compute-strategy-matrix.outputs.strategy-matrix }}
 
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     needs:
       - compute-build-strategy-matrix
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-pr.yml
```diff
@@ -31,10 +31,10 @@ jobs:
       - name: Generate Build Matrix
         id: compute-strategy-matrix
         run: |
-          echo "::set-output name=strategy-matrix::$(python3 .github/workflows/github_job_scheduler.py \
```
rashidnhm, line 34: This is being done because `::set-output` [has been deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

.github/workflows/build-pr.yml
```diff
@@ -149,14 +154,21 @@ jobs:
...
+            sphinx-${{ hashFiles('matrix_info.txt') }}-${{ github.ref_name }}-
+            sphinx-${{ hashFiles('matrix_info.txt') }}-
+
+      - name: Clear Cache
+        if: env.ignore_cache == 'true'
+        env:
+          sphinx_cache_filename: sphinx_cache-${{ hashFiles('matrix_info.txt') }}
+        run: |
+          if [ -d demos ]; then rm -rf demos; fi
+          if [ -d ${{ env.sphinx_cache_filename }} ]; then rm -rf ${{ env.sphinx_cache_filename }}; fi
```
rashidnhm, line 171: Instead of not restoring the cache at all if the `ignore-qml-cache` label is attached (previous behavior). Now the cache is restored under all circumstances.

And then this step will delete the cached paths if the `ignore-qml-cache` label is attached.

This allows subsequent steps to build fresh and the new values are then cached since we did call `action/cache` on the paths. 

And voila! You have yourself a cache refresh

.github/workflows/build-pr.yml
```diff
@@ -55,22 +55,15 @@ jobs:
...
-          # However, caching the venv may cause broken symlinks sporadically when restored.
-          # Once the following issues are resolved, the action version should be updated to provide a performance boost
-          # Ref:
-          # - https://github.com/actions/setup-python/issues/276
-          # - https://github.com/actions/setup-python/issues/330
-          # ---
-          # cache: pip
-          # cache-dependency-path: |
-          #   requirements.txt
-          #   requirements_no_deps.txt
```
rashidnhm, line 73: Removing the caching mention block since we won't be caching the python environment anymore.

josh146 (approved):
Looks good to me @rashidnhm, although better if someone with more github actions knowledge than me can also review 🙂

.github/workflows/build-pr.yml
```diff
@@ -138,30 +115,27 @@ jobs:
...
-          key: gallery-${{ hashFiles('matrix_info.txt') }}-${{ github.ref_name }}-${{ github.sha }}
-          restore-keys: |
-            gallery-${{ hashFiles('matrix_info.txt') }}-${{ github.ref_name }}-
-
       - name: Sphinx Cache
-        if: steps.python_cache.outputs.cache-hit == 'true' && env.ignore_cache != 'true'
         uses: actions/cache@v3
         with:
           path: sphinx_cache-${{ hashFiles('matrix_info.txt') }}
-          key: sphinx-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('matrix_info.txt') }}-${{ github.ref_name }}-${{ github.sha }}
```
smokingroosters, line 156: this has been removed since now it's included in the matrix_info.txt, correct?

smokingroosters (approved):
Just some comment for clarity, but looks goood Rashid!

.github/workflows/build-pr.yml
```diff
@@ -224,13 +196,14 @@ jobs:
             !_build/html/*.xml
             !_build/html/_static
             !_build/html/glossary
-            
 
-      # These two steps are required as the subsequent workflow_run will not have
-      # the current context available to it.
-      # Will run to create an artifact containing key pull_request event information
+  save-build-context:
```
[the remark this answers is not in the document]
rashidnhm: Correct, it's more clear this way on what it does vs having to figure out what `matrix.offset == 0`

.github/workflows/build-pr.yml
rashidnhm: Yup! I added a lot more metadata to the matrix_info file so an update to any of them invalidates the cache.

smokingroosters:
Thanks for the replies @rashidnhm, good to merge for me!

PR #609, opened 2022-10-18
Fix workflows by removing ` --use-feature=2020-resolver`

albi3ro:
Pip 22.3 just dropped, and it removed support for ` --use-feature=2020-resolver`, see the changelog https://pip.pypa.io/en/stable/news/#v22-3 .

To get the workflows running again, this removes that flag from the workflows.

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -33,8 +33,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip wheel
-          pip install -r requirements.txt --use-feature=2020-resolver
-          pip install --no-deps -r requirements_no_deps.txt --use-feature=2020-resolver
+          pip install -r requirements.txt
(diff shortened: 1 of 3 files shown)


rashidnhm:
> This is a PR that might be worth running with a fresh `pip` cache --- I have a suspicious that the new dependency resolver doesn't work with our `requirements.txt`.
> 
> If not, we may have to pin to an earlier version of `pip`

@josh146 The build-pr.yml has not been using the the 2020-resolver and still working. Check [this build](https://github.com/PennyLaneAI/qml/actions/runs/3274753188) as an example.

Also note, once #605 is merged, there won't be a python cache.

josh146:
Thanks @rashidnhm! That's good to know, the reason we originally added `--use-feature=2020-resolver` is that the new resolver previously wasn't working 😅

rashidnhm (approved):

PR #612, opened 2022-10-21
Remove manual brackets

josh146:
Once https://github.com/XanaduAI/xanadu-sphinx-theme/pull/27 is merged, we should remove manual brackets from recent demos, to avoid brackets appearing twice.


KetpuntoG (approved):
Thanks! :)

PR #619, opened 2022-11-07
QSim demo: use gates from PennyLane core instead of PL-Cirq

antalszava:
PennyLane-Cirq had some custom gates that are now in PennyLane. Those gates were recently removed from PennyLane-Cirq and so this PR changes the QSim demo. https://github.com/PennyLaneAI/pennylane-cirq/pull/115


timmysilv:
nice, love to see the plugins being used in all their ability! why are you giving the full urls in requirements.txt instead of the released/pinned versions from before?

antalszava:
@timmysilv great catch :slightly_smiling_face: The PR was supposed to target `dev`, where we pull in the dev version for each PennyLane package.

PR #620, opened 2022-11-07
Ivana adding author bios v02

ikurecic:
Including @KetpuntoG 's for today's demo.


demonstrations/tutorial_multiclass_classification.py
```diff
@@ -14,7 +14,7 @@
    tutorial_variational_classifier Variational classifier
    tutorial_data_reuploading_classifier Data-reuploading classifier
 
-*Author: PennyLane dev team. Posted: 9 Apr 2020. Last updated: 28 Jan 2021.*
+*Author: Safwan Hossein — Posted: 09 April 2020. Last updated: 28 January 2021.*
 
 In this tutorial, we show how to use the PyTorch interface for PennyLane
 to implement a multiclass variational classifier. We consider the iris database
```
KetpuntoG, line 20: Does this demo exist? I dont find it

demonstrations/adjoint_diff_benchmarking.py
```diff
@@ -10,6 +10,7 @@
     :property="og:image": https://pennylane.ai/qml/_static/thumbs/code.png
 
 
+*Author: Christina Lee. Posted: 23 November 2021. Last updated: 23 November 2021.*
 
 """
```
KetpuntoG, line 16: I think the first link should end in html. It has nothing to do with PR but it can be corrected here

demonstrations/tutorial_trapped_ions.py
```diff
@@ -1102,9 +1102,4 @@ def cnot_gate(basis_state):
 #
 # About the author
 # ----------------
```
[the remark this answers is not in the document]
ikurecic: You're right, thanks! It was incorrect in the original file and I didn't check. Thought Alvaro wouldn't make a mistake. 😉

demonstrations/tutorial_multiclass_classification.py
ikurecic: It's outdated and not actively linked anywhere. Accessible [here](https://pennylane.ai/qml/demos/tutorial_multiclass_classification.html).

demonstrations/qnspsa.py
```diff
@@ -6,7 +6,7 @@
     :property="og:description": Introduction to the Quantum natural SPSA optimizer, which reduces the number of quantum measurements in the optimization.
     :property="og:image": https://pennylane.ai/qml/_images/qnspsa_cover.png
 
-*Author: Yiheng Duan (yiheng@amazon.com). Posted: 18 July 2022. Last updated: 5 Sep 2022.*
+*Author: Yiheng Duan (yiheng@amazon.com) — Posted: 18 July 2022. Last updated: 05 September 2022.*
```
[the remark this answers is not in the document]
ikurecic: I left it because the author is an Amazon employee. You can let me know what you think. :)

demonstrations/adjoint_diff_benchmarking.py
ikurecic: Great point! I'll fix it.

ikurecic:
Thanks, @KetpuntoG ! Should be sorted now.

demonstrations/qnspsa.py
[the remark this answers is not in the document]
KetpuntoG: At the end of the bio it would be better maybe I think

KetpuntoG (approved):
After the change of the mailing address (if it's ok with you) it would be ready!

PR #625, opened 2022-11-08
bump mpl; remove duplicate sklearn

timmysilv:
can't install requirements.txt on mac, but I believe CI on pull_request will do it for me to confirm all is well.

Bumping MPL because many demos show warnings with 3.3.3, but they're resolved with 3.5.3 as pointed out by @eddddddy

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,9 +37,8 @@ toml==0.10.2
 torch==1.10.0+cpu
 torchvision==0.11.1+cpu
 dask[delayed]==2.10.1
-scikit-learn==0.23.2
 pyparsing==2.4.7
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -37,9 +37,8 @@ toml==0.10.2
 torch==1.10.0+cpu
 torchvision==0.11.1+cpu
 dask[delayed]==2.10.1
-scikit-learn==0.23.2
```
timmysilv, line 40: duplicate of line 26

eddddddy:
Thanks for this @timmysilv!

albi3ro (approved):
Thanks 🚀

PR #628, opened 2022-11-09
Update Pytorch and noisy devices demo in preparation for 0.27

trbromley:
Note, we should revisit whether the following makes sense to run on a computer that does not have a GPU:
```python
phi = Variable(torch.tensor(1.0, device="cuda"), requires_grad=True)
theta = Variable(torch.tensor(0.05, device="cuda"), requires_grad=True)
```
I get an error on my laptop.

### DIFF
--- a/demonstrations/pytorch_noise.py
+++ b/demonstrations/pytorch_noise.py
@@ -13,7 +13,7 @@ PyTorch and noisy devices
 
    tutorial_noisy_circuit_optimization Optimizing noisy circuits with Cirq
 
-*Author: Josh Izaac — Posted: 11 October 2019. Last updated: 14 September 2022.*
+*Author: Josh Izaac — Posted: 11 October 2019. Last updated: 9 November 2022.*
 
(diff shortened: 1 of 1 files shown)


Jaybsoni (approved):
Looks good to me 👍🏼. Is this device no longer active?

Jaybsoni:
> Note, we should revisit whether the following makes sense to run on a computer that does not have a GPU:
> 
> ```python
> phi = Variable(torch.tensor(1.0, device="cuda"), requires_grad=True)
> theta = Variable(torch.tensor(0.05, device="cuda"), requires_grad=True)
> ```
> 
> I get an error on my laptop.

Were you able to run it elsewhere?

josh146:
Note this is a really old demo, written around PL ~0.3 when we only had 5 or so demos. The purpose was to show that hybrid GPU-QPU computing is possible with PennyLane

josh146:
So less about wanting users to use a GPU specifically, and more about having at least one demo to show how to do it/that it is possible

PR #630, opened 2022-11-10
Qubit tapering demo fails with new validation checks 

Jaybsoni:
**Summary:**
New validation checks in PL cause this demo to fail when constructing sparse hamiltonians. Explicitly added wire value.

### DIFF
--- a/demonstrations/tutorial_qubit_tapering.py
+++ b/demonstrations/tutorial_qubit_tapering.py
@@ -177,8 +177,8 @@ print(H_tapered)
 # representation of Hamiltonians. This allows us to directly diagonalize them to obtain exact values
 # of the ground-state energies.
 
-H_sparse = qml.SparseHamiltonian(qml.utils.sparse_hamiltonian(H), wires=all)
-H_tapered_sparse = qml.SparseHamiltonian(qml.utils.sparse_hamiltonian(H_tapered), wires=all)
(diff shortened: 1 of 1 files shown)


demonstrations/tutorial_qubit_tapering.py
```diff
@@ -177,8 +177,8 @@
 # representation of Hamiltonians. This allows us to directly diagonalize them to obtain exact values
 # of the ground-state energies.
 
-H_sparse = qml.SparseHamiltonian(qml.utils.sparse_hamiltonian(H), wires=all)
-H_tapered_sparse = qml.SparseHamiltonian(qml.utils.sparse_hamiltonian(H_tapered), wires=all)
+H_sparse = qml.SparseHamiltonian(qml.utils.sparse_hamiltonian(H), wires=H.wires)
```
soranjh, line 180: This is the only way qml.SparseHamiltonian works now? Previously, the number of wires was not important, right?

soranjh (approved):

demonstrations/tutorial_qubit_tapering.py
Jaybsoni: Correct, there were some errors raised in other functions when the wires didn't match the matrix provided, so validation checks were added. A check was added to determine if `len(wires)` is not compatible with the matrix shape.

PR #635, opened 2022-11-14
Set up authorship data for Jack Ceroni

ikurecic:
Placeholder image was authored by @branedrane
Actual image taken (with author's permission) from Twitter.


KetpuntoG (approved):
Great! I think you can delete the png image. Otherwise perfect 😄

PR #639, opened 2022-11-15
[sc-29561]: Updated QML pipeline to retain downloadable python files included in demos

rashidnhm:
**Title:** Updated QML pipeline to retain downloadable python files included in demos

**Summary:**
There is a bug currently in the QML pipeline where downloadable content put in _downloads/ by sphinx (with the usage of the `:download:`) is not retained and deployed by the pipeline. This results in errors on the site when downloading files that should exist.

This PR fixes the bug by attempting to retain the downloadable files for each demo.

**Relevant references:**
[sc-29561](https://app.shortcut.com/xanaduai/story/29561/qml-ci-cd-unexpected-deletion-of-downloadable-files)

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Generate Build Matrix
         id: compute-strategy-matrix
         run: |
-          echo "strategy-matrix=$(python3 .github/workflows/github_job_scheduler.py \
+          echo "strategy-matrix=$(python3 .github/workflows/qml_pipeline_utils.py \
             build-strategy-matrix \
             ${{ github.workspace }} \
             --num-workers=${{ env.NUM_WORKERS }})" >> $GITHUB_OUTPUT
@@ -83,12 +83,12 @@ jobs:
           requirements: ${{ hashFiles('requirements.txt') }}
           requirements_no_deps: ${{ hashFiles('requirements_no_deps.txt') }}
           EOL
-          python3 .github/workflows/github_job_scheduler.py \
+          python3 .github/workflows/qml_pipeline_utils.py \
            remove-executable-code \
            ${{ github.workspace }} \
            --num-workers=${{ env.NUM_WORKERS }} \
            --offset=${{ matrix.offset }} \
-           --dry-run | tee -a matrix_info.txt
+           --dry-run >> matrix_info.txt
           cat matrix_info.txt
 
       - name: Install OS build dependencies
@@ -99,7 +99,7 @@ jobs:
       # See documentation in github_job_scheduler.py for more details.
       - name: Remove extraneous executable code from demos
         run: |
-          python3 .github/workflows/github_job_scheduler.py \
+          python3 .github/workflows/qml_pipeline_utils.py \
            remove-executable-code \
            ${{ github.workspace }} \
            --num-workers=${{ env.NUM_WORKERS }} \
@@ -142,7 +142,7 @@ jobs:
       # are published to the live website.
       - name: Update sitemap.xml
         run: |
-          python3 .github/workflows/github_job_scheduler.py \
+          python3 .github/workflows/qml_pipeline_utils.py \
             clean-sitemap \
             ${{  github.workspace }} \
             --html-files="demos/sg_execution_times.html" \
@@ -153,7 +153,7 @@ jobs:
       - name: Clean HTML Files
         if: matrix.offset == 0
         run: |
-          python3 .github/workflows/github_job_scheduler.py \
+          python3 .github/workflows/qml_pipeline_utils.py \
            remove-html \
            ${{ github.workspace }} \
            --num-workers=${{ env.NUM_WORKERS }} \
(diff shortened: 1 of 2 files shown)


.github/workflows/qml_pipeline_utils.py
```diff
@@ -261,6 +261,60 @@ def _calculate_files_to_retain(
...
+        sphinx_role_name: A string to indicate the sphinx role to parse for. Ex: 'doc', 'download'.
+        sphinx_file_encoding: The encoding to use while reading the sphinx file.
+        re_flags: Additional flags to passed to re.compile. Pass flags the same way you would to re.
+                  Ex: flag1 | flag2.
+    """
+    with sphinx_file_location.open("r", encoding=sphinx_file_encoding) as fh:
+        sphinx_file_content = fh.read()
+
+    sphinx_role_pattern = re.compile(
+        fr":{sphinx_role_name}: ?`(.+ ?<(?P<hyperlinked_target>.+)>|(?P<direct_target>.+))`",
```
smokingroosters, line 303: could we add a comment on the regexp? reg how we formulated to extract what we need. pretty hard to read in general, can be good for future maintenance

.github/workflows/qml_pipeline_utils.py
```diff
@@ -454,12 +510,23 @@ def remove_extraneous_html(
             if verbose:
                 print("Deleted", file_name)
 
+    files_downloadable_artifact_targets = [
+        # Get the last part (the actual file name, instead of full path to file)
+        target.split("/")[-1]
```
smokingroosters, line 515: any reason why you are not directly returning the actual file name from the function?

smokingroosters (approved):

.github/workflows/build-pr.yml
```diff
@@ -83,12 +83,12 @@ jobs:
...
           requirements_no_deps: ${{ hashFiles('requirements_no_deps.txt') }}
           EOL
-          python3 .github/workflows/github_job_scheduler.py \
+          python3 .github/workflows/qml_pipeline_utils.py \
            remove-executable-code \
            ${{ github.workspace }} \
            --num-workers=${{ env.NUM_WORKERS }} \
            --offset=${{ matrix.offset }} \
-           --dry-run | tee -a matrix_info.txt
+           --dry-run >> matrix_info.txt
```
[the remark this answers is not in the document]
rashidnhm: Both append to the file but tee also shows what it is appending as log output. This caused a duplicate for certain lines since the file is catted to log anyways.

.github/workflows/qml_pipeline_utils.py
rashidnhm: Yes fair point, I will add that

.github/workflows/qml_pipeline_utils.py
rashidnhm: When sphinx puts a file in the _downloads folder, it does not retain the path that the file came from.

So if you have: `path/to/file/in/subdir/file.py`
It will end up as: `_downloads/{hash generated by sphinx}/file.py`

Due to that, caring about the whole file path does not add much value.

.github/workflows/qml_pipeline_utils.py
josh146: @rashidnhm there are two ways the role could look in sphinx:

```rest
:download:`path/to/file.ext`
```

or 

```rest
:download:`link text <path/to/file.ext>`
```

Does the regex catch both of them? I can't tell 😆

josh146 (approved):
Thanks Rashid!

.github/workflows/qml_pipeline_utils.py
```diff
@@ -261,6 +261,60 @@ def _calculate_files_to_retain(
...
+
+    It behaves as the following:
+
+    :role: `target`
+               ^
+            The value of this target is returned
+
+    :role: `title <target>`
+                    ^
+                The value of this target is returned
```
rashidnhm, line 289: @josh146

.github/workflows/qml_pipeline_utils.py
rashidnhm: Yup! As a matter of fact if you check the docstrings for this function it outlines which value it extracts based on the role syntax used. I've left a comment [here as reference](https://github.com/PennyLaneAI/qml/pull/639/files#r1024717433)

rashidnhm: That being said, and this is something I've had in the back of my mind and also something Alberto has mentioned as well; we should start having unittests for this qml utility file. And also split it up into smaller files.

I will probably make it into a small package and have unittests for all the functions, including tests for this regex

.github/workflows/qml_pipeline_utils.py
josh146: Thanks Rashid, I shouldn't code review so late at night 🤦‍♂️

.github/workflows/qml_pipeline_utils.py
rashidnhm: Comment added which should explain this regex expression cleanly

PR #651, opened 2022-11-24
Update tutorial_noisy_circuits.py

co9olguy:
The current version of the noisy circuits tutorial begins at a critical point in the optimization landscape. Consequently, the optimization never progresses (or converges!). This can be fixed by starting slightly away from a critical point


KetpuntoG (approved):
Thanks @co9olguy ! Now it works fine 💪

PR #654, opened 2022-11-29
fixing python version

KetpuntoG:
It has been noted that some demos are currently failing tests. This is due to some new changes introduced that work from python 3.9 onwards. I have updated our repository version.

### DIFF
--- a/.github/workflows/build-dev.yml
+++ b/.github/workflows/build-dev.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/build-master.yml
+++ b/.github/workflows/build-master.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.9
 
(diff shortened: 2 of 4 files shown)


rashidnhm:
> Looks good! slightly_smiling_face Thanks for catching this.
> 
> There's a comment in the `workflows/qml_pipeline_utils.py` file:
> 
> > ```
> > # Prefixing the default namespace here to the xml tag as it seems to work more steadily
> > # across both Python 3.7 and above. Once QML starts using a higher version of Python then this
> > # method can be converted to passing the namespace to findall using the `namespaces` parameter.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > xml_search_paths = {
> > "url": "url" if not default_ns else f"{{{default_ns}}}url",
> > "loc": "loc" if not default_ns else f"{{{default_ns}}}loc",
> > }
> > @rashidnhm any thoughts on this in light of switching to Python 3.9?

Great catch @antalszava! We should be able to refactor this portion to use the `namespaces` paramater introduced in 3.8+. That being said I wouldn't do it as part of this PR. I am currently working on a re-design of the qml_pipeline_utils as part of [sc-29911](https://app.shortcut.com/xanaduai/story/29911/split-qml-pipeline-utils). I will update that code portion as part of that story :+1: 

The way the it is coded now should continue to work in 3.9. I had tested it up to 3.10 during development.

PR #660, opened 2022-12-09
Remove the build-master and build-dev workflows and introduce a new build-branches workflow

rashidnhm:
**Title:** Remove the build-master and build-dev workflows and introduce a new build-branches workflow

**Summary:**
This pull request removes the existing build-master and build-dev workflows and adds a new build-branches workflow.

This is being done since the build-master and build-dev workflows are identical with the exception of the branch they checkout, this is much better achieved with a strategy matrix as it avoids the code duplication.

It also adds a new step to the workflows where the `sg_execution_times.html` file that is generated by sphinx is parsed. This is a file that contains how long it took sphinx to execute the code in each demo. This information is converted to a easily digestable JSON format and uploaded as an artifact. In a subsequent PR, `build-pr.yml` will be uploaded to use this information to better distribute the demo load across workers.

**Relevant references:**
- [sc-29723](https://app.shortcut.com/xanaduai/story/29723/qml-implement-job-scheduling-based-on-load)

**Possible Drawbacks:**
No new drawbacks other than the current workflows are being deleted. However the functionality in all aspect is retained.

**Related GitHub Issues:**
None.


.github/workflows/build-branches.yml
```diff
@@ -0,0 +1,62 @@
+name: Build QML Branches
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 0,2,4,6' # At 00:00 on Sunday, Tuesday, Thursday, and Saturday.
+
+concurrency:
+  group: build-docs-default-branches-${{ github.ref }}
+  cancel-in-progress: true
```
smokingroosters, line 9: def nicer

.github/workflows/build-dev.yml
```diff
@@ -1,50 +0,0 @@
...
-
-      - uses: actions/checkout@v2
-        with:
-          ref: dev
-
-      - name: Run Rigetti Quilc
-        run: docker run --rm -d -p 5555:5555 rigetti/quilc -R
-
-      - name: Run Rigetti QVM
-        run: docker run --rm -d -p 5000:5000 rigetti/qvm -S
```
smokingroosters, line 26: is this not needed anymore?

.github/workflows/build-branches.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Install QML Pipeline Utils
+        run: |
+          cd .github/workflows/qml_pipeline_utils
+          pip install .
+
+      - name: Install Python Dependencies
+        run: |
+          pip install -r requirements.txt
+          pip install --no-deps -r requirements_no_deps.txt
+          pip install --upgrade pip 'setuptools<65' cmake
```
smokingroosters, line 43: this setuptools restriction seems new, what's the reason behind it?

.github/workflows/build-dev.yml
```diff
@@ -1,50 +0,0 @@
...
-        run: |
-          python -m pip install --upgrade pip wheel
-          pip install -r requirements.txt
-          pip install --no-deps -r requirements_no_deps.txt
-
-      - name: Build tutorials
-        run: |
-          make download
-          make html
-          zip -r /tmp/qml_demos.zip demos
```
smokingroosters, line 43: zip not needed?

.github/workflows/build-dev.yml
rashidnhm: Nope, it was removed from the PR builds as well

.github/workflows/build-branches.yml
rashidnhm: We had an issue with numpy installation and this was the quickest workaround, see #646 for the details. Once the numpy version is upgraded and tested, then this workaround can be removed.

.github/workflows/build-dev.yml
rashidnhm: `actions/upload-artifact` already zips so nested zips are not needed, BUT ALSO, I am no longer uploading the entire demo being built as I did not see that being used anywhere and it is a very large file. Though if needed I can add it back

.github/workflows/build-branches.yml
antalszava, line 9: What exactly is concurrency used for here?

.github/workflows/build-branches.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Generate Execution Time Map
+        run: |
+          mkdir /tmp/execution_times
+
+          qml_pipeline_utils \
+          parse-execution-times \
+          --build-dir="${{ github.workspace }}/_build/html" > /tmp/execution_times/execution_times.json
+
+      - name: Upload Execution Times
+        uses: actions/upload-artifact@v3
```
antalszava, line 59: We're moving to v3 here, any significance to that? Is anything expected to work differently?

.github/workflows/qml_pipeline_utils/qml_pipeline_utils/cli.py
```diff
@@ -122,6 +123,13 @@ def cli_parser():
         subparsers_show_worker_files, "num-workers", "offset", "examples-dir", "glob-pattern"
     )
 
+    subparsers_parse_execution_times = subparsers.add_parser(
+        "parse-execution-times",
+        description="Parse the `sg_exection_times.html` file generated by sphinx "
+        "to get the time to execute each example",
```
antalszava, line 129: Why do we do this?

.github/workflows/qml_pipeline_utils/qml_pipeline_utils/services/parse_execution_times.py
```diff
@@ -0,0 +1,53 @@
...
+    execution_time_parts = re.split(r"[:.]", execution_time)
+
+    execution_time_min = int(execution_time_parts[0])
+    execution_time_sec = int(execution_time_parts[1])
+    execution_time_dsc = int(execution_time_parts[2])
+
+    return (execution_time_min * MS_IN_MIN) + (execution_time_sec * MS_IN_SEC) + (execution_time_dsc * MS_IN_DSC)
+
+
+def parse_execution_times(sphinx_build_directory: "Path", sphinx_gallery_dir_name: str) -> Dict[str, int]:
```
antalszava, line 31: Maybe good to add a docstring here too.

antalszava (commented):
Looking great overall, just had some questions. Great to have this be cleaner :heart_eyes:

.github/workflows/build-branches.yml
rashidnhm: The core functionality is the same, v3 is the current major release of this action using newer node version and actions/core as well. Brings along bug fixes etc.

But the way we use it (at least of basic usage) is the same.

.github/workflows/build-branches.yml
rashidnhm: The [concurrency group](https://docs.github.com/en/actions/using-jobs/using-concurrency) ensures multiple instances of the workflow are not run from the same ref at the same time. 

It is the equivalent of doing:
```yaml
- name: Cancel Previous Runs
  uses: styfle/cancel-workflow-action@0.4.1
  with:
    access_token: ${{ github.token }}
```
The main difference is that concurrency groups are a built-in functionality of GitHub actions and (in my opinion) superseed the need of using a third party action to achieve the same results.

.github/workflows/qml_pipeline_utils/qml_pipeline_utils/cli.py
rashidnhm: Good question.

The `sg_execution_times` a file that sphinx generates after it executes all the demos and 
 it contains how long it took sphinx to execute the code in each demo. This information is converted from an html page to an easily digestable JSON format by using the pipeline utils and uploaded as an artifact. In a subsequent PR, `build-pr.yml` will be updated to use this information to better distribute the demo load across workers. 

Currently the demos are split across workers by `build-pr.yml` by name, this is not ideal as some workers finish very fast while others take a while.

.github/workflows/qml_pipeline_utils/qml_pipeline_utils/services/parse_execution_times.py
rashidnhm: Good point, let me do that

rashidnhm: Done!

.github/workflows/qml_pipeline_utils/qml_pipeline_utils/cli.py
antalszava: Thank you! Apologies, I missed this in the PR description.

antalszava (approved):
Thanks for the answers! This looks good to me. :+1:

smokingroosters (approved):

PR #662, opened 2022-12-12
Add `print`

AlbertMitjans:


demonstrations/tutorial_adjoint_diff.py
```diff
@@ -395,7 +395,7 @@ def circuit_adjoint(a):
     qml.RZ(a[2], wires=1)
     return qml.expval(M)
 
-qml.grad(circuit_adjoint)(x)
+print(qml.grad(circuit_adjoint)(x))
```
Jaybsoni, line 398: This will probably change the expected output of the demo correct? Might need to update the output if its manually listed out later in the file?

Jaybsoni (approved):
Nice catch 👍🏼

PR #665, opened 2022-12-13
Update the quantum volume demo before `v0.28.0`

antalszava:
Move away from using `with qml.tape.QuantumTape() as tape` in the quantum volume demo.


trbromley:
Thanks @antalszava! Please could we hold off on this PR for now - we're still working out the details around changes to `QuantumTape`, including things like naming (e.g., `qml.queuing.AnnotatedQueue` doesn't feel like something we want to make particularly user facing).

antalszava:
Sure! I will actually be closing this PR.

The original motivation here was to silence the following warning:
```
/xanadu/pennylane/pennylane/_qubit_device.py:776: UserWarning: Using a list of observables in ``QubitDevice.statistics`` is deprecated. Please use a ``QuantumScript`` instead.
```
For some reason, however, this warning only arose once and cannot reproduce anymore. Hence, I'm closing this.

PR #678, opened 2023-01-05
external lib - thumbnail fix

KetpuntoG:
It was found that the meta image was not loading when posting the link. This is because the image added in meta was not being used anywhere (usually it is used in the thumbnail but this time it was not). The same thumbnail image has been added to fix this problem.


ikurecic (approved):
That's one way of doing that. 😂
Let's find a good process for the next one. :)

PR #679, opened 2023-01-12
QML tutorial for Quantum Variational Rewinding.

jackbaker1001:

**Title:** Quantum Variational Rewinding for Time Series Anomaly Detection

**Summary:** A simple and didactic/pedagogical look into the quantum variational rewinding (QVR) algorithm, also acting as a first-stepping stone on the way to using Covalent: the heterogeneous workflow management/scheduling tool. The reader is first introduced to the algorithm in the simplest way and then  uses QVR to learn the normal behaviour of a univariate (one dimensional) time series using a 1-qubit neural network. The threshold value is then tuned to maximize the accuracy score, then the model is compared against random models (which it convincingly beats) to demonstrate that the normal behaviour of the time series was indeed learnt in the training process.

**Relevant references:**  

(1) Baker, Jack S. et al. “Quantum Variational Rewinding for Time Series
    Anomaly Detection.” arXiv, <https://arxiv.org/abs/2210.16438>, 2022

 (2) Stone, Marshall H. “On one-parameter unitary groups in Hilbert
     space.” Annals of Mathematics, 643-648,
     <https://doi.org/10.2307/1968538>, (1932).

 (3) Welch, Jonathan et al. “Efficient quantum circuits for diagonal
     unitaries without ancillas”, New Journal of Physics, **16**, 033040
     <https://doi.org/10.1088/1367-2630/16/3/033040>, (2014).

 (4) Cîrstoiu, Cristina et al. “Variational fast forwarding for quantum
    simulation beyond the coherence time”, npj Quantum Information,
     **6**,  <https://doi.org/10.1038/s41534-020-00302-0>, (2020)

**Possible Drawbacks:**

1. To run the tutorial script, a Covalent server needs to be running. This has already been added to `requirements.txt` if it is to be run at build time, (during the call to `sphinx`), the below command need to be issued in a terminal:

```bash
covalent start
```

2. Presently, the first image is a placeholder for a nice cartoon in-keeping with the Pennylane style. While we could do this ourselves, it probably wouldn't be as nice as those made by @josh146 . If you lot would like to try drawing, please let us know.

**Related GitHub Issues:**
N/A


KetpuntoG:
Hey! Thanks for working on this demo 😄  Just to let you know that all our demos work with the same  [requirements.txt](https://github.com/PennyLaneAI/qml/blob/master/requirements.txt) file so make sure there are no incompatibilities with it

KetpuntoG:
On the other hand, the python version used to build the demos is `3.9`

jackbaker1001:
Hey @KetpuntoG. I'm working on having things working in Python 3.9. I am finding a few problems with installing the requirements here though.

For example, for PyTorch:

```
ERROR: Could not find a version that satisfies the requirement torch==1.10.0+cpu (from versions: 1.7.1, 1.8.0, 1.8.1, 1.9.0, 1.9.1, 1.10.0, 1.10.1, 1.10.2, 1.11.0, 1.12.0, 1.12.1, 1.13.0, 1.13.1)
ERROR: No matching distribution found for torch==1.10.0+cpu
```

Pip version is 22.3.1
Python version is 3.9.15

jackbaker1001:
@KetpuntoG after upgrading the Dask version in `requirements.txt` to `dask[delayed]==2022.9.0`, many more tests pass in the checks. Where it fails in build (78) is where I expect it to fail. 

As the PR states, Covalent needs to be running before this notebook is executed. Can you suggest to me where in the code I should be putting the `covalent start` command?

KetpuntoG:
I see you have managed to solve that problem! 💪 The new issue may come with the part of setting the connection with `http://localhost:48008/` . There is another way to write the demo in this situation. If when writing the name of the demo you avoid the `tutorial_` part it will not run, and you can manually put the outputs, as you can see in this [demo](https://pennylane.ai/qml/demos/vqe_parallel.html) (this is the [code](https://github.com/PennyLaneAI/qml/blob/master/demonstrations/vqe_parallel.py))
This is not ideal for the maintenance of the demo and may be changed in the future but I think it is a good solution for the moment 🙂

KetpuntoG:
> I see you have managed to solve that problem! 💪 The new issue may come with the part of setting the connection with `http://localhost:48008/` . There is another way to write the demo in this situation. If when writing the name of the demo you avoid the `tutorial_` part it will not run, and you can manually put the outputs, as you can see in this [demo](https://pennylane.ai/qml/demos/vqe_parallel.html) (this is the [code](https://github.com/PennyLaneAI/qml/blob/master/demonstrations/vqe_parallel.py))
> This is not ideal for the maintenance of the demo and may be changed in the future but I think it is a good solution for the moment 🙂

Although I'm not really sure if it would solve the problem 🤔 being able to put manual outputs could be useful for this demo? or precisely what you want to show is the access to covalent?

josh146:
Hey @jackbaker1001, looks like the build is still failing 🤔 

```
 /home/runner/work/qml/qml/demonstrations/tutorial_univariate_qvr.py failed leaving traceback:
  File "/home/runner/work/qml/qml/demonstrations/tutorial_univariate_qvr.py", line 781, in <module>
    plt.plot(results_dict["loss_history"], ".-")
TypeError: 'NoneType' object is not subscriptable
```

jackbaker1001:
Hey @josh146. Yeah we are having some problems getting this working on the the testing pipeline. If that takes too long to figure out we shall take @KetpuntoG 's suggestion to have it as a non-running demo. I will get back to you on which choice is made after consulting my colleagues :)

KetpuntoG:
> Hey @josh146. Yeah we are having some problems getting this working on the the testing pipeline. If that takes too long to figure out we shall take @KetpuntoG 's suggestion to have it as a non-running demo. I will get back to you on which choice is made after consulting my colleagues :)

Okey, I'll stay tuned! If we can, let's try to keep the non-running demo as the last option 💪

KetpuntoG:
> ok @josh146 @KetpuntoG Tests now pass for the tutorial running. I'm just gonna make some small edits to non-vital parts then there is the final matter of the placeholder figure and who is drawing it etc.

🤩 Awesome! Thanks @jackbaker1001

KetpuntoG:
Nice catch @josh146 ! Thanks for fixing that last problem 💪
@jackbaker1001, let me know if everything is done so I can start with the review process of the content (don't worry about the images).

jackbaker1001:
Hey @KetpuntoG and @josh146 . Firstly, thanks for the fix @josh146. I have made my changes now which, among other things, includes cartoons for the tutorial and the thumbnail. Provided the pipleline still passes (I think it will), we are good to move on to content review. Having looked at the previous version of the built wesbsite, it seemed there were a couple of rendering/formatting bugs, but i'm sure we can deal with these before release.

jackbaker1001:
@KetpuntoG I noticed a strange bug in the preview website deployment. On this tutorial, (not mine) https://qml-build-previews.pennylane.ai/pull_request_build_preview/679/demos/tutorial_vqt.html, the thumbnail becomes mine! It seems because both me and Jack Cernoni share a first name, something strange happens...

KetpuntoG:
Fixed, the problem was that the image name was already being used by that other demo. That's why we try to give unique names to the images 😄

jackbaker1001:
Good catch @KetpuntoG . Ok, so I think this is ready for content review now. Do you think you could provide an ETA for this please? Thanks again!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+plt.axhline(0.5, color="k", linestyle=":", label="Random accuracy")
+plt.xticks([1, 2, 3], ["Trained model", "Random model 1", "Random model 2"])
+plt.ylabel("Accuracy score")
+plt.title("Accuracy scores for trained and random models")
+plt.legend()
+
+# Shut down the covalent server
+stop = os.system("covalent stop")
+
+######################################################################
```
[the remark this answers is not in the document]
KetpuntoG: I think this may fix the rendering here

KetpuntoG (changes requested):
It looks great! Thanks for your work! Here are a few comments to work on 😄
The review process usually goes through two or three reviewers in which we will do some iterations. I attach a small document to keep in mind how to act in a review to facilitate the process

[review.pdf](https://github.com/PennyLaneAI/qml/files/10450080/review.pdf)

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
+"""
+Quantum Variational Rewinding for Time Series Anomaly Detection
+===============================================================
+
+The following demonstration is based upon the *Quantum Variational
+Rewinding* algorithm, or, QVR, proposed by `Baker, Horowitz, Radha et.
+al (2022) <https://arxiv.org/abs/2210.16438>`__ `[1] <#Baker2022>`__.
+
```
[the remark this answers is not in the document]
jackbaker1001: Got it. I will add sentence communicating this.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+"""
+######################################################################
+# .. figure:: ../demonstrations/univariate_qvr/cartoon_pennylane.png
+#    :width: 70%
+#    :align: center
+#    :alt: QVR cartoon
+#
+#
+#    A cartoon describing the quantum variational rewinding algorithm.
```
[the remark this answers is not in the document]
jackbaker1001: Yeah I agree. We can just have this as an informal explanation before the more technical background section. We have to be careful a little though, because the figure (and caption) has math in it which isn't formally described until the "background" section. In that sense, it is almost better to have the image after the "background" section, but then the image isn't immediately visible on clicking the tutorial, which it should be, as it adds a bit of a draw to read the rest of the tutorial. 

Do you have an opinion on this @KetpuntoG ?

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+#
+# To begin, let’s quickly recount the data that QVR handles: time series.
+# A general time series :math:`\boldsymbol{y}` can be described as an
+# integer :math:`p`-length sequence of observations of a process/system in
+# chronological order
+#
+# .. math::
+#
+#
+#    \boldsymbol{y} := (\boldsymbol{y}_t: t \in T), \quad T := (t_l: 1 \leq l \leq p).
```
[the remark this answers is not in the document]
jackbaker1001: Your second comment is correct. Normal parentheses are used here to denote a sequence (which is of course ordered). This part has to stay!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+# dropped. Also, we take :math:`y_t \in \mathbb{R}` and
+# :math:`t_l \in \mathbb{R}_{>0}`
+#
+# The goal of QVR and many other (classical) machine learning algorithms
+# for time series anomaly detection is to determine a suitable *anomaly
+# score* function :math:`a_{X}`. When passed a general time series
+# :math:`y`, This function produces a real number:
+# :math:`a_X(y) \in \mathbb{R}`. :math:`a_{X}` is learned from a training
+# set of time series instances known to exhibit normal behaviour
+# :math:`x \in X` (:math:`x` is defined analogously to :math:`y` in the
```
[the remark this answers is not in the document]
jackbaker1001: Good catch. I will add a definition.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+       W(callable): a parameterized quantum circuit representing the matrix
+       of eigenvectors in the eigendecomposition of e^{-iHt}
+       D(callable): a parameterized quantum circuit representing the diagonal
+       matrix in the eigendecomposition of e^{-iHt}
+       n_qubits(int): the number of qubits for the quantum circuit
+
+    Returns:
+        probabilities(torch.tensor): a 1D tensor of the probabilities of
+        measuring each bitstring. Probabilities are in given in
+        lexographical order or the bit strings.
```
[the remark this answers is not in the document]
jackbaker1001: Good catch!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+# :math:`\zeta` later in this tutorial.
+#
+# The first step for doing all of this *quantumly* is to generate a
+# sequence :math:`\mathcal{S} := (|x_{t} \rangle| t \in T)` of
+# :math:`n`-qubit quantum states corresponding to a classical time series
+# instance in the training set. Now, we suppose that each
+# :math:`|x_t \rangle` is a quantum state evolved to a time :math:`t` as
+# generated by an *unknown embedding Hamiltonian* :math:`H_E`. That is,
+# each element of :math:`\mathcal{S}` is defined by
+# :math:`|x_t \rangle = e^{-iH_E(x_t)}|0\rangle^{\otimes n} = U(x_t)|0\rangle^{\otimes n}`
```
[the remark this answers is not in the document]
jackbaker1001: Yeah, second comment is correct!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+# for embedding unitary operator :math:`U(x_t)` implementing a quantum
+# feature map (see the `Pennylane embedding
+# templates <https://docs.pennylane.ai/en/stable/introduction/templates.html#embedding-templates>`__
+# for efficient quantum circuits for doing so). Next, We operate on each
```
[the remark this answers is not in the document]
jackbaker1001: hmmm, i'm not sure what you mean by "after" here. This is all part of the first step. Any unitary operator $U$ can be written as $e^{-iM}$ where $M$ is hermitian. In this case, $U(x_t) = e^{-iH_E(x_t)}$. The operation that "actually happens" is the application of $U(x_t)$ to $|0\rangle^{\otimes n}$. It just so happens that we can write this unitary as the matrix exponential (although we don't actually calculate $H_E$).  This fact is useful for the later math.

demonstrations/tutorial_univariate_qvr.py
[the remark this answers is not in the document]
KetpuntoG: Good point, maybe the best thing here is to put the picture with a short description ("This picture describes the quantum variational rewinding algorithm, which will be developed in this demo").
and then at the end of "Background", say, "as an example of this, let's look at the start picture again (without repainting it), <and here you put the long description>.

demonstrations/tutorial_univariate_qvr.py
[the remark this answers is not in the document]
KetpuntoG: I got the feeling that I had to know why you were multiplying by that operator, but then you explain it. Maybe comment it out, "for the next step, we'll need to multiply by this operator and now we'll see why." or something like that. Feel free to ignore this comment if you think it is correct :)

KetpuntoG:
> Good catch @KetpuntoG . Ok, so I think this is ready for content review now. Do you think you could provide an ETA for this please? Thanks again!

If all goes well, everything should be ready in ~3 weeks

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+# unitary groups `[2] <#Stone1932>`__.
+#
+# We now ask the question: *What condition is required for*
+# :math:`|x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma} \rangle = |0 \rangle^{\otimes n}` *for all time?* The answer can be
+# formalized in math as
+#
+# .. math::
+#
+#
+#    \forall t, |0\rangle^{\otimes n} \langle 0|^{\otimes n}e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}e^{-iH_E(x_t)}|0\rangle^{\otimes n} = 1 \iff \forall t, H(\boldsymbol{\alpha}, \boldsymbol{\gamma})t = -H_E(x_t).
```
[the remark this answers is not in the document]
jackbaker1001: You're correct that this is wrong. To make sense with the previous paragraph we need either to (i) removed the first $|0 \range^{\otimes n}$ on the left or change the $1$ to $|0 \range^{\otimes n}$.  To make sense with the next time this is wrong (which you caught below), I will go with your suggestion.

Correct math:

$P(\)$

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+# Hamiltonian will not be able to successfully rewind every
+# :math:`x \in X` (in fact this is impossible to do so exactly if each
+# :math:`x` is unique, which is usually true), we consider the average effect of many
+# Hamiltonians generated by drawing :math:`\boldsymbol{\gamma}` from a
+# normal distribution :math:`\mathcal{N}(\mu, \sigma)`
+#
+# .. math::
+#
+#
+#    F(\boldsymbol{\phi}, x_t) = \mathop{\mathbb{E}_{\boldsymbol{\gamma} \sim \mathcal{N}(\mu, \sigma)}}\left[|0\rangle^{\otimes n} \langle 0|^{\otimes n} |x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma}\rangle  \right], \quad \boldsymbol{\phi} = [\boldsymbol{\alpha}, \mu, \sigma].
```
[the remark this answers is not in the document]
jackbaker1001: Correct. I  removed it

demonstrations/tutorial_univariate_qvr.py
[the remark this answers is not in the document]
jackbaker1001: So I think in this case what is written is correct. So if it's ok with you, I shall leave it put!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+# prepare the states
+#
+# .. math::
+#
+#
+#    |x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma}\rangle := e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}|x_t\rangle
+#
+# where we write
+# :math:`e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}` as an
+# eigendecomposition
```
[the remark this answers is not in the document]
jackbaker1001: I'm not sure what is meant here. What particular way are you saying H should be defined? In our view, the decomposition $e^{-iH(\alpha, \gamma)} = W^{\dagger}(\alpha)D(\gamma, r)W(\alpha)$ can generalize to any $n$-qubit time evolution unitary provided enough variational layers are used to implement $W$ and $D$ is fully connected ($k=n$).

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+#    catepillar, a chrysalis and a butterfly: the life cycle of a butterfly.
+#    This data is then encoded into quantum states and passed to a
+#    time machine which time devolves the states as generated by a learnt
+#    Hamiltonian operator (in practice, there are a distribution of such operators). After the devolved state is measured and the
+#    anomaly score :math:`a_X(y)` calculated, the time series is recognized as
+#    normal if :math:`a_X(y)` is smaller than the threshold and anomalous if the
+#    threshold is exceeded. In the first case, the time series is considered
+#    rewindable, correctly recovering the initial condition for the life cycle
+#    of a butterfly: eggs on a leaf.
+
```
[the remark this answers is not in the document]
jackbaker1001: I've gone through and put imports as they are used.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+                cnots = [comb[i : i + 2] for i in range(len(comb) - 1)]
+                for j in cnots:
+                    qml.CNOT(wires=j)
+                qml.RZ(gamma[cnt], wires=[comb[-1]])
+                cnt += 1
+                for j in cnots[::-1]:
+                    qml.CNOT(wires=j)
+
+
+######################################################################
```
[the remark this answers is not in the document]
jackbaker1001: I added this. Bear in mind since we are only using one qubit, the resulting circuit is mere a R_z gate. I think it's nice for the general function to stay though in case people want to experiement with/modify stuff. I added this in the text

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+        durations = torch.randint(low=1, high=max_duration, size=(spike_num,))
+        spike_start_idxs = torch.randperm(p - max_duration)[:spike_num]
+        for start_idx, duration in zip(spike_start_idxs, durations):
+            y[start_idx : start_idx + duration] += torch.normal(0.0, spike_amp, (duration,))
+    T = torch.linspace(t_init, t_end, p)
+    return Y, T
+
+
+######################################################################
+# Like many machine learning algorithms, training is done in mini-batches.
```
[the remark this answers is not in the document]
jackbaker1001: Added!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1416 @@
...
+    gammas = sigma.abs() * torch.randn((n_samples, gamma_length)) + mu
+    expectation = torch.empty(n_samples)
+    for i, gamma in enumerate(gammas):
+        expectation[i] = callable_proj(xt, t, alpha, gamma)
+    return expectation.mean()
+
+
+######################################################################
+# We now return to the matter of the penalty function :math:`P_{\tau}`.
+# While any sigmoidal-type function can be used, we choose
```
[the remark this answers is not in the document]
jackbaker1001: I'll add some clarification. But basically, it's because, for positive sigma (sigma is always positive) the arctan function we use is bounded in (0, 0.5) (we have a factor of 1/2 in there). This means that the loss function is defined in (0, 1).

jackbaker1001 (commented):
LGTM!

jackbaker1001:
I think I'm done here now @KetpuntoG . Thanks again for your detailed feedback. I think the whole thing reads better now.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+al (2022) <https://arxiv.org/abs/2210.16438>`__ `[1] <#Baker2022>`__ : a quantum machine learning algorithm for gate model quantum computers. QVR leverages the power of unitary time evolution/devolution operators to
+learn a model of *normal* behaviour for time series data. Given a new (i.e, unseen in training) time
+series, the normal model produces a value, which beyond a threshold,
+defines anomalous behaviour. In this tutorial, we’ll be showing you how
+all of this works, combining elements from
+`Covalent <https://www.covalent.xyz/>`__,
+`Pennylane <https://pennylane.ai/>`__ and
+`PyTorch <https://pytorch.org/>`__.
+
+Before getting into the technical details of the algorithm, let's get high level overview with the help of the below cartoon
```
KetpuntoG, line 26:
```suggestion
Before getting into the technical details of the algorithm, let's get high level overview with the help of the below cartoon:
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+# .. math::
+#
+#
+#    \boldsymbol{y} := (\boldsymbol{y}_t: t \in T), \quad T := (t_l: 1 \leq l \leq p).
+#
+# In the simple and didactic case treated in this tutorial, we deal with
+# the case where :math:`\boldsymbol{y}` is univariate (i.e, is a
+# one-dimensional time series) so bold-face for :math:`\boldsymbol{y}` is
+# dropped. Also, we take :math:`y_t \in \mathbb{R}` and
+# :math:`t_l \in \mathbb{R}_{>0}`
```
KetpuntoG, line 67:
```suggestion
# :math:`t_l \in \mathbb{R}_{>0}`.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+#
+# The goal of QVR and many other (classical) machine learning algorithms
+# for time series anomaly detection is to determine a suitable *anomaly
+# score* function :math:`a_{X}` where :math:`X` is a training dataset of 
+# *normal* time series instances :math:`x \in X` (:math:`x` is defined 
+# analogously to :math:`y` in the above) from which the anomaly score function 
+# was learnt. When passed a general time series
+# :math:`y`, This function produces a real number:
```
KetpuntoG, line 75:
```suggestion
# :math:`y`, this function produces a real number:
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+# :math:`x \in X` produce :math:`F(\boldsymbol{\phi}^{\star}, x_t)`
+# clustered about a center at :math:`1`. Given a new time series
+# :math:`y`, should :math:`F(\boldsymbol{\phi}^{\star}, y_t)` venture far
+# from the normal center at 1, we are observing anomalous behaviour!
+#
+# Take the time now to have another look at the cartoon at the start of this tutorial.
+# Hopefully things should start making sense now.
+#
+# Now with our algorithm defined, let’s stitch this all together: enter
+# `Covalent <https://www.covalent.xyz/>`__
```
KetpuntoG, line 187:
```suggestion
# `Covalent <https://www.covalent.xyz/>`__.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+# .. figure:: ../demonstrations/univariate_qvr/covalent_platform.png
+#    :align: center
+#    :width: 70%
+#    :alt: The Covalent platform
+#
+#
+#    A schematic demonstrating the different platforms Covalent can interact with.
+#
+# Now is a good time to import Covalent and launch the Covalent server!
+#
```
KetpuntoG, line 224: Very wise to put this code here separately👌

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+        for start_idx, duration in zip(spike_start_idxs, durations):
+            y[start_idx : start_idx + duration] += torch.normal(0.0, spike_amp, (duration,))
+    T = torch.linspace(t_init, t_end, p)
+    return Y, T
+
+######################################################################
+# Let's do a quick sanity check and plot a couple of these series. Despite the
+# above function's ``@ct.electron`` decorators, these can still be used as normal
+# Python functions without using the Covalent server. This is useful
+# for quick checks like this
```
KetpuntoG, line 330:
```suggestion
# for quick checks like this:
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1506 @@
...
+                for j in cnots[::-1]:
+                    qml.CNOT(wires=j)
+    if get_probs:
+        return qml.probs(wires=range(n))
+
+######################################################################
+# While the above may seem a little complicated, since we only use a single
+# qubit in this tutorial, the resulting circuit is merely a single :math:`R_z(\theta)` gate.
+
+# use Qiskit device here because the circuit rendering is nice with mpl
```
KetpuntoG, line 531: in pennylane we have two ways to paint circuits:

`qml.draw` and `qml.draw_mpl`. The first one paints it in text and the second one with matplotlib. In this case the code would look like this:

```python

dev = qml.device("default.qubit", wires = 1)

D_one_qubit = qml.qnode(dev)(D)
qml.draw_mpl(D_one_qubit, decimals = 2)(torch.tensor([1,0]), 1, 1, True)
```

Check to see if you like it

KetpuntoG (commented):
Good job 💪 with these last changes that I suggest, I will finish my review, I will contact the second reviewer to take a look at the demo :)

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Thanks. It looks better that way right?

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: That's certainly more elegant than the present code. First thing that came up in google wsa the solution I put from this blog: https://pennylane.ai/blog/2021/05/how-to-visualize-quantum-circuits-in-pennylane/ 

One thing I always found a little frustrating in Pennylane  was drawing sections of circuits. Is there a way of doing this with having any measurements?

jackbaker1001:
Hey @KetpuntoG I made your second set of changes. Just waiting on second reviewer now!

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+right, which you *promised* you would clean yesterday, is slowly growing
+(seriously, clean it). In many situations, it is important to know when these
+systems start behaving abnormally. For example, if the pressure 
+inside a nuclear fission reactor starts violently fluctuating, you may wish
+to be alerted of that. The task of identifying such temporally abnormal
+behaviour is known as time series anomaly detection and is well known in machine
+learning circles. 
+
+In this tutorial, we take a stab at time series anomaly detection using the *Quantum Variational
+Rewinding* algorithm, or, QVR, proposed by `Baker, Horowitz, Radha et.
```
alvaro-at-xanadu, line 16:
```suggestion
Rewinding* algorithm, or QVR, proposed by `Baker, Horowitz, Radha et.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+al (2022) <https://arxiv.org/abs/2210.16438>`__ [#Baker2022]_ : a quantum machine learning algorithm for gate model quantum computers. QVR leverages the power of unitary time evolution/devolution operators to
+learn a model of *normal* behaviour for time series data. Given a new (i.e, unseen in training) time
```
alvaro-at-xanadu, line 18:
```suggestion
learn a model of *normal* behaviour for time series data. Given a new (i.e., unseen in training) time
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+series, the normal model produces a value, which beyond a threshold,
```
alvaro-at-xanadu, line 19:
```suggestion
series, the normal model produces a value that, beyond a threshold,
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+defines anomalous behaviour. In this tutorial, we’ll be showing you how
+all of this works, combining elements from
+`Covalent <https://www.covalent.xyz/>`__,
+`Pennylane <https://pennylane.ai/>`__ and
+`PyTorch <https://pytorch.org/>`__.
+
+Before getting into the technical details of the algorithm, let's get a high level overview with the help of the below cartoon:
```
alvaro-at-xanadu, line 26:
```suggestion
Before getting into the technical details of the algorithm, let's get a high-level overview with the help of the cartoon below:
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+"""
+######################################################################
+# .. figure:: ../demonstrations/univariate_qvr/cartoon_pennylane.png
+#    :width: 70%
+#    :align: center
+#    :alt: QVR cartoon
+#
+#
+# 
+# Going left-to-right, a time series is sampled at three points in time, corresponding to a
```
alvaro-at-xanadu, line 36:
```suggestion
# Going left-to-right, a time series is sampled at three points in time, corresponding to
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# different stages in the life cycle of a butterfly: a catepillar, a chrysalis and a butterfly.
+# This information is then encoded into quantum states and passed to a
+# time machine which time devolves the states as generated by a learnt
+# Hamiltonian operator (in practice, there are a distribution of such operators).
```
alvaro-at-xanadu, line 40:
```suggestion
# Hamiltonian operator (in practice, there is a distribution of such operators).
```

demonstrations/tutorial_univariate_qvr.py
KetpuntoG: Currently we can't but I'm going to suggest them to create the functionality since as the code stands, it shouldn't be difficult to do 😊

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+#
+#    \boldsymbol{y} := (\boldsymbol{y}_t: t \in T), \quad T := (t_l: 1 \leq l \leq p).
+#
+# In the simple and didactic case treated in this tutorial, we deal with
+# the case where :math:`\boldsymbol{y}` is univariate (i.e, is a
+# one-dimensional time series) so bold-face for :math:`\boldsymbol{y}` is
+# dropped. Also, we take :math:`y_t \in \mathbb{R}` and
+# :math:`t_l \in \mathbb{R}_{>0}`.
+#
```
alvaro-at-xanadu, line 68: I'm a bit confused by this definition and I think it could be stated a bit more clearly. I don't know what is an _integer length-p sequence_. Is it a series that is indexed by integers (this is true for all sequences, by [definition](https://en.wikipedia.org/wiki/Sequence))? Or does the variable take integer values (does not make much sense if it's a vector, actually). The last sentence also seems to imply that $t_\l$ (the index?) is a real number, which doesn't make much sense to me.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+        p(int): length of each time series instance
+        num_series(int): the total number of time series in the set
+        noise_amp(float): standard deviation of normal distribution data is
+        drawn from
+        t_init(float): time time each time series starts from
+        t_end(float): the time each time series ends at
+        seed(int): random seed for torch for reproducibility
+    Returns:
+        tuple(torch.Tensor, torch.Tensor): the time series set X and the time
+        set T
```
josh146, line 276: @jackbaker1001 a minor suggestion here, but you'll notice that our other demos don't have docstrings (or when they do, the docstrings are very minimal).

The reason for this is that, in tutorial formats, we aim to ensure that the text adequately describes the function -- so that after reading the preceding paragraph, it is clear what the function is doing without needing a large docstring, which can disrupt the 'flow' of the story.

Having function arguments with descriptive names (e.g., `time_series_len` rather than `p`) can also help in reducing the need to have large docstrings in tutorials :)

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# the case where :math:`\boldsymbol{y}` is univariate (i.e, is a
+# one-dimensional time series) so bold-face for :math:`\boldsymbol{y}` is
+# dropped. Also, we take :math:`y_t \in \mathbb{R}` and
+# :math:`t_l \in \mathbb{R}_{>0}`.
+#
+# The goal of QVR and many other (classical) machine learning algorithms
+# for time series anomaly detection is to determine a suitable *anomaly
+# score* function :math:`a_{X}` where :math:`X` is a training dataset of 
+# *normal* time series instances :math:`x \in X` (:math:`x` is defined 
+# analogously to :math:`y` in the above) from which the anomaly score function
```
alvaro-at-xanadu, line 73:
```suggestion
# analogously to :math:`y` in the above), from which the anomaly score function 
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# The goal of QVR and many other (classical) machine learning algorithms
+# for time series anomaly detection is to determine a suitable *anomaly
+# score* function :math:`a_{X}` where :math:`X` is a training dataset of 
+# *normal* time series instances :math:`x \in X` (:math:`x` is defined 
+# analogously to :math:`y` in the above) from which the anomaly score function 
+# was learnt. When passed a general time series
+# :math:`y`, this function produces a real number:
+# :math:`a_X(y) \in \mathbb{R}`. The goal is to have
+# :math:`a_X(x) \rightarrow 0, \forall x \in X`. Then, for an unseen time
```
alvaro-at-xanadu, line 77:
```suggestion
# :math:`a_X(x) \approx 0, \forall x \in X`. Then, for an unseen time
```

alvaro-at-xanadu: $\rightarrow$ makes sense in the context of limits of sequences. This is not the case here.

alvaro-at-xanadu: Let's avoid the use of $\forall$, etc. It's good practice for math readability.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# series :math:`y`, the series is said to be anomalous should
```
alvaro-at-xanadu, line 78:
```suggestion
# series :math:`y` and a threshold :math:`\xi \in \mathbb{R}`, the series is said to be anomalous should
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# :math:`a_X(y) > \zeta` and normal otherwise for threshold
```
alvaro-at-xanadu, line 79:
```suggestion
# :math:`a_X(y) > \zeta` and normal otherwise.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# :math:`\zeta \in \mathbb{R}`. We show a strategy for setting
```
alvaro-at-xanadu, line 80:
```suggestion
# series :math:`y` and a threshold $\zeta \in \mathbb{R}$, the series is said to be anomalous should
# :math:`a_X(y) > \zeta,` and normal otherwise. We show a strategy for setting
```

alvaro-at-xanadu: Rewritten for clarity.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# The first step for doing all of this *quantumly* is to generate a
+# sequence :math:`\mathcal{S} := (|x_{t} \rangle: t \in T)` of
+# :math:`n`-qubit quantum states corresponding to a classical time series
+# instance in the training set. Now, we suppose that each
+# :math:`|x_t \rangle` is a quantum state evolved to a time :math:`t` as
+# generated by an *unknown embedding Hamiltonian* :math:`H_E`. That is,
+# each element of :math:`\mathcal{S}` is defined by
+# :math:`|x_t \rangle = e^{-iH_E(x_t)}|0\rangle^{\otimes n} = U(x_t)|0\rangle^{\otimes n}`
+# for embedding unitary operator :math:`U(x_t)` implementing a quantum
```
alvaro-at-xanadu, line 91:
```suggestion
# for an embedding unitary operator :math:`U(x_t)` implementing a quantum
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# templates <https://docs.pennylane.ai/en/stable/introduction/templates.html#embedding-templates>`__
+# for efficient quantum circuits for doing so). Next, We operate on each
+# :math:`|x_t\rangle` with a parameterized
+# :math:`e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}` operator to
+# prepare the states
+#
+# .. math::
+#
+#
+#    |x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma}\rangle := e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}|x_t\rangle
```
alvaro-at-xanadu, line 102:
```suggestion
#    |x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma}\rangle := e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}|x_t\rangle,
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#    V_t(\boldsymbol{\alpha}, \boldsymbol{\gamma}) := W^{\dagger}(\boldsymbol{\alpha})D(\boldsymbol{\gamma}, t)W(\boldsymbol{\alpha}) = e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}
+#
+# for :math:`\boldsymbol{\alpha}`-parametrized unitary matrix of
+# eigenvectors :math:`W(\boldsymbol{\alpha})` and
+# :math:`\boldsymbol{\gamma}`-parametrized diagonal unitary operator
+# :math:`D(\boldsymbol{\gamma}, t)` where both can be implemented
+# efficiently using parameterized quantum circuits. The above equality
+# with :math:`e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}` is a
+# consequence of Stone’s theorem for strongly continuous one-parameter
+# unitary groups [#Stone1932]_.
```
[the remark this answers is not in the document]
alvaro-at-xanadu: Rewritten for clarity, but please check that it renders properly since suggestions ruin my #

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# We now ask the question: *What condition is required for*
+# :math:`|x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma} \rangle = |0 \rangle^{\otimes n}` *for all time?*
+# To answer this, we must find
```
alvaro-at-xanadu, line 124:
```suggestion
# To answer this, we impose
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# :math:`D(\boldsymbol{\gamma}, t)` where both can be implemented
+# efficiently using parameterized quantum circuits. The above equality
+# with :math:`e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}` is a
+# consequence of Stone’s theorem for strongly continuous one-parameter
+# unitary groups [#Stone1932]_.
+#
+# We now ask the question: *What condition is required for*
+# :math:`|x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma} \rangle = |0 \rangle^{\otimes n}` *for all time?*
+# To answer this, we must find 
+# :math:`P(|0\rangle^{\otimes n}) = |\langle 0|^{\otimes n}|x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma} \rangle|^2 = 1`
```
alvaro-at-xanadu, line 125:
```suggestion
# :math:`P(|0\rangle^{\otimes n}) = |\langle 0|^{\otimes n}|x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma} \rangle|^2 = 1.`
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# Playing with the algebra a little we have the condition
```
alvaro-at-xanadu, line 126:
```suggestion
# Playing with the algebra a little, we have the condition
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# .. math::
+#
+#
+#    \forall t, \langle 0|^{\otimes n}e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}e^{-iH_E(x_t)}|0\rangle^{\otimes n} = 1 \iff \forall t, H(\boldsymbol{\alpha}, \boldsymbol{\gamma})t = -H_E(x_t).
```
alvaro-at-xanadu, line 131:
```suggestion
#   \langle 0|^{\otimes n}e^{-iH(\boldsymbol{\alpha}, \boldsymbol{\gamma})t}e^{-iH_E(x_t)}|0\rangle^{\otimes n} = 1 \iff  H(\boldsymbol{\alpha}, \boldsymbol{\gamma})t = -H_E(x_t).
```

alvaro-at-xanadu: for all symbols crowd notation.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# In other words, for the above to be true, the parameterized unitary
+# operator :math:`V_t(\boldsymbol{\alpha}, \boldsymbol{\gamma})` should be
+# able to reverse or *rewind* :math:`|x_t\rangle` to its initial
+# condition before the embedding unitary operator :math:`U(x_t)` was
+# applied: :math:`|0\rangle^{\otimes n}`.
```
alvaro-at-xanadu, line 137:
```suggestion
# state :math:`|0\rangle^{\otimes n}` before the embedding unitary operator :math:`U(x_t)` was
# applied.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# We are nearly there! Because it is reasonable to expect that a single
+# Hamiltonian will not be able to successfully rewind every
+# :math:`x \in X` (in fact this is impossible to do so exactly if each
```
alvaro-at-xanadu, line 141:
```suggestion
# :math:`x \in X` (in fact, this is impossible to do if each
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# Hamiltonians generated by drawing :math:`\boldsymbol{\gamma}` from a
+# normal distribution :math:`\mathcal{N}(\mu, \sigma)` with mean :math:`\mu` and
+# standard deviation :math:`\sigma`
+#
+# .. math::
+#
+#
+#    F(\boldsymbol{\phi}, x_t) = \mathop{\mathbb{E}_{\boldsymbol{\gamma} \sim \mathcal{N}(\mu, \sigma)}}\left[\langle 0|^{\otimes n} |x_t, \boldsymbol{\alpha}, \boldsymbol{\gamma}\rangle  \right], \quad \boldsymbol{\phi} = [\boldsymbol{\alpha}, \mu, \sigma].
+#
+# The goal is for the above to be :math:`\approx 1`
```
alvaro-at-xanadu, line 152:
```suggestion
# The goal is for the function :math:`F` defined above to be as close to :math:`1` as possible,
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# :math:`\forall x \in X` and :math:`\forall t`. With this in mind, we can
```
alvaro-at-xanadu, line 153:
```suggestion
#  for all :math:`x \in X.` With this in mind, we can
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# define the loss function to minimize as the mean square error (MSE)
+# regularized by a penalty function :math:`P_{\tau}(\sigma)` with a single
+# hyperparameter :math:`\tau`.
```
alvaro-at-xanadu, line 156:
```suggestion
# hyperparameter :math:`\tau:`
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# .. math::
+#
+#
+#    \mathcal{L(\boldsymbol{\phi})} = \frac{1}{2|X||T|}\sum_{x \in X} \sum_{t \in T}[1 - F(\boldsymbol{\phi}, x_t)]^2 + P_{\tau}(\sigma)
```
alvaro-at-xanadu, line 161:
```suggestion
#    \mathcal{L(\boldsymbol{\phi})} = \frac{1}{2|X||T|}\sum_{x \in X} \sum_{t \in T}[1 - F(\boldsymbol{\phi}, x_t)]^2 + P_{\tau}(\sigma).
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# Where we shall show the exact form of :math:`P_{\tau}(\sigma)` later.
```
alvaro-at-xanadu, line 163:
```suggestion
# We will show the exact form of :math:`P_{\tau}(\sigma)` later.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# The general purpose of the penalty function, however, is to penalize
```
alvaro-at-xanadu, line 164:
```suggestion
# The general purpose of the penalty function is to penalize
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# Now with our algorithm defined, let’s stitch this all together: enter
+# `Covalent <https://www.covalent.xyz/>`__.
+#
+
+
+######################################################################
+# Covalent: heterogeneous workflow orchestration
+# ----------------------------------------------
+#
+# Presently, many QML algorithms are *heterogeneous* in nature. That is,
```
alvaro-at-xanadu, line 195:
```suggestion
# Presently, many QML algorithms are *heterogeneous* in nature. This means that
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# they require compute resources from different computation paradigms:
```
alvaro-at-xanadu, line 196:
```suggestion
# they require computational resources from different computation paradigms:
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# classical and quantum computation. Covalent is a tool that can be used
```
alvaro-at-xanadu, line 197:
```suggestion
# Covalent is a tool that can be used
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# to manage their interaction by sending different tasks to different
+# compute resources and stiching them together as a workflow. While you
+# will be introduced to other concepts in Covalent throughout this
+# tutorial, to begin we define two key components.
```
alvaro-at-xanadu, line 201:
```suggestion
# tutorial, we define two key components to being with.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# 1. **Electrons**. Decorate regular Python functions with
+#    ``@ct.electron`` to desginate a *task*. These are the atoms of
+#    computation.
```
alvaro-at-xanadu, line 205:
```suggestion
#  a computation.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#
+# 2. **Lattices**. Decorate a regular Python function with ``@ct.lattice``
+#    to designate a *workflow*. These contain electrons stitched together to do something useful.
+#
+#    It should now be noted that while different electrons can be run remotely on
+#    different hardware and multiple compute paridigms (classical, quantum etc.: see the `Covalent executors <https://covalent.readthedocs.io/en/stable/plugins.html>`__),
```
alvaro-at-xanadu, line 211:
```suggestion
#    different hardware and multiple computational paradigms (classical, quantum etc.: see the `Covalent executors <https://covalent.readthedocs.io/en/stable/plugins.html>`__),
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+#    in this tutorial, to keep things simple, things are run on a local
```
alvaro-at-xanadu, line 212:
```suggestion
#    In this tutorial, to keep things simple, things are run on a local
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+time.sleep(2)  # give the Dask cluster some time to launch
+
+
+######################################################################
+# Generating univariate synthetic time series
+# -------------------------------------------
+#
+# In this tutorial, we shall deal with a simple and didactic example.
+# Normal time series instances are chosen to be noisy low-amplitude
+# signals normally distributed about the origin: :math:`x_t \sim \mathcal{N}(0, 0.1)` and the
```
alvaro-at-xanadu, line 242:
```suggestion
# signals normally distributed about the origin, for example, :math:`x_t \sim \mathcal{N}(0, 0.1).` The
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# series we deem to be anomalous are the same but with randomly inserted
+# spikes with random duration and amplitude.
```
alvaro-at-xanadu, line 244:
```suggestion
# spikes with random durations and amplitudes.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+plt.plot(T_anom, Y_anom[1], label="Anomalous")
+plt.ylabel("$y(t)$")
+plt.xlabel("t")
+plt.grid()
+leg = plt.legend()
+
+######################################################################
+# Taking a look at the above, the generated series are what we wanted. We have
+# a simple human-parsable notion of what it is for a time series to be anomalous
+# (big spikes). We of course don't need a complicated algorithm to be able to detect
```
alvaro-at-xanadu, line 349:
```suggestion
# (big spikes). Of course, we don't need a complicated algorithm to be able to detect
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# such anomalies but this is just a didactic example remember!
+#
+# Moving on, like many machine learning algorithms, training is done in mini-batches.
```
alvaro-at-xanadu, line 352:
```suggestion
# Like many machine learning algorithms, training is done in mini-batches.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+    X_flat = torch.flatten(X)
+    T_flat = T.repeat(X.size()[0])
+    atomized = [(xt, t) for xt, t in zip(X_flat, T_flat)]
+    return atomized
+
+
+######################################################################
+# We now wish to pass this to a cycled ``torch.utils.data.DataLoader``.
+# However, this object is not
+# `pickleable <https://docs.python.org/3/library/pickle.html#:~:text=%E2%80%9CPickling%E2%80%9D%20is%20the%20process%20whereby,back%20into%20an%20object%20hierarchy.>`__
```
alvaro-at-xanadu, line 382:
```suggestion
# `pickleable, <https://docs.python.org/3/library/pickle.html#:~:text=%E2%80%9CPickling%E2%80%9D%20is%20the%20process%20whereby,back%20into%20an%20object%20hierarchy.>`__
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+        seed(int): random seed for torch for reproducibility
+    Returns:
+        DataGetter: an instance of the DataGetter helper class
+    """
+    return DataGetter(Xtr, batch_size, seed)
+
+
+######################################################################
+# Now we have the means to create synthetic data and cycle through a
+# training set, we need to build our loss function
```
alvaro-at-xanadu, line 470:
```suggestion
# training set. We need to build our loss function
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+    qml.adjoint(W)(alpha, wires=range(n_qubits))
+    return qml.probs(range(n_qubits))
+
+
+######################################################################
+# To take the projector
+# :math:`|0\rangle^{\otimes n} \langle 0 |^{\otimes n}`, we consider only
+# the probability of measuring the bit string of all zeroes, which is the
+# :math:`0^{th}` element of the probabilities (bit strings are returned in
+# lexicographic order)
```
alvaro-at-xanadu, line 595:
```suggestion
# lexicographic order).
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+    Returns:
+        callable_proj(callable): a function which when suppied data, a time point
+        and the variational parameters returns the probability of measuring the
+        bit string of all zeroes.
+    """
+    callable_proj = lambda xt, t, alpha, gamma: probs_func(
+        xt, t, alpha, gamma, k, U, W, D, n_qubits
+    )[0]
+    return callable_proj
+
```
alvaro-at-xanadu, line 628: I wonder if you need to put these docstrings in every single function, since a lot of them have already been defined before.

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+
+######################################################################
+# Training the normal model
+# -------------------------
+#
+# Now equipped with a loss function, we need to minimize it with a
+# classical optimization routine. To start this optimization, however, we
+# need some initial parameters. We can generate them with the below
+# electron
```
alvaro-at-xanadu, line 764:
```suggestion
# electron.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+        batch_iterations,
+        callable_proj,
+        gamma_length,
+        print_intermediate=False,
+    )
+    return results_dict
+
+
+######################################################################
+# before running this workflow, we define all of the input parameters.
```
alvaro-at-xanadu, line 939:
```suggestion
# Before running this workflow, we define all of the input parameters.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+# We can now dispatch the lattice to the Covalent server.
+#
+
+tr_dispatch_id = ct.dispatch(training_workflow)(**training_options)
+
+
+######################################################################
+# If you are running the notebook version of this tutorial, if you
+# navigate to http://localhost:48008/ you can view the workflow on the
+# Covalent GUI. It will look like the below, showing nicely how all of the
```
alvaro-at-xanadu, line 980:
```suggestion
# Covalent GUI. It will look like the screenshot below, showing nicely how all of the
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+
+
+######################################################################
+# Tuning the threshold :math:`\zeta`
+# ----------------------------------
+#
+# When we have access to labelled anomalous series (as we do in our toy
+# problem here, often not the case in reality), we can tune the threshold
+# :math:`\zeta` to maximize some success metric. We choose to maximize the
+# accuracy score as defined using the below three electrons.
```
alvaro-at-xanadu, line 1021:
```suggestion
# accuracy score as defined using the three electrons below.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+        truth(torch.Tensor): the truth class values for each series
+    Returns:
+        accuracy_score(torch.Tensor): the accuracy score
+    """
+    return torch.sum(pred == truth) / truth.size()[0]
+
+
+######################################################################
+# Then, knowing the anomaly scores :math:`a_X(y)` for a validation data
+# set, we can scan through various zetas on a fine 1D grid and calcuate
```
alvaro-at-xanadu, line 1074:
```suggestion
# set, we can scan through various values of $:math:`zeta`$ on a fine 1D grid and calcuate
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+        scores = get_norm_and_anom_scores(
+            X_val_norm, X_val_anom, T, callable_proj, rand_params, gamma_length, n_samples
+        )
+        accs_list.append(threshold_scan_acc_score(scores, truth_labels, zeta_min, zeta_max, steps))
+        scores_list.append(scores)
+    return accs_list, scores_list
+
+
+######################################################################
+# and set the input parameters.
```
alvaro-at-xanadu, line 1277:
```suggestion
# We set the input parameters.
```

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1508 @@
...
+almost everywhere we look. The temperature changes as day turns to night, 
+stock markets fluctuate and the bacteria colony living in the coffee cup to your
+right, which you *promised* you would clean yesterday, is slowly growing
+(seriously, clean it). In many situations, it is important to know when these
+systems start behaving abnormally. For example, if the pressure 
+inside a nuclear fission reactor starts violently fluctuating, you may wish
+to be alerted of that. The task of identifying such temporally abnormal
+behaviour is known as time series anomaly detection and is well known in machine
+learning circles. 
+
```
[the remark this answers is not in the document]
jackbaker1001: Thanks!

demonstrations/tutorial_univariate_qvr.py
alvaro-at-xanadu: This happens in many places and needs fixing. "Cartoon below" stands for "the cartoon (that is shown) below", so _below_ does not modify cartoon and must not go before it. It is an adverb, and it must go after.

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: I think there is a typo above, I think you mean "begin" with.

alvaro-at-xanadu (changes requested):
Great tutorial over all, I learned a lot. There are quite a few grammar/style fixes that need to be addressed, and a few definitions that are unclear. Excellent work!

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Added manually with new commit

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: I don't think "for example" makes sense here. There are not other examples, the normal signals are always drawn from N(0, 0.1).

I will change to:

"signals normally distributed about the origin. In our case, :math:`x_t \sim \mathcal{N}(0, 0.1)."

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Made this change in another commit

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: We don't want to have the comma hyperlinked no? Added it after

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Having taken a look through a fair few, it seems some tutorials include docstrings and others don't. I can certainly reduce the size of these docstrings by making sure there are no repeat definitions, but, I do think docstrings in general need to stay. Although variables are already named to match the math in the text, I think it just easier to see them defined in the function rather than back-track through the tutorial to find definitions.

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Ok since you and @alvaro-at-xanadu made similar comments, I will slash the variable definitions in docstrings entirely. I will try and use informative variable names where possible, but please note, things like `p` were used to match the math notation in the text cells. Does your opinion change with this or do you still want these change to `time_series_len` etc? @josh146

josh146: I'm happy with `p` as long as it is explained in the text/maths!

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Added in another commit

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Thanks for pointing this out. I will clarify things here.

It indeed is indexed by an integer. What I mean is the sequence has p-many elements where p is an integer. I don't understand your last comment though, it is stated that $t_l \in \mathbb{R}_{>0}$ not that $l$ is a real number. $l$ is an integer in $1 \leq l \leq p$.

I think I will change this last inequality to $l \in \mathbb{Z}^{+}_{\leq p}$ to avoid ambiguity

jackbaker1001:
Ok @alvaro-at-xanadu , thank you for your detailed review, it was very helpful. I have made your changes and the PR should now be up to date. 

Where do we go from here @KetpuntoG @josh146 ?

KetpuntoG:
> Ok @alvaro-at-xanadu , thank you for your detailed review, it was very helpful. I have made your changes and the PR should now be up to date.
> 
> Where do we go from here @KetpuntoG @josh146 ?

Hey @jackbaker1001 , I would like to take a last review to check that everything is ok but the work is going fast so I think it can be published before QHack 😊

jackbaker1001:
Thanks @KetpuntoG . I'll standby for your last review then!

jackbaker1001:
@KetpuntoG are there any action items for me on here?

KetpuntoG:
> @KetpuntoG are there any action items for me on here?

Hi @jackbaker1001 ! @ikurecic will be the person who will do the final review, soon you will have her suggestions :)
On the other hand, the estimated release date of the demo is Tuesday February 7th

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1252 @@
...
+threshold_tuning_options.update(general_options)
+
+
+######################################################################
+# As before, we dispatch the lattice to the ``Covalent`` server.
+#
+
+val_dispatch_id = ct.dispatch(threshold_tuning_workflow)(**threshold_tuning_options)
+ct_val_results = ct.get_result(dispatch_id=val_dispatch_id, wait=True)
+accs_list, scores_list = ct_val_results.result
```
ikurecic, line 1051: Unfortunately, I got stuck here. Is it something on my end?

> Node get_norm_and_anom_scores failed:
File "/home/ikurecic/miniconda3_9/lib/python3.9/site-packages/covalent_dispatcher/_core/execution.py", line 353, in _run_task
output, stdout, stderr = await execute_callable()
File "/home/ikurecic/miniconda3_9/lib/python3.9/site-packages/covalent/executor/base.py", line 387, in execute
result = await self.run(function, args, kwargs, task_metadata)
File "/home/ikurecic/miniconda3_9/lib/python3.9/site-packages/covalent/executor/executor_plugins/dask.py", line 112, in run
result, worker_stdout, worker_stderr = await dask_client.gather(future)
File "/home/ikurecic/miniconda3_9/lib/python3.9/site-packages/distributed/client.py", line 2089, in _gather
raise exception.with_traceback(traceback)
distributed.scheduler.KilledWorker: ('dask_wrapper-af3cea404e772fc9c72d220417abe2b7', <WorkerState 'tcp://127.0.0.1:38265', name: 13, status: closed, memory: 0, processing: 1>)

ikurecic (commented):
Very cool demo!
I made just very simple language correction suggestions.
However, I got stuck around the end ([line 1050](https://github.com/PennyLaneAI/qml/blob/bfa280726304bcc91adf91782d1454d39c78229f/demonstrations/tutorial_univariate_qvr.py#L1049-L1051)). Is it on my end?

ikurecic (commented):
Very cool demo!
I made just very simple language correction suggestions.
However, I got stuck around the end ([line 1050](https://github.com/PennyLaneAI/qml/pull/679/files#diff-783d37dde04a8455b37a02139dd341b4171f1783f51cb4745764ab77df3ff8ecR1049-R1051)). Is it on my end?

demonstrations/tutorial_univariate_qvr.py
```diff
@@ -0,0 +1,1252 @@
...
+    k: int,
+    U: callable,
+    W: callable,
+    D: callable,
+    n_qubits: int,
+    random_model_seeds: torch.Tensor,
+    W_layers: int,
+) -> list:
+    """A workflow for calculating anomaly scores for a set of testing time series
+     given an optimal model and set of random models. We use the optimal zetas found in threshold tuning.
```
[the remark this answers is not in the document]
jackbaker1001: I think you may have accidentally introduced a typo here. I will keep the old version!

demonstrations/tutorial_univariate_qvr.py
jackbaker1001: Hey @ikurecic. So, this happened because one of Dask workers ran out of memory during this part of the tutorial. Under the hood, when you run Covalent locally, it attempts to parallelize tasks using workers/threads in Dask. It just so happens that on your local machine, the default Dask cluster had workers (which have their memory managed by Dask) which didn't have enough memory.

You can try running again by replacing the `covalent start` at the beginning of the script specifically requesting memory:

`covalent start -m "2GiB" -n 2`

This starts a a covalent server with two processes and 2 Gigs of memory each.

Long story short, this only happens when RAM on somebody's laptop becomes a factor. Does this make sense @ikurecic ?

jackbaker1001:
Thank you so much @ikurecic! You certainly have a keen eye for typos and language. Things look much better now!

@KetpuntoG Does this conclude review?

demonstrations/tutorial_univariate_qvr.py
ikurecic: That makes sense, thanks @jackbaker1001 — do you mind adding that note to the demo? It's a simple enough fix.

KetpuntoG:
> Thank you so much @ikurecic! You certainly have a keen eye for typos and language. Things look much better now!
> 
> @KetpuntoG Does this conclude review?

Yes, the only thing left to do is to write the little note that has been suggested about the dask problem, and then the review would be finished 💯

jackbaker1001:
@KetpuntoG @ikurecic  I put a comment near where the Covalent server is started on how to troubleshoot memory issues.

jackbaker1001:
@KetpuntoG Could you please confirm the URL for this demo once it goes live? (are we still planning on Feb 7th?)

KetpuntoG:
> @KetpuntoG Could you please confirm the URL for this demo once it goes live? (are we still planning on Feb 7th?)

No problem, I'll share the link right here when it's published 😄
Yes, the date is still the 7th. We will do the merge the day before for security but we will not announce it until the 7th

josh146:
@jackbaker1001 the URL is based on the filename, so assuming no changes, it _should_ be https://pennylane.ai/qml/demos/tutorial_univariate_qvr.html. But I would still double check before release :)

KetpuntoG:
Hey @jackbaker1001 , the title is too long and fits too tightly inside the card. Do you have any suggestion how to shorten it?

jackbaker1001:
@KetpuntoG I also noticed this some time back but didn't know if it was an issue or not (it seemed to be that it did just about fit in the box).

I struggled to think of a shorter title. But perhaps: "Time series anomaly detection, Quantumly". Is good?

jackbaker1001:
@KetpuntoG what's the character limit?

KetpuntoG:
> @KetpuntoG what's the character limit?

not sure, what about "Quantum detection of time series anomalies"?

KetpuntoG (approved):
Amazing job in this demo 🚀💯

KetpuntoG:
There you go @jackbaker1001 ! https://pennylane.ai/qml/demos/tutorial_univariate_qvr.html
Tomorrow we will announce it on our social networks :)

jackbaker1001:
Thanks again to everybody who reviewed this demo! @KetpuntoG @ikurecic @alvaro-at-xanadu @josh146

jackbaker1001:
FYI @KetpuntoG it appears that the link is broken!

KetpuntoG:
Oh that's weird, It is working for me and I have tried it in two different browsers @jackbaker1001 . Could you double check? Try clearing the cache in case you are getting an old version 🤔

jackbaker1001:
@KetpuntoG Clearing the cache fixed it. A few others were having this issue on my side too haha. I will tell them to clear the cache also

PR #682, opened 2023-01-18
Sc 30753 Add Quantum Machine Learning Subhub

Anakhom:
[sc-30753-add-quantum-machine-learning-hub-page-in-qml-projec](https://app.shortcut.com/xanaduai/story/30753/add-quantum-machine-learning-hub-page-in-qml-project)
Added QML Subhub

A few important things to note:
- This PR partially includes copy from the QML page we already have in pennylane.ai 
- The videos for QML and QC hubs are not yet filmed. I followed up with Ivana, she let me know that Lara would need at least two weeks to film the videos.
- This PR has a placeholder image in place of where the video thumbnail image should be. The reason is that we don't currently have any high resolution images of any video thumbnails. That also would require at least to weeks to be delivered by the community team. 
- As far as I know, Tarik is making some new icons for the hubs. The ETA for that is unknown and currently all hubs have the same icons in the cards. 

Demo with new graphics and commented out video section:

https://user-images.githubusercontent.com/58717051/215795282-c1603b73-d7a5-435e-b1db-075d7ae15cb2.mov

PR #688, opened 2023-01-26
Updated 'What is QML' file name, switched images

ikurecic:
In preparation for the PennyLane website update. Second image will be completed and switched out later.


BM7878 (dismissed):

josh146 (changes requested):
Thanks @ikurecic! Only one request here, could you rename `the page back to `whatisqml.rst`?

This is because the name of the rst file becomes the URL, and we want to preserve the same URL for search purposes

ikurecic:
@josh146 I thought the solution was that we'd double up the pages, run both whatisqml.html and what-is-quantum-machine-learning.html for a few months. I can return it, or make duplicates from here, or leave that to the dev end of things?

BM7878:
Search engines might not look favourably on us for doubling-up the pages. Let's keep the current file name & URL for now - we'll do the change later (along with a few other pages) and set up a redirect.

josh146 (dismissed):
💯

Anakhom:
Closing as changes are implemented in https://github.com/PennyLaneAI/qml/pull/682

PR #690, opened 2023-01-30
Remove `collections` usage from demos

AlbertMitjans:


demonstrations/tutorial_rosalin.py
```diff
@@ -199,7 +191,12 @@ def cost(params):
 
     result = 0
 
-    for h, c, p, s in zip(qnodes, coeffs, prob_shots, shots_per_term):
+    for o, c, p, s in zip(obs, coeffs, prob_shots, shots_per_term):
+
+        @qml.qnode(non_analytic_dev, diff_method="parameter-shift")
+        def h(weights):
+            StronglyEntanglingLayers(weights, wires=non_analytic_dev.wires)
+            return qml.expval(o)
```
trbromley, line 199: Minor point, but couldn't we also evaluate `h` outside of the `for` loop but with `return [qml.expval(o) for o in obs]`, and we then zip over the outputs of calling this function in the `for` loop.

trbromley (commented):
Thanks @AlbertMitjans! Should we also update the "Last updated" field for both demos?

demonstrations/tutorial_rosalin.py
AlbertMitjans: I don't think so, because the number of shots for each observable is different.

AlbertMitjans:
> Should we also update the "Last updated" field for both demos?

Yes, sorry. Updated! 👍

demonstrations/tutorial_QGAN.py
```diff
@@ -275,11 +274,18 @@ def gen_cost(gen_weights):
 
 obs = [qml.PauliX(0), qml.PauliY(0), qml.PauliZ(0)]
```
lillian542, line 275: This doesn't seem to be used now, they're listed out explicitly in both functions

demonstrations/tutorial_rosalin.py
```diff
@@ -130,9 +130,7 @@
 
 
 ##############################################################################
-# We can now create our quantum device (let's use the ``default.qubit`` simulator),
-# and begin constructing some QNodes to evaluate each observable. For our ansatz, we'll use the
-# :class:`~.pennylane.templates.layers.StronglyEntanglingLayers`.
+# We can now create our quantum device (let's use the ``default.qubit`` simulator).
 
 from pennylane.templates.layers import StronglyEntanglingLayers
```
lillian542, line 135: I guess not related to the `collections` module, but I think it would be better if this import were moved to the code block where it is used

demonstrations/tutorial_rosalin.py
```diff
@@ -235,13 +232,13 @@ def cost(params):
 
 ##############################################################################
 # Let's compare this against an optimization not using weighted random sampling.
-# Here, we will split the 8000 total shots evenly across all Hamiltonian terms,
-# also known as *uniform deterministic sampling*.
 
-non_analytic_dev.shots = int(total_shots / len(coeffs))
```
lillian542, line 241: Why don't we want to distribute the shots anymore?

demonstrations/tutorial_QGAN.py
AlbertMitjans: Now the `qnodes` use `obs` to remove copied code 👍

demonstrations/tutorial_rosalin.py
AlbertMitjans: Moved!

demonstrations/tutorial_rosalin.py
AlbertMitjans: I did it to avoid using multiple `qnodes`. But maybe it's better to mantain the demo functionality.

I added multiple qnodes using `total_shots / len(coeffs)`. Let me know what you think!

demonstrations/tutorial_rosalin.py
mudit2812: If `shots` are overridden inside the loop anyway, `h` can be defined before the loop and the observable can be made a parameter that is given to `h`

demonstrations/tutorial_rosalin.py
```diff
@@ -187,19 +176,28 @@
...
+
+
 def cost(params):
     # sample from the multinomial distribution
     shots_per_term = si.rvs()[0]
 
     result = 0
 
-    for h, c, p, s in zip(qnodes, coeffs, prob_shots, shots_per_term):
+    for o, c, p, s in zip(obs, coeffs, prob_shots, shots_per_term):
```
mudit2812: As far as I can tell, `prob_shots` aren't being used anywhere inside this loop.

mudit2812 (commented):
The demos mostly look good to me. I've added some minor suggestions, but the demos seem correct :)

demonstrations/tutorial_rosalin.py
AlbertMitjans: Good call! Added your suggestion

demonstrations/tutorial_rosalin.py
```diff
@@ -436,7 +435,12 @@ def estimate_hamiltonian(self, params, shots):
         shots_per_term = si.rvs()[0]
 
         results = []
-        for h, c, p, s in zip(self.qnodes, self.coeffs, prob_shots, shots_per_term):
+        for o, c, p, s in zip(self.obs, self.coeffs, prob_shots, shots_per_term):
+
+            @qml.qnode(rosalin_device, diff_method="parameter-shift")
+            def h(weights):
```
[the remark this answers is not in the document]
AlbertMitjans: Moved `h` outside of the `cost` function!

demonstrations/tutorial_rosalin.py
AlbertMitjans: Removed!

lillian542 (approved):
Looks good to me! 🚀

demonstrations/tutorial_rosalin.py
```diff
@@ -436,7 +447,12 @@ def estimate_hamiltonian(self, params, shots):
         shots_per_term = si.rvs()[0]
 
         results = []
-        for h, c, p, s in zip(self.qnodes, self.coeffs, prob_shots, shots_per_term):
+        for o, c, p, s in zip(self.obs, self.coeffs, prob_shots, shots_per_term):
+
+            @qml.qnode(rosalin_device, diff_method="parameter-shift")
+            def h(weights):
+                StronglyEntanglingLayers(weights, wires=rosalin_device.wires)
+                return qml.sample(o)
```
mudit2812, line 455:
```suggestion
        @qml.qnode(rosalin_device, diff_method="parameter-shift")
        def h(weights, obs):
            StronglyEntanglingLayers(weights, wires=rosalin_device.wires)
            return qml.sample(obs)

        for o, c, p, s in zip(self.obs, self.coeffs, prob_shots, shots_per_term):

```

And update the code inside the loop to pass in the observable to `h` accordingly

demonstrations/tutorial_rosalin.py
```diff
@@ -238,10 +235,23 @@ def cost(params):
...
+    for o, c in zip(obs, coeffs):
+
+        @qml.qnode(non_analytic_dev, diff_method="parameter-shift")
+        def h(weights):
+            StronglyEntanglingLayers(weights, wires=non_analytic_dev.wires)
+            return qml.expval(o)
+
+        # evaluate the QNode corresponding to
+        # the Hamiltonian term, and add it on to our running sum
+        result += c * h(params, shots=shots_per_term)
```
mudit2812, line 252:
```suggestion
    @qml.qnode(non_analytic_dev, diff_method="parameter-shift")
    def h(weights, obs):
        StronglyEntanglingLayers(weights, wires=non_analytic_dev.wires)
        return qml.expval(obs)

    for o, c in zip(obs, coeffs):
        # evaluate the QNode corresponding to
        # the Hamiltonian term, and add it on to our running sum
        result += c * h(params, o, shots=shots_per_term)

```

mudit2812 (changes requested):
A couple of more suggestions, but overall, ready to go!

demonstrations/tutorial_rosalin.py
AlbertMitjans: Thanks for noticing this!

demonstrations/tutorial_rosalin.py
AlbertMitjans: Added 👍

PR #691, opened 2023-02-01
Adding documentation for the demonstrations metadata

BM7878:
This pull request contains the more formal technical documentation for the demos metadata JSON.


demonstrations_metadata.md
```diff
@@ -0,0 +1,151 @@
+# Demonstrations Metadata
+
+This document describes the structure of the JSON files that hold the metadata for the demonstrations.
+
+The metadata JSON file for a given demo should be stored in the same folder as the Python demo file. It should have exactly the same name, except ending with `.metadata.json` instead of `.py`.
```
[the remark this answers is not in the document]
BM7878: This is an excellent point. From talking to @josh146, it sounds like it would be quite hard to tidy up the demos folder in a nice way, because the limitations of Sphinx. In the long-run, my preference would be that each demo has a single folder, named well, in which the Python file, JSON metadata file, and all images and other resources reside. But it sounds like this would take some Sphinx-wrangling, so we thought we'd leave this as a separate project.

demonstrations_metadata.md
```diff
@@ -0,0 +1,151 @@
...
+            "title": "Addition on a Quantum Computer",
+            "authors": "Thomas G. Draper",
+            "year": "2000",
+            "journal": "",
+            "doi": "arXiv:quant-ph/0008033"
+        }
+    ],
+    "basedOnPapers": [],
+    "referencedByPapers": []
+}
```
KetpuntoG, line 45: Oh I don't know why I had in mind that the related demos would also be included here 🤔

josh146: oh! @BM7878 including the related demos would be a good idea, how did we miss this

demonstrations_metadata.md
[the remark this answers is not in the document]
ikurecic: I'd love to know exactly what specific idiotic thing Sphinx is being stubborn about this time around... 😅 But yeah, then I guess the meta files will just get dumped somewhere in the 'main' demonstrations folder.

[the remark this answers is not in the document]
josh146: It's because sphinx has a 1-1 mapping between source directory <---> website URLs :(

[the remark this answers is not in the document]
ikurecic: Yuck... Naturally. That's the best we can do, then. :)

demonstrations_metadata.md
BM7878: It's something that came up. I wasn't sure if this was the best place to add it, but now I think it is. Have added it.

demonstrations_metadata.md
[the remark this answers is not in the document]
rashidnhm: In theory, we can put the demos in their own folder for a cleaner directory and have the CI pipeline flatten it during runtime. 

But this comes with some issues of it's own:
- Heavier reliance on the CI pipeline, harder to test demos locally
- References to other material can potentially break as the demos are moved during the sphinx build process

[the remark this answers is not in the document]
josh146: I think in a sense this will be 'resolved' once the react port is done. If sphinx is no longer generating the whole website - just the content - we can choose any source directory organization we want!

demonstrations_metadata.md
```diff
@@ -0,0 +1,153 @@
...
+The metadata JSON file for a given demo should be stored in the same folder as the Python demo file. It should have exactly the same name, except ending with `.metadata.json` instead of `.py`.
+
+
+
+## Example
+
+Below is given an example of a complete metadata file for a demonstration. 
+
+```json
+{
```
ashishks0522, line 14: Viewing it from the perspective of using in react project, would we want to include data related to the downloads section?


Screenshot 2023-02-02 at 9 28 28 AM

demonstrations_metadata.md
```diff
@@ -0,0 +1,153 @@
...
+    "title": "Basic arithmetic with the quantum Fourier transform (QFT)",
+    "authors": [
```
ashishks0522, line 16: Again checking form the perspective of using in react project. Since we can have multiple authors, how do we decide which one to show at the end in `About the author` section? 

Screenshot 2023-02-02 at 9 32 16 AM

Also I saw that right now we have a folder which contains author bio's (name, bio, image) should we also consolidate that into this json file or keep that seperate?

ashishks0522: Thanks @ikurecic, I agree keeping single source for authors data would be easier to manage.

BM7878: Should we have an `id` field in each author object then, rather than a `name` field, so that it refers directly to the author bio files?

demonstrations_metadata.md
BM7878: Would we include a list of URLs to these resources? Or were you thinking something else?

ashishks0522: Yes URLs to populate them with each demo

demonstrations_metadata.md
ashishks0522: I think `id` might be better to keep data consistent. Not entirely sure how we will populate ID's for authors, but we can figure that out later.

demonstrations_metadata.md
josh146: So basically, no need to include this in the custom written metadata, since it is not hardcoded URLs. We will find a different approach to providing this info to react.

ashishks0522 (approved):
Thanks @BM7878!

josh146 (approved):
Looks great from my end! Excited to have this added 🤓

demonstrations_metadata.md
```diff
@@ -0,0 +1,153 @@
+# Demonstrations Metadata
+
+This document describes the structure of the JSON files that hold the metadata for the demonstrations.
+
+The metadata JSON file for a given demo should be stored in the same folder as the Python demo file. It should have exactly the same name, except ending with `.metadata.json` instead of `.py`.
```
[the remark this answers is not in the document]
BM7878: I find adding this sort of thing in tends to future-proof things. It also means that anyone glancing at the directory for the first time immediately knows how these files relate to others in the directory.

demonstrations_metadata.md
```diff
@@ -0,0 +1,153 @@
...
+    "dateOfPublication": "2022-11-07T00:00:00",
+    "dateOfLastModification": "2023-01-20T00:00:00",
+    "categories": ["Getting Started"],
+    "tags": ["quantum Fourier transforms", "qft"],
+    "previewImages": [
+        {
+            "type": "thumbnail",
+            "uri": "/qml/_images/qft_arithmetics_thumbnail.png"
+        }
+    ],
```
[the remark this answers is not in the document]
BM7878: These are defined below. So far I've only defined 'thumbnail', because it's one that we definitely know we need, but I imagine very quickly we'll start adding more.

PR #692, opened 2023-02-02
Update the "Optimization with SPSA" demo

dwierichs:
**Title:**
Optimization using SPSA

**Summary:**
This updates the existing demo with the same title.
- PennyLane's own `qml.SPSAOptimizer` is used instead of `noisyopt` optimization. This also reduces required reshaping hustle.
- Some other minor simplifications have been done. 
- A bug reporting smaller execution numbers for SPSA was fixed
- The text was updated corresponding to the newly used PL feature instead of `noisyopt`.

**Relevant references:**

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**


demonstrations/spsa.py
```diff
@@ -171,86 +174,66 @@
...
 # Once we have defined each piece of the optimization, there's only one
 # remaining component required: the *SPSA optimizer*.
-# We'll use the SPSA optimizer provided by the ``noisyopt`` package. Once
-# imported, we can initialize parts of the optimization such as the number of
-# iterations, a collection to store the cost values, and a callback function.
-# Once the optimization has concluded, we save the number of device executions
-# required for completion using the callback function. This will be an
-# interesting quantity!
-from noisyopt import minimizeSPSA
+# We'll use the ``qml.SPSAOptimizer`` built into PennyLane, 200 iterations in total,
```
josh146, line 198:
```suggestion
# We'll use the :class:`~pennylane.SPSAOptimizer` built into PennyLane, 200 iterations in total,
```

Hopefully this works 🤞

demonstrations/spsa.py
```diff
@@ -13,14 +13,15 @@
    tutorial_vqe A brief overview of VQE
    tutorial_vqe_qng Accelerating VQEs with quantum natural gradient
 
-*Author: Antal Szava — Posted: 19 March 2021. Last updated: 13 September 2022.*
+*Author: Antal Szava — Posted: 19 March 2021. Last updated: xx February 2023.*
```
dwierichs, line 16: Should I modify the author info and add myself? Or is this still a minor edit?

demonstrations/spsa.py
dwierichs: Yes, this works @josh146  Thanks :)

demonstrations/spsa.py
```diff
@@ -171,86 +169,80 @@
...
+    for step in range(num_steps):
+        # Monitor the device executions, deducting the cost evaluations for monitoring
+        exec_history.append(dev.num_executions - step)
+        # Monitor the cost value
+        cost_history.append(fun(param))
+        # Perform an update step
+        param = opt.step(fun, param)
+
+        # Print out the status of the optimization
+        if step % int(np.ceil(num_steps/20)) == 0:
```
timmysilv, line 217: the `int(np.ceil(num_steps/20))` loop invariant can be evaluated before the for loop

demonstrations/spsa.py
```diff
@@ -284,63 +268,42 @@ def callback_fn(xk):
 #
 #  .. code-block:: none
 #
-#     Iteration = 10, Number of device executions = 14, Cost = 0.09
```
timmysilv, line 287: curious what changed here. used to be 1.5 device executions per iteration (with some minus-one shift), now it's 2. i'm mostly wondering about this because we are taking special care to point out the total device executions but it went up using the SPSAOptimizer!

demonstrations/spsa.py
```diff
@@ -414,109 +374,88 @@ def cost_gd(params):
 #
 #  .. code-block:: none
 #
-#     Device execution ratio: 4.161375661375661
+#     Device execution ratio: 10.083
```
timmysilv, line 377: on the other hand... nice!

demonstrations/spsa.py
```diff
@@ -414,109 +374,88 @@ def cost_gd(params):
...
-
-##############################################################################
 # SPSA and the variational quantum eigensolver
 # --------------------------------------------
 #
-# Now that we've explored the theoretical underpinnings of SPSA, let's use it
-# to optimize a real chemical system, that of the hydrogen molecule :math:`H_2`.
+# Now that we've explored the theoretical underpinnings of SPSA and its use for a
+# toy problem optimization, let's use it
+# to optimize a real chemical system, namely that of the hydrogen molecule :math:`H_2`.
```
timmysilv, line 388: I've heard the qchem folks call the H2 molecule a toy problem lol 😄 luckily I think their purposes differ sufficiently from an SPSA demo

demonstrations/spsa.py
```diff
@@ -414,109 +374,88 @@ def cost_gd(params):
...
 # IBMQ.providers()    # List all available providers
+# dev = qml.device("qiskit.ibmq", wires=num_qubits, backend="ibmq_lima")
 
-dev = qml.device("qiskit.ibmq", wires=num_qubits, backend="ibmq_lima")
-noise_model = noise.NoiseModel.from_backend(dev.backend)
-dev_noisy = qml.device(
-    "qiskit.aer", wires=dev.num_wires, shots=1000, noise_model=noise_model
-)
+from qiskit.providers.aer import noise
+from qiskit.providers.fake_provider import FakeLima
```
timmysilv, line 432: was FakeLima just not available before? are we making this switch because it's easier on users?

demonstrations/spsa.py
```diff
@@ -171,86 +169,80 @@
...
 
-def cost_spsa(params):
-    return qnode_spsa(params.reshape(num_layers, num_wires, 3))
+##############################################################################
+# We will execute a few optimizations in this demo, so let's prepare a convenience
+# function that runs an optimizer instance and records the cost values and the
+# number of device executions along the way. The latter will be an interesting
+# quantity to evaluate the optimization cost on hardware!
+
+def run_optimizer(opt, fun, init_param, num_steps):
```
rmoyard, line 196: Minor: Just to make it a be more clear; you could name the arg `optimizer` and `cost_function`

demonstrations/spsa.py
```diff
@@ -284,63 +268,42 @@ def callback_fn(xk):
...
-# Create a device, qnode and cost function specific to gradient descent
-dev_sampler_gd = qml.device("qiskit.aer", wires=num_wires, shots=1000)
-qnode_gd = qml.QNode(circuit, dev_sampler_gd)
-
+# convergence. Note that we also create a new device in order to reset the
+# execution count to 0. With the new device, we recreate the cost function as well.
 
-def cost_gd(params):
-    return qnode_gd(params)
+# Create a new device and qnode
```
rmoyard, line 300:
```suggestion
# Create a new device and a qnode as cost function
```

demonstrations/spsa.py
```diff
@@ -398,14 +357,15 @@ def cost_gd(params):
...
 # compared to gradient descent!
 #
-# Let's take a deeper dive to see how much better it actually is.
+# Let's take a deeper dive to see how much better it actually is by computing
+# the ratio of required device executions to reach an absolute accuracy of 0.01.
 #
-grad_desc_exec_min = device_execs_grad[np.argmin(cost_store_grad)]
-spsa_exec_min = device_execs_spsa[np.argmin(cost_store_spsa)]
-print(f"Device execution ratio: {grad_desc_exec_min/spsa_exec_min}.")
+grad_execs_to_prec = exec_history_grad[np.where(np.array(cost_history_grad) < 0.800)[0][0]]
```
rmoyard, line 366: Where does the 0.800 come from?

demonstrations/spsa.py
rmoyard: I think it is the new recommended way of getting the noise model

demonstrations/spsa.py
```diff
@@ -525,23 +464,32 @@ def exp_val_circuit(params):
...
 #  .. code-block:: none
 #
-#     Iteration = 0, Number of device executions = 333,  Energy = -0.75215552 Ha
-#     Iteration = 5, Number of device executions = 1998,  Energy = -1.07779739 Ha
-#     Iteration = 10, Number of device executions = 3663,  Energy = -1.06789006 Ha
-#     Iteration = 15, Number of device executions = 5328,  Energy = -1.07183096 Ha
-#
-#     Final estimated value of the ground-state energy = -1.07234725 Ha
-#     Accuracy with respect to the true energy: 0.06384221 Ha
+# Running the GradientDescentOptimizer optimizer for 15 iterations.
```
rmoyard, line 467: Given the numbers it might be not worth to display all of them

demonstrations/spsa.py
```diff
@@ -635,47 +543,40 @@ def callback_fn(xk):
...
-#     Iteration = 150, Number of device executions = 3360,  Energy = -1.05679187 Ha
-#     Iteration = 160, Number of device executions = 3585,  Energy = -1.06176909 Ha
-#     Iteration = 170, Number of device executions = 3810,  Energy = -1.04974101 Ha
-#     Iteration = 180, Number of device executions = 4035,  Energy = -1.06182034 Ha
-#     Iteration = 190, Number of device executions = 4260,  Energy = -1.05587549 Ha
-#     Iteration = 200, Number of device executions = 4485,  Energy = -1.05229175 Ha
-#
-#     Final estimated value of the ground-state energy = -1.06069987 Ha
-#     Accuracy with respect to the true energy: 0.07548958 Ha
+# Running the SPSAOptimizer optimizer for 200 iterations.
```
rmoyard, line 546: Here you also reduce the number of executions, as you have more executions per iteration

demonstrations/spsa.py
```diff
@@ -739,6 +638,12 @@ def callback_fn(xk):
...
 #    vol. 45, no. 10, pp. 1839-1853, Oct 2020, doi:
 #    10.1109/TAC.2000.880982.
+#
+# .. [#qnspsa]
+#
+#    J. Gacon, C. Zoufal, G. Carleo, and S. Woerner "Simultaneous Perturbation
+#    Stochastic Approximation of the Quantum Fisher Information",
+#    `Quantum, 5, 567 <https://quantum-journal.org/papers/q-2021-10-20-567/>`__, Oct 2021
 
 ##############################################################################
```
rmoyard, line 648: https://github.com/PennyLaneAI/qml/blob/71dc81cc508036b7d8474adf3825ed37fa321f21/demonstrations/spsa.py#L588

I would suggest to match the number of executions on this graph

demonstrations/spsa.py
```diff
@@ -739,6 +638,12 @@ def callback_fn(xk):
...
 # About the author
```
rmoyard, line 649: You can add yourself here :100:

rmoyard (commented):
Great job @dwierichs :100: I have left some comments and suggestions

demonstrations/spsa.py
josh146: Yep agreed!

demonstrations/spsa.py
josh146: Huh. Is noisy opt more efficient than ours?

dwierichs: So the previous version of the demo divided the iteration count for spsa (and only for spsa!) by 2.
It had a [comment associated with that line](https://github.com/PennyLaneAI/qml/blob/e50d7b0483252d31f20296ae35fc2402caa77436/demonstrations/spsa.py#L232) stating that this was in order to correct for the execution of the device to monitor the cost function itself. However, that was incorrect, the correct mitigation for these evaluations is to subtract one device execution per iteration. 
SPSA uses two evaluations for each gradient estimate, so having `1.5*num_epochs` evaluations overall is not possible.
This is also seen further down where the text of the demo made the prognosis that we will need 6000 executions and then the number of executions reported by the log is smaller than that.

dwierichs: Note that this bugfix gets us into trouble regarding the performance in the VQE example. I guess mostly because it has fewer parameters.

demonstrations/spsa.py
```diff
@@ -13,14 +13,16 @@
    tutorial_vqe A brief overview of VQE
    tutorial_vqe_qng Accelerating VQEs with quantum natural gradient
 
-*Author: Antal Szava — Posted: 19 March 2021. Last updated: 13 September 2022.*
+*Author: Antal Szava — Posted: 19 March 2021. Last updated: xx February 2023.*
 
-In this tutorial, we investigate using a gradient-free optimizer called
```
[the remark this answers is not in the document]
dwierichs: Ah, no, we're still using the `SPSAOptimizer` built into PennyLane, replacing `noisyopt`.
However, I find it a bit confusing to say "gradient-free" because `SPSA` _does_ estimate the gradient. 
I mean, it is not an analytic gradient but a stochastic estimate, but that's true in quantum in many other settings.
If `SPSA` is "gradient-free", by that logic, gradient descent with a shot-based simulator also is "gradient-free" :grimacing:

demonstrations/spsa.py
```diff
@@ -171,86 +169,80 @@
...
-# We seed so that we can simulate the same circuit every time.
-np.random.seed(50)
+device = qml.device("qiskit.aer", wires=num_wires, shots=1000)
 
-all_pauliz_tensor_prod = qml.operation.Tensor(
-    *[qml.PauliZ(i) for i in range(num_wires)]
-)
+ansatz = qml.StronglyEntanglingLayers
 
+all_pauliz_tensor_prod = qml.operation.Tensor(*[qml.PauliZ(i) for i in range(num_wires)])
```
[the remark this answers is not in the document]
dwierichs: Ah, I'll check :+1:

demonstrations/spsa.py
dwierichs: Oh, this is outdated. Thanks!

demonstrations/spsa.py
dwierichs: The previous code did not execute on my machine, asking for IBM Q account things. I thought it's nicest if one does not have to bother with that just to get the noise model. Web search suggested that this indeed is the current way of obtaining the noise model :)

demonstrations/spsa.py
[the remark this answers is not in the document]
dwierichs: Works.

demonstrations/spsa.py
dwierichs: Went back to every 5th.

demonstrations/spsa.py
dwierichs: This is actually wrong again, I have patched the `run_optimizer` function to account for Hamiltonian decompositions.

demonstrations/spsa.py
dwierichs: As there was an outdated threshold in here, the 10 was wrong.
We're back to about 4 :D

demonstrations/spsa.py
dwierichs: We're now at 4800 (SPSA) vs 4995 (GradDesc).

dwierichs:
Given that we switch the IBMQ import to a `fake_providers` noise model, one could just have this tutorial run like the others, without hardcoding outputs and plots.
On my laptop it took about 8 minutes to run in its latest form.
@josh146  @rmoyard @timmysilv what do you think?

josh146:
> Given that we switch the IBMQ import to a fake_providers noise model, one could just have this tutorial run like the others, without hardcoding outputs and plots.

I am in favour!

rmoyard:
> Given that we switch the IBMQ import to a fake_providers noise model, one could just have this tutorial run like the others, without hardcoding outputs and plots.

@dwierichs I am also in favor, can we move it to tutorial then? It will be run automatically and remove some burden on the core team :)

dwierichs:
I moved it to `tutorial_spsa.py`. The one main issue that needs addressing is that SPSA does not outperform Gradient descent anymore, now that we updated the way we count the executions.
One could test with an enhanced ansatz, trying to get better energies for both optimizers.

dwierichs:
>The one main issue that needs addressing is that SPSA does not outperform Gradient descent anymore, now that we updated the way we count the executions.

"fixed" this by using a different ansatz.

demonstrations/spsa.py
timmysilv: but that's because we were.. lying before? so the new code isn't worse, just more honest?

dwierichs: If I'm not entirely misunderstanding what was going on before, yes.

rmoyard (approved):
Thanks it looks good to me :100: Very great update and good catch for the ansatz!

timmysilv (approved):
love it, thanks for getting this done! don't forget to update the date on the demo. also, should we add QNSPSA to the Related section for SPSA? I'm fine either way, just figured since QNSPSA relates to SPSA.

dwierichs:
I was wondering about this as well. I thought it's not a causal dependence or anything like that, so having a cycle of ref links seems ok :)

demonstrations/tutorial_spsa.py
```diff
@@ -0,0 +1,555 @@
...
+        approximation algorithm to optimize variational circuits in PennyLane.
+    :property="og:image": https://pennylane.ai/qml/_images/spsa_mntn.png
+
+.. related::
+
+   tutorial_vqe A brief overview of VQE
+   tutorial_vqe_qng Accelerating VQEs with quantum natural gradient
+   qnspsa Quantum natural SPSA optimizer
+
+*Authors: Antal Szava & David Wierichs — Posted: 19 March 2021. Last updated: xx February 2023.*
```
dwierichs, line 17:
```suggestion
*Authors: Antal Szava & David Wierichs — Posted: 19 March 2021. Last updated: 10 February 2023.*
```

PR #696, opened 2023-02-03
Added links to docs and GitHub

ikurecic:
Linking opportunity spotted by @alvaro-at-xanadu 

(I'm tired and bounced from the wrong branch, but these caps need fixing anyway, so I'll leave both commits here.)


josh146 (approved):
Brilliant, thanks @ikurecic! Super fast 🚄

PR #700, opened 2023-02-14
Update Autograd Requirement

AmintorDusko:
**Title:** Update Autograd Requirement.

**Summary:** Upgrading Autograd version to keep up with the Numpy deprecation cycle. 
This PR also unpin the numpy version.

**Relevant references:** 

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 --find-links https://download.pytorch.org/whl/torch_stable.html
 appdirs==1.4.4
-autograd==1.3
-numpy~=1.21,<1.24
+autograd==1.5
+numpy~=1.24
 numba==0.53.1
 networkx==2.5.1
 nlopt==2.6.2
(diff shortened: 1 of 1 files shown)


josh146:
@AmintorDusko I've added the `#ignore-qml-cache` label -- this will force the PR to build _all_ demos from scratch, to ensure it still works. However, you may need to push a new commit to enable this!

AmintorDusko:
Hi @josh146, We can see that the building is falling because of the numpy version restriction on PennyLane. I'm preparing a PR there to remove it.

AmintorDusko:
We can't unpin the numpy version on PennyLane because numpy removed support for ragged arrays.
I will downgrade the numpy minimal version here.

josh146 (approved):
Thank you @AmintorDusko!

PR #701, opened 2023-02-21
Replaced Grouping module with Pauli module from all demos

Jaybsoni:
**Title: Stop using the now deprecated grouping module**

**Summary:**
We are removing the grouping module in favour of the Pauli module. We remove uses of it from the demos and instead use the same functionality from the Pauli module.


Jaybsoni:
[sc-33546]

mudit2812 (approved):
Looks good!

rmoyard:
Lgtm!

rmoyard (approved):

PR #703, opened 2023-02-21
Change ibm device in Quantum volume

rmoyard:


timmysilv:
is it expected that this device expands the circuit differently? the results look completely different, just wondering if that's alright for the results being demonstrated.

josh146:
Might be good to tag @glassnotes in that case :)

demonstrations/quantum_volume.py
```diff
@@ -557,7 +557,7 @@ def heavy_output_set(m, probs):
 #    Users can get a list of available IBM Q backends by importing IBM Q,
 #    specifying their provider and then calling: ``provider.backends()``
 #
-dev_ourense = qml.device("qiskit.ibmq", wires=5, backend="ibmq_bogota")
+dev_ourense = qml.device("qiskit.ibmq", wires=5, backend="ibm_oslo")
```
mudit2812, line 560: Why are we switching to `ibm_oslo`? Oslo is a 7 qubit device. We could use a 5 qubit device like manila, lima, etc.

mudit2812 (approved):
Just curious about why we switched to a 7-qubit device, but other than that, looks good

glassnotes:
> is it expected that this device expands the circuit differently? the results look completely different, just wondering if that's alright for the results being demonstrated.

@timmysilv Thanks for keeping this up to date, it's a challenging one to maintain! This is no problem, the protocol works by generating random SU(4) *and* random permutations, which is why it looks so different. It looks like the results at the end are sensible (log2(QV) = 3), though it is probably worth checking that the QV of Oslo is, in fact, reported as 8 and not significantly higher or lower. 

Side note: there are numerous explicit references to the previous device (Ourense) in the demo, which would have to be updated in addition to numerical results.

mudit2812:
@glassnotes `ibm_oslo` has QV=32 😅

mudit2812:
I might have approved prematurely. As @glassnotes pointed out, there are a lot of references to `ourense` in the demo itself. Moreover, the QV is incorrect as `ibm_oslo` has QV 32, so our result should be `log2(32)=5`. I think that can be fixed by changing the number of wires to 7.

glassnotes:
> I might have approved prematurely. As @glassnotes pointed out, there are a lot of references to `ourense` in the demo itself. Moreover, the QV is incorrect as `ibm_oslo` has QV 32, so our result should be `log2(32)=5`. I think that can be fixed by either changing `num_wires` to 7, or changing the backend to a 5-qubit device.

Part of the reason for the discrepancy is that we're using a simulated noise model in the demo, and only 1000 shots.

But @mudit2812 as you point out, `num_wires` should be 7. QV takes into account the compilation process, which involves qubit placement and routing. If not enough wires are specified, it might not be able to access to the "good" qubits and full connectivity of the hardware. I don't think that will totally account for the jump from 3 to 5, but it should at least help.

demonstrations/quantum_volume.py
rmoyard: Change to Lima, same # qubits, same connectivity and same QV!

demonstrations/quantum_volume.py
```diff
@@ -697,16 +694,16 @@ def heavy_output_set(m, probs):
...
 #
 #     Device mean probabilities:
-#     m = 2: 0.773760 above threshold.
-#     m = 3: 0.794875 above threshold.
-#     m = 4: 0.722860 above threshold.
-#     m = 5: 0.692935 above threshold.
+#     m = 2: 0.765920 above threshold.
+#     m = 3: 0.773985 above threshold.
+#     m = 4: 0.674380 above threshold.
+#     m = 5: 0.639500 below threshold.
```
mudit2812, line 706: Since the mena probability for `m=5` is below the threshold, I would update the documentation below (lines 710-719) accordingly as well, as they currently state that ALL noisy device probabilities are above the threshold.

mudit2812 (commented):
This looks much better. Thanks for the changes! Some of the documentation needs minor updates, but other than that, this looks good!

mudit2812 (changes requested):
I think this is ready to merge once the updates from my review are added

mudit2812 (approved):
Thanks for the changes @rmoyard

timmysilv (approved):
looks great!

PR #704, opened 2023-02-21
Update chemical reactions demo to use pyscf for openshell H3

soranjh:
**Summary:**
The demo on _modelling chemical reactions on a quantum computer_ uses the differentiable Hartree-Fock solver for constructing a molecular Hamiltonian. The demo fails after merging this [PR](https://github.com/PennyLaneAI/pennylane/pull/3748) which raises an error when the differentiable Hartree-Fock solver is requested with systems with unpaired electrons. 

The call to `molecular_hamiltonian`  is updated to use `method='pyscf'` until we extend the differentiable Hartree-Fock solver to treat systems with unpaired electrons. The results in the demo are not affected but, in general, it is better to use `method='pyscf'` for these type of systems.

### DIFF
--- a/demonstrations/tutorial_chemical_reactions.py
+++ b/demonstrations/tutorial_chemical_reactions.py
@@ -129,7 +129,7 @@ for r in r_range:
     coordinates = np.array([0.0, 0.0, 0.0, 0.0, 0.0, r])
 
     # Obtain the qubit Hamiltonian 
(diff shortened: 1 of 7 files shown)


rmoyard (approved):
Thanks @soranjh for fixing that issue!

timmysilv (approved):

PR #706, opened 2023-02-23
Update SPSA demo to count circuit executions manually.

dwierichs:
The way the Qiskit aer plugin device counts executions was changed in (this patch)[https://github.com/PennyLaneAI/pennylane-qiskit/pull/259].
This PR accounts for this change by switching to manual counting of executions.


demonstrations/tutorial_spsa.py
```diff
@@ -191,52 +191,38 @@ def circuit(param):
 
 ##############################################################################
 # We will execute a few optimizations in this demo, so let's prepare a convenience
-# function that runs an optimizer instance and records the cost values and the
-# number of device executions along the way. The latter will be an interesting
-# quantity to evaluate the optimization cost on hardware!
+# function that runs an optimizer instance and records the cost values
+# along the way. Together with the number of device executions, this will be an
```
rmoyard, line 195: I think here and in general we do not have device executions but circuit executions.

rmoyard (commented):
Thanks @dwierichs for fixing that it looks great :+1: One thing here: I think we should replace device executions by circuits executions

demonstrations/tutorial_spsa.py
dwierichs: Done.

mudit2812 (commented):
Looks good. Thanks for the changes! I just left a few non-blocking questions

rmoyard (approved):
Thanks ! :+1:

demonstrations/tutorial_spsa.py
```diff
@@ -191,52 +191,38 @@ def circuit(param):
...
         cost_history.append(cost_function(param))
 
         # Print out the status of the optimization
         if step % interval == 0:
-            print(
-                f"Iteration = {step:3d}, "
-                f"Device executions = {exec_history[step]:4d}, "
-                f"Cost = {cost_history[step]}"
-            )
+            print(f"Iteration = {step:3d}, Cost = {cost_history[step]}")
```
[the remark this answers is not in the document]
dwierichs: Because it is manually accounted for now, and for the execution of the optimizer we use the same function four times. Making the manual accounting available within that function would require more argument passing around and so on, and I thought it's not crucial.

demonstrations/tutorial_spsa.py
```diff
@@ -285,16 +272,16 @@ def run_optimizer(optimizer, cost_function, init_param, num_steps, interval):
 
 num_steps_grad = 15
 opt = qml.GradientDescentOptimizer(stepsize=0.3)
-cost_history_grad, exec_history_grad = run_optimizer(
-    opt, cost_function, init_param, num_steps_grad, 3
-)
+cost_history_grad = run_optimizer(opt, cost_function, init_param, num_steps_grad, 3)
+# We have 16 energy recordings, including the initial value, and spent 2 circuit
+# evaluations *per parameter* in each step
+exec_history_grad = np.arange(num_steps_grad + 1) * (2 * np.prod(param_shape))
```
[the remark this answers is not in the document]
dwierichs: see above.

demonstrations/tutorial_spsa.py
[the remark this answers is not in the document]
dwierichs: I thought this would take much rewriting, but actually it is just one more parameter. I conceptually prefer that we have the `run_optimizer` function return the execution number history, so I changed it.
Accoridngly, it now also is printed again :)

dwierichs:
For some reason, in the second example (VQE), gradient descent was reporting 531 circuit executions per step, which I don't think is correct:
We have 15 Hamiltonian terms (and no grouping for the simple demo), parameter shape `(2, 4, 3)` and 2 circuit executions per parameter (parameter shift of Pauli rotations). This leads to `15*2*4*3*2=720` circuits per step, which is the reported number now.

demonstrations/tutorial_spsa.py
```diff
@@ -414,7 +425,7 @@ def circuit(param):
 plt.ylabel("Energy (Ha)", fontsize=14)
 plt.grid()
 
-plt.axhline(y=true_energy, color="black", linestyle="dashed", label="True energy")
+plt.axhline(y=true_energy, color="black", linestyle="--", label="True energy")
```
mudit2812, line 428: For the latter two plots, the number of executions I'm seeing locally for gradient descent is ~10,500 as compared to ~8,000 on master. Would you say that's within the margin of error? There's always some uncertainty in the number of steps needed for convergence with gradient descent due to random initialization, so I would imagine it's fine.

mudit2812 (approved):
Thanks for the changes David! The local number of executions I saw for gradient descent was a bit higher than master, but I think it should be fine.

demonstrations/tutorial_spsa.py
mudit2812: Just saw your comment from a few hours ago. The number of executions I see locally is consistent with 720 executions per step, so it seems like it is expected. Thanks

dwierichs: Yeah, I'm not sure what created this number of executions per step :thinking:

PR #708, opened 2023-02-23
Update classical shadows

rmoyard:
Cannot trustworthy set the device if the interface is auto.


Qottmann (approved):
Amazing @rmoyard 🥳

mudit2812 (approved):
Thanks for updating

rmoyard:
@rashidnhm The save build context  is not running, anything I can do?

PR #710, opened 2023-02-27
[sc-34057]: Parallelised building of QML Branches

rashidnhm:
**Title:** Parallelised building of QML Branches

**Summary:**
This PR updates the existing re-usable workflow to build branches to support parallelisation. 

Note:
You may notice that the re-usable workflow looks very close to identical to the build-pr.yml workflow now. The re-usable workflow is indeed almost identical to the existing build-pr.yml workflow, this is due to that they both essentially do the same thing now. I will port over build-pr.yml workflow to use this re-usable workflow so that there is no duplication of the workflow code, but have deemed it out of scope for this story.

**Relevant references:**
- [sc-34057](https://app.shortcut.com/xanaduai/story/34057/qml-fix-time-limit-issue-on-build-master-and-build-dev)

**Possible Drawbacks:**
N/A. Possible bugs in the new workflow, will tackle as any come up.

**Related GitHub Issues:**
Time limit failures in existing branch builds:
- https://github.com/PennyLaneAI/qml/actions/runs/4189316884
- https://github.com/PennyLaneAI/qml/actions/runs/3897838260
- https://github.com/PennyLaneAI/qml/actions/runs/4248212489
- https://github.com/PennyLaneAI/qml/actions/runs/4088978040


.github/workflows/build-branch.yml
```diff
@@ -1,59 +1,328 @@
 name: Build QML Branch
 on:
-  workflow_dispatch:
-    inputs:
-      branch:
-        description: The QML branch to checkout and build demos for
-        required: true
-        type: string
```
josh146, line 8: out of curiosity how come this was removed?

.github/workflows/build-branch.yml
```diff
@@ -1,59 +1,328 @@
...
+      enable_sphinx_cache:
+        description: Use actions/cache for sphinx and sphinx-gallery
+        required: false
+        type: boolean
+        default: false
+      refresh_sphinx_cache:
+        description: Build QML Demos without using cache and create new caches after the build
+        required: false
+        type: boolean
+        default: false
```
josh146, line 23: I'm curious if saving this cache for PRs to use will have any unintended circumstances. Tagging @trbromley and @rmoyard here

josh146 (approved):
Thank you @rashidnhm! did a quick skim but also good to get an approval from someone with more CI/CD knowledge than me :)

Just to double check, but this new action will still be triggered twice weekly?

Also, do the readme badges need to be updated?

.github/workflows/build-branch.yml
rmoyard: What cache are we talking about here? Not the builds of the website?

.github/workflows/build-branch.yml
rashidnhm: Due to inconsistencies and nuances within GitHub Actions itself, there were issues handling both `workflow_dispatch` and `workflow_call` within the same re-usable action. That being said, I didn't want workflow_dispatch to be added directly to the re-usable action itself. We have the existing [build-branch-dev](https://github.com/PennyLaneAI/qml/blob/0e76f071313360d58078779eb8362bdbed5963df/.github/workflows/build-branch-dev.yml) and [build-branch-master](https://github.com/PennyLaneAI/qml/blob/0e76f071313360d58078779eb8362bdbed5963df/.github/workflows/build-branch-master.yml) workflows that have workflow_dispatch added to them that users can use to trigger a build of said branch.

.github/workflows/build-branch.yml
rashidnhm: > I'm curious if saving this cache for PRs to use will have any unintended circumstances. 

We already do this in [build-pr.yml](https://github.com/PennyLaneAI/qml/blob/5469e25aec0221997b4faef45ba0a5dfc9be4891/.github/workflows/build-pr.yml#L126-L142) where the cache from master/dev is used for the PRs. This `refresh_sphinx_cache` parameters mimicks what happens today when we add the `ignore-qml-cache` label to PRs.

rashidnhm:
> Thank you @rashidnhm! did a quick skim but also good to get an approval from someone with more CI/CD knowledge than me :)
> 
> Just to double check, but this new action will still be triggered twice weekly?
> 
> Also, do the readme badges need to be updated?


> this new action will still be triggered twice weekly

@josh146 that is correct, it will be triggered from either [build-branch-dev.yml](https://github.com/PennyLaneAI/qml/blob/0e76f071313360d58078779eb8362bdbed5963df/.github/workflows/build-branch-dev.yml) or [build-branch-master.yml](https://github.com/PennyLaneAI/qml/blob/0e76f071313360d58078779eb8362bdbed5963df/.github/workflows/build-branch-master.yml).


> Also, do the readme badges need to be updated?

The badges on the [plugin test matrix](https://github.com/PennyLaneAI/plugin-test-matrix/pull/29) was updated back in December to show build result of the two new workflows.
... QML repo itself needs to be updated though that is a quick fix.

.github/workflows/build-branch.yml
josh146: @rmoyard, builds of the tutorials :)

smokingroosters (approved):

PR #713, opened 2023-03-02
Added more logging to parse_execution_times if regex error is triggered

rashidnhm:
**Title:** Added more logging to parse_execution_times if regex error is triggered

**Summary:**
When parsing and generating the execution time, there seems be an occasional case where the execution time for all demos cannot be parsed out by the regex from the sphinx generated html.

This case has not been reproducible locally, this PR adds more verbose logging for when this happens.

**Relevant references:**
https://github.com/PennyLaneAI/qml/actions/runs/4309304990/jobs/7516549630

**Possible Drawbacks:**
N/A.

**Related GitHub Issues:**
N/A.


smokingroosters (approved):

josh146 (approved):
Thanks @rashidnhm!

trbromley (approved):

PR #716, opened 2023-03-15
Fixed thumbnail issue in demonstrations.rst

ikurecic:
Copied the thumbnail to the correct folder, fixed the call in the demonstrations.rst file.


josh146 (approved):
Looks good @ikurecic!

I would suggest though _not_ putting anything in `_static/thumbs`, and instead having the URL for the image being `_images/thumbnail_vqt.png`.

This is because I have a feeling `static/thumbs` is autogenerated, so it might be overwritten...

ikurecic:
> Looks good @ikurecic!
> 
> I would suggest though _not_ putting anything in `_static/thumbs`, and instead having the URL for the image being `_images/thumbnail_vqt.png`.
> 
> This is because I have a feeling `static/thumbs` is autogenerated, so it might be overwritten...

That wouldn't surprise me, but instead of trawling through the code I tagged @Anakhom for a review. 😅
I deleted the file from the _static folder — Ana, can you sort us out here? :) Does this update work?

Anakhom:
> > Looks good @ikurecic!
> > I would suggest though _not_ putting anything in `_static/thumbs`, and instead having the URL for the image being `_images/thumbnail_vqt.png`.
> > This is because I have a feeling `static/thumbs` is autogenerated, so it might be overwritten...
> 
> That wouldn't surprise me, but instead of trawling through the code I tagged @Anakhom for a review. 😅 I deleted the file from the _static folder — Ana, can you sort us out here? :) Does this update work?

@josh146, @ikurecic  I fixed the path, everything looks good locally!

josh146 (approved):
Thank you @Anakhom!

Anakhom (approved):

PR #726, opened 2023-03-29
Demo metadata schema and validation

doctorperceptron:
This PR adds a [JSON schema](https://json-schema.org/) defining the structure of metadata files for demos.

## Usage

To install the necessary dependencies, use

```console
$ poetry install --only metadata-validation
```

Then, demo metadata files can be validated against the schema manually using:

```console
$ cd metadata_schemas
$ poetry run check-jsonschema \
  --schemafile demo.metadata.schema.0.1.0.json \
  ../demonstrations/<your_demo_name>.metadata.json
```

## Notes

* The JSON schema is based on `demonstrations_metadata.md`.
* There is good reason to make this the source of truth for the schema. It can replace the markdown definition.
* There is a GitHub Actions workflow for validating metadata files.
* In the future, we'll want to change the `$id` field for the schemas to be GitHub. This makes them more portable.
* Some fields (e.g., datetimes and categories) were cleaned up to be consistent with the schema; however, most of the metadata files don't pass validation because they either lack references or are missing required reference fields.

[demo_validation.webm](https://user-images.githubusercontent.com/50180049/228660149-0af0854a-3326-439b-9a2f-397b9d200e52.webm)


metadata_schemas/demo.metadata.schema.0.1.0.json
```diff
@@ -0,0 +1,105 @@
...
+        "type": "string",
+        "format": "date-time",
+        "minLength": 25,
+        "maxLength": 25
+      },
+      "categories": {
+        "description": "An array of the categories that this demo is in.",
+        "type": "array",
+        "items": {
+          "enum": ["Getting Started", "Optimization", "QML", "Quantum Chemistry", "Quantum Computing", "Community"]
```
BM7878, line 39: We'll need to update these to the new categories, which are in `demonstrations/demonstrations-categories.metadata.json`.

metadata_schemas/objects/author.schema.0.1.0.json
```diff
@@ -0,0 +1,21 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "author.schema.0.1.0.json",
+    "title": "Authors",
```
BM7878, line 4: 'Author Object'?

metadata_schemas/objects/preview.image.schema.0.1.0.json
```diff
@@ -0,0 +1,21 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "preview.image.schema.0.1.0.json",
+    "title": "Preview Image Object",
+    "description": "The different images that can be used as previews for the demo - e.g., thumbnails, social media cards (perhaps of different aspect ratios).",
+    "type": "object",
+    "properties": {
+        "type": {
+            "description": "What type of preview image this is. At the moment, the only value this can take is thumbnail, which refers to the image used on the QML part of pennylane.ai when browsing through lists of demos.",
+            "enum": ["thumbnail"]
```
BM7878, line 10: This can now also be 'hero_image' as well.

BM7878 (changes requested):

metadata_schemas/demo.metadata.schema.0.1.0.json
Mandrenkov: Done!

metadata_schemas/objects/preview.image.schema.0.1.0.json
Mandrenkov: Thanks! This can also be `large_thumbnail` too, right?

metadata_schemas/objects/author.schema.0.1.0.json
Mandrenkov: Good catch! 🎣

Mandrenkov:
Thanks for the review, @BM7878! 🚀 

Note that there are still a number of validation errors. These are:

1. Two demos (**Adjoint Differentiation** and **Multiclass margin classifier**) are not associated with any categories.
2. One demo (**The stochastic parameter-shift rule**) lists a type of related content as `glossary_entry`.
3. Some demos are missing one or more of the following related content fields:
    * `authors`
    * `id`
    * `title`
    * `type`
    * `year`
4. Some demos have blank strings (`""`) for one or more of the following related content fields:
    * `authors`
    * `id`
    * `title`
    * `year`

What should we do to resolve these issues? Try to fix the demos? Or adapt the schema accordingly?

Mandrenkov:
Oh, and another (minor) issue is that some of the canonical URLs only include the path component e.g.
```
/qml/demos/ahs_aquila
```
instead of the full URL
```
https://pennylane.ai/qml/demos/ahs_aquila
```
Is this intentional?

josh146:
> Two demos (Adjoint Differentiation and Multiclass margin classifier) are not associated with any categories.

Oh I can comment on this one. This is because we want the demos on the website, but we don't want them appearing within any of the categories.

For Adjoint diff, this is 'supplementary' information that is linked from one demo.

For Multiclass, I believe there is a bug in the demo, and we want to 'hide' it from the demos page without it coming down completely.

doctorperceptron:
> > Two demos (Adjoint Differentiation and Multiclass margin classifier) are not associated with any categories.
> 
> Oh I can comment on this one. This is because we want the demos on the website, but we don't want them appearing within any of the categories or carousels for end-users who are browsing.
> 
> For Adjoint diff, this is 'supplementary' information that is linked from one demo.
> 
> For Multiclass, I believe there is a bug in the demo, and we want to 'hide' it from the demos page without it coming down completely.

Maybe we could add something like a "Hidden" category? Allowing no category makes it tricky to enforce having a category for "regular" demos.

requirements-dev.txt
rashidnhm, line 1: This PR was likely opened while the QML repo used requirements files, however it's been ported over to using Poetry. Would you be able to update this branch with the changes from master, and add these as a poetry dev group?

Mandrenkov: Good call, @rashidnhm! I've created a Poetry group `metadata-validation` and updated the lock file.

.github/workflows/validate-demo-metadata.yml
```diff
@@ -0,0 +1,34 @@
...
+# Consider combining this with the build-pr
+# Action if we want to keep it.
+name: Validate Demo Metadata
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - dev
+
```
rashidnhm, line 11: Let's add a concurrency group here as well so older buids are auto-cancelled.

```suggestion
concurrency:
  group: metadata-validation-${{ github.ref }}
  cancel-in-progress: true
```

pyproject.toml
```diff
@@ -85,3 +85,8 @@ qulacs = {version = "0.1.10.1", markers = "platform_machine == 'x86_64'" }
 
 # The following packages are only installed on MacOS for compatibility
 tensorflow-macos = { version = "2.8.0", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" }
+
```
rashidnhm, line 88:
```suggestion

[tool.poetry.group.metadata-validation]
optional = true

```

So the metadata validation group isn't installed by default

Mandrenkov:
Looks like the CI is finally passing! 🥳 

Are there any other changes you would like to see, @BM7878 or @rashidnhm? The PR is still in need of an approval.

BM7878 (approved):

BM7878:
All's good from my end!

metadata_schemas/README.md
```diff
@@ -0,0 +1,19 @@
...
+Install the necessary dependencies using:
+
+```console
+$ poetry install --only metadata-validation
+```
+
+Then run the following command **from this directory**:
+
+```console
+$ poetry run check-jsonschema --schemafile demo.metadata.schema.[VERSION].json ../demonstrations/*.metadata.json
```
rashidnhm, line 18: Let's update this to exclude the metadata categories file

rashidnhm (approved):

metadata_schemas/README.md
Mandrenkov: Good idea! 💡 

Instead of listing the complicated `find` command from the GitHub Actions workflow though, I chose to simplify the README to show how to validate a single metadata JSON file.

PR #734, opened 2023-04-11
Sc 32461 notebook demo conveter

rashidnhm:
**Title:** Added new script that aids conversion from Jupyter Notebook to QML Demo

**Summary:**
This PR adds a new script that is able to convert a Jupyter Notebook to QML Demo.

As noted in the README, there would not be a scenario where a notebook is converted using this script and is instantly ready as a Demo. In most cases, some manual updates will be required, but this script attempts to do most of the initial conversion, which includes:

- Place the first cell of the Notebook as raw docstring.
- Format subsequent non-code cells as Python comments
- For non-executable demos, extract the output from the notebook and add that as a commented output code block in the demo.
- For non-executable demos, extract code image outputs and place them as demo assets and link them in the demo.
- Take author info as an input and automatically add author block at the end of the demo.

**Relevant references:**
[sc-32461](https://app.shortcut.com/xanaduai/story/32461/qml-develop-tool-to-be-able-to-convert-jupyter-notebook-to-python-files-and-vice-versa)

**Possible Drawbacks:**
If there are any major bugs, it can be addressed in subsequent PRs.

**Related GitHub Issues:**
None.


KetpuntoG:
Thanks for opening this PR! What does this error mean? I have deleted the images, probably this is a plot of the demo.

Captura de pantalla 2023-04-13 a las 10 53 36

notebook_converter/README.md
```diff
@@ -0,0 +1,97 @@
...
+  Information about the Author to put in their bio section
+```
+
+#### Options available in author file
+
+##### `name` (required)
+The display name for the Author
+
+#### `profile_picture` (required)
+The path to the display picture for the author.
```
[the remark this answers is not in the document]
josh146: Also once I got it working, I got an error 😆 

```python
shutil.SameFileError: PosixPath('/home/josh/xanadu/qml/notebook_converter/../_static/authors/josh_izaac.png') and PosixPath('/home/josh/xanadu/qml/notebook_converter/../_static/authors/josh_izaac.png') are the same file
```

I guess because the script won't work with existing authors, only new ones?

josh146 (changes requested):
Nice work @rashidnhm! Unfortunately I couldn't get the script to work with my notebook:

```console
  File "notebook_converter/notebook_to_demo.py", line 260, in <module>
    nb_py = convert_notebook_to_python(
  File "notebook_converter/notebook_to_demo.py", line 194, in convert_notebook_to_python
    with open(image_save_loc, "wb") as ifh:
IsADirectoryError: [Errno 21] Is a directory: '/home/josh/xanadu/qml/notebook_converter/../demonstrations/Optimization transform/Optimization transform_91527bf2_1.png'
```

notebook_converter/README.md
```diff
@@ -0,0 +1,97 @@
...
+## Setup
+`pip install -r notebook_converter/requirements.txt`
+
+Also, ensure [pandoc](https://pandoc.org/installing.html) is installed.
+
+## Running the Converter
+```bash
+python3 notebook_converter/notebook_to_demo.py \
+   path/to/my/notebook.ipynb \
+   --author-file=path/to/author_file.yaml
```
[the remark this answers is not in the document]
rashidnhm: @josh146 Top level of the repo. I will add this to the README. 

The reason for this is since it extracts images from the notebook (for non executable ones), and put in the link to it on the demo.

The path to the notebook and author file can be anywhere on the PC and it should be fine! But the script itself should be invoked from the root of the repo, I can potentially fix that by hardcoding certain values within the script.

notebook_converter/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Details on author file shown below
```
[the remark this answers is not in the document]
rashidnhm: Required, error will be raised by argparse otherwise

notebook_converter/README.md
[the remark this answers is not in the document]
rashidnhm: I thought about doing this, would we accept image this way too then? The other issue I was running into was determining the name the files would be saved as in the authors/ directory, which was easier to implement with the file.

Would you rather scrap the author file and pass everything by cli?

notebook_converter/README.md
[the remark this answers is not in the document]
rashidnhm: Oh that has been missed! I can definitely add a check to see if the file exists .. a lot of the issue you and Guillermo have run into seems to be something that occurs on Windows that I was not able to run into in linux. I will try to update the script and address these issues.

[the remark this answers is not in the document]
josh146: Oh I am using WSL 😆

KetpuntoG:
I had time to do a little test today and everything rendered 99% as expected. Great job! 🚀

It took me a while to get the paths right initially. On the other hand I think we could skip the author part for the moment or at least have the option of do it manually . It is clear enough how it is incorporated by looking at any other demo from the user's point of view and the creation of the yaml has confused me.

I still need to do more testing by adding images and non-executable tutorials but so far it is on the right track 😊

notebook_converter/README.md
```diff
@@ -0,0 +1,97 @@
...
+it must be relative to the script's location.
+
+#### `--author-file` (required)
+Path to a YAML file with information about the author. See information above to what this file has to contain.
+If a relative path is used, it must be relative to the script's location, absolute path is used verbatim.
+
+#### `--is-executable` (optional)
+Indicate if the notebook is intended to be an executable demo or non-executable. If this is not passed,
+the information is inferred from the notebook name. If the notebook name startswith `tutorial_` then it is
+considered executable.
```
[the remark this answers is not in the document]
KetpuntoG: Is there a reason other than time for making a demo non-executable? 
If it is because of credentials that you do not want to show, this would not be valid, right? 🤔
(I am not sure if this situation can occur)

notebook_converter/README.md
```diff
@@ -0,0 +1,97 @@
...
+```yaml
+name: Rashid N H M  # This would by default format to `rashid_n_h_m`
+formatted_name: rashid_nhm
+```
+
+### CLI Flags
+
+#### `/path/to/notebook` (required)
+The path to the notebook, if absolute path is passed, it will be used verbatim. If relative path is passed,
+it must be relative to the script's location.
```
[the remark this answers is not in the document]
KetpuntoG: I was in the root of qml and the path was relative to it

notebook_converter/README.md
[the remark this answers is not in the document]
josh146: @KetpuntoG time is one thing, the other is dependencies! E.g., if they are using a package, or even hardware, we don't support or have access to.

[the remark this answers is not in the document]
KetpuntoG: Ah I had misunderstood you, I thought we would run the demo internally to get the outputs but you are referring to his notebook. That makes sense! 👌😊

rashidnhm:
@josh146 @KetpuntoG Thank you for all the feedback! Thought I'd address most of the commennts here.

I will be going back to the drawing board and make the following updates:
- Make it so that the script can be called from anywhere and not require to be at the root of the repo.
- Make it so the author file is not required. Instead, introduce `--author-name` `--author-bio` & `--author-picture` cli parameters and make them optional.
- Handle the `SameFileError` that Josh pointed out.

An extra note:
@josh146 the point you made about the script running the non-executable demo locally and capturing the output ... though it is a great idea it might not be very trivial to add :sweat_smile: .  For now, I will add a note in the README that the user would need to run their notebook locally before running the conversion script.

I will convert this PR back to draft, feel free to unsubscribe from the notifications, I will add you all back once the updates are all complete!

rashidnhm:
@josh146 @KetpuntoG 

The path issue has been fixed! I have updated the README as well.

- The script can now be called from anywhere, you do not need to be within the qml repo
- The path to the notebook can be an absolute path or relative to where you are
- The author-file parameter has been removed, instead there is now `--author-name` `--author-bio` `--author-profile-picture` 
  - The path for author-profile-picture can be absolute or relative to wherever the user is at the time of calling the script.

These changes should make the script a lot easier and intuitive to use. Please let me know on the further experience.

notebook_converter/notebook_to_demo.py
```diff
@@ -0,0 +1,290 @@
...
+    parser.add_argument(
+        "--is-executable",
+        help="Indicate if the notebook is executable. "
+        "If not passed, this information is inferred "
+        "from the notebook file name. "
+        "If the notebook name startswith tutotial_, "
+        "then it is treated as executable.",
+        type=str_to_bool,
+        default=None
+    )
```
josh146, line 251: Would it make more sense to use
```python
parser.add_argument('--executable', action='store_true')
```

here? It avoids edge case issues with things like "true", "TRUE", etc.

notebook_converter/notebook_to_demo.py
```diff
@@ -0,0 +1,290 @@
...
+    parser = argparse.ArgumentParser(description="Convert Jupyter Notebook to QML Demo")
+
+    parser.add_argument("notebook", help="Path to file that needs to be converted")
+
+    parser.add_argument(
+        "--is-executable",
+        help="Indicate if the notebook is executable. "
+        "If not passed, this information is inferred "
+        "from the notebook file name. "
+        "If the notebook name startswith tutotial_, "
```
josh146, line 247:
```suggestion
        "If the notebook name starts with tutorial_, "
```

notebook_converter/notebook_to_demo.py
```diff
@@ -0,0 +1,290 @@
...
+    import argparse
+
+    parser = argparse.ArgumentParser(description="Convert Jupyter Notebook to QML Demo")
+
+    parser.add_argument("notebook", help="Path to file that needs to be converted")
+
+    parser.add_argument(
+        "--is-executable",
+        help="Indicate if the notebook is executable. "
+        "If not passed, this information is inferred "
```
josh146, line 245: This is a nice touch!

notebook_converter/notebook_to_demo.py
```diff
@@ -0,0 +1,290 @@
...
+        "If the notebook name startswith tutotial_, "
+        "then it is treated as executable.",
+        type=str_to_bool,
+        default=None
+    )
+
+    author_parser = parser.add_argument_group("Author")
+    author_parser.add_argument("--author-name", help="Full name of author", type=str, required=True)
+    author_parser.add_argument("--author-bio", help="Small description of author", type=str)
+    author_parser.add_argument("--author-profile-picture", help="Path to profile picture for author", type=str)
```
josh146, line 256: Nice, this was working well for me 🙏 

@rashidnhm I have two questions here:

- What should we do if the author already exists in `_static/authors`? I noticed that the current behaviour is to overwrite the file. It would be super nice to also have `--author-file=_static/authors/josh_izaac.txt` which could replace `--author-name` and `--author-bio`. Would this be easy to do?

- It occurred to me that some demos might have multiple authors - but maybe this is a problem for another day 😅

notebook_converter/notebook_to_demo.py
```diff
@@ -0,0 +1,290 @@
...
+                                image_file_dir / image_filename
+                            )
+                            image_file_link_path = DEMO["link-dir"] / notebook_assets_folder_name / image_filename
+                            role_text = generate_sphinx_role_comment(
+                                "figure", image_file_link_path.as_posix(), align="center", width="80%"
+                            )
+                            image_data = b64decode(output_data["image/png"])
+
+                            if not image_file_path.exists():
+                                image_file_path.mkdir(parents=True)
```
josh146, line 221: This branch is causing issues for me, because it is creating a _a directory with the image name_. That is, it is creating a directory of the form

```
demonstrations/notebook_name/notebook_name_3224.png/..
```

Which then causes the line 223 to fail for me, since Python can't write image data to a directory:

```
  File "notebook_converter/notebook_to_demo.py", line 222, in convert_notebook_to_python
    with open(image_file_path, "wb") as ifh:
IsADirectoryError: [Errno 21] Is a directory: '/home/josh/xanadu/qml/demonstrations/Optimization transform/Optimization transform_91527bf2_1.png'
```

To get this working, I commented out this branch and manually created the directory for images 🤔

josh146 (changes requested):
> the point you made about the script running the non-executable demo locally and capturing the output ... though it is a great idea it might not be very trivial to add 😅 . For now, I will add a note in the README that the user would need to run their notebook locally before running the conversion script.

Yep no worries at all, this is definitely a P2 in the grand scheme of things 😅 

@rashidnhm the new version is working really well for me, barring one bug which required me to comment out a bit of the script. Aside from that, I love the ability to capture output as codeblocks!!

notebook_converter/notebook_to_demo.py
rashidnhm: Yep this was a bug from my end, going to fix shortly!

notebook_converter/notebook_to_demo.py
rashidnhm: I can change it back to this, the downside to store_true is that since it defaults to `False` if the notebook starts with tutorial_, it would always be treated as executable.

The I have it now is if you pass `--is-executable=<boo>` then it overrides the assumption behaviour.

Also the str_to_bool function definitely handles the true,True,True cases, you can see that [here](https://github.com/PennyLaneAI/qml/blob/1f8146c951d0ccbe95a5a45ed7ef8ca2f2a67505/notebook_converter/notebook_to_demo.py#L26-L38)

notebook_converter/notebook_to_demo.py
rashidnhm: > What should we do if the author already exists in _static/authors? I noticed that the current behaviour is to overwrite the file. It would be super nice to also have --author-file=_static/authors/josh_izaac.txt which could replace --author-name and --author-bio. Would this be easy to do?

I would say very easy to do since if you have already made a authors file, there is no parsing to information required by the script! I can take the file drop it in demo in the authors section!

> It occurred to me that some demos might have multiple authors - but maybe this is a problem for another day

Oooo this might a bit tougher to handle with how it is designed now, I can say with the previous yaml setup multiple authors would be a lot simpler to handle, but I can think of something

notebook_converter/notebook_to_demo.py
josh146: oh perfect! Ignore my comment then @rashidnhm

notebook_converter/notebook_to_demo.py
rashidnhm: Should be fixed now, let me know

notebook_converter/README.md
```diff
@@ -0,0 +1,87 @@
+# QML Notebook to Demo Converter
+This converter aids the Jupyter Notebook to QML Demo conversion. Though this converter will not completely convert
```
josh146, line 2:
```suggestion
# Notebook to PennyLane Demo Converter
This converter aids the Jupyter Notebook to Pennylane Demo conversion. Though this converter will not completely convert
```

notebook_converter/README.md
```diff
@@ -0,0 +1,87 @@
...
+   --author "Jane Doe" "Some info about author" "some/path/to/profile/picture/jane_doe.png" \
+   --is-executable=False
+```
+
+#### `--author-file` (optional)
+If the notebook being converted is being used by an existing author, this option can be used to pass the location of the existing author-file.
+
+```bash
+--author-file "qml/_static/authors/john_doe.txt"
+```
```
josh146, line 69: Minor comment @rashidnhm: I ran the script from the root directory of the repo, and got the error

```
  File "/Users/josh/pennylane/qml/notebook_converter/notebook_to_demo.py", line 365, in <module>
    author_sphinx = set_authors(*authors)
  File "/Users/josh/pennylane/qml/notebook_converter/notebook_to_demo.py", line 120, in set_authors
    author_files = [
  File "/Users/josh/pennylane/qml/notebook_converter/notebook_to_demo.py", line 121, in <listcomp>
    AUTHORS['link-dir'] / set_author_info(author).name
  File "/Users/josh/pennylane/qml/notebook_converter/notebook_to_demo.py", line 89, in set_author_info
    shutil.copy(profile_picture_loc, new_profile_picture_save_loc)
  File "/Users/josh/miniconda3/lib/python3.10/shutil.py", line 417, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/Users/josh/miniconda3/lib/python3.10/shutil.py", line 254, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: '../_static/authors/josh_izaac.png'
```

after I passed

```
--author-file "_static/authors/josh_izaac.txt"
```

Maybe an issue with relative directories?

josh146 (approved):
Apart from a small issue with the author file (I had to manually edit the author file to allow the script to work), everything worked really well @rashidnhm!

notebook_converter/README.md
rashidnhm: @josh146 taking a look into this!

rashidnhm: Fixed!

rashidnhm:
Thank you @josh146 , @KetpuntoG for all the help in testing and reviewing the converter!

PR #737, opened 2023-04-14
Run CI to find what tutorials are broken

albi3ro:
The plugin test matrix cancels all other runners when it finds the first failure. This makes it impossible for us to have a concrete list of what tutorials need to be updated.

This PR toggles the "cancel-in-progress" on the CI workflow, so we can find all failures on the dev branch.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -11,7 +11,7 @@ env:
 
 concurrency:
   group: build-docs-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-pr.yml
```diff
@@ -11,7 +11,7 @@ env:
 
 concurrency:
   group: build-docs-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
```
rashidnhm, line 14:
```suggestion
  cancel-in-progress: true
```

rashidnhm: This can cause issues in the pipeline so better to leave it as true

albi3ro:
#741 is now passing. This PR can be closed.

PR #738, opened 2023-04-17
Update function_fitting_qsp.py

KetpuntoG:
eliminating the global macros, in this case \ket and \bra

It solves part of the issue #739


Jaybsoni (approved):
Thanks !! 💯

alvaro-at-xanadu (approved):
OK!

PR #742, opened 2023-04-17
FIx photonics and noisy demos

rmoyard:
**Changes:**

1) Photonics demo: remove `.numpy()` and separate the defautl Gaussian qnode in two because (multiple measurement not supported on default Gaussian).

2) Noisy circuits: Add array on target to get the right array box

### DIFF
--- a/demonstrations/tutorial_noisy_circuits.py
+++ b/demonstrations/tutorial_noisy_circuits.py
@@ -231,7 +231,7 @@ for p in ps:
 # ensure that the trainable parameters give rise to a valid channel parameter, i.e., a number
 # between 0 and 1.
 #
-ev = np.tensor(0.7781, requires_grad=False)  # observed expectation value
+ev = 0.7781  # observed expectation value
 
 def sigmoid(x):
     return 1/(1+np.exp(-x))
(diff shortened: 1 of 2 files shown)


demonstrations/tutorial_noisy_circuits.py
```diff
@@ -250,7 +250,7 @@ def damping_circuit(x):
 
 
 def cost(x, target):
-    return (damping_circuit(x) - target)**2
+    return (damping_circuit(x) - qml.numpy.array(target, requires_grad=True))**2
```
timmysilv, line 253: the second arg passed to `cost` below (`ev`) is explicitly marked as `requires_grad=False`. Should we reflect that change here?
```suggestion
    return (damping_circuit(x) - qml.numpy.array(target, requires_grad=False))**2
```
not sure why it doesn't come in with the right shape. The originally passed value is an `np.tensor`, any reason to change that?

rmoyard: I changed everything to float and it works, autograd mysteries

timmysilv (approved):

rmoyard:
[sc-37260]

PR #743, opened 2023-04-17
wrap relevant results in np.array

timmysilv:
as title says. the demo used numpy array indexing (like `res[:, :3]`) but now QNode results are vanilla lists, so they don't work. The values I wrap are either just for display, or they will be converted to numpy soon after anyway, so this is fine.


timmysilv:
[sc-37267]

rmoyard (approved):
Thanks @timmysilv :+1:

Issue #739, opened 2023-04-17
Update custom macros


KetpuntoG:
When updating the website it has been found that the macros do not work. 
This is the list of macros to change -> [link](https://github.com/XanaduAI/xanadu-sphinx-theme/blob/237ab9e456717a40e0725eabe02a278e552a689e/xanadu_sphinx_theme/layout.html#L37)
The following demos need to be updated:

- [x] function_fitting_qsp.py
- [x] tutorial_diffable_shadows.py
- [x] tutorial_gbs.py
- [x] tutorial_givens_rotations.py
- [x] tutorial_here_comes_the_sun.py
- [x] tutorial_plugins_hybrid.py
- [x] tutorial_pulse_programming101.py

They will be corrected in different PRs that will address this issue

PR #745, opened 2023-04-18
Replace macros in demo "Here comes the SU(N)"

dwierichs:
Following #739 , this removes the macro `\bm` from `tutorial_here_comes_the_sun` and replaces it with the explicit `\boldsymbol`. This is the only macro in the list I could find in that demo.


alvaro-at-xanadu (approved):

isaacdevlugt:
Hey sorry! Was going to get to this by EOD I swear 😅. Anywho, it looks fine to me! 😁

PR #752, opened 2023-04-20
macros tutorial_plugins

KetpuntoG:
#739


demonstrations/tutorial_plugins_hybrid.py
```diff
@@ -66,7 +66,7 @@
 
 4. **Finally, we measure the mean photon number** :math:`\left\langle \hat{n}\right\rangle` of the second wire, where
 
-   .. math:: \hat{n} = \ad\a
+   .. math:: \hat{n} = \hat{a}^{\dagger}\hat{a}
```
isaacdevlugt, line 69: 🎩 🗡️

isaacdevlugt (approved):
LGTM!

demonstrations/tutorial_plugins_hybrid.py
CatalinaAlbornoz: It looks fine in Safari but it looks weird in Chrome for me. I seem to be the only one having weird rendering issues in Chrome so if nobody else sees this then you can approve this. However if this isn't just me then we should fix it.

image

KetpuntoG: Oh you are right, in chrome is weird. Good catch, I'll ask in the channel, thanks!

CatalinaAlbornoz (approved):

PR #753, opened 2023-04-20
Update local_cost_functions and geometric_qml demos for new return types

lillian542:
Updated `tutorial_local_cost_functions.py` and `tutorial_geometric_qml.py` to work with the new return types


lillian542:
Closing and trying again but without the mess

PR #755, opened 2023-04-24
remove nlopt from list of requirements

albi3ro:
NLOpt is causing the CI to fail due to the absence of wheels (https://github.com/PennyLaneAI/qml/actions/runs/4790041189/jobs/8518643923), but only the manually run `qonn.py` demo uses it.

Since no demos are actually using `nlopt`, we remove it from the list of dependencies.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,6 @@ autograd==1.5
 numpy~=1.21
 numba==0.53.1
 networkx==2.5.1
-nlopt==2.6.2
 semantic_version==2.10
 strawberryfields==0.22
 sympy==1.6.2


timmysilv:
can we merge the same thing to master as well? or is it best to merge this to master, then use the bot to merge master to dev. Asking bc master will need this for that CI check to pass

albi3ro:
> can we merge the same thing to master as well? or is it best to merge this to master, then use the bot to merge master to dev. Asking bc master will need this for that CI check to pass

Next week we will merge `dev` into master.

lillian542 (approved):

PR #757, opened 2023-04-24
use lightning in quantumvolume; fix qnspsa for new return; use H.sparse_matrix

timmysilv:
as the title states.

I checked the QNSPSA one with an older braket plugin (1.10.3) because it's what I had installed and didn't complain about me using the latest master of PennyLane. Perhaps it's too soon to bump this demo, because the braket plugin isn't even compatible with the latest pennylane (which is when the new return system first becomes default)


rmoyard (approved):
Thanks @timmysilv it looks good to me!

timmysilv:
updated the description with a word of warning about bumping qnspsa now

mudit2812 (approved):

PR #759, opened 2023-04-25
Use dev lightning on demo diff checker

albi3ro:
See failure on https://github.com/PennyLaneAI/qml/actions/runs/4791273197/jobs/8521476423

The demo diff checker needs to build the dev branch with the updated dev version of lightning.


timmysilv (approved):
thanks for the quick fix!

lillian542 (approved):

PR #760, opened 2023-04-25
Demonstrations mathematics testing

BM7878:
Some code for automatically grabbing all of the mathematical expressions used in demonstrations, and using them to create a test demonstration containing lots of mathematical expressions only.


demonstrations/mathematics_test_demonstration.py
```diff
@@ -0,0 +1,438 @@
+r"""
+
+"""
+# Mathematics Test Demonstration
+# ==============================
```
josh146, line 5:
```suggestion
r"""
# Mathematics Test Demonstration
# ==============================
"""
```

josh146: I think maybe why the build is failing?

josh146 (commented):
Thanks @BM7878! Just holding back on approval until I can see the build. One question - do we need the `all_mathematical_expressions.txt` file committed? Since it can be extracted from the statistics script.

PR #762, opened 2023-04-25
Improve formatting of output in toric code demo

albi3ro:
The output of the toric code demo was no longer rounding for easy readability.

This change rounds outputs so they appear as actual zeros and ones instead of `5.551115123125783e-17` and `array(1.)`.

### DIFF
--- a/demonstrations/tutorial_toric_code.py
+++ b/demonstrations/tutorial_toric_code.py
@@ -436,8 +436,8 @@ single_x = [(1, 2)]
 
 x_expvals, z_expvals = separate_expvals(excitations(single_x, []))
 
-print("XGroup: ", [np.round(val) for val in xgroup_expvals])
-print("ZGroup: ", [np.round(val) for val in zgroup_expvals])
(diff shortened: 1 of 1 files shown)


demonstrations/tutorial_toric_code.py
```diff
@@ -436,8 +436,8 @@ def excitations(x_sites, z_sites):
 
 x_expvals, z_expvals = separate_expvals(excitations(single_x, []))
 
-print("XGroup: ", x_expvals)
-print("ZGroup: ", z_expvals)
+print("XGroup: ", [np.round(val) for val in xgroup_expvals])
+print("ZGroup: ", [np.round(val) for val in zgroup_expvals])
```
mudit2812, line 440:
```suggestion
print("XGroup: ", [np.round(val) for val in x_expvals])
print("ZGroup: ", [np.round(val) for val in z_expvals])
```

mudit2812 (approved):
One change, overall looks good :)

timmysilv (approved):

PR #763, opened 2023-04-26
Fix `HCN` qubit count in the dataset page

obliviateandsurrender:
**Title:** Fix `HCN` qubit count in the dataset page

**Summary:** Number of qubits required to build `HCN` in the minimal basis set is `22` instead of `20`.

**Relevant references:** 
```
>>> qml.qchem.Molecule(["H", "C", "N"], np.array([[0.0, 0.0, 0.0], [0.0, 0.0, 1.0], [0.0, 0.0, 2.0]]))
<Molecule = CHN, Charge: 0, Basis: STO-3G, Orbitals: 11, Electrons: 14>
```

**Possible Drawbacks:** N/A

**Related GitHub Issues:** N/A


soranjh (approved):

datasets_qchem.rst
```diff
@@ -143,12 +143,12 @@ please see the `PennyLane Documentation <https://docs.pennylane.ai/en/stable/int
...
     | H\ :math:`_{10}`             | STO\ :math:`\text{-}`\3G     | 20           | | H\ :math:`_A-`\ H\ :math:`_B` :math:`= 1.0` Å,                          |          N/A                                                               |
     |                              |                              |              | | :math:`\measuredangle` HHH :math:`= 180^{\circ}`                        |                                                                            |
     +------------------------------+------------------------------+--------------+---------------------------------------------------------------------------+----------------------------------------------------------------------------+
+    | HCN                          | STO\ :math:`\text{-}`\3G     | 22           | | :math:`\measuredangle` HCN :math:`\in\ [0, \pi]^{\circ}`,               | | :math:`\measuredangle` HCN :math:`= 0.0^{\circ}`,                        |
+    |                              |                              |              | | C\ :math:`-`\ N :math:`= 1.156` Å                                       | | C\ :math:`-`\ N :math:`=1.156` Å                                         |
```
soranjh, line 150: The optimal angle is `0°`? It shouldn't be `180°`?
Also, the PES is available for the whole range `[0,𝜋]`?

DSGuala (approved):
Looks good. It's a small change, so I'm approving.

I noticed that HCN is not included at the very top under "organic molecules"
Additionally, "inorganic molecules" contains two organic molecules: CO and CO2. Could be good to use this PR as an opportunity to change those too.

datasets_qchem.rst
obliviateandsurrender: The optimal angle is `180°`, and the PES data is not available.

PR #765, opened 2023-04-27
Adding the category definitions

BM7878:
The categories for the demonstrations are now defined in demonstrations_categories.json.

This includes the recently-created categories.

A description of the contents of this file can be found in demonstrations_categories.md.

The exact titles of the categories have been updated in the demonstrations metadata files.


demonstrations_categories.json
rashidnhm, line 1: Question regarding `orderIndex` ... would we want to have this attribute or can we imply it from the index of the category in the file itself.

As in, if "Getting Started" is the first item in the list, it is index 0. So each category gets the orderIndex of where it is in this list.

BM7878: Good point. I'll remove that property.

demonstrations_categories.json
```diff
@@ -0,0 +1,37 @@
+[
+    {
+        "title": "Getting Started",
+        "urlFragment": "getting-started",
```
ashishks0522, line 4: Do we want to use these URL fragments in demo category page? instead of the existing ones (`/demos_getting-started`, `/demos_optimization`, ...)

That would be a nice change for URL structure 😄, if we are open to URL change would it be okay if we fix the category page URL to include home page fragment (`/demonstrations`) as well like:

Getting started demos at - `/qml/demonstrations/getting-started`

We can keep the individual demo pages URLs as they are since their google indexing might be more important for traffic. 

This is more of a demo redesign implementation question, tagging @josh @jen-dodd @Lorraine

demonstrations_categories.json
```diff
@@ -0,0 +1,37 @@
+[
```
ashishks0522, line 1: @rashidnhm we would be needing this file as well from the demo pipeline output

ashishks0522 (approved):
Looks good overall!

Anakhom (approved):
Looks good to me! Thank you Ben

demonstrations_categories.json
rashidnhm: Not an issue, @ashishks0522 

... Wanted to ask @BM7878 & @ashishks0522 

Can we have these files is a predetermined situation or with some predetermined prefix/suffix?

As it stands, when a new metadata file is added with a demo, that file will be exported and be available automatically. However, if I add a one off line to export this file, any time there is a new file, we will need workflow code updates, which is not ideal.


One solution I can think of is moving this file to `demonstrations/demonstrations_categories.metadata.json` This will make it so this file is also auto exported along with all the metadata files ... open to other suggestions too

demonstrations_categories.json
josh146: I have no strong opinion here if we wish to change the URL of the category pages, as long as:

- there is a redirect set up
- the redirect is set up in a way that makes [google happy](https://developers.google.com/search/docs/crawling-indexing/301-redirects) 😆

ikurecic: That looks good (better) to me in theory, but we would need to do a gradual transition of the URL and make sure to update all the places where we've used the 'original' URLs in links. This includes stuff like blog posts (the v0.30 release post draft I saw today links to a category page, for example).
The individual demos should be locked in where they are... that would probably be the simplest, because there's a non-negligible amount of people who use them as references in papers and theses.

demonstrations_categories.json
```diff
@@ -0,0 +1,37 @@
...
+    },
+    {
+        "title": "Quantum Chemistry",
+        "urlFragment": "quantum-chemistry",
+        "description": "Study the properties of molecules and materials using quantum computing."
+    },
+    {
+        "title": "Devices and Performance",
+        "urlFragment": "devices-and-performance",
+        "description": "Here you will discover and learn how to use many of the devices offered by PennyLane. From built in devices and its different interfaces to external devices."
```
ikurecic, line 25:
```suggestion
        "description": "Here you will discover and learn how to use many of the devices offered by PennyLane. From built-in devices and their different interfaces to external devices."
```
Small correction. But this tagline doesn't really pack a punch. The new descriptions also don't match the ones from the 'old' categories.

Maybe we can revisit the descriptions here once all the demos have been packed into their drawers?

demonstrations_categories.json
BM7878: Yep

- Individual demo pages staying where they are for now.
- Since we're adding new categories for the demos, these new categories will need to have their own `single category view` pages - I think it makes sense to update the old page URLs at the same time.
- I think each `single category view` page should be at `/qml/demonstrations/categories/<category-url-fragment>` - later down the line, individual demos will probably be at `/qml/demonstrations/<demo-url-fragment>`, and while it's unlikely for a category title to be the same as the title of a demo itself, it seems prudent to have the `/categories/` fragment to avoid any accidental clashes.
- Then just having redirects in place for all the old category page URLs.
- The 'Quantum Computing' category is being removed, so the redirect for `https://pennylane.ai/qml/demos_quantum-computing` should be to the main demos page.

demonstrations_categories.json
BM7878: Yep - we can update these as often as we like - the changes should propagate through to the live React site.

demonstrations_categories.json
BM7878: Yep - I'm fine with the `.metadata` fragment. Will add.

demonstrations_categories.json
ashishks0522: That also works, one thing to note - to populate this type of back links in the new design, it helps to have each fragment in the URL correspond to a valid page. Otherwise we would need to hard-code for fragments which do not have a valid page (e.g. `/categories`)


Screenshot 2023-04-28 at 1 06 03 PM

Also @BM7878 can you help in finalizing these URLs, then I can add that spec in the technical stories

BM7878: @ashishks0522 ahh I see. If it's easier to use `/qml/demonstrations/<category-url-fragment>` then, we could do that - the odds of a category title being the same as a demo title are quite low - it might just be something we have to remember about in future.

PR #766, opened 2023-04-27
Update manually run demos

mudit2812:
* Updated `ensemble_multi_qpu.py` to run parallel code without `QNodeCollection` with `dask`.
* Updated an image generated by `adjoint_diff_benchmarking.py`.
* Fixed `tutorial_quantum_transfer_learning.py` to use `torch.hstack` to preserve trainability. Predictions should now be as accurate as before.


demonstrations/ensemble_multi_qpu.py
```diff
@@ -245,19 +246,23 @@ def circuit1(params, x=None):
...
 
-
 def decision(softmax):
     return int(torch.argmax(softmax))
 
 
 def predict_point(params, x_point=None, parallel=True):
-    results = qnodes(params, x=x_point, parallel=parallel)
+    if parallel:
+        results = tuple(dask.delayed(q)(params, x=x_point) for q in qnodes)
```
timmysilv, line 261: i didn't see any notes when removing QNodeCollection, but I feel like the intention wasn't for users to have to call dask directly? would need to ask someone who knows better

josh146: Yep the intention was to show users, through demos, how to call Dask and use it with PennyLane :)

josh146: This is something we realized we should have done from the beginning -- it is quite simple, highlights how well PL works with libraries like Dask (as long as there are no race conditions 😬), and sets up users to use Dask where they need to.

timmysilv: ah, ignore me then! looks excellent

timmysilv (approved):

demonstrations/ensemble_multi_qpu.py
mudit2812: @trbromley is the author. He might have some insight.

albi3ro (approved):

PR #769, opened 2023-05-03
Demonstrations metadata updates

BM7878:
Adding metadata for five files that previously lacked it.

One file is a supplementary page to another demonstration. Whether this can be considered a 'true' demonstration in its own right is debatable, but unless it has a metadata file, it won't get served on the site, so I've added one for now.

Two files are for demonstrations published very recently.


ashishks0522 (approved):
Thanks @BM7878!

rashidnhm (commented):
Do we want to move the metadata categories to the demonstrations folder as part of this PR?

ikurecic (approved):
lgtm

BM7878:
> 

Yep, done.

PR #775, opened 2023-05-04
Tutorial compilation

BorjaRequena:
**Title:**
Compilation of quantum circuits

**Summary:** 
In this tutorial we introduce the most fundamental principles of circuit compilation. We briefly introduce the idea of circuit transformations, and we illustrate the functionalities of the ones involved in the `qml.compile` default pipeline, namely: `qml.transforms.commute_controlled`, `qml.transforms.merge_rotations`, and `qml.transforms.cancel_inverses`. Then, we explain the role they play in circuit compilation and we show how `qml.compile` works. We introduce its different parameters and how they relate to different key aspects of circuit compilation, such as the number of compiler passes, the transforms to apply at every pass, and the basis gates to represent the circuit. 

**Relevant references:**
https://docs.pennylane.ai/en/stable/code/qml_transforms.html 

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

---------------------------------

I have placed some provisional images: the first image in the tutorial showing simple circuit transformations should be reformatted, and I put an arbitrary circuit plot as thumbnail for the tutorial that should be entirely redesigned.

Let me know what you think! :)

PR #779, opened 2023-05-05
Replace deprectaed `qml.ultis.sparse_hamiltonian`

BorjaRequena:
Uploading my tutorial in PR #775 I realized the test were not passing due to a deprecation error in the Classically boosted VQE tutorial. There is a call to `qml.utils.sparse_hamiltonian(H)` which is now deprecated. I replaced it by `H.sparse_matrix().todense()` (line 310) which provides the same result. I tested it and the code should run now.

I tried solving it with `H.matrix()` or `H.compute_matrix()`, but I didn't fully understand how these methods work.


timmysilv:
do you mind replacing its mention in [the adaptive circuit demo](https://github.com/BorjaRequena/qml/blob/47a3d52ff977caec68dcd65bd3b36e5e094e88a5/demonstrations/tutorial_adaptive_circuits.py#LL306C30-L306C30) to say "the Hamiltonian function :func:\`~.pennylane.Hamiltonian.sparse_matrix\`" too? otherwise lgtm!

KetpuntoG:
Thank you for the work here, I close the PR

PR #780, opened 2023-05-05
pin pl and plugins to stable versions; remove sparse_hamiltonian; cast hamiltonian coeffs to real

timmysilv:
1. these should be pinned to PyPI releases on master. They were accidentally set this way when merging dev to master for the 0.30 release
2. sparse_hamiltonian is gone, so it should be removed from qml as well
3. fix tutorial_spsa by casting hamiltonian coeffs

### DIFF
--- a/demonstrations/tutorial_adaptive_circuits.py
+++ b/demonstrations/tutorial_adaptive_circuits.py
@@ -303,7 +303,7 @@ for n in range(20):
 #
 # Molecular Hamiltonians and quantum states are sparse. For instance, let’s look at the Hamiltonian
 # we built for LiH. We can compute its matrix representation in the computational basis using the
-# Hamiltonian function :func:`~.pennylane.Hamiltonian.sparse_matrix`. This function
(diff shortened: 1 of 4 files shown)


albi3ro (approved):

lillian542 (approved):

demonstrations/tutorial_adaptive_circuits.py
```diff
@@ -303,7 +303,7 @@ def circuit_2(params, excitations, gates_select, params_select):
 #
 # Molecular Hamiltonians and quantum states are sparse. For instance, let’s look at the Hamiltonian
 # we built for LiH. We can compute its matrix representation in the computational basis using the
-# PennyLane function :func:`~.pennylane.utils.sparse_hamiltonian`. This function
+# Hamiltonian function :func:`~.pennylane.Hamiltonian.sparse_matrix`. This function
```
timmysilv, line 306: this didn't work, fixing rn

timmysilv: should be `:meth:`...
image

PR #798, opened 2023-05-16
Update ahs demo

lillian542:
makes some edits requested to the AHS hardware demo


demonstrations/ahs_aquila.py
```diff
@@ -49,17 +49,17 @@
 hardware when unperturbed, and a time-dependent drive.
 
 Pulse control gives some insight into the low-level implementation of more abstract quantum
-computations. In most digital quantum architectures, the native gates of the computer are, at the
-implementation level, electromagnetic control pulses that have been finely tuned to perform a
-particular logical gate.
+computations. In most digital (as opposed to continuous-variable) quantum architectures, the
+native gates of the computer are, at the implementation level, electromagnetic control pulses
+that have been finely tuned to perform a particular logical gate.
```
Qottmann, line 54: I thought the point was to distinguish between measurement-based (xanadu) and non-measurement based (superconducting, neutral atoms, ion traps) approaches. I wouldnt call measurement based approaches in general continuous variable, though in Xanadu's case there is some CV stuff going on internally.

Qottmann: The abstract above already makes the relevant point by talking about systems where the qubits are realized through discrete energy levels. So in fact this paragraph is repeating most of what is said in the previous one, so maybe just scratching it? I find the continuous variable part more confusing than helpful.

demonstrations/ahs_aquila.py
```diff
@@ -119,10 +122,14 @@
...
 .. figure:: ../demonstrations/ahs_aquila/rydberg_blockade_diagram.png
     :align: center
-    :width: 50%
+    :figwidth: 55%
+    :width: 95%
     :alt: A diagram of the energy levels for the ground, single excitation, and double excitation states
     :target: javascript:void(0);
 
+    Energy levels for the ground (:math:`\ket{gg}`), single Rydberg excitation (:math:`\ket{gr}`, :math:`\ket{rg}`),
+    and double Rydberg excitation (:math:`\ket{rr}`) states
```
Qottmann, line 131: 👍

Qottmann (commented):
Thanks for the update @lillian542 🚀 

Not sure about the CV comparison, wouldnt take that over from their suggestions. Rest looks good!

One small thing: can we actually move this demo next to the other pulse demo on the overview panel? By moving it before the qutrit demo in the quantum computing section. CC @KetpuntoG

demonstrations/ahs_aquila.py
lillian542: The comment she left was "“In most digital quantum architectures” <— let’s define digital in 
this context. Does this mean as opposed to CV?". 

To be honest I don't know what we mean here by 'digital quantum architectures' or how they are different from measurement-based approaches. I pretty much copy-pasted this part from your demo because I recall you all discussing it and landing on the (original) formulation here. So I'm not sure how to incorporate the feedback.

lillian542: Should I just remove this change?

demonstrations/ahs_aquila.py
```diff
@@ -49,17 +49,17 @@
 hardware when unperturbed, and a time-dependent drive.
 
 Pulse control gives some insight into the low-level implementation of more abstract quantum
-computations. In most digital quantum architectures, the native gates of the computer are, at the
-implementation level, electromagnetic control pulses that have been finely tuned to perform a
-particular logical gate.
+computations. In most digital (as opposed to continuous-variable) quantum architectures, the
```
lillian542, line 52:
```suggestion
computations. In most digital quantum architectures, the
```

demonstrations/ahs_aquila.py
lillian542: Removed

trbromley (approved):
Thanks @lillian542!

Qottmann (approved):

PR #800, opened 2023-05-17
Replacing the triple and double hyphens in demos to em and en dashes

BM7878:
We've had some issues with the conversion of triple and double hyphens to em and en dashes, so I've simply swapped them to their Unicode counterparts.


KetpuntoG (approved):
Nice! Thanks for these changes

Anakhom (approved):
All looking good to me! Thank you, Ben

PR #801, opened 2023-05-18
Updated CI workflow

rashidnhm:
### Before submitting

**Title:**

**Summary:**

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -8,9 +8,16 @@ on:
         type: string
       num_workers:
         description: The number of workers to use for building the QML demos
+        required: true
+        type: string
+      sphinx_examples_to_build:
+        description: |
+          Instead of building all Sphinx Examples, only build the ones passed.
+          To build multiple demos, pass the filenames separated by space.
+          Leave as blank to build all demos (default behavior)
         required: false
         type: string
-        default: '15'
+        default: ''
       enable_sphinx_cache:
         description: Use actions/cache for sphinx and sphinx-gallery
         required: false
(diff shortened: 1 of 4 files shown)


KetpuntoG:
oh that’s amazing! Thanks for work on this @rashidnhm

.github/workflows/build-branch.yml
```diff
@@ -8,9 +8,16 @@ on:
         type: string
       num_workers:
         description: The number of workers to use for building the QML demos
+        required: true
+        type: string
```
rashidnhm, line 12: Made this parameter required now. Since we very much are changing this value depending on varying contexts

.github/workflows/teardown-pr.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         id: check_deployment
         run: |
           num_obj=$(aws s3 ls s3://${{ secrets.AWS_PR_S3_BUCKET_ID }}/${{ secrets.AWS_PR_BUCKET_BUILD_DIR }}/${{ github.event.pull_request.number }}/index.html --summarize | grep "Total Objects: " | sed 's/[^0-9]*//g')
-          echo "::set-output name=exists::$num_obj"
+          echo "exists=$num_obj" >> $GITHUB_OUTPUT
```
rashidnhm, line 20: The old set-output method will start raising an error starting end of May

josh146 (approved):
Thanks Rashid!! A quick question - is the build using the cache for the non-built demos for the PR preview?

.github/workflows/teardown-pr.yml
smokingroosters: Should we go around workflows to check this? 🤔 
maybe good to send a message in some channels

smokingroosters (approved):

rashidnhm:
> Thanks Rashid!! A quick question - is the build using the cache for the non-built demos for the PR preview?

Hey Josh, not exactly. 

The cache isn't a singular one, due to the parallelisation, each worker checks which demos they are building and checks to see if a cache entry exists for those sets of demos (hash of the files are part of the cache key). So we cannot use those caches in this case.

The non-built demos simply won't exist on the PR preview. The menu will show them but trying to go to them will result in a 404.

PR #803, opened 2023-05-18
Format testing demonstration

BM7878:
A demonstration containing a variety of the different formatting blocks and options close together - making it a bit quicker to identify issues in future.

### DIFF
--- a/demonstrations/test_demonstration_1.py
+++ b/demonstrations/test_demonstration_1.py
@@ -1,0 +1,47 @@
+r"""
+
+Formatting Test Demonstration
+=============================
+
+.. meta::
+    :property="og:description": A demonstration for testing different formatting features.
+
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
+
+"""
+
+##############################################################################
+# .. important::
+#
+#     This is an 'important' block. It should have an orange background.
+#
+#     * List Item 1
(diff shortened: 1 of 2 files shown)


BM7878:
We can remove this demo from the live system later simply by changing the file name endings (perhaps to just .txt).

demonstrations/test_demonstration_1.py
```diff
@@ -0,0 +1,147 @@
...
+    :scale: 75%
+    :alt: PennyLane can leverage Braket for parallelized gradient calculations
+
+"""
+
+##############################################################################
+# .. important::
+#
+#     This is an 'important' block. It should have an orange background.
+#
```
josh146, line 111: Something a bit odd is happening here 🤔 

image


I have an odd feeling it is because of the transition from a string to a comment, I've encountered this before. I wonder if it is resolved if you put some code after the docstring?

josh146 (approved):
Looks great @BM7878! Nothing comes to mind in terms of other formatting that I can think of, except potentially trying different image formats.

josh146:
Oh, actually a question - should this be merged into `dev` or `master`? I recall @rashidnhm  saying `dev` but I can't remember off the top of my head.

rashidnhm:
> Oh, actually a question - should this be merged into `dev` or `master`? I recall @rashidnhm  saying `dev` but I can't remember off the top of my head.

We can merge it into dev if we don't want it to show up on the prod site. But in the future if we merge dev into master as part of a release it'll show up in prod then

ikurecic:
Same comment as Josh, otherwise looks fine.
An unrelated thought: is there a reason we didn't want to create author pages (as in, make author names clickable, lead to a list of all the demos they created)? I wasn't involved directly in the 'revamp' of the demo part of the website, but this functionality was something I assumed will be included.

josh146:
> An unrelated thought: is there a reason we didn't want to create author pages (as in, make author names clickable, lead to a list of all the demos they created)? I wasn't involved directly in the 'revamp' of the demo part of the website, but this functionality was something I assumed will be included.

Definitely something we want to do, but this I would consider as a 'new feature' so has been out of scope so far. Ideally, we would want to combine author pages for _all_ content. So that there is a single author page that lists blog posts, demos, etc.

PR #804, opened 2023-05-19
Remove deprecated ::set-output.

vincentmr:
**Context:**
`save-state` and `set-output` commands will be deprecated. 
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

**Description of the Change:**
Use new env files instead of `save-state` and `set-output` commands.

**Benefits:**
CI will keep running.

**Possible Drawbacks:**
None.

**Related GitHub Issues:**
None.


vincentmr:
> @vincentmr #801 is already open for this repo to address

NP, will close it.

PR #806, opened 2023-05-24
Added some utility code to ensure bidirectional links between demos, and ran it

BM7878:
A longstanding issue is that newer demos have tended to link to older demos (in the 'related content'), but older demos haven't been updated to link them to the newer demos.

These links are now specified in the metadata. A script has been run to ensure that these links go in both directions (updating the metadata for the older demos to include the links).


KetpuntoG:
This will not be needed :) We'll reduce the related demo to the necessary ones [here](https://github.com/PennyLaneAI/qml/pull/988).

PR #809, opened 2023-05-30
Fix to note in neutral atoms demo

alvaro-at-xanadu:
### Before submitting

**Title:**

**Summary:**

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**


KetpuntoG (approved):
Thanks for the updated!

Qottmann (commented):
The easiest and quickest change from the current online version would be to just add the $\mathcal{T}$ for the time-ordering in the exponential expression (and maybe adding that this is the time-ordered exponential, realized by the [Dyson series](https://en.wikipedia.org/wiki/Dyson_series)).

I.e. have the formula be:
$\vert \psi(t)\rangle = \mathcal{T}\exp\left(-i\int_{0}^{t}H(\tau)d\tau\right)\vert \psi(0)\rangle$
(no bracket from $\mathcal{T}(..)$ enclosing the exponential!)

If you want to keep the simple case of just a matrix exponential as per this current PR, I left two suggestions to make it mathematically more precise. I think it can be more confusing though.

PR #811, opened 2023-05-31
extra metadata categories

KetpuntoG:
adding category to some demos that did not have it yet

"Multiclass_classification" is not a demo that we have active right now.


ashishks0522 (approved):
Thanks @KetpuntoG 🙌

PR #819, opened 2023-06-08
Removed category from supplementary demonstration

BM7878:
We have one demonstration that is a 'supplementary' article. This one we don't want to appear in the category view, as otherwise we have two demonstrations with the same name, so I've simply removed the category from the metadata file.


ashishks0522 (approved):

KetpuntoG (approved):
Nice, thanks!

PR #821, opened 2023-06-09
No longer build HTML docs for push to master/dev

rashidnhm:
**Title:** No longer build HTML docs for push to master/dev

**Summary:** As part of the new pennylane.ai website release. The Demos will be served by react, which are generated from this repo as JSON. Thus, only need HTML docs being built for PR-previews and not on push to master/dev anymore.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -1,10 +1,6 @@
 name: Build Website
 on:
   pull_request:
-  push:
-    branches:
-      - master
-      - dev
 
 concurrency:
   group: build-docs-${{ github.ref }}
--- a/.github/workflows/update-dev.yml
+++ b/.github/workflows/update-dev.yml
@@ -2,7 +2,7 @@ name: update-dev
 on:
   workflow_run:
     workflows:
-      - Build Website
+      - Build Website - JSON
     types:
       - completed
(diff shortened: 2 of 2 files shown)


smokingroosters (approved):
LGTM, just one question: I guess we want to leave PR-previews so contributors will be able to see their changes before merging, but the real visualization will be on the pl-react repo, right? Could there be a mismatch between what gets shown here in a PR-preview and what will endup on the pl website?

josh146:
@smokingroosters yep there will be a mismatch :( we need to:
- communicate this to authors, and
- start to strip out the theme elements from the PR preview, so we are left with just the content.

josh146 (approved):

rashidnhm:
> LGTM, just one question: I guess we want to leave PR-previews so contributors will be able to see their changes before merging, but the real visualization will be on the pl-react repo, right? Could there be a mismatch between what gets shown here in a PR-preview and what will endup on the pl website?

Exactly as Josh mentioned. I do think if it occurs too often that the PR preview looks fine but the demo rendering breaks on the react side, then we may need to re-design this solution. But making it a simpler stripped version should be okay

PR #822, opened 2023-06-12
Add Github Action for running image and text diff check

mudit2812:
**Summary:**
* Added a new Github action that uses the `build-branch-master` and `build-branch-dev` artifacts to do an image and text diff check on the master and dev branches.
* Added image comparison script from the `demo_image_checker` repository.
* Added new branch `demo_output_comparison_new` to save diff results.

See [here](https://github.com/PennyLaneAI/qml/actions/workflows/all_diffs.yml) to check previous runs of this workflow.


mudit2812:
[sc-38737]

rashidnhm (commented):
Great stuff @mudit2812!

A few notes:
- Caching the zip file the way you are doing it won't work long term, a cache key cannot be updated, so once you have a build saved under `master_build-ubuntu22.04`, then this workflow ran again, it will fail as it will try to save the new cache under the same key. However, I do not think caching is needed at all in this scenario, each of those files are 200MB+ and could cause other sphinx cache to get dropped.

- A better solution would be to do the following:
  - Instead of downloading the zip file in get-dev-build or get-master-build. Have those jobs only output the workflow ID from where the artifact needs to be downloaded.
  - Then in check-diffs, download both artifacts. 
  This way there is no need of caching.

Also we should avoid running another long lived branch `demo_output_comparison_new` unless there is good reason to do so.

.github/workflows/all_diffs.yml
```diff
@@ -0,0 +1,166 @@
...
+          artifact_download_dir: ${{ github.workspace }}/dev-build
+
+      - name: Unpack Dev Artifact
+        run: |
+          cd dev-build
+          FILE='html.zip.zip'
+          if [ -f "$FILE" ]; then
+            echo "$FILE exists."
+            unzip html.zip
+          fi
```
rashidnhm, line 103: Let's also add a rm step at the bottom so we know the next step is downloading the correct new file, since both have the same name.
```suggestion
      - name: Unpack Dev Artifact
        run: |
          cd dev-build
          FILE='html.zip.zip'
          if [ -f "$FILE" ]; then
            echo "$FILE exists."
            unzip html.zip
          fi
          rm $FILE
```

.github/workflows/all_diffs.yml
```diff
@@ -0,0 +1,166 @@
...
+          git commit -m "Update the demonstration text differences found"
+          git push
+
+      - name: Update Image Diff File
+        run: |
+          mkdir -p "$HOME/.local/bin"
+          curl -O 'https://download.imagemagick.org/ImageMagick/download/binaries/magick'
+          chmod u+x magick
+          mv magick "$HOME/.local/bin"
+          echo "$HOME/.local/bin" >> "$GITHUB_PATH"
```
rashidnhm, line 145: Where is this binary being used?

mudit2812: `imagemagick` is used by the image diff script.

PR #827, opened 2023-06-16
Optimal control for gate compilation

dwierichs:
**Title:**
Optimal control for gate compilation

**Summary:**
This demo showcases the optimization of pulse programs to produce digital quantum gates.
To this end, a brief introduction to pulse programming and optimal control is given, followed by
the definition of a useful pulse shape and two exemplary optimizations to compile a `CNOT` and ` Toffoli`
gate, respectively.

**Relevant references:**
TBD


demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+Instead, it allows us to play sequences of analogue electromagnetic pulses,
+for example laser pulses that interact with trapped ions or Rydberg atoms, or microwave
+pulses acting on superconducting qubits.
+These pulses need to be calibrated to yield the desired digital gates, and in
+this tutorial we will be concerned with exactly that step.
+For this, we parametrize a pulse sequence, which yields a whole *space* 
+of sequences, and then optimize the pulse parameters to find a configuration
+in the space that acts as closely to the gate we are interested in
+as possible. This training of control parameters to achieve a specific time
+evolution falls under the field of *quantum optimal control*.
```
Qottmann, line 34: A lot of convoluted Schachtelsätze, I think this could do with some grammatical simplifications. 100% subjective though 😇

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+
+For an introduction see
+:doc:`the demo on differentiable pulse programming </demos/tutorial_pulse_programming101>`
+in PennyLane.
+Instead of optimizing pulses to yield digital gates that are used in quantum circuits,
+we may use them directly to solve minimization problems, as is also showcased in the demo.
+If you are interested in specific hardware pulses, take a look at
+:doc:`an introduction to neutral-atom quantum computing </demos/tutorial_neutral_atoms>`
+or :doc:`the tutorial on the QuEra Aquila device </demos/ahs_aquila>`, which treat pulse
+programming with Rydberg atoms.
```
Qottmann, line 47: 👌

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+or
+"What is the shortest time in which the system can reach a specific state, given some
+initial state?".
+
+In this tutorial, we consider the control of few-qubit systems through pulse sequences,
+with the goal to produce a given target, namely a digital gate, to the highest-possible
+precision.
+To do this, we will choose an ansatz for the pulse sequence that contains
+free parameters and define a cost function that measures the deviation of the qubit
+system from the target gate.
```
Qottmann, line 70: not sure about "to measure" and "qubit system"
```suggestion
free parameters and define a cost function that quantifies the deviation of the qubit
operation from the target gate.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+
+In this tutorial we will describe the distance of two unitary matrices :math:`U` and
+:math:`V` based on the (average) gate fidelity:
+
+.. math::
+
+    d(U,V) = 1 - \frac{1}{2^N}\big|\operatorname{tr}[U^\dagger V]\big|.
+
+It is similar to the distance measure obtained from the
+[Frobenius norm]
```
Qottmann, line 142: ?

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+How can we tell whether the evolution of the qubit system is close to the digital gate
+we aim to produce? We will need a distance measure!
+
+|
+
+.. figure:: ../demonstrations/optimal_control/distance.png
+    :align: center
+    :width: 80%
+    :alt: Illustration of landscape with a path drawn between markers for a pulse unitary and a CNOT gate
+    :target: javascript:void(0);
```
Qottmann, line 130: lol! love it!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+
+#. The Heaviside step function is not differentiable with respect
+   to the times :math:`t_0` and :math:`t_1` in the conventional sense (but
+   only if we were to consider distributions in addition to functions), and in
+   particular we cannot differentiate the resulting :math:`U(\boldsymbol{p},T)`
+   within the automatic differentiation framework provided by JAX.
+
+#. The instantaneous change in the amplitude will not be realizable in practice.
+   In reality, the pulses describe some electromagnetic control field that only
+   can be changed at a bounded rate and in a smooth manner. :math:`R_\infty` is not
```
Qottmann, line 204:
```suggestion
   In reality, the pulses describe some electromagnetic control field that can only
   be changed at a bounded rate and in a smooth manner. :math:`R_\infty` is not
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+
+def sigmoid(t, k=1.0):
+    """Sigmoid function with steepness parameter ``k``."""
+    return 1 / (1 + jnp.exp(-k * t))
+
+
+def normalize(t, k=1.0):
+    """Smoothly normalize a real input value to the interval (-1, 1) using ``sigmoid``
+    with steepness parameter ``k``."""
+    return 2 * sigmoid(t, k) - 1.0
```
Qottmann, line 263: move to the section where it is used later

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+    f"which has an infidelity of {arb_cost:.6f}."
+)
+
+#############################################################################
+# Before we can start the optimization, we require initial parameters.
+# We set small alternating amplitudes and evenly distributed start and end times
+# for :math:`P=3` smoothened rectangles. This choice leads to
+# a total of :math:`21P=63` parameters in the pulse sequence.
+
+num_pulses = 3  # Number of rectangles P
```
Qottmann, line 450: why not call it `P` like in the text?

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+# As we can see, adam runs into an oscillating behaviour during the optimization. The
+# precision with which it can minimize the cost function depends on the expressivity
+# of the pulses we use, but also on the precision with which we run the ODE solver.
+#
+# Let's pick those parameters with the smallest cost function we observed during
+# the training and take a look at the pulses we found. We again write a function
+# that plots the pulse sequence, to be reused later on.
+
+
+def plot_optimal_pulses(hist, pulse_fn, ops, T, target_name, separate_two_qubit=False):
```
Qottmann, line 504: purely subjective, almost philosophical: Personally, I'd just do two separate plot codes instead of one big function that takes into account the two cases. Makes it more readable and more copy-pastable/modifiable as no one is going to go into the depths of deciphering this 😅

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+# maximal amplitudes and bounded rates of change, and we tried to use only few
+# interaction terms between qubits. However, it is important to note
+# that the shown optimization remains a toy model for calibration of
+# quantum hardware: We did not take into account the interaction terms
+# or pulse shapes available on realistic devices and their control electronics.
+# We also did not consider a unit system tied to real devices, and we
+# ignored noise, which plays a very important role in today's quantum devices
+# and in quantum optimal control.
+# We leave the extension to real-world pulse Hamiltonians and noisy systems
+# to a future tutorial--or maybe your work?
```
Qottmann, line 673: genius 😆 !

Qottmann (commented):
This already looks very good @dwierichs ! Mostly left minor and very subjective comments that are free to be ignored 😇 

I like the conclusion that clarifies the limitations of these toy models. I think the demo could benefit from already making that disclaimer somewhere early / in the intro to set the expectations right.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+    ahs_aquila Pulse programming on Rydberg atom hardware
+
+*Author: David Wierichs. Posted: xx June, 2023.*
+
+
+Today, quantum computations largely are phrased as quantum circuits--or gate 
+sequences more generally--that are composed of digital quantum gates.
+However, most quantum hardware does not come with such digital
+gates as elementary native operations.
+Instead, it allows us to play sequences of analogue electromagnetic pulses,
```
dwierichs, line 25:
```suggestion
Instead, the hardware allows us to play sequences of analogue electromagnetic pulses,
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+for example laser pulses that interact with trapped ions or Rydberg atoms, or microwave
+pulses acting on superconducting qubits.
```
dwierichs, line 27:
```suggestion
for example by shining laser pulses on trapped ions or Rydberg atoms, or by sending microwave
pulses onto superconducting qubit cavities.
```

demonstrations/tutorial_optimal_control.py
dwierichs, line 34:
```suggestion
For this, we parametrize a pulse sequence, which leads to a whole *space* 
of possible sequences. Then we optimize the pulse parameters in order to
find a configuration in this space that behaves as closely to the gate 
of interest as possible.
This training of control parameters to achieve a specific time
evolution is a standard task in the field of *quantum optimal control*.
```

dwierichs: Changed a bit, I hope it is better now :D :)

demonstrations/tutorial_optimal_control.py
dwierichs: whoops, missing wiki link.

dwierichs, line 142:
```suggestion
`Frobenius norm <https://en.wikipedia.org/wiki/Matrix_norm#Frobenius_norm>`__
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,694 @@
...
+.. math::
+
+    d(U,V) = 1 - \frac{1}{2^N}\big|\operatorname{tr}[U^\dagger V]\big|.
+
+It is similar to the distance measure obtained from the
+[Frobenius norm]
+but allows us to ignore differences in the global phase.
+
+With a distance measure in our hands, we can write out the cost function to be minimized 
+while training the pulse parameters:
```
dwierichs, line 146:
```suggestion
With a distance measure in our hands, we can write out the cost function that we want to
minimize by training the pulse parameters:
```

demonstrations/tutorial_optimal_control.py
dwierichs: Very good call

demonstrations/tutorial_optimal_control.py
dwierichs: :stuck_out_tongue:

dwierichs (commented):
Thanks for the first comments, @qottmann! 
I'll incorporate all of them, they all seem very reasonable to me :)

demonstrations/tutorial_optimal_control.py
dwierichs: Actually, what do you think about always plotting the two-qubit terms in a separate axis? Might be a better visualization?

Qottmann: sure :)

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,707 @@
...
+for example by shining laser pulses on trapped ions or Rydberg atoms, or by sending microwave
+pulses onto superconducting qubit cavities.
+These pulses need to be calibrated to yield the desired digital gates, and in
+this tutorial we will be concerned with exactly that step.
+For this, we parametrize a pulse sequence, which leads to a whole *space*
+of possible sequences. Then we optimize the pulse parameters in order to
+find a configuration in this space that behaves as closely to the gate
+of interest as possible.
+This training of control parameters to achieve a specific time
+evolution is a standard task in the field of *quantum optimal control*.
```
Qottmann, line 32: Very nice 🙂

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,707 @@
...
+    R_k(t, (\Omega, t_0, t_1), k)=
+    \Omega [1+\exp(-k (t-t_0))+\exp(-k (t_1-t))+\exp(-k(t_1-t_0))]^{-1}.
+"""
+import jax
+from jax import numpy as jnp
+
+jax.config.update("jax_enable_x64", True)  # Use float64 precision
+jax.config.update("jax_platform_name", "cpu")  # Disables a warning regarding device choice
+
+def sigmoid_rectangle(t, t_0, t_1, k=1.0):
```
Qottmann, line 234: 👌

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,707 @@
...
+
+params_hist, cost_hist = list(zip(*hist))
+min_params = params_hist[jnp.argmin(jnp.array(cost_hist))]
+
+#############################################################################
+# This looks promising: adam minimized the cost function successfully and we thus compiled
+# a pulse sequence that implements a Toffoli gate! Let's look at the pulse
+# sequence itself:
+
+plot_optimal_pulses(hist, f, ops_param, T, target_name)
```
Qottmann, line 646: In my preview this plot is capped at the bottom, can you check if it is also for you?

Qottmann (commented):
Thanks for the updates @dwierichs looks good !

I was wondering if it could be nice to add explicit prints of the execution times in the optimization loops to showcase how fast the jax jit business is :-) Comparable demos with qutip took much longer, right?

demonstrations/tutorial_optimal_control.py
dwierichs: Seems to be fine for me: :grimacing: 
image

Qottmann: Okay was just my cache that is sometimes acting strange on demo builds, with emptied cache it looks fine as well 👍

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+.. related::
+
+    tutorial_pulse_programming101 Introduction to pulse programming in PennyLane
+    tutorial_neutral_atoms Introduction to neutral-atom quantum computers
+    ahs_aquila Pulse programming on Rydberg atom hardware
+
+*Author: David Wierichs. Posted: xx July, 2023.*
+
+
+Today, quantum computations largely are phrased in the language of quantum circuits
```
alvaro-at-xanadu, line 18:
```suggestion
Today, quantum computations are largely phrased in the language of quantum circuits
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+that are composed of digital quantum gates.
```
alvaro-at-xanadu, line 19:
```suggestion
composed of digital quantum gates.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+For this, we will parametrize a pulse sequence, which leads to a whole *space*
+of possible sequences. Then we optimize the pulse parameters in order to
+find a configuration in this space that behaves as closely to the target gate
+as possible.
+This training of control parameters to achieve a specific time
+evolution is a standard task in the field of *quantum optimal control*.
+
+More concretely, we will optimize simple pulse programs on two and three qubits to
+obtain a CNOT and a Toffoli gate.
+
```
alvaro-at-xanadu, line 37: One line paragraph - should be merged with paragraph above.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+For an introduction see
+:doc:`the demo on differentiable pulse programming </demos/tutorial_pulse_programming101>`
+in PennyLane.
+Instead of optimizing pulses to yield digital gates that are used in quantum circuits,
```
alvaro-at-xanadu, line 41:
```suggestion
Instead of optimizing pulses to yield digital quantum gates,
```

alvaro-at-xanadu: Too lengthy and you defined this before, so let's use the word!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+or :doc:`the tutorial on the QuEra Aquila device </demos/ahs_aquila>`, which treat pulse
+programming with Rydberg atoms.
+
+Quantum optimal control
+-----------------------
+
+The overarching goal of quantum optimal control is to find the best way to steer
+a microscopical physical system such that its dynamics matches a desired behaviour.
+The meaning of "best" and "desired behaviour" will depend on the specific
+task, and it is important to specify underlying assumptions and constraints on
```
alvaro-at-xanadu, line 54:
```suggestion
task, and it is important to specify the underlying assumptions and constraints on
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+Once we specified all these details, optimal control theory is concerned with
+questions like
+"How close can the system get to modelling the desired behaviour?",
+"How can we find the best (sequence of) control parameters to obtain the desired behaviour?",
+or
+"What is the shortest time in which the system can reach a specific state, given some
+initial state?" (controlling at the so-called quantum speed limit) [#CanevaMurphy09]_.
+
+In this tutorial, we consider the control of few-qubit systems through pulse sequences,
+with the goal to produce a given target, namely a digital gate, to the highest-possible
```
alvaro-at-xanadu, line 65:
```suggestion
with the goal to produce a given target, namely a digital gate, to the highest possible
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+precision.
+To do this, we will choose an ansatz for the pulse sequence that contains
+free parameters and define a cost function that quantifies the deviation of the qubit
+operation from the target gate.
+Then we minimize the cost function by optimizing the pulse parameters until we
+find the desired gate to a sufficient precision--or can no longer improve on the
+approximation we found.
+For the training phase we will make use of fully-differentiable classical simulations
```
alvaro-at-xanadu, line 73:
```suggestion
For the training phase, we will make use of fully-differentiable classical simulations
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+of the qubit dynamics, allowing us to make use of backpropagation, the efficient
+differentiation technique widely used in machine learning, and of gradient-based
```
alvaro-at-xanadu, line 75:
```suggestion
of the qubit dynamics, allowing us to make use of backpropagation -- an efficient
differentiation technique widely used in machine learning -- and of gradient-based
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+optimization.
+At the same time we attempt to find pulse shapes and control parameters that are
+(to some degree) realistic to be implemented, including bounded
```
alvaro-at-xanadu, line 78:
```suggestion
(to some degree) realistically feasible, including bounded
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+
+For this, we will parametrize a pulse sequence, which leads to a whole *space*
+of possible sequences. Then we optimize the pulse parameters in order to
+find a configuration in this space that behaves as closely to the target gate
+as possible.
+This training of control parameters to achieve a specific time
+evolution is a standard task in the field of *quantum optimal control*.
+
+More concretely, we will optimize simple pulse programs on two and three qubits to
+obtain a CNOT and a Toffoli gate.
```
dwierichs, line 36:
```suggestion
More concretely, we will optimize simple pulse programs on two and three qubits to
obtain a CNOT and a Toffoli gate.
This training of control parameters to achieve a specific time
evolution is a standard task in the field of *quantum optimal control*.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+(to some degree) realistic to be implemented, including bounded
+pulse amplitudes and rates of change of the amplitudes.
+
+Tutorials that use other techniques are available for example for the
+`open-source quantum toolbox QuTiP <https://qutip.org/qutip-tutorials/#optimal-control>`__.
+
+Gate calibration via pulse programming
+--------------------------------------
+
+Here we briefly discuss the general setup of pulse programs that we will use for our
```
alvaro-at-xanadu, line 87:
```suggestion
Here, we briefly discuss the general setup of pulse programs that we will use for our
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,736 @@
...
+optimal control application. For more details also consider the related
+tutorials focusing on pulse programming.
+
+Consider a quantum system comprised of :math:`N` two-level systems, or qubits, described
+by a Hamiltonian
+
+.. math::
+
+    H(\boldsymbol{p}, t) = H_d + \sum_{i=1}^K f_i(\boldsymbol{p_i}, t) H_i.
```
alvaro-at-xanadu, line 96: $p_i$ has a subscript $i$ here, which makes the left and right hand side notations inconsistent. Should it just be $\boldsymbol{p}$ on the right hand side?

dwierichs: I meant to express that $\mathbf{p}$ is a vector of vectors, where we call the inner vectors $\mathbf{p_i}$. Does that make sense?

alvaro-at-xanadu: Sounds good, makes sense, but it would be good to state it explicitly!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+we aim to produce? We will need a distance measure!
+
+|
+
+.. figure:: ../demonstrations/optimal_control/OptimalControl_distance.png
+    :align: center
+    :width: 80%
+    :alt: Illustration of landscape with a path drawn between markers for a pulse unitary and a CNOT gate
+    :target: javascript:void(0);
+
```
alvaro-at-xanadu, line 142: The calligraphy in this drawing needs a bit of love (namely, the p looks weird 😅). Lmk if you would like help with this!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+In this tutorial we will describe the distance of two unitary matrices :math:`U` and
+:math:`V` based on the (average) gate fidelity:
+
+.. math::
+
+    d(U,V) = 1 - \frac{1}{2^N}\big|\operatorname{tr}[U^\dagger V]\big|.
+
+It is similar to the distance measure obtained from the
+`Frobenius norm <https://en.wikipedia.org/wiki/Matrix_norm#Frobenius_norm>`__
+but allows us to ignore differences in the global phase.
```
alvaro-at-xanadu, line 154:
```suggestion
It is similar to the distance measure obtained from the
`Frobenius norm <https://en.wikipedia.org/wiki/Matrix_norm#Frobenius_norm>`__
but it allows us to ignore differences in the global phase.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+With a distance measure in our hands, we can write out the cost function that we want to
+minimize by training the pulse parameters:
+
+.. math::
+
+    C(\boldsymbol{p}) = d(U_\text{target}, U(\boldsymbol{p}, T)).
+
+Here :math:`U_\text{target}` is the unitary matrix of the gate that we want to compile
+and we exclude the total duration :math:`T` from :math:`C` of the pulse sequence because we will
+consider it as a constraint to the optimization problem, rather than a free variable.
```
alvaro-at-xanadu, line 165:
```suggestion
Here :math:`U_\text{target}` is the unitary matrix of the gate that we want to compile.
We exclude the total duration :math:`T` from :math:`C` of the pulse sequence because we will
consider it as a constraint to the optimization problem, rather than a free variable.
```

alvaro-at-xanadu: So you mean to say we consider $T$ is fixed?

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+
+We can then minimize the cost function :math:`C`, for example using gradient-based
```
alvaro-at-xanadu, line 167:
```suggestion
We can then minimize the cost function :math:`C`, for example, using gradient-based
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+optimization algorithms like adam [#KingmaBa14]_.
```
alvaro-at-xanadu, line 168:
```suggestion
optimization algorithms like Adam [#KingmaBa14]_.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+But how do we obtain the gradient of a cost function that requires us to run an ODE solver
+to obtain its value? We are in luck! The implementation of pulse programming in PennyLane is
+fully differentiable via backpropagation thanks to its backend based on the machine
+learning library `JAX <https://jax.readthedocs.io/en/latest/>`__.
+This enables us to optimize the gate sequences using efficiently computed gradients
+(provided the target gate is not too large).
+
+Before we dive into the task of calibrating a gate, let's briefly talk about
+the pulse shape that we will use:
```
alvaro-at-xanadu, line 177:
```suggestion
the pulse shape that we will use.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+-----------------------
+
+|
+
+.. figure:: ../demonstrations/optimal_control/OptimalControl_Smoother_Rectangles.png
+    :align: center
+    :width: 60%
+    :alt: Sketch of converting a rectangular pulse shape into a smoothened rectangular pulse shape
+    :target: javascript:void(0);
+
```
alvaro-at-xanadu, line 189: Not sure this is the best part to include this figure. Maybe later is better once we've defined $R_{\infty}$ and $R_k$. It could be good to plot them using pyplot, even if we don't show the code. Also, it would be great to label the plots ($R_{\infty}$ and $R_k$).

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+    R_k(t, (\Omega, t_0, t_1)) &= \Omega S(t-t_0, k) S(t_1-t, k)\\
+    S(t, k) &= (1+\exp(-k t))^{-1}.
+
+We introduced an additional parameter, :math:`k`, that controls the steepness of the sigmoid
+functions and can be adapted to the constraints posed by hardware on the maximal rate of change.
+In contrast to :math:`R_\infty`, its sister :math:`R_k` is smooth in all three arguments
+:math:`\Omega`, :math:`t_0` and :math:`t_1`, and training these three parameters with
+automatic differentiation will not be a problem.
+
+Let's implement the smooth-rectangle function using JAX's ``numpy``. We
```
alvaro-at-xanadu, line 233:
```suggestion
Let's implement the smooth rectangle function using JAX's ``numpy``. We
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+#
+# In this first example we will tune a two-qubit pulse to produce a standard CNOT gate.
+#
+# We start by choosing a system Hamiltonian.
+# It contains the drift term :math:`H_d = Z_0 + Z_1`, i.e. a Pauli :math:`Z` operator
+# acting on each qubit, with a constant unit amplitude.
+# The parametrized part uses five generating terms: Pauli :math:`Z` acting on the
+# first qubit (:math:`Z_0`), all three Pauli operators acting on the second qubit
+# (:math:`X_1, Y_1, Z_1`) and a single interaction term :math:`Z_0X_1`, resembling an
+# abstract cross resonance driving term. Due to this choice, the :math:`Z_0` term
```
alvaro-at-xanadu, line 296: Can we write this Hamiltonian explicitly, for clarity?

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# commutes with all other terms, including the drift term, and can be considered a
+# correction of the drive term to obtain the correct action on the first qubit.
+#
```
alvaro-at-xanadu, line 299: Does this Hamiltonian correspond to anything in reality, or is it a toy model?

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# :math:`[\epsilon, T-\epsilon]`, which makes sure that the pulse amplitudes are
+# close to zero at :math:`t=0` and :math:`t=T`. Without this step, we might be
+# tuning the pulses to be turned on (off) instantaneously at the beginning (end) of the
+# sequence, negating our effort on the pulse shape itself not to vary too quickly.
+# Second, we normalize the final output value to the interval
+# :math:`(-\Omega_\text{max}, \Omega_\text{max})`, which
+# allows us to bound the maximal amplitudes of the pulses to a realizable range while
+# maintaining differentiability.
+#
+# For the normalization steps we define a ``sigmoid`` and a ``normalize`` function.
```
alvaro-at-xanadu, line 319:
```suggestion
# For the normalization steps, we define a ``sigmoid`` and a ``normalize`` function.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# The first is a straight-forward implementation of :math:`R_k` whereas the second
```
alvaro-at-xanadu, line 320:
```suggestion
# The first is a straightforward implementation of :math:`R_k` whereas the second
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+#
+# The five coefficient functions :math:`f_i` are sums of multiple smooth-rectangle
+# pulse shapes :math:`R_k`
+# (see the section above) each, using distinct parameters for each generating term.
+# Each coefficient function sums :math:`P` smooth rectangles :math:`R_k` with individual
+# amplitudes and start and end times. Overall, this leads to
+# :math:`n=5\cdot 3\cdot P=15P` parameters in :math:`H`.
+#
+# Before we define this Hamiltonian, we implement the sum over multiple
+# ``sigmoid_rectangle`` functions, including two normalization steps:
```
alvaro-at-xanadu, line 308:
```suggestion
# ``sigmoid_rectangle`` functions, including two normalization steps.
```

demonstrations/tutorial_optimal_control.py
dwierichs: Was this remark for the original ugly sketch of mine or is it for the updated illustration? :)

demonstrations/tutorial_optimal_control.py
dwierichs: Yes :laughing:

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+plt.show()
+
+#############################################################################
+# Note that the rectangles are rather round for these generic parameters.
+# The optimized parameters in the training workflows below will lead to more
+# sharply defined pulses that resemble rectangles more closely. The amplitude normalization
+# step in ``smooth_rectangles`` enables us to produce them in a differentiable manner,
+# as was our goal with introducing :math:`R_k`.
+# Also note that the normalization of the final output value is not a simple clipping
+# step, but again a smooth function. As a consequence, the values ``1.9`` and ``-2.``
```
alvaro-at-xanadu, line 399:
```suggestion
# step, but again a smooth function. As a consequence, the amplitudes ``1.9`` and ``-2.``
```

demonstrations/tutorial_optimal_control.py
dwierichs: Mhm. Yeah I thought it might be nice as a "teaser" for this rather dry step of making the pulse work :thinking:

demonstrations/tutorial_optimal_control.py
dwierichs: Rephrased.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+
+
+@jax.jit
+def cost(params):
+    """Compute the infidelity cost function for given parameters."""
+    # Compute the unitary time evolution of the pulse Hamiltonian
+    op_mat = pulse_matrix(params)
+    # Compute the infidelity between the target and the pulse evolution
+    return 1 - jnp.abs(jnp.trace(target @ op_mat)) / 2**num_wires
+
```
alvaro-at-xanadu, line 443: This is the first time I see the word infidelity. We called it "average gate fidelity before".

demonstrations/tutorial_optimal_control.py
alvaro-at-xanadu: Ah, the build I was seeing had the previous one. Sorry to bash on it!

demonstrations/tutorial_optimal_control.py
alvaro-at-xanadu: Ok! I think we can keep it, but maybe not right at the start of the section!

demonstrations/tutorial_optimal_control.py
dwierichs: It would look like this:
$H(\boldsymbol{p}, t) = Z_0+Z_1+F(\boldsymbol{p_1}, t) Z_0 + F(\boldsymbol{p_2}, t) X_1+F(\boldsymbol{p_3}, t) Y_1 + F(\boldsymbol{p_4}, t) Z_1+F(\boldsymbol{p_5}, t) Z_0X_1$
I'm not sure how informative this is? :thinking:

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# acting on each qubit, with a constant unit amplitude.
+# The parametrized part uses five generating terms: Pauli :math:`Z` acting on the
+# first qubit (:math:`Z_0`), all three Pauli operators acting on the second qubit
+# (:math:`X_1, Y_1, Z_1`) and a single interaction term :math:`Z_0X_1`, resembling an
+# abstract cross resonance driving term. Due to this choice, the :math:`Z_0` term
+# commutes with all other terms, including the drift term, and can be considered a
+# correction of the drive term to obtain the correct action on the first qubit.
+#
+# The five coefficient functions :math:`f_i` are sums of multiple smooth-rectangle
+# pulse shapes :math:`R_k`
```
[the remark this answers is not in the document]
dwierichs: I think it's good to answer this question, but probably this needs to be done consistently across the demo. I'll incorporate that. :)

demonstrations/tutorial_optimal_control.py
dwierichs: It's a toy model. Added a comment :)

demonstrations/tutorial_optimal_control.py
dwierichs: Haha no, no worries, it _was_ ugly! :D

demonstrations/tutorial_optimal_control.py
dwierichs: Moved it to inbetween the math and the code implementation :)

demonstrations/tutorial_optimal_control.py
dwierichs: Ah, this is `1-fidelity`, hence infidelity :) (It's used in the literature, e.g. [here](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.105.022437) but if it is too uncommon I can try to replace :) )

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+
+P = 3  # Number of rectangles P
+# Initial parameters for the start and end times of the rectangles
+times = [jnp.linspace(eps, T - eps, P * 2) for op in ops_param]
+# All initial parameters: small alternating amplitudes and times
+params = [jnp.hstack([[0.1 * (-1) ** i for i in range(P)], time]) for time in times]
+
+#############################################################################
+# Now we are all set up to train the parameters of the pulse sequence to produce
+# our target gate, the CNOT. We will use the adam optimizer [#KingmaBa14]_, implemented in the
```
alvaro-at-xanadu, line 473:
```suggestion
# our target gate, the CNOT. We will use the Adam optimizer [#KingmaBa14]_, implemented in the
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+    print(f"The optimization took {end_time-start_time:.1f} (CPU) seconds.")
+    return hist
+
+
+learning_rate = 0.2
+num_steps = 500
+hist = run_adam(cost, grad, params, learning_rate, num_steps, target_name)
+
+#############################################################################
+# As we can see, adam steadily reduces the cost function, bringing the pulse program
```
alvaro-at-xanadu, line 517:
```suggestion
# As we can see, Adam steadily reduces the cost function, bringing the pulse program
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# closer and closer to the target unitary. On its way, the optimizer produces a mild
+# oscillating behaviour. The precision to which the optimization can produce the
+# target unitary depends on the expressivity of the pulses we use,
+# but also on the precision with which we run the ODE solver and the hyperparameters
+# of the optimizer.
+#
+# Let's pick those parameters with the smallest cost function we observed during
+# the training and take a look at the pulses we found. We again prepare a function
+# that plots the pulse sequence which we can reuse later on.
```
alvaro-at-xanadu, line 526:
```suggestion
# that plots the pulse sequence, which we can reuse later on.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# For the single-qubit terms we encode their qubit in the color and the type of Pauli
```
alvaro-at-xanadu, line 527:
```suggestion
# For the single-qubit terms, we encode their qubit in the color and the type of Pauli
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+    # Compute the infidelity between the target and the pulse evolution
+    return 1 - jnp.abs(jnp.trace(target @ op_mat)) / 2**num_wires
+
+
+grad = jax.jit(jax.grad(cost))
+
+#############################################################################
+# We create initial parameters similar to above but allow for a larger number
+# of :math:`1200` optimization steps and use a reduced learning rate
+# in the optimization with adam. Our ``run_adam`` function from above comes
```
alvaro-at-xanadu, line 647:
```suggestion
# in the optimization with Adam. Our ``run_adam`` function from above comes
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+
+num_steps = 1200
+learning_rate = 2e-3
+hist = run_adam(cost, grad, params, learning_rate, num_steps, target_name)
+
+params_hist, cost_hist = list(zip(*hist))
+min_params = params_hist[jnp.argmin(jnp.array(cost_hist))]
+
+#############################################################################
+# This looks promising: adam minimized the cost function successfully and we thus compiled
```
alvaro-at-xanadu, line 662:
```suggestion
# This looks promising: Adam minimized the cost function successfully and we thus compiled
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# sequence itself:
+
+plot_optimal_pulses(hist, f, ops_param, T, target_name)
+
+#############################################################################
+# As we can see, the optimized smooth rectangles do not fill out the time at maximal
+# amplitudes. This means that we probably can find shorter pulse sequences with
+# larger amplitudes that produce a Toffoli with the
+# same fidelity. If you are interested, take a shot at it and try to
+# optimize the sequence regarding the number of generators and pulse duration!
```
alvaro-at-xanadu, line 673:
```suggestion
# optimize the sequence with respect to the number of generators and pulse duration!
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,745 @@
...
+# the chopped random-basis quantum optimization (CRAB) algorithm [#DoriaCalarco11]_.
+# It uses a different parametrization altogether, exploiting randomized basis functions
+# for the pulse envelopes.
+#
+# While setting up the application examples, we accomodated for
+# some requirements of realistic hardware, like smooth pulse shapes with bounded
+# maximal amplitudes and bounded rates of change, and we tried to use only few
+# interaction terms between qubits. However, it is important to note
+# that the shown optimization remains a toy model for calibration of
+# quantum hardware: We did not take into account the interaction terms
```
alvaro-at-xanadu, line 703:
```suggestion
# quantum hardware. We did not take into account the interaction terms
```

alvaro-at-xanadu (changes requested):
Amazing demo! Only minor grammar suggestions and comments, and we'll be good to go!

demonstrations/tutorial_optimal_control.py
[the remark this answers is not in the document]
dwierichs: Done :) It ended up being an explanation at this point in the demo mostly, with two comments in other places.

dwierichs:
Thanks @alvaro-at-xanadu for the thorough review! :) 
I incorporated all but a few changes, the others I answered in the comments here. Let me know how to proceed with those :)

demonstrations/tutorial_optimal_control.py
alvaro-at-xanadu: More informative than just explaining through text, which is super confusing! If you feel it's too long, write the drift and parametrized parts separately.

demonstrations/tutorial_optimal_control.py
alvaro-at-xanadu: Thank! It seems you did have that in the conclusion, but it's good to say this from the start!

alvaro-at-xanadu (approved):
All good on my side, just some minor things pending to address

demonstrations/tutorial_optimal_control.py
dwierichs: Oh, sorry, I misunderstood your initial comment on this!

After looking at this again, I think I'd prefer to switch to _fidelity_ altogether. It's anyways the nicer concept than infidelity, leads to a maximization instead of a minimization problem for a change, and it better matches the first illustration in the demo :)

dwierichs:
Thanks a lot @alvaro-at-xanadu 
Left an answer to the last open point (which should also be resolved now) and resolved the other closed ones.
I switched from infidelity/cost functions/minimization to fidelity/profit/maximization, matching the image of "climbing mount fidelity" :)
Also made the images a bit larger, as they were already made with the total size in mind, unlike the previous sketches.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+
+
+def sigmoid(t, k=1.0):
+    """Sigmoid function with steepness parameter ``k``."""
+    return 1 / (1 + jnp.exp(-k * t))
+
+
+def normalize(t, k=1.0):
+    """Smoothly normalize a real input value to the interval (-1, 1) using ``sigmoid``
+    with steepness parameter ``k``."""
```
KetpuntoG, line 349:
```suggestion
    """Smoothly normalize a real input value to the interval (-1, 1) using 'sigmoid'
    with steepness parameter k."""
```

KetpuntoG: This applies to everything after that. In pennylane it is defined the other way because it renders on the web but here I would try to do it the way it reads best in code

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+    return qml.evolve(H, atol=atol, rtol=rtol)(params, T).matrix()
+
+
+@jax.jit
+def cost(params):
+    """Compute the infidelity cost function for given parameters."""
+    # Compute the unitary time evolution of the pulse Hamiltonian
+    op_mat = pulse_matrix(params)
+    # Compute the infidelity between the target and the pulse evolution
+    return 1 - jnp.abs(jnp.trace(target @ op_mat)) / 2**num_wires
```
KetpuntoG, line 460: In this case you have done it this way because target = target ^dagg but it might not be the case. Maybe it is more accurate to make the adjoint or leave a note to be faithful to the definition of distance

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+
+plot_optimal_pulses(hist, f, ops_param, T, target_name)
+
+#############################################################################
+# As we can see, the optimized smooth rectangles do not fill out the time at maximal
+# amplitudes. This means that we probably can find shorter pulse sequences with
+# larger amplitudes that produce a Toffoli with the
+# same fidelity. If you are interested, take a shot at it and try to
+# optimize the sequence with respect to the number of generators and pulse duration!
+#
```
KetpuntoG, line 692: How cool! I've missed a simple example where you can see it works. Could you do something like this?

```python
def circuit():
   qml.PauliX(wires = 0)
   qml.PauliX(wires = 1)
   qml.evolve(H,.... params)....
   return qml.probs(wires = [0,1,2])

```

to see that the output is indeed |111> as a toffoli would do.
Maybe it's not that easy, in that case, ignore this comment

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+    :target: javascript:void(0);
+
+|
+
+In this tutorial we will describe the distance of two unitary matrices :math:`U` and
+:math:`V` based on the (average) gate fidelity:
+
+.. math::
+
+    d(U,V) = 1 - \frac{1}{2^N}\big|\operatorname{tr}[U^\dagger V]\big|.
```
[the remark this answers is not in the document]
dwierichs: > I thought it was a commutator, could you use tr() instead of tr[]?

Sure :)

> we usually put n as the number of qubits and N = 2^n, do you think it makes sense to change this?

Oh, I didn't know there was a convention! I'll update.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+    \Omega [1+\exp(-k (t-t_0))+\exp(-k (t_1-t))+\exp(-k(t_1-t_0))]^{-1}.
+"""
+import jax
+from jax import numpy as jnp
+
+jax.config.update("jax_enable_x64", True)  # Use float64 precision
+jax.config.update("jax_platform_name", "cpu")  # Disables a warning regarding device choice
+
+
+def sigmoid_rectangle(t, t_0, t_1, k=1.0):
```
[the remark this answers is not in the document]
dwierichs: That's a fair point. I think there is no particular reason, I can change that :)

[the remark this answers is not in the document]
dwierichs: :heavy_check_mark:

demonstrations/tutorial_optimal_control.py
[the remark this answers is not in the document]
dwierichs: :heavy_check_mark:

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+# allows us to bound the maximal amplitudes of the pulses to a realizable range while
+# maintaining differentiability.
+#
+# For the normalization steps, we define a ``sigmoid`` and a ``normalize`` function.
+# The first is a straightforward implementation of :math:`R_k` whereas the second
+# uses the ``sigmoid`` function to normalize real numbers to the interval :math:`(-1, 1)`.
+
+
+def sigmoid(t, k=1.0):
+    """Sigmoid function with steepness parameter ``k``."""
```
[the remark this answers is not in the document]
dwierichs: :heavy_check_mark:

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+#############################################################################
+# For the arbitrary parameters from above, of course we get a rather arbitrary unitary
+# time evolution, which does not match the CNOT at all:
+
+params = [params] * len(ops_param)
+arb_mat = jnp.round(pulse_matrix(params), 4)
+arb_cost = cost(params)
+print(
+    f"The arbitrarily chosen parameters yield the unitary\n{arb_mat}\n"
+    f"which has an infidelity of {arb_cost:.6f}."
```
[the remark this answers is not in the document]
dwierichs: Sorry, I had switched to fidelity instead of infidelity already ,but missed to push the change. Let me know what you think of the current phrasing :)

demonstrations/tutorial_optimal_control.py
dwierichs: There is a number of options here. I'll try something :)

_static/large_demo_thumbnails/thumbnail_large_optimal_control.png
Qottmann, line 1: absolutely adore this 🤩 ❤️

demonstrations/optimal_control/OptimalControl_control_quantum_systems.png
Qottmann, line 1: [Optional] can we add two horizontal qubit lines for the CNOT gate, at first glance this looked like an upside down ♀️ to me

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,763 @@
...
+            print(f"Step {step+1:4d}: {c:.6f}")
+    _, cost_hist = list(zip(*hist))
+    plt.plot(list(range(num_steps + 1)), cost_hist)
+    ax = plt.gca()
+    ax.set(
+        xlabel="Iteration", ylabel=f"Infidelity $d(U_{{{target_name}}}, U(p))$", yscale="log"
+    )
+    plt.show()
+    end_time = time.process_time()
+    print(f"The optimization took {end_time-start_time:.1f} (CPU) seconds.")
```
Qottmann, line 526: 🚀 nice!

Qottmann (approved):
Great work @dwierichs 😇

demonstrations/tutorial_optimal_control.py
dwierichs: I added the example you provided, together with an additional plot that shows that the pulse sequence looks like a Toffoli gate :)

demonstrations/optimal_control/OptimalControl_control_quantum_systems.png
dwierichs: Done :)

Qottmann: So sorry but I meant to comment on the demo image, not the thumbnail, i.e. https://github.com/PennyLaneAI/qml/pull/827/files#diff-566fa175ca0060d0897c4a6f38c95d4e8aa785834b5a0f61b67e25c6418738db

dwierichs: All good :) Updated that one as well :+1:

Qottmann: Looks great! 👌

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+# parameter :math:`k=20` for all rectangles in the sum:
+
+from functools import partial
+
+T = 2 * jnp.pi  # Total pulse sequence time
+k = 20.0  # Steepness parameter
+max_amp = 1.0  # Maximal amplitude \Omega_{max}
+eps = 0.1 * T  # Margin for the start/end times of the rectangles
+# Bind hyperparameters to the smooth_rectangles function
+f = partial(smooth_rectangles, k=k, max_amp=max_amp, eps=eps, T=T)
```
BorjaRequena, line 410: Maybe we could give this partial a more meaningful name. When I reached the next code cell, in the `[f for op in ops_param]` I knew this was supposed to be the sum of smooth rectangles, but still I scrolled back up to check what `f` was again. Maybe something like `sr` would already work (?) It is not a critical thing though.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+    axs[0].set(ylabel=r"Pulse function $f(p, t)$", title=title)
+    axs[1].set(xlabel="Time $t$", ylabel=r"Pulse function $f(p, t)$")
+    plt.show()
+
+
+plot_optimal_pulses(hist, f, ops_param, T, target_name)
+
+#############################################################################
+# We observe that a single rectangular pulse is sufficient for most of the
+# generating terms in the Hamiltonian, and we see that their shape is closer to
```
BorjaRequena, line 605: I am not sure I would say "most terms". If I understand the plot correctly, only the $Z_0X_1$ and the $Z_1$ terms have a single rectangle, and the rest have two at least. Maybe we could soften this sentence?
```suggestion
# We observe that a single rectangular pulse is sufficient for many of the
# generating terms in the Hamiltonian, and we see that their shape is closer to
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+# actual rectangles now, in particular for those with a saturated amplitude.
+#
+# The final fidelity tells us that we achieved our goal of finding a pulse
+# sequence that implements a unitary close to a CNOT gate.
+# It could be optimized further, for example by running the optimization for more
+# training iterations, by tuning the optimizer further to avoid oscillations,
+# or by increasing the precision with which we run the ODE solver.
```
BorjaRequena, line 612: The result is pretty nice! Fidelity 1 up to 1e-6 is pretty impressive 🚀

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+num_wires = 3
+# New pulse hyperparameters
+T = 3 * jnp.pi  # Longer total duration
+eps = 0.1 * T
+P = 5  # More rectangles in sum: P=5
+f = partial(smooth_rectangles, k=k, max_amp=max_amp, eps=eps, T=T)
+
+# Hamiltonian terms of the drift and parametrized parts of H
+ops_H_d = [Z(0), Z(1), Z(2)]
+ops_param = [P(w) for P in [X, Y, Z] for w in range(num_wires)]
```
BorjaRequena, line 642: Maybe we should avoid using `P` here since it is the number of rectangles in every sum.
```suggestion
ops_param = [pauli(w) for pauli in [X, Y, Z] for w in range(num_wires)]
```
I am not very happy with the `pauli` name. It's mainly to avoid any possible confusion. Feel free to use any other name 😄 Again, not too critical.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+ops_param += [Z(0) @ X(1), Z(1) @ X(2), Z(2) @ X(0)]
+
+# Coefficients: 1. for drift Hamiltonian and smooth rectangles for parametrized part
+coeffs = [1.0, 1.0, 1.0] + [f for op in ops_param]
```
BorjaRequena, line 646: If you decide to change the name of the previous `f` in the end, we should also update this one.

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+max_params = params_hist[jnp.argmax(jnp.array(profit_hist))]
+
+#############################################################################
+# This looks promising: Adam maximized the fidelity successfully and we thus compiled
+# a pulse sequence that implements a Toffoli gate!
+# To inspect how close the compiled pulse sequence is to the Toffoli gate,
+# we can apply it to an exemplary quantum state, say :math:`|110\rangle`,
+# and investigate the returned probabilities. A perfect Toffoli gate would
+# flip the third qubit, returning a probability of one in the last entry
+# and zeros else.
```
BorjaRequena, line 699: I am not sure this is a proper suggestion. It just sounded a bit odd to me, take it or leave it 😄
```suggestion
# flip the third qubit, returning a probability of one in the last entry
# and zeros elsewhere.
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+    ylabel="Output state",
+)
+# Add axes for the colorbar
+cax = plt.axes([0.85, 0.15, 0.02, 0.62])
+sc = mpl.cm.ScalarMappable(cmap=mpl.cm.turbo, norm=color_norm)
+sc.set_array([])
+# Plot colorbar
+plt.colorbar(sc, cax=cax, ax=ax)
+plt.show()
+
```
BorjaRequena, line 758: Badass plot! 🔥

BorjaRequena (commented):
Pretty cool demo! Thanks @dwierichs for putting together such a great work 🔥 

I left a few minor comments. The most important one is regarding the $S_k$ notation in the first parametrized Hamiltonian.

demonstrations/tutorial_optimal_control.py
dwierichs: It's way easier without hardware constraints than with :P :) Thanks though!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+`open-source quantum toolbox QuTiP <https://qutip.org/qutip-tutorials/#optimal-control>`__.
+
+Gate calibration via pulse programming
+--------------------------------------
+
+Here, we briefly discuss the general setup of pulse programs that we will use for our
+optimal control application. For more details, you may peruse the related
+tutorials focusing on pulse programming.
+
+Consider a quantum system comprised of :math:`N` two-level systems, or qubits, described
```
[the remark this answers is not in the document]
dwierichs: Yes, it should indeed be lower case! Great catch!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,847 @@
...
+# abstract cross resonance driving term. That is, our Hamiltonian is
+#
+# .. math::
+#
+#     H(\boldsymbol{p}, t) = \underset{H_d}{\underbrace{Z_0 + Z_1}}
+#     + S_k(\boldsymbol{p_1}, t) Z_0
+#     + S_k(\boldsymbol{p_2}, t) X_1
+#     + S_k(\boldsymbol{p_3}, t) Y_1
+#     + S_k(\boldsymbol{p_4}, t) Z_1
+#     + \underset{\text{interaction}}{\underbrace{S_k(\boldsymbol{p_5}, t) Z_0X_1}}
```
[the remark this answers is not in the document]
dwierichs: Thanks, good call! 
I now did the following: Before the equation you marked, I explicitly say $f_i=S_k \forall i$, because previously, the coefficient functions are called $f_i$ and in the large equation, they suddenly are called $S_k$.
Essentially, I moved a bit of the commentary on the coefficient function above the equation, which to me seems to fix the problem.
Let me know what you think :)

demonstrations/tutorial_optimal_control.py
dwierichs: How about `S_k`? :)

demonstrations/tutorial_optimal_control.py
dwierichs: Yeah, you're right! I'll modify it further, thanks! :)

demonstrations/tutorial_optimal_control.py
dwierichs: Went for `pauli_op` :)

BorjaRequena: Much better!

demonstrations/tutorial_optimal_control.py
BorjaRequena: I like it!

demonstrations/tutorial_optimal_control.py
BorjaRequena: Oh ye! 😄

demonstrations/tutorial_optimal_control.py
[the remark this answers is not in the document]
BorjaRequena: Nice, I think it is much clearer now!

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+#
+# The idea behind using the sum of smooth rectangles function for the parametrization
+# is the following:
+# Many methods in quantum optimal control work with discretized pulse shapes that keep
+# the pulse envelope constant for short time bins. This approach leads to a large number
+# of parameters that need to be trained, and it requires us to manually enforce that the
+# values do not differ by too much between neighbouring time bins.
+# The smooth rectangles introduced above have a limited rate of change by design, and
+# the number of parameters is much smaller than in generic discretization approaches.
+# Each coefficient function :math:`f_i` sums :math:`P` smooth rectangles
```
BorjaRequena, line 334: Just this sneaky boi here 🥷 
```suggestion
# Each coefficient function :math:`S_k` sums :math:`P` smooth rectangles
```

BorjaRequena (approved):
Thanks for the changes @dwierichs!! Looking good 🚀

demonstrations/tutorial_optimal_control.py
dwierichs: Hahaha.
Yeah, I was unsure about this one, because before this point, the Hamiltonian functions are still called $f_i$ :thinking:

demonstrations/tutorial_optimal_control.metadata.json
```diff
@@ -0,0 +1,101 @@
...
+            "publisher": "",
+            "journal": "Phys. Rev. Lett. 106, 190501",
+            "doi": "10.1103/PhysRevLett.106.190501",
+            "url": "https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.106.190501",
+            "preprint": "https://arxiv.org/abs/1003.3750"
+        }
+    ],
+    "basedOnPapers": [],
+    "referencedByPapers": [],
+    "relatedDemonstrations": [
```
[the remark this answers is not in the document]
dwierichs: Not sure, I don't remember changing this. :thinking:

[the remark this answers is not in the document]
dwierichs: It's actually coming from [`demonstrations_metadata.md`](https://github.com/PennyLaneAI/qml/blob/c185dd4a90ab3e2dc47d318d193cebfbb87fb4bd/demonstrations_metadata.md?plain=1#L55)

demonstrations/tutorial_optimal_control.py
dwierichs: Ah no, you are right, I confused them! :)

demonstrations/tutorial_optimal_control.metadata.json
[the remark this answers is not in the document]
dwierichs: @KetpuntoG I modified the linked line in `demonstrations_metadata.md` as well as in the metadata for the demo added here.

trbromley:
[SC-37123]

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+    ahs_aquila Pulse programming on Rydberg atom hardware
+
+*Author: David Wierichs. Posted: xx July, 2023.*
+
+
+Today, quantum computations are largely phrased in the language of quantum circuits,
+composed of digital quantum gates.
+However, most quantum hardware does not come with such digital
+gates as elementary native operations.
+Instead, the hardware allows us to play sequences of analogue electromagnetic pulses,
```
KetpuntoG, line 22:
```suggestion
Instead, the hardware allows us to play sequences of analog electromagnetic pulses,
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+    starting and ending times, as well as the total output amplitude.
+
+    Args:
+        params (tensor_like): Amplitudes and start and end times for the rectangles,
+            in the order '[amp_1, ... amp_P, t_{1, 0}, t_{1, 1}, ... t_{P, 0}, t_{P, 1}]'.
+        t (float): Time at which to evaluate the pulse function.
+        k (float): Steepness of the sigmoid functions that delimit the rectangles
+        max_amp (float): Maximal amplitude of the rectangles. The output will be normalized
+            to the interval '(-max_amp, max_amp)'.
+        eps (float): Margin to beginning and end of the pulse sequence within which the
```
KetpuntoG, line 378:
```suggestion
        eps (float): Margin to the beginning and end of the pulse sequence within which the
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+
+    Returns:
+        float: Value of sum of smooth-rectangle pulses at 't' for the given parameters.
+    """
+    P = len(params) // 3
+    # Split amplitudes from times
+    amps, times = jnp.split(params, [P])
+    # Normalize times to be sufficiently far away from 0 and T
+    times = sigmoid(times - T / 2, k=1.0) * (T - 2 * eps) + eps
+    # Extract start and end times of single rectangles
```
KetpuntoG, line 390:
```suggestion
    # Extract the start and end times of single rectangles
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+# Note that the rectangles are rather round for these generic parameters.
+# The optimized parameters in the training workflows below will lead to more
+# sharply defined pulses that resemble rectangles more closely. The amplitude normalization
+# step in ``smooth_rectangles`` enables us to produce them in a differentiable manner,
+# as was our goal with introducing :math:`R_k`.
+# Also note that the normalization of the final output value is not a simple clipping
+# step, but again a smooth function. As a consequence, the amplitudes ``1.9`` and ``-2.``
+# in the example above, which are not in the interval ``[-1, 1]``,
+# are not set to ``1`` and ``-1`` but take smaller absolute values.
+# Finally, also not that the start and end times of the smooth rectangles are being
```
KetpuntoG, line 435:
```suggestion
# Finally, also note that the start and end times of the smooth rectangles are being
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+    # Compute the unitary time evolution of the pulse Hamiltonian
+    op_mat = pulse_matrix(params)
+    # Compute the fidelity between the target and the pulse evolution
+    return jnp.abs(jnp.trace(target.conj().T @ op_mat)) / 2**num_wires
+
+
+grad = jax.jit(jax.grad(profit))
+
+#############################################################################
+# We create initial parameters similar to above but allow for a larger number
```
KetpuntoG, line 683:
```suggestion
# We create initial parameters similar to the above but allow for a larger number
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+# very good approximations to the target gates.
+# Thanks to JAX, just-in-time (JIT) compiling and the PennyLane ``pulse``
+# module, training the pulse sequences was simple to implement and fast to run.
+#
+# There are many different techniques in quantum optimal control that can be
+# used to calibrate pulse sequences, some of which include gradient-based
+# training. A widely-used technique called GRAPE [#KhanejaReiss05]_
+# makes use of discretized pulses, which leads to a large number of free parameters
+# to be optimized with gradient ascent.
+# The technique shown here reduces the parameter count signficantly
```
KetpuntoG, line 799:
```suggestion
# The technique shown here reduces the parameter count significantly
```

KetpuntoG (approved):

demonstrations/tutorial_optimal_control.metadata.json
```diff
@@ -0,0 +1,101 @@
+{
+    "title": "Optimal control for gate compilation",
+    "authors": [
+        {
+            "id": "david_wierichs"
+        }
+    ],
+    "dateOfPublication": "2023-07-25T00:00:00+00:00",
+    "dateOfLastModification": "2023-07-25T00:00:00+00:00",
```
KetpuntoG, line 9:
```suggestion
    "dateOfPublication": "2023-08-08T00:00:00+00:00",
    "dateOfLastModification": "2023-08-08T00:00:00+00:00",
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+|
+
+For an introduction, see
+:doc:`the demo on differentiable pulse programming </demos/tutorial_pulse_programming101>`
+in PennyLane.
+Instead of optimizing pulses to yield digital quantum gates,
+we may use them directly to solve optimization problems, as is also showcased in this
+introductory demo. If you are interested in specific hardware pulses, take a look at
+:doc:`an introduction to neutral-atom quantum computing </demos/tutorial_neutral_atoms>`
+or :doc:`the tutorial on the QuEra Aquila device </demos/ahs_aquila>`, which treat pulse
```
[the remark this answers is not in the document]
dwierichs: I think both mentioned tutorials treat Rydberg atoms, so I'd keep the plural :)

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,854 @@
...
+    :property="og:description": Optimize pulse programs to obtain digital gates.
+    :property="og:image": https://pennylane.ai/qml/_images/thumbnail_tutorial_optimal_control.png
+
+.. related::
+
+    tutorial_pulse_programming101 Introduction to pulse programming in PennyLane
+    tutorial_neutral_atoms Introduction to neutral atom quantum computers
+    ahs_aquila Pulse programming on Rydberg atom hardware
+
+*Author: David Wierichs. Posted: xx July, 2023.*
```
KetpuntoG, line 15:
```suggestion
```

demonstrations/tutorial_optimal_control.py
```diff
@@ -0,0 +1,853 @@
...
+# that the shown optimization remains a toy model for calibration of
+# quantum hardware. We did not take into account the interaction terms
+# or pulse shapes available on realistic devices and their control electronics.
+# We also did not consider a unit system tied to real devices, and we
+# ignored noise, which plays a very important role in today's quantum devices
+# and in quantum optimal control.
+# We leave the extension to real-world pulse Hamiltonians and noisy systems
+# to a future tutorial--or maybe your work?
+#
+# Bibliography
```
KetpuntoG, line 819:
```suggestion
# References
```

PR #832, opened 2023-06-20
Change private methods in adjoint diff tutorial

albi3ro:
The current adjoint diff tutorial uses private methods for default qubit.

 We have ported all the simulation tools over to a new `qubit` submodule, and we are deprecating the methods of the old default.qubit device.

So this updates the demos to use the newer `qubit` submodule functions, instead of the private default qubit methods.

### DIFF
--- a/demonstrations/tutorial_adjoint_diff.py
+++ b/demonstrations/tutorial_adjoint_diff.py
@@ -110,21 +110,26 @@ ops = [
 M = qml.PauliX(wires=1)
 
 ##############################################################################
-# We create our state by using the ``"default.qubit"`` methods ``_create_basis_state``
-# and ``_apply_operation``.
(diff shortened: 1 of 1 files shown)


lillian542 (approved):
Much nicer this way! 🎉

mudit2812 (approved):

demonstrations/tutorial_adjoint_diff.py
```diff
@@ -347,25 +352,24 @@ def circuit(a):
 # function spits back out a matrix instead of an operation,
 # we have to use ``dev._apply_unitary`` instead to create :math:`|\tilde{k}_i\rangle`.
```
timmysilv, line 353:
```suggestion
# we have to use :class:`~pennylane.QubitUnitary` instead to create :math:`|\tilde{k}_i\rangle`.
```

PR #835, opened 2023-06-22
Fixes for 0.31 release

mudit2812:
**Summary:**
* `ensemble_multi_qpu.py` prints a warning due to permuting a zipped array of features and classes with ragged dimensions. This PR updates the code so that the permutation for the feature array and class array happens separately so that the warning is not raised.
* Updated `tutorial_quantum_circuit_cutting.py` to remove use of the `do_queue` keyword.


timmysilv (approved):
snazzy fix 🕴️

demonstrations/tutorial_quantum_circuit_cutting.py
```diff
@@ -658,28 +658,29 @@ def make_kraus_ops(num_wires: int):
 # In code, this looks like:
 #
 
-with QuantumTape(do_queue=False) as tape0, QuantumTape(do_queue=False) as tape1:
-    # Record on new "fragment" tapes
+with qml.queuing.QueuingManager.stop_recording():
```
timmysilv, line 661: i wonder if this is even necessary?

mudit2812: We need to stop recording somewhere since there are two operations, each one being appended to a different queuing context. This solution is pretty bad, so I've rewritten this part so that the `stop_recording` is a bit more fine grained.

albi3ro: We can just rewrite all of this to use the "construct upon initialization" syntax and avoid all the queuing confusion.

demonstrations/tutorial_quantum_circuit_cutting.py
albi3ro, line 1:
```
def channel_from_wire_cut(op: qml.WireCut, i: int) -> qml.QubitChannel:
    k = len(op.wires)
    d = 2**k

    kraus_ops = make_kraus_ops(k)  # Generate Kraus operators on the fly
    probs = (d + 1) / (2 * d + 1), d / (2 * d + 1)  # Probabilities of the two channels

    return qml.QubitChannel(kraus_ops[i], wires=op.wires, do_queue=False)

ops0 = [channel_from_wire_cut(op, 0) if isinstance(op, qml.WireCut) else op for op in tape.operations]
ops1 = [channel_from_wire_cut(op, 1) if isinstance(op, qml.WireCut) else op for op in tape.operations]

tape0 = QuantumTape(ops0, tape.measurements)
tape1 = QuantumTape(ops1, tape.measurements)
```

A potential option.

demonstrations/tutorial_quantum_circuit_cutting.py
```diff
@@ -658,28 +658,31 @@ def make_kraus_ops(num_wires: int):
...
+probs = (d + 1) / (2 * d + 1), d / (2 * d + 1)  # Probabilities of the two channels
 
-            psi_0 = qml.QubitChannel(K0, wires=op.wires, do_queue=False)
-            psi_1 = qml.QubitChannel(K1, wires=op.wires, do_queue=False)
+psi_0 = qml.QubitChannel(K0, wires=wire_cut.wires)
+psi_1 = qml.QubitChannel(K1, wires=wire_cut.wires)
 
-            qml.apply(psi_0, context=tape0)
-            qml.apply(psi_1, context=tape1)
+ops_0 = tape.operations
```
timmysilv, line 679: this performs a copy because `tape.operations` is `tape._prep + tape._ops`, so it works fine. figured i'd point it out because if one day that changes, this might change lists in place which could be bad. probably won't though

demonstrations/tutorial_quantum_circuit_cutting.py
```diff
@@ -689,6 +692,8 @@ def make_kraus_ops(num_wires: int):
 print(f"Cut size: k={k}")
 print(f"Channel probabilities: p0={probs[0]:.2f}; p1={probs[1]:.2f}", "\n")
 
+print(tape0.operations)
+
```
timmysilv, line 696:
```suggestion
```

timmysilv (approved):
thanks for getting rid of queuing there! I suggested the removal of an extra print, but it didn't remove the line below it so black might fail. lgtm otherwise

BorjaRequena (approved):
Nice fix! It even looks cleaner than before 😄

PR #845, opened 2023-06-29
Fix and re-add embedding demo 

mariaschuld:
The embedding_metric_learning demo was removed a few years back because of a bug that I could not trace (and no priority at the time to fix it). 

This PR fixes the bug and speeds up the code using jax.


KetpuntoG:
Oh! this is a great hackweek project 🤩

josh146:
@mariaschuld for some reason it is giving the error
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/qml/qml/demonstrations/embeddings_metric_learning/training.png'
```

which is super strange to me 🤔

mariaschuld:
Thanks @KetpuntoG and Josh! I will try to find a bit of time this week to finish the PR. To be honest, the model does train more or less, but not in every run - I need to think about how to communicate this honestly in the demo.

KetpuntoG:
Hey @mariaschuld , is the demo ready for review?

mariaschuld:
I am so sorry for the delay, I want to explore a bit more. The model sometimes trains and often just performs sub-par. I need to go back and compare with the paper code...

KetpuntoG:
> I am so sorry for the delay, I want to explore a bit more. The model sometimes trains and often just performs sub-par. I need to go back and compare with the paper code...

Don't worry, there's no rush 😄 I just wanted to make sure you weren't waiting for us

mariaschuld:
I am closing this PR because after fixing the bug (and a few others along the way), the issue of overfitting persists: The metric learning objective manages to nicely arrange the training data on a periodic grid that helps the quantum embedding to separate the two image classes. However, the optimal parameters do not generalise, i.e. they do not have the same effect on the test set. I assume this is due to an interplay between a very small dataset and the metric learning objective itself (where overfitting seems to be a known issue). 

The demo itself is otherwise publication-ready except from an issue with the meta-data that I couldn't fix. Given that we close the PR I won't spend more time on the attempt.

KetpuntoG:
This thumbnail has been used for another demo.

PR #846, opened 2023-06-29
Minor corrections for pulse programming on Rydberg atom arrays

BorjaRequena:
Going through [the tutorial](https://pennylane.ai/qml/demos/ahs_aquila), I spotted some minor things that could be amended.


demonstrations/ahs_aquila.py
```diff
@@ -575,9 +574,8 @@ def circuit(params):
 # effect of Rydberg blockade arising from the interaction term.
 #
 #
-# The radius within which two neighboring atoms are effectively prevented from both being excited is referred to as the
-# *blockade radius* :math:`R_b`. The blockade radius is proportional to the :math:`C_6^{1/6}` value for the transition
```
lillian542, line 579: Why delete the first sentence? Is this already defined earlier?

demonstrations/ahs_aquila.py
```diff
@@ -349,17 +350,17 @@
...
-#        'rabiFrequencySlewRateMax': Decimal('250000000000000.0'),
-#        'detuningRange': (Decimal('-125000000.0'), Decimal('125000000.0')),
-#        'detuningResolution': Decimal('0.2'),
-#        'detuningSlewRateMax': Decimal('2500000000000000.0'),
-#        'phaseRange': (Decimal('-99.0'), Decimal('99.0')),
-#        'phaseResolution': Decimal('5E-7'),
-#        'timeResolution': Decimal('1E-9'),
-#        'timeDeltaMin': Decimal('5E-8'),
-#        'timeMin': Decimal('0.0'),
-#        'timeMax': Decimal('0.000004')}}
```
lillian542, line 362: The spacing above is correct - this is a nested dictionary, so these elements are all indented by one space compared to 'c6Coefficient' and 'rydbergGlobal' when printed in a notebook

demonstrations/ahs_aquila.py
```diff
@@ -409,7 +410,7 @@ def angular_SI_to_MHz(angular_SI):
 #     | .. centered:: | .. centered::    | .. centered::    | .. centered:: | .. centered:: | .. centered::           |
 #     |  Parameter    |  Minimum value   |   Maximum value  |   Resolution  |  PWC or PWL   |  Maximum rate of change |
 #     +===============+==================+==================+===============+===============+=========================+
-#     | Amplitude     | 0 MHz            |  2.51465 MHz     | 64 MHz        |  PWL          | 39788735 MHz/s          |
+#     | Amplitude     | 0 MHz            |  2.51465 MHz     | 6.4e-5 MHz    |  PWL          | 39788735 MHz/s          |
```
lillian542, line 413: I don't think this change is correct. The resolution is 400 rad/s, which converts to 0.000064 MHz, or 6.4e-5 MHz. I believe the original text is accurate.

demonstrations/ahs_aquila.py
```diff
@@ -479,7 +480,7 @@ def gaussian_fn(p, t):
 ######################################################################
 # With only amplitude as non-zero, the overall driven Hamiltonian in this case simplifies to:
 #
-# .. math::  \hat{H} = \sum_{k=1}^N \frac{\Omega(t)}{2} (| g_k \rangle \langle r_k | - | r_k \rangle \langle g_k |) + \sum_{j=1}^{N-1}\sum_{k=j+1}^{N} \frac{C_6}{R^6_{jk}}\hat{n}_j\hat{n}_k
+# .. math::  \hat{H} = \sum_{k=1}^N \frac{\Omega(t)}{2} (| g_k \rangle \langle r_k | + | r_k \rangle \langle g_k |) + \sum_{j=1}^{N-1}\sum_{k=j+1}^{N} \frac{C_6}{R^6_{jk}}\hat{n}_j\hat{n}_k
```
lillian542, line 483: Nice catch! 🚀

demonstrations/ahs_aquila.py
```diff
@@ -608,13 +606,14 @@ def circuit(params):
...
-times = np.linspace(0, 1.75, 1000)
+times = np.linspace(0, 1.75, 36)
+timestep = (times[1] - times[0]) * 1e-6  # Time step in seconds
 amplitude = [gaussian_fn(amplitude_params, t) for t in times]
 
 start_val = amplitude[0]
 stop_val = amplitude[-1]
 max_val = np.max(amplitude)
-max_rate = np.max([(amplitude[i + 1] - amplitude[i]) / 50e-9 for i in range(999)])
+max_rate = np.max([(amplitude[i + 1] - amplitude[i]) / timestep for i in range(len(times))])
```
lillian542, line 616: ❤️

demonstrations/ahs_aquila.py
```diff
@@ -108,16 +108,16 @@
 and Rydberg states (for Aquila, :math:`5.24 \times 10^{-24} \text{rad m}^6 / \text{s}`, referring to the
 :math:`| 70S_{1/2} \rangle` state of the Rb-87 atom).
 
-There are two key things to be aware of in the interaction term. First, the energy contribution of
+There are two key elements in the interaction term. First, the energy contribution of
```
lillian542, line 111: Personally I prefer the previous formulation, 'elements in the interaction term' sounds to me more like you are going to talk about different variables that contribute to it, rather than a description of its implications for the system's behaviour. But I don't have a strong opinion here. Could maybe also be 'things to notice' if that sounds better?

lillian542 (commented):
Looks good Borja, thanks for catching these things! 

There are a couple of the changes where I believe the original text is more correct than the updates; have a look a see what you think.

demonstrations/ahs_aquila.py
BorjaRequena: I like the old formulation too. Let's keep it as it was originally, this was mainly just a minor style suggestion😄

demonstrations/ahs_aquila.py
BorjaRequena: 🔥

demonstrations/ahs_aquila.py
BorjaRequena: I thought I was missing something when I went through this initially 😅

demonstrations/ahs_aquila.py
BorjaRequena: I am not sure I understand this. In the original text, it said 64 MHz, and I changed it to be 6.4e-5 MHz, which is what you say would be correct. Maybe it was intended to be 64 Hz? 🤔 It wouldn't make sense to have a resolution of 64 MHz if the maximum is 2.5 MHz, right?

demonstrations/ahs_aquila.py
BorjaRequena: Oh, I entirely missed it! I let my OCD take over here 😝

demonstrations/ahs_aquila.py
BorjaRequena: I moved the blockade radius definition [here](https://github.com/PennyLaneAI/qml/blob/030ec4694d5292d998c587f1d2bc01c37ab2f9f4/demonstrations/ahs_aquila.py#L135) because we were using the concept before reaching this far in the demo but we had not properly stated what it was. Let me know what you think 😄

demonstrations/ahs_aquila.py
lillian542: Hahah... I'm sorry, I read this backwards, as if it had been changed to 64 MHz 🙈 Good catch, thank you!

demonstrations/ahs_aquila.py
lillian542: Awesome, looks great :)

lillian542 (approved):
Thanks for catching these @BorjaRequena ! Looks good 🎉

KetpuntoG (approved):
🚀

PR #848, opened 2023-07-03
udate demo date

KetpuntoG:
It seems that React does not show the demos before the date. To make final checks I will mark from now on the publication date on Mondays

### DIFF
--- a/demonstrations/tutorial_grovers_algorithm.metada.json
+++ b/demonstrations/tutorial_grovers_algorithm.metada.json
@@ -5,8 +5,8 @@
             "id": "ludmila_botelho"
         }
     ],
-    "dateOfPublication": "2023-07-04T00:00:00",
-    "dateOfLastModification": "2023-06-29T00:00:00",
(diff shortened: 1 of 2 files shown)


ikurecic (approved):

josh146:
@KetpuntoG is this behaviour we should change/a bug?

KetpuntoG:
> @KetpuntoG is this behaviour we should change/a bug?

hmm it really makes sense that future dates cannot be published. I would simply have to post the merge date. Although I would suggest being able to do that in case in the future some date is more important to respect. I can include it in the feedback log with a P1 

(could you approve the PR @josh146 ? not a lot of people online today)

KetpuntoG:
> @KetpuntoG is this behaviour we should change/a bug?

hmm it makes sense that future dates cannot be published. I would simply have to post the merge date.
Although I would suggest being able to do that in case in the future some date is more important to respect.

(could you approve the PR josh? not a lot of people online today)

larawatson (approved):

PR #865, opened 2023-07-17
fixes W(a) --> U(a) typo

ixfoduap:
An operation first defined as U(a) was incorrectly referred to as W(a) afterwards. This fixes the issue


KetpuntoG (approved):
Good catch, thanks!  🚀

Jaybsoni (approved):
Looks good 💯

PR #871, opened 2023-07-21
Fixes to metadata

BM7878:
This consists of, mostly, just adding "type": "article" to lots of the references, and fixing datetimes.

There are also a few adjustments to the schema to allow for some of the more unusual references that are used.

### DIFF
--- a/demonstrations/tutorial_classical_shadows.metadata.json
+++ b/demonstrations/tutorial_classical_shadows.metadata.json
@@ -26,6 +26,7 @@
     "references": [
         {
             "id": "Mauro2003",
+            "type": "article",
             "title": "Quantum Tomography",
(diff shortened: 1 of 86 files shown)


demonstrations/tutorial_classical_shadows.metadata.json
```diff
@@ -42,9 +44,10 @@
         },
         {
             "id": "Gottesman1997",
+            "type": "article",
             "title": "Stabilizer Codes and Quantum Error Correction",
             "authors": "Gottesman, Daniel",
-            "year": "",
+            "year": "1997",
             "journal": "",
```
Mandrenkov, line 51: Should we do anything about these empty `journal` entries?

Mandrenkov (approved):
Awesome, @BM7878! 🥳

PR #876, opened 2023-07-27
removing files

KetpuntoG:
In this PR I delete files that are not needed in the QML repo. Among them I include everything related to the Glossary, Quantum Datasets, "What is QML/Qchem/QC", community demos and videos.
Most of the deleted pages are because I have seen react have them [here](https://github.com/XanaduAI/pennylane.ai-react/tree/dev/apps/pennylane-website/src/pages/qml) (I'm not very familiar with React so I'm not sure if the fact that it's there implies that it can be deleted, so please, double check 🙂)

It has also been changed to default to deploy on the demos page instead of the main page. (This will not be visible until we merge it)


rashidnhm:
@KetpuntoG I've added the ci label to build this PR. Please push another commit to trigger CI

KetpuntoG:
> @KetpuntoG I've added the ci label to build this PR. Please push another commit to trigger CI
> 
> EDIT: You can push an empty commit with: `git commit --allow-empty -m "Trigger CI"`

Thanks @rashidnhm !
Is it possible to make the deployment point to `/demonstrations.html` instead of `/index.html`?

rashidnhm:
> Thanks @rashidnhm ! Is it possible to make the deployment point to `/demonstrations.html` instead of `/index.html`?

Yes, you can [update this line](https://github.com/PennyLaneAI/qml/blob/9d8db2fbbda8c73737d3d200d083fe6911ebcff9/.github/workflows/deploy-pr.yml#L80) and switch out `index.html` to `demonstrations.html` ... But do note that the change will not take affect until the PR is merged.

ashishks0522 (commented):
This is great! 🧹 

@rashidnhm @KetpuntoG Would it be better to try this out in `dev` first before merging to `master`? Just to see all things work and render correctly?

KetpuntoG:
> This is great! 🧹
> 
> @rashidnhm @KetpuntoG Would it be better to try this out in `dev` first before merging to `master`? Just to see all things work and render correctly?

The dev version would be in react?
I think it worth it!

rashidnhm:
> This is great! 🧹 
> 
> @rashidnhm @KetpuntoG Would it be better to try this out in `dev` first before merging to `master`? Just to see all things work and render correctly?

Yup I agree to this

ashishks0522 (approved):

PR #890, opened 2023-08-14
lightning performance

KetpuntoG:
In this PR I will update the demos based on the feedback received by the performance team.
(still work in progress)


KetpuntoG:
In 0.33 this will work without having to make any changes. It has been decided to leave the demos as they are

PR #895, opened 2023-08-21
Remove some stuff dev

KetpuntoG:
Removing demo QSVT


KetpuntoG:
I'm going to leave the demo on dev as it's not bad there either

PR #901, opened 2023-08-23
learning2learn executable

KetpuntoG:
This PR is intended to convert the `learning2learn` demo to an executable demo

### DIFF
--- a/demonstrations/tutorial_givens_rotations.py
+++ b/demonstrations/tutorial_givens_rotations.py
@@ -308,7 +308,7 @@ print(states)
 #      0 & \cos (\theta) & e^{i\phi}\sin (\theta) & 0\\
 #      0 & e^{-i\phi}\sin(\theta) & -\cos(\theta) & 0\\
 #      0 & 0 & 0 & 1
-#      \end{pmatrix},\\[12pt]
+#      \end{pmatrix},\\
 #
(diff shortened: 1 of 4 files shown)


isaacdevlugt (commented):
Looks good but there are math issues

Screenshot 2023-08-23 at 4 04 50 PM

KetpuntoG:
> Looks good but there are math issues
> 
> Screenshot 2023-08-23 at 4 04 50 PM

Oh that's weird, can you reload the page? 

Captura de pantalla 2023-08-23 a las 16 10 46

isaacdevlugt (approved):
Math errors are gone. Not sure what happened 🤔. LGTM!

ikurecic (approved):
@KetpuntoG I'd probably change the 'last updated' line in the PY file. Otherwise, I _think_ everything is good like this. :)

KetpuntoG:
> @KetpuntoG I'd probably change the 'last updated' line in the PY file. Otherwise, I _think_ everything is good like this. :)

I have thought about it but it doesn't really change anything for the user point of view. That's why I decided to leave it. Do you think it should be changed in that case?

ikurecic:
You're right about the content, but the URL will change, right?
I've seen at least one research paper that I can remember off the top of my head that referenced that demo. So I would say this is a modification. Actually, maybe you should check about the URL change with Alex.

PR #903, opened 2023-08-24
Avoid index out of range error

lillian542:
Found a small mistake when checking the demo. Replacing `range(len(times))` with `range(len(times)-1)` avoids an index out of range error.


lillian542:
from wrong base

PR #905, opened 2023-08-24
Add warning filter to `tutorial_measurement_optimize`

mudit2812:
**Summary:**
Operator `__hash__` and `__eq__` raise warnings to indicate upcoming changes to PennyLane. This causes deprecation warnings to be raised in `tutorial_measurement_optimize.py` when operators are placed in a `networkx` graph. The new behaviour will not break the demo in any way, so I've added a warning filter for now, which should be removed for the 0.33 release.


KetpuntoG (approved):
Great! I see no warning 🚀

lillian542 (approved):

PR #914, opened 2023-08-29
Upgrade TensorFlow version

mudit2812:
As title says.

Also fixed `tutorial_pasqal.py` and `tutorial_QGAN.py` to be compatible with the new TF version


timmysilv (approved):
I triggered CI so it builds all demos to make sure this fixes it. Assuming it passes, :shipit: :shipit: :shipit:

mudit2812:
Build failing because plugins master haven't been updated to get rid of `qml.active_return`

isaacdevlugt (approved):

PR #917, opened 2023-09-06
Deploy prod to staging

doctorperceptron:
Depends on PR #920 being merged first.

This PR adds two new features to the existing CI workflows:
1. Merges to `master` will now deploy the demos to the [staging environment](https://staging.pennylane.ai) rather than into production.
2. The demos from `master` can be deployed to [production](https://pennylane.ai) via a manually triggered workflow (`deploy-demos-to-prod`).

The purpose for this is so that after reviewing new demos via PR preview, they can be QA-ed one final time on the staging website to ensure everything works correctly before promoting them to production. This PR requires [this pennylane.ai-react PR](https://github.com/XanaduAI/pennylane.ai-react/pull/541) to be merged as well.

NOTE: Won't update the README as this is for external users.


.github/workflows/deploy-demos-to-prod.yml
```diff
@@ -0,0 +1,55 @@
+name: Deploy Demos to Production
+on:
+    workflow_dispatch:
+
+jobs:
```
Mandrenkov, line 5: Having three separate jobs is overkill - just one (with three steps) would do!

.github/workflows/deploy-demos-to-prod.yml
```diff
@@ -0,0 +1,55 @@
...
+      AWS_ACCESS_KEY_ID: ${{ secrets.PL_SITE_STAGING_NON_REACT_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.PL_SITE_STAGING_NON_REACT_SECRET_ACCESS_KEY }}
+      AWS_S3_BUCKET_ID: ${{ secrets.PL_SITE_STAGING_QML_JSON_S3_BUCKET_ID }}
+
+    steps:
+      - name: Pull staging bucket
+        run: |
+          aws s3 cp s3://$AWS_S3_BUCKET_ID/master ./local-master
+          aws s3 cp s3://$AWS_S3_BUCKET_ID/qml/searchindex.js searchindex.js
+          rm -rf ./local-master/robots.txt
```
Mandrenkov, line 20: Why are we deleting this file?

Mandrenkov (approved):
Looks good to me! 👍

.github/workflows/deploy-demos-to-prod.yml
rashidnhm, line 1: I don't think the way this workflow is currently structured will work .... each job is deployed on a fresh runner, and will not have access to the GITHUB_WORKSPACE of other runners. The easiest way to do this would be to actually have 1 job but with each stage being a step:

(This is pseudocode)
```yaml
sync_staging_to_production:
  - name: Pull from Staging Bucket
    run: |
       aws s3 sync s3://staging local
       <Other commands>
  
  - name: Push to Production
    run: aws s3 sync local s3://production

  - trigger website build
```
The third can or cannot be it's own job (this is up to preference). But the first two steps has to be under one job.

.github/workflows/deploy-demos-to-prod.yml
doctorperceptron: This prevents Google from crawling the staging site. We want Google to crawl the production site though, so we need to delete it first.

doctorperceptron: ... but, I don't think this is actually necessary after all. This is just the demos bucket, not the website, so there wouldn't be a robots.txt file. Remove this step.

.github/workflows/deploy-demos-to-prod.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    steps:
+      - name: Pull staging bucket
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.PL_SITE_STAGING_NON_REACT_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.PL_SITE_STAGING_NON_REACT_SECRET_ACCESS_KEY }}
+          AWS_S3_BUCKET_ID: ${{ secrets.PL_SITE_STAGING_QML_JSON_S3_BUCKET_ID }}
+        run: |
+          aws s3 cp s3://$AWS_S3_BUCKET_ID/master ./local-master
+          aws s3 cp s3://$AWS_S3_BUCKET_ID/qml/searchindex.js searchindex.js
```
rashidnhm, line 20: The searchindex unfortunately does NOT reside with the demos ... it is synced to the "non-react" bucket as /search.html and /searchindex.js are still served by sphinx .. 

Ref:
https://github.com/XanaduAI/xanadu-terraform-provisioning/blob/33f2edadaca267540a5b539344cd0db3781b0317/workloads/software/quantum_software/pennylane/pennylane_ai_website_staging/locals.tf#L48-L49


There are two options on how to handle this:
1) Pull down the searchindex.js from `${{ secrets.PL_SITE_<ENV>_S3_BUCKET_NAME }}/qml/searchindex.js`. This bucket exists in all envs alongside the JSON qml bucket.

2) Pull down **searchindex.json** from the JSON bucket and then [formatting it](https://github.com/PennyLaneAI/qml/blob/161db02c434899c33d5dcd5ca1d47bf19ed6233b/.github/workflows/upload-json.yml#L73-L78) to be same as searchindex.js


Either way ... if this file is uploaded to the react prod bucket, it won't be consumed as it needs to still exist in the non-react website since search.html will be served from there.


As an aside,
... Writing this comment made me notice a bug in upload-json which I will go fix now

Mandrenkov (approved):
Looks good to me once @rashidnhm's comment is resolved!

rashidnhm (approved):
:tada: Great stuff @doctorperceptron !!

PR #918, opened 2023-09-06
testing style

KetpuntoG:
Pl style of matplotlib does not work in the repository. 
Testing locally has shown that updating matplotlib does work.


timmysilv:
can you re-run CI with the build-all-demos label? bumping matplotlib is a major change for qml, I just wanna make sure everything still works as expected

KetpuntoG:
> can you re-run CI with the build-all-demos label? bumping matplotlib is a major change for qml, I just wanna make sure everything still works as expected

Oh good point!!

timmysilv (approved):
awesome! would probably take too long to spot-check every demo, but the fact that they all built is pretty good imo

KetpuntoG:
> awesome! would probably take too long to spot-check every demo, but the fact that they all built is pretty good imo

I have tested 15 random demos and everything seems to be fine

ikurecic (approved):
Nice! Thanks for checking everything, @KetpuntoG , glad it works with a higher version. :)
This looks good to me!

PR #920, opened 2023-09-08
Fix bug in upload-json where searchindex.js was not being synced to the correct bucket

rashidnhm:
**Title:** Fix bug in upload-json where searchindex.js was not being synced to the correct bucket

**Summary:**
The searchindex.js is currently being synced to the incorrect bucket .. meaning new demos cannot be searched from the qml website. This PR fixes the bug.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/.github/workflows/upload-json.yml
+++ b/.github/workflows/upload-json.yml
@@ -49,6 +49,7 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.aws_access_key_id }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.aws_secret_access_key }}
       AWS_S3_BUCKET_ID: ${{ secrets.aws_json_s3_bucket_id }}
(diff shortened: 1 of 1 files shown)


Mandrenkov (approved):
Is there a way we can test if this change worked before merging to `master`?

doctorperceptron:
> Is there a way we can test if this change worked before merging to `master`?

Good call. @rashidnhm can you manually move the searchindex.js to the correct location and verify you can search new demos in prod?

rashidnhm:
@Mandrenkov @doctorperceptron good call on the verification and sorry for not providing it earlier. The following tests were performed on pennylane dev website.

Searching for the recently released `Contexuality and inductive bias` demo:
image

Moving file to the correct location it will be once this PR is merged:
image

Search starts working:
image


It should also be noted that this workflow was working before. However, during the changes for staging, I moved these steps out to their own workflow and forgot to update that step (mistake on my part).

doctorperceptron (approved):

PR #925, opened 2023-09-13
Update tutorial_qubit_rotation.py

KetpuntoG:
This is a little test in which I have told to a certain large language model to change `numpy` to `jax` in a demo.
 I have chosen qubit rotation for the experiment.


KetpuntoG:
Not interesting right now

PR #927, opened 2023-09-14
Add hardware section to metadata

LucasSilbernagel:
## Related epic
https://app.shortcut.com/xanaduai/epic/45372/users-can-see-example-code-that-demonstrates-how-demos-run-on-real-quantum-hardware-from-aws

## Changes
- Updated the demo metadata schema to include an optional hardware section (only to include AWS for the time being, but can accommodate more vendors in future)
- Added the hardware metadata to 1 demo for testing purposes
- Added AWS logo image asset


ashishks0522 (approved):
LGTM!

ottoak (approved):
LGTM - let's just remember that Amazon will be creating new demos to host on our site to link to, so we'll have to change the demos we add these too for launch.

LucasSilbernagel:
> LGTM - let's just remember that Amazon will be creating new demos to host on our site to link to, so we'll have to change the demos we add these to for launch.

For testing, should I maybe remove the hardware metadata from all but one of the four demos I added it to on this PR?

ottoak:
> > LGTM - let's just remember that Amazon will be creating new demos to host on our site to link to, so we'll have to change the demos we add these to for launch.
> 
> For testing, should I maybe remove the hardware metadata from all but one of the four demos I added it to on this PR?

I think that's a good idea!

demonstrations/tutorial_qubit_rotation.metadata.json
```diff
@@ -39,5 +39,13 @@
             "id": "tutorial_state_preparation",
             "weight": 1.0
         }
+    ],
+    "hardware": [
+        {
+            "id": "aws",
+            "name": "Amazon",
```
doctorperceptron, line 46: What's the distinction between these two? What are they for?

LucasSilbernagel: `id` is used to grab the corresponding data from the CMS. I guess `name` is not used, I can remove that.

doctorperceptron: We should make "id" an enum as well. For now it can only be "aws" but in the future we can add other hardware providers as you said. Making it an enum will prevent people misspelling the id and breaking the front end.

josh146 (approved):
Thanks @LucasSilbernagel!

doctorperceptron (approved):

rashidnhm (approved):

PR #929, opened 2023-09-15
Fix circuit compilation tutorial to use new transforms syntax

mudit2812:
**Summary:**
As title says. `tutorial_circuit_compilation.py` was failing because of deprecated transforms syntax for the transforms used in the demo. I replaced the deprecated syntax with the new API, and it works locally. ([See failure here](https://github.com/PennyLaneAI/qml/actions/runs/6179181147/job/16773713171#step:18:1518))

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**


rmoyard (approved):
:100: Thanks it looks great

KetpuntoG (approved):
Awesome, thanks!

PR #943, opened 2023-10-02
Add block encoding with matrix access oracles demo

soranjh:
**Title:**
Block encoding with matrix access oracles

**Summary:**
This demo explains a general block encoding method that can be very efficient for sparse and structured matrices. It complements block encoding with linear combination of unitaries.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**


demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,117 @@
...
+# ------------------------------
+# Sparse matrices that have specific structures can be efficiently block encoded. A general
+# circuit for block encoding s-sparce matrices can be constructed as
+#
+# Fig 5 of https://arxiv.org/abs/2203.10236
+#
+# The circuit has n qubits and m + 1 ancilla qubits.
+# Ds is defined as HxHxH
+# O_A, O_s can be constructed for structured sparse matrices
+# Let's look at O_A, O_s for the Banded circulant matrix
```
soranjh, line 70: @Jaybsoni , Could you please expand these points?
We also need a small code example.

demonstrations/tutorial_block_encoding.metadata.json
```diff
@@ -0,0 +1,59 @@
...
+    "basedOnPapers": [],
+    "referencedByPapers": [],
+    "relatedContent": [
+        {
+            "type": "demonstration",
+            "id": "tutorial_local_cost_functions",
+            "weight": 1.0
+        }
+    ]
+}
```
Jaybsoni, line 59:
```suggestion
}

```

Jaybsoni:
Added some more theory, just needs to be polished once the CI runs, then I will approve 👍🏼

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+#
+# Summary and conclusions
+# -----------------------
+# Block encoding is a powerful technique in quantum computing that allows us to implement a non-unitary
+# operation in a quantum circuit by embedding the operation in a larger unitary gate.
+# In this demo, we reviewed two important block encoding methods with code examples using PennyLane.
+# The block encoding functionality provided in PennyLane allows you to explore and benchmark several
+# block encoding approaches for a desired problem. The efficiency of the block encoding methods
+# typically depends on the sparsity and structure of the original matrix. We hope that you can use 
+# these tips and tricks to find a more efficient block encoding your matrix!
```
Jaybsoni, line 289:
```suggestion
# these tips and tricks to find a more efficient block encoding for your matrix! 
```

Jaybsoni (approved):
Looks good from my side 👍🏼

KetpuntoG (commented):
Great job team! It's great to see how good the results are 🚀
Here I leave a few comments :)
I missed the "why is it done this way" but I assume that is outside the scope of this tutorial

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+# the FABLE circuit.
+
+def UA(ancilla, wire_l, theta):
+    qml.ctrl(qml.RY, control=wire_l, control_values=[0, 0])(theta[0], wires=ancilla)
+    qml.ctrl(qml.RY, control=wire_l, control_values=[1, 0])(theta[1], wires=ancilla)
+    qml.ctrl(qml.RY, control=wire_l, control_values=[0, 1])(theta[2], wires=ancilla)
+
+##############################################################################
+# The :math:`U_B` oracle is defined in terms of the so-called "Left" and "Right" shift operators
+# ([#sparse]_).
```
Jaybsoni, line 236:
```suggestion
# The :math:`U_B` oracle is defined in terms of the so-called "Left" and "Right" shift operators.
# They correspond to the modular arithmetic operations :math:`+1` or :math:`-1` respectively ([#sparse]_).
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+print(f"Original A:\n{A}", "\n")
+
+##############################################################################
+# The :math:`U_A` oracle for this matrix is constructed from controlled rotation gates, similar to
+# the FABLE circuit.
+
+def UA(ancilla, wire_l, theta):
+    qml.ctrl(qml.RY, control=wire_l, control_values=[0, 0])(theta[0], wires=ancilla)
+    qml.ctrl(qml.RY, control=wire_l, control_values=[1, 0])(theta[1], wires=ancilla)
+    qml.ctrl(qml.RY, control=wire_l, control_values=[0, 1])(theta[2], wires=ancilla)
```
Jaybsoni, line 232:
```suggestion
def UA(ancilla, wire_i, theta):
    qml.ctrl(qml.RY, control=wire_i, control_values=[0, 0])(theta[0], wires=ancilla)
    qml.ctrl(qml.RY, control=wire_i, control_values=[1, 0])(theta[1], wires=ancilla)
    qml.ctrl(qml.RY, control=wire_i, control_values=[0, 1])(theta[2], wires=ancilla)
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+def shift_op(s_wires, shift="Left"):
+    control_values = [1, 1] if shift == "Left" else [0, 0]
+    qml.ctrl(qml.PauliX, control=s_wires[:2], control_values=control_values)(wires=s_wires[2])
+    qml.ctrl(qml.PauliX, control=s_wires[0], control_values=control_values[0])(wires=s_wires[1])
+    qml.PauliX(s_wires[0])
+
+
+def UB(wires_l, wires_j):
+    qml.ctrl(shift_op, control=wires_l[0])(wires_j, shift="Left")
+    qml.ctrl(shift_op, control=wires_l[1])(wires_j, shift="Right")
```
Jaybsoni, line 247:
```suggestion
def UB(wires_i, wires_j):
    qml.ctrl(shift_op, control=wires_i[0])(wires_j, shift="Left")
    qml.ctrl(shift_op, control=wires_i[1])(wires_j, shift="Right")
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+
+##############################################################################
+# We now construct our circuit to block encode the sparse matrix.
+
+dev = qml.device("default.qubit", wires=["ancilla", "l1", "l0", "j2", "j1", "j0"])
```
Jaybsoni, line 252:
```suggestion
dev = qml.device("default.qubit", wires=["ancilla", "i1", "i0", "j2", "j1", "j0"])
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+
+@qml.qnode(dev)
+def complete_circuit(theta):
+    for w in ["l0", "l1"]:  # hadamard transform over |l> register
```
Jaybsoni, line 256:
```suggestion
    for w in ["i0", "i1"]:  # hadamard transform over |i> register
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+        qml.Hadamard(w)
+
+    UA("ancilla", ["l0", "l1"], theta)
+
+    UB(["l0", "l1"], ["j0", "j1", "j2"])
```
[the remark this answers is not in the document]
Jaybsoni: @soranjh or @DSGuala is it possible to reactor the code above for this? Otherwise I can change this for consistency.

Jaybsoni, line 261:
```suggestion
    UA("ancilla", ["i0", "i1"], theta)

    UB(["i0", "i1"], ["j0", "j1", "j2"])
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+
+    for w in ["l0", "l1"]:  # hadamard transform over |l> register
```
Jaybsoni, line 263:
```suggestion
    for w in ["i0", "i1"]:  # hadamard transform over |l> register
```

demonstrations/tutorial_block_encoding.py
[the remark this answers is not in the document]
Jaybsoni: That's the index name they used in the paper, changed it to index "I".

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,316 @@
...
+def UA(ancilla, wire_l, theta):
+    qml.ctrl(qml.RY, control=wire_l, control_values=[0, 0])(theta[0], wires=ancilla)
+    qml.ctrl(qml.RY, control=wire_l, control_values=[1, 0])(theta[1], wires=ancilla)
+    qml.ctrl(qml.RY, control=wire_l, control_values=[0, 1])(theta[2], wires=ancilla)
+
+##############################################################################
+# The :math:`U_B` oracle is defined in terms of the so-called "Left" and "Right" shift operators
+# ([#sparse]_).
+
+def shift_op(s_wires, shift="Left"):
```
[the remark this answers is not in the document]
Jaybsoni: Added 👍🏼

DSGuala (changes requested):
Left several comments. I'm not sure I'm happy with the overall structure of the demo. The first section: `Circuits with matrix access oracles` doesn't connect well to the next section: `Block encoding with FABLE`.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+code = gray_code(2*np.sqrt(len(A)))
+n_selections = len(code)
+
+##############################################################################
+# The next step is to identify and prepare the qubit registers used in the oracle access framework.
+# There are three registers :code:`("ancilla", "wires_i", "wires_j")`: 
+# 
+# The :code:`"ancilla"` register will always contain a single qubit, this is the target where we
+# apply the rotation gates. The :code:`"wires_i"` and :code:`"wires_j"` registers are the same size
+# and need to be able to encode :math:`A` itself, so they will both have :math:`4` qubits for our
```
KetpuntoG, line 85: maybe 2 instead of 4?

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+s = int(np.log2(A.shape[0]))  # number of qubits needed to encode A
+wires_i = [f"i{index}" for index in range(s)]  # depend on the size of A 
+wires_j = [f"j{index}" for index in range(s)]
+
+##############################################################################
+# Finally, we obtain the control wires for the C-NOT gates and a wire map that we later use to
+# translate the :code:`control_wires` into the wire registers we prepared.
+
+control_wires = [int(np.log2(int(code[i], 2) ^ int(code[(i + 1) %
+                 n_selections], 2))) for i in range(n_selections)]
```
KetpuntoG, line 99: I don't understand this line, can it be expressed in another way?

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+
+tolerance= 0.01
+
+def UA(thetas, control_wires, ancilla):
+    for theta, control_index in zip(thetas, control_wires):
+        if abs(theta)>tolerance:
+            qml.RY(theta, wires=ancilla)
+        qml.CNOT(wires=[wire_map[control_index], ancilla])
+
+qml.draw_mpl(circuit, style='pennylane')()
```
KetpuntoG, line 165: at the end of all plots, put plt.show()
(import matplotlib if necessary)
This will prevent unnecessary text from coming out when displaying

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+Circuits with matrix access oracles [[#fable]_, [#sparse]_] can block encode an arbitrary matrix
+:math:`A`. These circuits can be constructed as shown in the figure below.
+
+.. figure:: ../demonstrations/block_encoding/general_circuit.png
+    :width: 50%
+    :align: center
+
+The :math:`U_A` and :math:`U_B` operations are oracles which provide access to the elements of the
+matrix we wish to block encode and the :math:`H^{\otimes n}` operation is a Hadamard transformation
+on :math:`n` qubits. Finding the optimal sequence of the quantum gates that implement :math:`U_A`
```
KetpuntoG, line 35: Is it possible to explain super quickly what each one does? I understand that the part of b(i,j) has been removed because it was confusing but maybe you can comment on the idea.

I haven't read the paper so maybe this is wrong but I imagine something like:

U_A is able to do the necessary rotation in the auxiliary qubit for a specific i,j coordinates of the matrix.

U_B selects exactly which elements are those we want to encode. This oracle is important since it is not always necessary to encode all the elements and in sparse matrices it will play will considerably boost the algorithm.

For more information on its construction, I invite you to read the original paper.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+# translate the :code:`control_wires` into the wire registers we prepared.
+
+control_wires = [int(np.log2(int(code[i], 2) ^ int(code[(i + 1) %
+                 n_selections], 2))) for i in range(n_selections)]
+
+wire_map = {control_index : wire for control_index, wire in enumerate(wires_j + wires_i)}
+
+##############################################################################
+# We now construct the :math:`U_A` and :math:`U_B` oracles as well as the operator representing the
+# tensor product of Hadamard gates. Note that :math:`U_B` in FABLE is constructed as a set of SWAP
```
KetpuntoG, line 105: I think something like this would be better:

Note that in the particular case where we have no non-zero elements in the matrix [...] as a set of SWAPs. I invite you to think about why this is so.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+# compare it with the original matrix.
+
+print(f"Original matrix:\n{A}", "\n")
+wire_order = [ancilla_wire] + wires_i[::-1] + wires_j[::-1] 
+M = len(A) * qml.matrix(circuit, wire_order=wire_order)().real[0:len(A),0:len(A)]
+print(f"Block-encoded matrix:\n{M}", "\n")
+
+##############################################################################
+# You can easily confirm that the circuit block encodes the original matrix defined above. Note that
+# the dimension of :math:`A` should be :math:`2^n, n = 1, 2, 3, ...`. For matrices with an arbitrary
```
KetpuntoG, line 148: It renders a little strange that n = 1,2,3. I think it would be better to eliminate it or say for n an integer.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+    for theta, control_index in zip(thetas, control_wires):
+        if abs(theta)>tolerance:
+            qml.RY(theta, wires=ancilla)
+        qml.CNOT(wires=[wire_map[control_index], ancilla])
+
+qml.draw_mpl(circuit, style='pennylane')()
+
+##############################################################################
+# Compressing the circuit by removing some of the rotations is an approximation. We can now see how
+# good this approximation is in the case of our example.
```
KetpuntoG, line 169: It would be good to clarify what is going to be done so as not to scare the reader. Something like:

We will make a small modification to UA so that it removes those CNOTs that, as we have seen, we do not need

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+# The :code:`"ancilla"` register will still contain a single qubit, the target where for the
+# controlled rotation gates. The :code:`"wires_i"` register needs to be large enough to binary
+# encode the maximum number of non-zero entries in any column or row. Given the structure of
+# :math:`A` defined above, we have at most 3 non-zero entries, thus this register will have 2
+# qubits. Finally, the :code:`"wires_j"` register will be used to encode :math:`A` itself, so it
+# will have 3 qubits. We prepare the wires below:
+
+s = int(np.log2(A.shape[0]))  # number of qubits needed to encode A
+
+ancilla_wires = ["ancilla"]   # always 1 qubit for controlled rotations
```
KetpuntoG, line 244: In the previous example it was `ancilla_wire = "ancilla" ` and it was pass in the device as [ancilla_wire]. I like the current notation better, so I would change the above notation to consistency.

KetpuntoG (commented):
A big difference from the last time I reviewed it 💯 
Much clearer, and the drawings look great in PL style 🚀
Here are some suggestions

demonstrations/tutorial_block_encoding.py
Jaybsoni: natural number :)

demonstrations/tutorial_block_encoding.py
Jaybsoni: Yes correct!

Jaybsoni, line 85:
```suggestion
# and need to be able to encode :math:`A` itself, so they will both have :math:`2` qubits for our
```

demonstrations/tutorial_block_encoding.py
Jaybsoni: Fixed the previous case to use `ancilla_wires = ["ancilla"]` as well!

demonstrations/tutorial_block_encoding.py
Jaybsoni: Add plt.show()

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+that can be very efficient for matrices with specific sparsity and structure.
+
+Block encoding with FABLE
+-------------------------
+The Fast Approximate BLock Encodings (FABLE) technique is a general method for block encoding dense
+and sparse matrices [#fable]_. The level of approximation in FABLE can be adjusted to simplify the
+resulting circuit. For matrices with specific structures, FABLE provides an efficient circuit
+without reducing accuracy.
+
+The FABLE circuit is constructed from a set of rotation and C-NOT gates. The rotation angles,
```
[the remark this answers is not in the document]
Jaybsoni: Added 👍🏼

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,340 @@
...
+thetas = compute_theta(alphas)
+
+code = gray_code(2*np.sqrt(len(A)))
+n_selections = len(code)
+
+##############################################################################
+# The next step is to identify and prepare the qubit registers used in the oracle access framework.
+# There are three registers :code:`("ancilla", "wires_i", "wires_j")`: 
+# 
+# The :code:`"ancilla"` register will always contain a single qubit, this is the target where we
```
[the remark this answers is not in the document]
Jaybsoni: Added 👍🏼

demonstrations/tutorial_block_encoding.py
Jaybsoni: Added 👍🏼

demonstrations/tutorial_block_encoding.py
Jaybsoni: Added 👍🏼

Jaybsoni:
@KetpuntoG, I addressed some of your comments and added a thumbnail. We also have a large thumbnail file, which I am not sure where it goes, please add it into the right place 👍🏼

demonstrations/tutorial_block_encoding.py
DSGuala: This is copied from [MottonenStatePreparation](https://github.com/PennyLaneAI/pennylane/blob/3dee10187e0698d521daa841e5d146e92a50f343/pennylane/templates/state_preparations/mottonen.py#L133).

I don't think we know a better way to express it. It creates the following kind of patterns:
[0, 1, 0, 2, 0, 1, 0, 2]
[0, 1, 0, 2, 0, 1, 0, 3, 0, 1, 0, 2, 0, 1, 0, 3]

I'm thinking of another way to express it, but it's difficult.

soranjh: If possible, we can add 1-2 sentences to briefly explain what this does.

KetpuntoG (approved):
A great improvement! Everything much clearer and cleaner than the first review, well done! 💯 I did a final commit with some [smalls modifications](https://github.com/PennyLaneAI/qml/pull/943/commits/aadd90e2ddd7bfa4b5ffb9ac11bdc82b9cdce381)
It's ready for me 😄 (but don't merge it)

demonstrations/block_encoding/general_circuit.png
Jaybsoni, line 1: @DSGuala, could you update this image by adding a subscript "anc" to the |0> qubit? So it looks something like:  `|0>_anc` I refer to it as an ancilla qubit and denote it as such in the math sections.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,415 @@
...
+
+control_wires = [int(np.log2(int(code[i], 2) ^ int(code[(i + 1) %
+                 n_selections], 2))) for i in range(n_selections)]
+
+wire_map = {control_index : wire for control_index, wire in enumerate(wires_j + wires_i)}
+
+##############################################################################
+# We now construct the :math:`U_A` and :math:`U_B` oracles as well as the operator representing the
+# tensor product of Hadamard gates. Note in the case where we have no non-zero elements in the matrix
+# then :math:`U_B` in FABLE is constructed as a set of SWAP gates. We invite you to think about why this is so.
```
KetpuntoG, line 132: the invitation to deduce the SWAP was put in the previous review because b(i,j) was previously defined. Now that it's not there, it's not something I think the user can infer so you can remove it.

KetpuntoG:
From what we have discussed in person, the confusion may come since you don't see the clear separation between the two algorithms. It may seem that the spare matrix case is a particular case of Fable but they work differently. For that reason, I would clarify it in the introduction. After showing the structure of this type of encodings, say:


"In this demo we are going to work with two particular algorithms based on oracles: the FABLE method and another method that focuses on sparse matrices."

So here you already start talking about FABLE, you talk about UA with the formulas and say that we will always take the same size |i> and |j>, representing the row and column indexes (I think). Remark that in this **algorithm** you always take UB as a SWAP gate whose goal is to order the indices in the right way. 


Then you get to the sparse matrix block. Here you start by saying that this is a different method that seeks to optimize the algorithm for situations where the matrix is sparse. You also say something like "Unlike the FABLE algorithm, in this case the |j> register will not store the column "j" but will store b(i,j). b(i, j) means [...]. In this way we can greatly improve the efficiency since the number of qubits will not grow with the size of the matrix but with the number of non-empty entries. With this in mind, note that we will find a small difference in the construction of UA, we will simply have to adapt the indexes to the new definition:


[formula]

On the other hand UB is in charge of returning precisely that b(i,j) value we were commenting:

[formula]

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,415 @@
...
+
+Where the :math:`H^{\otimes n}` operation is a Hadamard transformation on :math:`n` qubits. The
+:math:`U_A` and :math:`U_B` operations are oracles which accomplish a specific task, their form
+depends on the matrix we wish to block encode. See ([#fable]_, [#sparse]_) for an explanation 
+on the action of these oracles.
+
+Where the :math:`H^{\otimes n}` operation is a Hadamard transformation on :math:`n` qubits. The
+:math:`U_A` operation is an oracle which encodes the matrix element :math:`A_{i,j}` into the the 
+amplitude of the ancilla qubit. The :math:`U_B` oracle ensures that we iterate over every 
+combination of :math:`(i,j)`.
```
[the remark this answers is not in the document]
Jaybsoni: Ahh, meant to remove one. Nice catch!

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,415 @@
...
+    UA(thetas, control_wires, ancilla_wires)
+    
+    qml.Barrier()
+    
+    UB(wires_i, wires_j)
+    
+    qml.Barrier()
+
+    HN(wires_i)
+    return qml.probs(wires=ancilla_wires + wires_i)
```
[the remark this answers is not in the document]
Jaybsoni: We could remove it since we don't use the output of the qnode. I personally think it's nice to keep the measurements in because in the paper, we need to measure out those qubits to post select the states that ensure we correctly block-encoded. (ancilla_wires + wires_i need to be in the all |0> state to access A).

demonstrations/tutorial_block_encoding.py
Jaybsoni: Removed it.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,415 @@
...
+
+alphas = np.arccos(A).flatten()
+thetas = compute_theta(alphas)
+
+##############################################################################
+# The next step is to identify and prepare the qubit registers used in the oracle access framework.
+# There are three registers :code:`"ancilla"`, :code:`"wires_i"`, :code:`"wires_j"`. The
+# :code:`"ancilla"` register will always contain a single qubit, this is the auxillary qubit where we
+# apply the rotation gates methioned above. The :code:`"wires_i"` and :code:`"wires_j"` registers are 
+# the same size and need to be able to encode :math:`A` itself, so they will both have :math:`2` qubits
```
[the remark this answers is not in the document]
Jaybsoni: Added 👍🏼

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,415 @@
...
+this case simplifies to be the :math:`SWAP` gate:
+
+.. math:: U_B |i\rangle|j\rangle \ = \ |j\rangle |i\rangle
+
+The FABLE circuit is constructed from a set of rotation and C-NOT gates. The rotation angles,
+:math:`(\theta_1, ..., \theta_n)`, are obtained from a transformation of the elements of
+the block-encoded matrix.
+
+.. math:: \begin{pmatrix} \theta_1 \\ \cdots \\ \theta_n \end{pmatrix} =
+          M \begin{pmatrix} \alpha_1 \\ \cdots \\ \alpha_n \end{pmatrix},
```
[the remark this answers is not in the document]
Jaybsoni: Added 👍🏼

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,415 @@
...
+where
+
+.. math::
+
+    |A_{i,j}\rangle_{\text{anc}} \equiv A_{i,j}|0\rangle_{\text{anc}} + \sqrt{1 - |A_{i,j}|^2}|1\rangle_{\text{anc}}.
+
+:math:`U_A`, in the most general case, can be constructed from a sequence of uniformly controlled 
+rotation gates with rotation angles computed as :math:`\theta = \text{arccos}(A_{i,j})`. The 
+:math:`U_B` oracle is responsible for ensuring proper indexing of each entry in :math:`A` and for
+this case simplifies to be the :math:`SWAP` gate:
```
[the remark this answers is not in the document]
Jaybsoni: I added the link to the class :class:`~.pennylane.SWAP`

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+The :math:`U_B` oracle is responsible for ensuring proper indexing of each entry in :math:`A` 
+and for this algorithm, it simplifies to be the :class:`~.pennylane.SWAP` gate:
+
+.. math:: U_B |i\rangle|j\rangle \ = \ |j\rangle |i\rangle
+
+The naive approach is to construct :math:`U_A` using a sequence of multi-controlled :math:`Ry(\alpha)` 
+rotation gates with rotation angles computed as :math:`\alpha = \text{arccos}(A_{i,j})`. It turns out
+that this requires :math:`O(N^{4})` gates and is very inefficient. A more efficient approach is a 
+
+The Fast Approximate BLock Encodings (FABLE) technique is a method for block encoding matrices [#fable]_
```
KetpuntoG, line 65: Check this sentence

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+# We also compute the rotation angles.
+
+alphas = np.arccos(A).flatten()
+thetas = compute_theta(alphas)
+
+##############################################################################
+# The next step is to identify and prepare the qubit registers used in the oracle access framework.
+# There are three registers :code:`"ancilla"`, :code:`"wires_i"`, :code:`"wires_j"`. The
+# :code:`"ancilla"` register will always contain a single qubit, this is the auxillary qubit where we
+# apply the rotation gates methioned above. The :code:`"wires_i"` and :code:`"wires_j"` registers are
```
KetpuntoG, line 106:
```suggestion
# apply the rotation gates mentioned above. The :code:`"wires_i"` and :code:`"wires_j"` registers are 
```

demonstrations/tutorial_block_encoding.py
Jaybsoni: Updated

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+
+Finding an optimal quantum gate decomposition that implements :math:`U_A` and :math:`U_B` is not 
+always possible. We now explore two different approaches for constructing these oracles that can be 
+very efficient for matrices with specific structure or sparsity.
+
+Block encoding structured matricies 
+-----------------------------------
+In order to better understand the oracle access framework let us first define :math:`U_A` and :math:`U_B`
+For the exact block-encoding of :math:`A`. The :math:`U_A` oracle is responsible for encoding the 
+matrix entries of :math:`A` into the amplitude of an auxillary qubit :math:`|0\rangle_{\text{anc}}`:
```
KetpuntoG, line 44:
```suggestion
matrix entries of :math:`A` into the amplitude of an auxilary qubit :math:`|0\rangle_{\text{anc}}`:
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+##############################################################################
+# We also compute the rotation angles.
+
+alphas = np.arccos(A).flatten()
+thetas = compute_theta(alphas)
+
+##############################################################################
+# The next step is to identify and prepare the qubit registers used in the oracle access framework.
+# There are three registers :code:`"ancilla"`, :code:`"wires_i"`, :code:`"wires_j"`. The
+# :code:`"ancilla"` register will always contain a single qubit, this is the auxillary qubit where we
```
KetpuntoG, line 105:
```suggestion
# :code:`"ancilla"` register will always contain a single qubit, this is the auxilary qubit where we
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+
+##############################################################################
+# We construct the circuit using these oracles and draw it.
+
+dev = qml.device('default.qubit', wires=ancilla_wires + wires_i + wires_j)
+@qml.qnode(dev)
+def circuit():
+    HN(wires_i)
+
+    qml.Barrier()  # to seperate the sections in the circuit
```
KetpuntoG, line 155:
```suggestion
    qml.Barrier()  # to separate the sections in the circuit 
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+#
+#     U_A |0\rangle_{\text{anc}} |i\rangle |j\rangle = |A_{b(i,j),j}\rangle_{\text{anc}} |i\rangle |j\rangle,
+#
+# where
+#
+# .. math::
+#
+#     |A_{l,j}\rangle_{\text{anc}} \equiv A_{l,j}|0\rangle_{\text{anc}} + \sqrt{1 - A_{l,j}^2}|1\rangle_{\text{anc}}.
+#
+# In this case the :math:`U_B` oracle is responsible for implmenting the :math:`b(i,j)` function
```
KetpuntoG, line 265:
```suggestion
# In this case the :math:`U_B` oracle is responsible for implementing the :math:`b(i,j)` function 
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+# .. math::
+#
+#     |A_{l,j}\rangle_{\text{anc}} \equiv A_{l,j}|0\rangle_{\text{anc}} + \sqrt{1 - A_{l,j}^2}|1\rangle_{\text{anc}}.
+#
+# In this case the :math:`U_B` oracle is responsible for implmenting the :math:`b(i,j)` function 
+# and taking us from the column index to the row index in the qubit register:
+#
+# .. math:: U_B |i\rangle|j\rangle \ = \ |i\rangle |b(i,j)\rangle
+#
+# Lets work through an example. Consider the sparse matrix given by:
```
KetpuntoG, line 270:
```suggestion
# Let's work through an example. Consider the sparse matrix given by:
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+combination of :math:`(i,j)`.
+
+Finding an optimal quantum gate decomposition that implements :math:`U_A` and :math:`U_B` is not 
+always possible. We now explore two different approaches for constructing these oracles that can be 
+very efficient for matrices with specific structure or sparsity.
+
+Block encoding structured matricies 
+-----------------------------------
+In order to better understand the oracle access framework let us first define :math:`U_A` and :math:`U_B`
+For the exact block-encoding of :math:`A`. The :math:`U_A` oracle is responsible for encoding the
```
KetpuntoG, line 43:
```suggestion
for the exact block-encoding of :math:`A`. The :math:`U_A` oracle is responsible for encoding the 
```

is this change correct, right?

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+The :math:`U_B` oracle is responsible for ensuring proper indexing of each entry in :math:`A` 
+and for this algorithm, it simplifies to be the :class:`~.pennylane.SWAP` gate:
+
+.. math:: U_B |i\rangle|j\rangle \ = \ |j\rangle |i\rangle
+
+The naive approach is to construct :math:`U_A` using a sequence of multi-controlled :math:`Ry(\alpha)` 
+rotation gates with rotation angles computed as :math:`\alpha = \text{arccos}(A_{i,j})`. It turns out
+that this requires :math:`O(N^{4})` gates and is very inefficient. A more efficient approach is a 
+
+The Fast Approximate BLock Encoding (FABLE) technique is a method for block encoding matrices [#fable]_
```
KetpuntoG, line 65: I think you have not change this, make sure you pushed all your changes

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+the need for multi-controlled rotations using a special transformation of the angles (see [#fable]_ for details).
+The rotation angles, :math:`(\theta_1, ..., \theta_n)`, are obtained from a transformation of the elements 
+of the block-encoded matrix.
+
+.. math:: \begin{pmatrix} \theta_1 \\ \cdots \\ \theta_n \end{pmatrix} =
+          M \begin{pmatrix} \alpha_1 \\ \cdots \\ \alpha_n \end{pmatrix}.
+
+The angles :math:`\alpha` are obtained from the matrix elements of the matrix :math:`A` as
+:math:`\alpha_1 = \text{arccos}(A_{00}), ...,` and :math:`M` is the transformation matrix that can
+be obtained with the :func:`~.pennylane.templates.state_preparations.mottonen.compute_theta()`
```
KetpuntoG, line 80:
```suggestion
be obtained with the :func:`~.pennylane.templates.state_preparations.mottonen.compute_theta`
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+              [    0,     0,  beta, alpha, gamma,     0,     0,     0],
+              [    0,     0,     0,  beta, alpha, gamma,     0,     0],
+              [    0,     0,     0,     0,  beta, alpha, gamma,     0],
+              [    0,     0,     0,     0,     0,  beta, alpha, gamma],
+              [gamma,     0,     0,     0,     0,     0,  beta, alpha]])
+
+print(f"Original A:\n{A}", "\n")
+
+##############################################################################
+# Once again we identify and prepare the qubit registers used in the oracle access framework:
```
KetpuntoG, line 297:
```suggestion
# Once again we identify and prepare the qubit registers used in the oracle access framework.
```

KetpuntoG (commented):
Just some typos :)

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+Where the :math:`H^{\otimes n}` operation is a Hadamard transformation on :math:`n` qubits. The
+:math:`U_A` operation is an oracle which encodes the matrix element :math:`A_{i,j}` into the the 
+amplitude of the ancilla qubit. The :math:`U_B` oracle ensures that we iterate over every 
+combination of :math:`(i,j)`.
+
+Finding an optimal quantum gate decomposition that implements :math:`U_A` and :math:`U_B` is not 
+always possible. We now explore two different approaches for constructing these oracles that can be 
+very efficient for matrices with specific structure or sparsity.
+
+Block encoding structured matricies
```
soranjh, line 40:
```suggestion
Block encoding structured matrices 
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+amplitude of the ancilla qubit. The :math:`U_B` oracle ensures that we iterate over every 
+combination of :math:`(i,j)`.
+
+Finding an optimal quantum gate decomposition that implements :math:`U_A` and :math:`U_B` is not 
+always possible. We now explore two different approaches for constructing these oracles that can be 
+very efficient for matrices with specific structure or sparsity.
+
+Block encoding structured matricies 
+-----------------------------------
+In order to better understand the oracle access framework let us first define :math:`U_A` and :math:`U_B`
```
soranjh, line 42:
```suggestion
In order to better understand the oracle access framework, let us first define :math:`U_A` and :math:`U_B`
```

demonstrations/tutorial_block_encoding.py
soranjh: The link to compute_theta does not work for some reason. Maybe use " ... obtained with the compute_theta() function in PennyLane."

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+
+ancilla_wires = ["ancilla"]
+
+s = int(np.log2(A.shape[0]))
+wires_i = [f"i{index}" for index in range(s)]
+wires_j = [f"j{index}" for index in range(s)]
+
+##############################################################################
+# Finally, we obtain the control wires for the C-NOT gates and a wire map that we later use to
+# translate the :code:`control_wires` into the wire registers we prepared.
```
soranjh, line 118:
```suggestion
# translate the control wires into the wire registers we prepared.
```

soranjh: [optional suggestion]

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+
+dev = qml.device('default.qubit', wires=ancilla_wires + wires_i + wires_j)
+@qml.qnode(dev)
+def circuit():
+    HN(wires_i)
+
+    qml.Barrier()  # to seperate the sections in the circuit 
+
+    UA(thetas, control_wires, ancilla_wires)
+
```
soranjh, line 158: Please consider removing the empty lines for a more compact code.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+# circuit representation for :math:`U_A` and :math:`U_B` [#sparse]_ by only keeping track of the 
+# non-zero entries of the matrix. 
+#
+# Let :math:`b(i,j)` be a function such that it takes a column index :math:`j` and returns the
+# row index for the :math:`i^{th}` non-zero entry in that column of :math:`A`. Note, in this 
+# formulation, the :math:`|i\rangle` qubit register now refers to the number of non-zero entries 
+# in :math:`A`. For sparse matrices, this can be much smaller than :math:`N`, thus saving us many
+# qubits. We use this to define :math:`U_A` and :math:`U_B`.
+#
+# Like in the structured approach the :math:`U_A` oracle is responsible for encoding the matrix
```
soranjh, line 251:
```suggestion
# Like in the structured approach, the :math:`U_A` oracle is responsible for encoding the matrix 
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+              [    0,     0,     0,     0,  beta, alpha, gamma,     0],
+              [    0,     0,     0,     0,     0,  beta, alpha, gamma],
+              [gamma,     0,     0,     0,     0,     0,  beta, alpha]])
+
+print(f"Original A:\n{A}", "\n")
+
+##############################################################################
+# Once again we identify and prepare the qubit registers used in the oracle access framework:
+# 
+# The :code:`"ancilla"` register will still contain a single qubit, the target where for the
```
soranjh, line 299:
```suggestion
# The :code:`"ancilla"` register will still contain a single qubit, the target for the
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+
+##############################################################################
+# We now construct our circuit to block encode the sparse matrix and draw it.
+
+dev = qml.device("default.qubit", wires=(ancilla_wires + wires_i + wires_j))
+
+@qml.qnode(dev)
+def complete_circuit(thetas):
+    HN(wires_i)
+
```
soranjh, line 344: Might consider removing the blank lines to make the circuit compact.

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,413 @@
...
+and for this algorithm, it simplifies to be the :class:`~.pennylane.SWAP` gate:
+
+.. math:: U_B |i\rangle|j\rangle \ = \ |j\rangle |i\rangle
+
+The naive approach is to construct :math:`U_A` using a sequence of multi-controlled :math:`Ry(\alpha)` 
+rotation gates with rotation angles computed as :math:`\alpha = \text{arccos}(A_{i,j})`. It turns out
+that this requires :math:`O(N^{4})` gates and is very inefficient. A more efficient approach is a 
+
+The Fast Approximate BLock Encoding (FABLE) technique is a method for block encoding matrices [#fable]_
+using the oracle access framework and some clever approximations 🧠. The level of approximation in FABLE
```
soranjh, line 66: Note that the original paper uses "Encodings".

soranjh, line 66:
```suggestion
that this requires :math:`O(N^{4})` gates and is very inefficient. A more efficient approach is the 
Fast Approximate BLock Encodings (FABLE) technique [#fable]_, which
uses the oracle access framework and some clever approximations 🧠. The level of approximation in FABLE 
```

demonstrations/tutorial_block_encoding.py
```diff
@@ -0,0 +1,400 @@
...
+#
+# .. math::
+#
+#     U_A |0\rangle_{\text{anc}} |i\rangle |j\rangle = |A_{b(i,j),j}\rangle_{\text{anc}} |i\rangle |j\rangle,
+#
+# where
+#
+# .. math::
+#
+#     |A_{l,j}\rangle_{\text{anc}} \equiv A_{l,j}|0\rangle_{\text{anc}} + \sqrt{1 - A_{l,j}^2}|1\rangle_{\text{anc}}.
```
Jaybsoni, line 256:
```suggestion
```

KetpuntoG:
Great Job @Jaybsoni ! It was not an easy topic but now it's clear 🚀
You can have the honors of clicking on the merge button 😉

PR #946, opened 2023-10-04
Minor update "Computing gradients in parallel with Amazon Braket"

mbeach-aws:
you no longer need to explicitly specify a S3 bucket, so removing that from the demo to simplify.


KetpuntoG:
I can close this PR, right? @mbeach-aws

PR #949, opened 2023-10-05
Fix rosalin demo

albi3ro:
Still not entirely sure why the problem with the shots was occuring, but this fixes it.


KetpuntoG (approved):
Thanks for work on this!

isaacdevlugt (approved):
LGTM! Just noting that matplotlib is yelling twice

Screenshot 2023-10-05 at 4 28 59 PM

KetpuntoG:
> LGTM! Just noting that matplotlib is yelling twice
> 
> Screenshot 2023-10-05 at 4 28 59 PM

Yep, I'll take a look to silence that in all demos that uses seaborn

PR #952, opened 2023-10-10
Fixing 3 demos

KetpuntoG:
`dev.state` does not work with the new device:
`variational_classifier`
`tutorial_noisy_circuits`

Typo:
`circuit_cutting`


demonstrations/tutorial_noisy_circuits.py
```diff
@@ -75,12 +75,18 @@ def circuit():
 print(f"QNode output = {circuit():.4f}")
 
 ######################################################################
-# The device stores the output state as a density matrix. In this case, the density matrix is
+# With a small modification of the circuit we can also ask for the density matrix. In this case, the density matrix is
 # equal to :math:`|\psi\rangle\langle\psi|`,
 # where :math:`|\psi\rangle=\frac{1}{\sqrt{2}}(|00\rangle + |11\rangle)`.
 
+@qml.qnode(dev)
+def circuit():
```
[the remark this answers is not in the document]
KetpuntoG: Sure, done :)

demonstrations/tutorial_noisy_circuits.py
```diff
@@ -75,12 +75,18 @@ def circuit():
 print(f"QNode output = {circuit():.4f}")
 
 ######################################################################
-# The device stores the output state as a density matrix. In this case, the density matrix is
+# With a small modification of the circuit we can also ask for the density matrix. In this case, the density matrix is
 # equal to :math:`|\psi\rangle\langle\psi|`,
 # where :math:`|\psi\rangle=\frac{1}{\sqrt{2}}(|00\rangle + |11\rangle)`.
 
+@qml.qnode(dev)
+def density_matrix_circuit():
```
albi3ro, line 83: Oh sorry I read it wrong. I thought we were returning the density matrix.  

How about `state_circuit` instead? I think its confusing if we say "density matrix" and then return the state instead.

KetpuntoG: It is the density matrix because `dev = "default.mixed"`. I put `state = density_matrix()` because previously in the demo it was `dev.state`. But maybe I can just put density matrix everywhere

albi3ro: It might be clearer. I've always found it a bit confusing how `state` is sometimes the density matrix and sometimes not.

albi3ro (approved):

trbromley (approved):

PR #963, opened 2023-10-25
Minor fixes for version v0.33

albi3ro:
In version v0.33, we ported transforms over to a new syntax. This requires a few syntax changes in the demo.

We also switched from the old default qubit to the new one. The old device used the global random number generator each time it sampled. The new device will query the global random number generator once upon initialization if a seed is not specified. This caused the global random number generator to be in a different state when creating parameters.

In order to avoid coupling between random number generation in different parts of the workflow in the future, I switched to using local random number generation for parameters and setting a seed explicitly on the device. Future changes to random generation shouldn't have as much impact on the output of the demos.

A couple of deprecation warnings from matplotlib and torch also forced a couple of updates.


demonstrations/tutorial_backprop.py
```diff
@@ -84,7 +82,8 @@ def circuit(params):
 # Let's test the variational circuit evaluation with some parameter input:
 
 # initial parameters
-params = np.random.random([6], requires_grad=True)
+rng = np.random.default_rng(1949320)
```
timmysilv, line 85: I thought there was a message here. spent like 5 minutes trying to figure out who might have been born on 1949 march 20 (didn't find much unless you are a fan of Carl Palmer, drummer of Emerson, Lake and Palmer?)

demonstrations/tutorial_quantum_circuit_cutting.py
```diff
@@ -534,11 +536,13 @@ def qaoa_template(params):
...
 
-with QuantumTape() as tape:
+with qml.queuing.AnnotatedQueue() as q:
     qaoa_template(optimal_params)
     qml.sample(wires=all_wires)
 
-fig, _ = qml.drawer.tape_mpl(tape, expansion_strategy="device")
+tape = QuantumTape.from_queue(q)
+
+fig, _ = qml.drawer.tape_mpl(tape)
```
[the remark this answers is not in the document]
albi3ro: That keyword argument never existed on that function. It just happened to not raise an error. Maybe we should actually have keyword argument validation on that function.

demonstrations/tutorial_backprop.py
albi3ro: Keyboard mashing 😆

timmysilv (approved):

lillian542 (approved):

mudit2812 (approved):

PR #964, opened 2023-10-26
Update braket parallel gradients

lillian542:
Update times in demo to reflect speed-ups in both local and remote devices


lillian542:
[sc-48590]

demonstrations/braket-parallel-gradients.py
```diff
@@ -259,9 +259,9 @@ def circuit(params):
...
 #  .. code-block:: none
 #
-#      Gradient calculation time on local device (seconds): 941.8518133479993
+#      Gradient calculation time on local device (seconds): 102.44530320167542
 #
-# Wow, the local device needs around 15 minutes or more! Compare this to less than a minute spent
+# Wow, the local device needs around a couple of minutes! Compare this to a few seconds spent
 # calculating the gradient on SV1. This provides a powerful lesson in parallelization.
 #
 # What if we had run on SV1 with ``parallel=False``? It would have taken around 3 minutes—still
```
timmysilv, line 267: is this still true?

timmysilv (approved):
nice! except for that one slowdown... but it's no biggie 😄

demonstrations/braket-parallel-gradients.py
lillian542: Good question. Now I've checked, and no, not at all. 

Actually, with `parallel=False`, it takes around 7 seconds (compared to around 7 for `parallel=True`). The iterations below are also all less than 10 seconds even if I set `parallel=False` and remove the `max_parallel=20` kwarg. 

So maybe this demo doesn't actually demonstrate what it says it does anymore. It just seems to show that SV1 is faster than `default.qubit`. It doesn't seem to show that setting `parallel=True` makes things faster.

timmysilv: welp I regret asking now... idk what to do

PR #965, opened 2023-10-26
Fix QNSPSA demo

mudit2812:
Replace `op.adjoint()` with `qml.adjoint(op)`.
Got error about braket job name being taken, so changed name of job.


timmysilv (approved):

lillian542 (approved):

demonstrations/qnspsa.py
```diff
@@ -374,7 +374,7 @@ def get_overlap_tape(qnode, params1, params2):
         for op in op_forward:
             qml.apply(op)
         for op in reversed(op_inv):
-            qml.adjoint(copy(op))
```
albi3ro, line 377: Think you do need the copy. Otherwise it will remove `adjoint` will remove `op` from the queue.  Was this causing problems?

demonstrations/qnspsa.py
```diff
@@ -874,7 +873,7 @@ def __get_overlap_tape(self, cost, params1, params2):
             for op in op_forward:
                 qml.apply(op)
             for op in reversed(op_inv):
-                op.adjoint()
+                qml.adjoint(op)
```
albi3ro, line 876: As above, we want to be careful that we aren't removing `op` from the queue.  might be worth even just doing:
```
tape = qml.tape.QuantumTape(op_forward + [qml.adjoint(op) for op in reversed(op_inv)], [qml.probs(wires=cost.tape.wires)])
```

albi3ro (approved):
queuing be confusing. 👍

PR #966, opened 2023-10-26
fix parallel demo

albi3ro:
This update to the `vqe_parallel` demo:

* Changes the plot so that atomic separation is actually atomic separation
* removes use of non-existant `data` variable`
* Updates how to combine coefficients are results


demonstrations/vqe_parallel.py
```diff
@@ -273,48 +265,50 @@ def compute_energy_parallel_optimized(H, devs, param):
...
+#        8 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 1.7 Å
+#        9 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 1.9 Å
+#        10 / 10: Parallel execution and measurement optimization; Running for inter-atomic distance 2.1 Å
+#        Evaluation time: 78.62 s
 
 
 ##############################################################################
 # We have seen how Hamiltonian measurements can be parallelized and optimized at the same time.
+#
+# While parallelism did not provide any speed up in this circumstance, using dask with fewer available cores
```
timmysilv, line 309: this "while" makes it seem like you're going to explain when parallelism will provide a speed-up, so I'd just flip the language in the second part

timmysilv (commented):
just one concern, but it seems good otherwise!

demonstrations/vqe_parallel.py
```diff
@@ -273,48 +265,50 @@ def compute_energy_parallel_optimized(H, devs, param):
...
+# or a simulator that already uses multithreading can oversubscribe the machine, resulting in a net slowdown.
```
albi3ro, line 310:
```suggestion
# Using dask with fewer available cores or a simulator that already uses multithreading can oversubscribe the
# machine, resulting in a net slowdown. That may be occurring in this situation.
# 
```

demonstrations/vqe_parallel/vqe_parallel_001.png
josh146, line 1: @albi3ro how come the plot has changed?

albi3ro: For the old plot, the "bond length" was actually the index. Now it is the bond length. Two lines were also exactly the same color. And i added a line for the energy from the dataset.

demonstrations/vqe_parallel.py
```diff
@@ -332,9 +332,9 @@ def compute_energy_parallel_optimized(H, devs, param):
 
 np.savez("vqe_parallel", energies_seq=energies_seq, energies_par=energies_par, energies_par_opt=energies_par_opt)
 
-plt.plot(energies_seq, linewidth=2.2, marker="d", color="blue", label="sequential")
-plt.plot(energies_par, linewidth=2.2, marker="o", color="red", label="parallel")
-plt.plot(energies_par_opt, linewidth=2.2, marker="d", color="blue", label="paralell and optimized")
+plt.plot(bonds, energies_seq, linewidth=2.2, marker="d", color="green", label="sequential")
+plt.plot(bonds, energies_par, linewidth=2.2, marker="o", color="red", label="parallel")
+plt.plot(bonds, energies_par_opt, linewidth=2.2, marker="d", color="blue", label="paralell and optimized")
```
trbromley, line 337:
```suggestion
plt.plot(bonds, energies_par_opt, linewidth=2.2, marker="d", color="blue", label="parallel and optimized")
```

mlxd:
Validated locally on QVM/Quilc server. The only line that needs updating is `result = H.coeffs @ dask.compute(*results, scheduler="threads")`. Commit to follow.

mudit2812 (approved):

demonstrations/vqe_parallel.py
```diff
@@ -254,12 +264,17 @@ def compute_energy_parallel_optimized(H, devs, param):
     result = qml.math.sum(dask.compute(*results, scheduler="threads"))
     return result
 
-print("Evaluating the potential energy surface in parallel with measurement optimization")
+
```
trbromley, line 267:
```suggestion
```

trbromley (approved):

PR #970, opened 2023-10-31
Add how-to's on JAX optimization

josh146:
**Context**

The PennyLane demos have a lot of 'explanation' content -- good for when you are learning, and for understanding new theory and its context (alongside code). However, we are lacking in 'how to' content.

How-to content is a bit different from a tutorial (teaching someone the basics) or a demo (teaching concepts and understanding). If you think of a tutorial as a 'Basic cooking lesson', and a demo as a 'Book on the history of cooking', you can think of a how-to as a **recipe**.

A how-to teaches the reader practical knowledge, but is not presented as a lesson. Instead, it is content that is useful when working, and you need to remember how to do something.

Rather than teaching the basic concepts like a tutorial does, how-to's will:

- Assume basic understanding and familiarity.
- Provide concrete steps.
- Discuss options and practical alternates where appropriate, because it won’t often be clear what the users real-world situation is.

For these two how-tos, I tried to follow the following guides:

- **Do not teach.** Just like recipes don’t attempt to teach readers the basic cooking skills, how-to’s shouldn’t either. Instead, link out to tutorials and reference documentation if needed.
- Provide concrete steps, and assume familiarity with basic concepts and features
- Be practical. Don’t get distracted by explaining theoretical or abstract concepts. 
- Be concise. Use simple, straightforward language.
- Be consistent. Make sure that examples provided follow on and are consistent throughout all steps.

**Details**

The two how-to's are available here:

- [How to optimize a quantum machine learning model using JAX and Optax](https://qml-build-previews.pennylane.ai/pull_request_build_preview/970/demos/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.html)

- [How to optimize a quantum machine learning model using JAX and JAXopt](https://qml-build-previews.pennylane.ai/pull_request_build_preview/970/demos/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.html)

**Drawbacks**

For now, I have categorised these simply under existing categories. However:

- I have added a 'how to' tag.
- In the future, it might be nice to easily show all demos tagged as how-to.


demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+# -  We will embed our data through a series of rotation gates.
+# -  We will then have an ansatz of trainable rotation gates with parameters ``weights``; it is these
+#    values we will train to minimize our cost function.
+# -  We will train the QML model on ``data``, a ``(5, 4)`` array, and optimize the model to match
+#    target predictions given by ``target``.
+#
+
+import pennylane as qml
+import jax
+from jax import numpy as jnp
```
KetpuntoG, line 34: Is `jnp` a stsandar or do people use `np` as well?

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+    loss = jnp.sum((targets - predictions) ** 2 / len(data))
+    return loss
+
+######################################################################
+# Note that the model above is just an example for demonstration – there are important considerations
+# that must be taken into account when performing QML research, including methods for data embedding,
+# circuit architecture, and cost function, in order to build models that may have use. This is still
+# an active area of research; see our `demonstrations <https://pennylane.ai/qml/demonstrations>`__ for
+# details.
+#
```
KetpuntoG, line 84: I think it is worth commenting on what `jax.jit` does or at least providing a link

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+
+print(jax.grad(loss_fn)(params, data, targets))
+
+######################################################################
+# Create the optimizer
+# --------------------
+#
+
+######################################################################
+# We can now use Optax to create an Adam optimizer, and train our circuit.
```
KetpuntoG, line 115: Could you link to the other posible optimizers?
https://optax.readthedocs.io/en/latest/api.html

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+# -  Apply the update step of our optimizer via ``opt.update``
+#
+# -  Update the parameters via ``optax.apply_updates``
+#
+
+def update(params, opt_state, data, targets):
+    loss_val, grads = jax.value_and_grad(loss_fn)(params, data, targets)
+    updates, opt_state = opt.update(grads, opt_state)
+    params = optax.apply_updates(params, updates)
+    return params, opt_state, loss_val
```
KetpuntoG, line 136: It is about confusing to have `update`, `updates` and `opt.update`.
If this is the standard way to write it, ignore this comment.

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+
+    loss_val, grads = jax.value_and_grad(loss_fn)(params, data, targets)
+    updates, opt_state = opt.update(grads, opt_state)
+    params = optax.apply_updates(params, updates)
+
+    def print_fn():
+        jax.debug.print("Step: {i}  Loss: {loss_val}", i=i, loss_val=loss_val)
+
+    # if print_training=True, print the loss every 5 steps
+    jax.lax.cond((jnp.mod(i, 5) == 0) & print_training, print_fn, lambda: None)
```
KetpuntoG, line 172: First time I see "&" in the code 😄 is it exactly the same as "and"?

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+
+times = timeit.repeat("optimization(params, data, targets)", globals=globals(), number=num, repeat=reps)
+result = min(times) / num
+
+print(f"JIT compiling just the cost function (best of {reps}): {result} sec per loop")
+
+times = timeit.repeat("optimization_jit(params, data, targets)", globals=globals(), number=num, repeat=reps)
+result = min(times) / num
+
+print(f"JIT compiling entire optimization loop (best of {reps}): {result} sec per loop")
```
KetpuntoG, line 224: That's super cool! Could you finish with a sentence highlighting the difference between the two times?

KetpuntoG, line 224: Don't forget to put the authors at the end

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+
+    # trainable ansatz
+    for i in range(n_wires):
+        qml.RX(weights[i, 0], wires=i)
+        qml.RY(weights[i, 1], wires=i)
+        qml.RX(weights[i, 2], wires=i)
+        qml.CNOT(wires=[i, (i + 1) % n_wires])
+
+    # we use a sum of local Z's as an observable since a
+    # local Z would only be afffected by params on that qubit.
```
KetpuntoG, line 61:
```suggestion
    # local Z would only be affected by params on that qubit.
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,213 @@
...
+@jax.jit
+def loss_fn(params, data, targets):
+    predictions = my_model(data, params["weights"], params["bias"])
+    loss = jnp.sum((targets - predictions) ** 2 / len(data))
+    return loss
+
+######################################################################
+# Note that the model above is just an example for demonstration – there are important considerations
+# that must be taken into account when performing QML research, including methods for data embedding,
+# circuit architecture, and cost function, in order to build models that may have use. This is still
```
KetpuntoG, line 80:
```suggestion
# circuit architecture, and cost function, in order to build models that may have used. This is still
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,213 @@
...
+
+######################################################################
+# Jitting the optimization loop
+# -----------------------------
+#
+
+######################################################################
+# In the above example, we just-in-time (JIT) compiled our cost function ``loss_fn``. However, we can
+# also JIT compile the entire optimization loop; this means that the for-loop around optimization is
+# not happening in Python, but is compiled and executed native. This avoids (potentially costly) data
```
KetpuntoG, line 157:
```suggestion
# not happening in Python, but is compiled and executed natively. This avoids (potentially costly) data
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+
+######################################################################
+# Jitting the optimization loop
+# -----------------------------
+#
+
+######################################################################
+# In the above example, we just-in-time (JIT) compiled our cost function ``loss_fn``. However, we can
+# also JIT compile the entire optimization loop; this means that the for-loop around optimization is
+# not happening in Python, but is compiled and executed native. This avoids (potentially costly) data
```
KetpuntoG, line 156:
```suggestion
# not happening in Python, but is compiled and executed natively. This avoids (potentially costly) data
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # trainable ansatz
+    for i in range(n_wires):
+        qml.RX(weights[i, 0], wires=i)
+        qml.RY(weights[i, 1], wires=i)
+        qml.RX(weights[i, 2], wires=i)
+        qml.CNOT(wires=[i, (i + 1) % n_wires])
+
+    # we use a sum of local Z's as an observable since a
+    # local Z would only be afffected by params on that qubit.
```
KetpuntoG, line 60:
```suggestion
    # local Z would only be affected by params on that qubit.
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+@jax.jit
+def loss_fn(params, data, targets):
+    predictions = my_model(data, params["weights"], params["bias"])
+    loss = jnp.sum((targets - predictions) ** 2 / len(data))
+    return loss
+
+######################################################################
+# Note that the model above is just an example for demonstration – there are important considerations
+# that must be taken into account when performing QML research, including methods for data embedding,
+# circuit architecture, and cost function, in order to build models that may have use. This is still
```
KetpuntoG, line 81:
```suggestion
# circuit architecture, and cost function, in order to build models that may have used. This is still
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,213 @@
...
+#    values we will train to minimize our cost function.
+# -  We will train the QML model on ``data``, a ``(5, 4)`` array, and optimize the model to match
+#    target predictions given by ``target``.
+#
+
+import pennylane as qml
+import jax
+from jax import numpy as jnp
+import jaxopt
+
```
KetpuntoG, line 35: I use to put this to silence the messages in the demo:

jax.config.update("jax_platform_name", "cpu")
jax.config.update('jax_enable_x64', True)

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,224 @@
...
+    opt = optax.adam(learning_rate=0.3)
+    opt_state = opt.init(params)
+
+    args = (params, opt_state, data, targets, print_training)
+    (params, opt_state, _, _, _) = jax.lax.fori_loop(0, 100, update_jit, args)
+
+    return params
+
+######################################################################
+# Note that we use ``jax.lax.fori_loop`` and ``jax.lax.cond``, rather than a standard Python for loop
```
KetpuntoG, line 187: Could you add jax.debug.print too?

KetpuntoG (commented):
Great job! This will be very useful 😍
I leave a few comments

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
josh146: Yes good point, I will update this!

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
josh146: In this specific case yes, but in general it is different :) This is the _bitwise_ and operator, while `and` is the logical and operator.

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
josh146: I thought we didn't need this any more with the react website?

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
josh146: 'may have use' is the intended one here :)

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
josh146: jnp is the standard :) JAX recommends:

```python
from jax import numpy as jnp # for when you need JAX version
import numpy as np  # for when you don't need JAX
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
josh146:
```suggestion
jax.config.update("jax_platform_name", "cpu")
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
josh146: @KetpuntoG I've left in `updates` and `opt.update`, as this is what is used in the Optax docs

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
josh146: @KetpuntoG I've added a short sentence, but unlike a demo I'm not sure I want to draw any conclusions here -- I'm not trying to teach the reader the 'best' way, I'm just showing a comparison between the two methods I used.

In all honesty, I shouldn't even be including this appendix here, I just thought it was nice to show the times 😆

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,219 @@
+r"""How to optimize a quantum machine learning model using JAX and JAXopt
+=====================================================================
+"""
+
+######################################################################
+# Once you have set up your quantum machine learning model (which typically includes deciding on your
+# circuit architecture/ansatz, determining how you embed or integrate your data, and creating your
+# cost function to minimize a quantity of interest), the next step is **optimization**. That is,
```
mariaschuld, line 8:
```suggestion
# Once you have set up a quantum machine learning model, data to train with and 
# cost function to minimize as an objective, the next step is to perform the **optimization**. That is,
```

The brakets introduce a lot of bulky detail for the first sentence. I think if people don't know what data, cost and model are they are not on the right page, so this lean version should suffice?

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,234 @@
...
+# Here, we choose the Adam optimizer, however
+# `other available optimizers <https://optax.readthedocs.io/en/latest/api.html>`__
+# may be used here.
+#
+
+opt = optax.adam(learning_rate=0.3)
+opt_state = opt.init(params)
+
+######################################################################
+# We first define our ``update`` function, which needs to do a couple of things:
```
mariaschuld, line 125:
```suggestion
# We first define our ``update_step`` function, which needs to do a couple of things:
```
I would go with `update` everywhere - the `_step` does not add much information because an update is kind of a step?

mariaschuld: Unless you want to distinguish it from `opt.update`? Could then just use `step`?

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
```diff
@@ -0,0 +1,234 @@
+r"""How to optimize a quantum machine learning model using JAX and Optax
+====================================================================
+"""
+
+######################################################################
+# Once you have set up your quantum machine learning model (which typically includes deciding on your
+# circuit architecture/ansatz, determining how you embed or integrate your data, and creating your
+# cost function to minimize a quantity of interest), the next step is **optimization**. That is,
```
mariaschuld, line 8:
```suggestion
# Once you have set up a quantum machine learning model, data to train with and 
# cost function to minimize as an objective, the next step is to perform the **optimization**. That is,
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py
mariaschuld: I always found it super confusing that we did the `from pennylane import numpy as np` because one totally looses the idea of what is trainable and what not. Happy we are slowly changing standards as well!

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+######################################################################
+# Jitting the optimization loop
+# -----------------------------
+#
+
+######################################################################
+# In the above example, we JIT compiled our cost function ``loss_fn``. However, we can
+# also JIT compile the entire optimization loop; this means that the for-loop around optimization is
```
mariaschuld, line 158:
```suggestion
# In the above example, we JIT compiled our cost function ``loss_fn`` (and JAXopt automatically JIT compiled the `loss_and_grad` function behind the scenes). However, we can
# also JIT compile the entire optimization loop; this means that the for-loop around optimization is
```

This omission confused me, so maybe be explicit here?

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,219 @@
...
+        params, opt_state = opt.update(*args, i)
+        return (params, *args[1:])
+
+    args = (params, opt_state, data, targets, print_training)
+    (params, opt_state, _, _, _) = jax.lax.fori_loop(0, 100, update, args)
+
+    return params
+
+######################################################################
+# Note that we use ``jax.lax.fori_loop`` and ``jax.lax.cond``, rather than a standard Python for loop
```
mariaschuld, line 178:
```suggestion
# Note that -- similar to above -- we use ``jax.lax.fori_loop`` and ``jax.lax.cond``, rather than a standard Python for loop
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,219 @@
...
+# and if statement, to allow the control flow to be JIT compatible.
+#
+
+params = {"weights": weights, "bias": bias}
+optimization_jit(params, data, targets, print_training=True)
+
```
mariaschuld, line 184: Oh-oh: Check the print-out, somehow this example is not converging during training any more (while the above one is!)

mariaschuld (changes requested):
Very good how-to's, they contained a lot of information that I didn't know of, and which is very valuable for me!!

The only critical comment is that the JAXopt second example does not converge for some reason...

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,218 @@
...
+#
+
+@jax.jit
+def optimization_jit(params, data, targets, print_training=False):
+    opt = jaxopt.GradientDescent(loss_and_grad, stepsize=0.3, value_and_grad=True)
+    opt_state = opt.init_state(params)
+
+    def update(i, args):
+        params, opt_state = opt.update(*args, i)
+        return (params, *args[1:])
```
josh146, line 169:
```suggestion
        params, opt_state = opt.update(*args, i)
        return (params, opt_state, *args[2:])
```

demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.py
```diff
@@ -0,0 +1,218 @@
+r"""How to optimize a quantum machine learning model using JAX and JAXopt
+=====================================================================
+"""
+
+######################################################################
+# Once you have set up a quantum machine learning model, data to train with and 
+# cost function to minimize as an objective, the next step is to **perform the optimization**. That is,
```
josh146, line 7:
```suggestion
# Once you have set up a quantum machine learning model, data to train with, and 
# cost function to minimize as an objective, the next step is to **perform the optimization**. That is,
```

KetpuntoG (commented):
Thanks @josh146 ! Great job here :) Ideal for the performance year! For your part it is ready ✅ Do you mind if I work directly on the branch to correct some final typos and fixes, add thumbnails, etc?

KetpuntoG (approved):
Great job, ready to go! 🚀

mariaschuld (approved):

PR #972, opened 2023-11-01
removing interfaces qml demos

KetpuntoG:
indicating "interface" in the qnode is no longer necessary. The qml demos have been updated


demonstrations/ensemble_multi_qpu.py
```diff
@@ -232,13 +232,12 @@ def circuit1(params, x=None):
```
CatalinaAlbornoz, line 232: There's mention of the PyTorch interface at the beginning of the demo. Do we want to keep this mention?

demonstrations/tutorial_contextuality.py
```diff
@@ -512,7 +512,7 @@ def bias_inv_layer(weights, x):
 dev = qml.device("default.qubit", wires=3)
```
CatalinaAlbornoz, line 513: There's mention about the Jax interface at the beginning of the demo. Do we want to keep it?

demonstrations/tutorial_data_reuploading_classifier.py
```diff
@@ -262,7 +262,7 @@ def density_matrix(state):
 # Install any pennylane-plugin to run on some particular backend
```
CatalinaAlbornoz, line 263: There's mention about the PyTorch interface at the beginning of the demo. Do we want to keep it?

demonstrations/tutorial_kernel_based_training.py
```diff
@@ -254,7 +254,7 @@
 projector = np.zeros((2**n_qubits, 2**n_qubits))
 projector[0, 0] = 1
```
CatalinaAlbornoz, line 256: Another mention about the PyTorch interface at the beginning of the demo

demonstrations/tutorial_quantum_transfer_learning.py
```diff
@@ -307,7 +307,7 @@ def entangling_layer(nqubits):
 #   additional post-processing.
```
CatalinaAlbornoz, line 308: This mentions the PyTorch interface at the beginning too. Will people still need to install PyTorch separately to use it?

demonstrations/tutorial_quanvolution.py
```diff
@@ -145,7 +145,7 @@
 # Random circuit parameters
 rand_params = np.random.uniform(high=2 * np.pi, size=(n_layers, 4))
```
CatalinaAlbornoz, line 147: Mentions the Tensorflow interface at the beginning

CatalinaAlbornoz (approved):
It looks good to me. I'm not sure how much we want to make interfaces explicit or hidden but it's weird that only 1/3 of the demos mention interfaces explicitly at the beginning in the description of the demo.

KetpuntoG:
Hey @trbromley , It is correct that the demos talk about the interface, right? Because they are using torch, tensorflow or jax. Or does the word "interface" only make sense in connection with qnode?

demonstrations/ensemble_multi_qpu.py
trbromley: I think it is ok - we are still using PyTorch. It could be "Using a pre-trained model and PyTorch, ...".

demonstrations/tutorial_contextuality.py
trbromley: I think it's good. We can definitely still speak about the various ML libraries and even introduce them as an "interface" when used within the context of a QNode.

The goal of this PR is to have fewer demos using the `interfaces="..."` QNode argument. This can help establish that the non-explicit way is also supported, making it a bit of a lower overhead for users to think about when loading a QNode. For any performance-related demos though, we should still specify the interface.

DanielNino27 (approved):
Looks good, but we should monitor if this causes confusion in the use of these demos (e.g. if we get more questions about this in the discussion forum).

trbromley:
> Looks good, but we should monitor if this causes confusion in the use of these demos (e.g. if we get more questions about this in the discussion forum).

There is also the survey being shared in the discussion forum :+1:

PR #973, opened 2023-11-01
Test How-to category

KetpuntoG:
Trying to add category how to


demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_JAXopt.metadata.json
```diff
@@ -11,10 +11,9 @@
     "dateOfPublication": "2023-11-07T00:00:00+00:00",
     "dateOfLastModification": "2023-11-07T00:00:00+00:00",
     "categories": [
-        "Quantum Machine Learning",
-        "Optimization"
+        "How-to"
```
josh146, line 14: @KetpuntoG I would still keep the QML and optimization categories here, as users may be looking there for this type of content. cc @mariaschuld

PR #967, opened 2023-10-30
Align QML PR Previews with pennylane.ai

rashidnhm:
**Title:** Align QML PR Previews with React deployment

**Summary:**
**NOTE: This PR includes breaking changes and will not get deployed properly by current CI. The changes to deploy-pr.yml will take affect once this is merged**

This PR introduces the following changes:

- Aligns the QML PR Previews such that the deployment has parity with the react deployment once the PR is merged.
    - To achieve this, PRs will no longer build HTML versions of docs, rather it will build the demos as json, and in `deploy-pr.yml`, the source code from pennylane.ai-react will be pulled, modified slightly to fit the specific needs of QML PR previews, and then built.

- Removes many legacy code from deploy-py.yml.
    - When deploy-pr was originally written, sphinx was serving both the PR preview and production versions of the QML demos. However, as this is not the case, a lot of the code there (especially ones to differentiate between `pull_request` and `push` events) are not needed.
    - A lot of the actions/github-script are now part of the XanaduAI/cloud-actions repo, code blocks can are now removed and reference the re-usable actions within the cloud-actions repo.

**Relevant references:**
[sc-47381](https://app.shortcut.com/xanaduai/story/47381/align-pull-request-previews-in-qml-to-actual-deployment-version)

**Possible Drawbacks:**
- The pinned comment on existing PRs will now be duplicated due to the code changes on how it happens. This will only happen once for existing PRs, we can mark the older comment as outdated.
- This is overall a large change, if bugs are found, they will need to be addressed according. The following scenarios have been tested:
    - PR where only one or few demos are updated. Site was built successfully and only showed the demos in scope.
    - PR where all demos are re-built. Site was built successfully and showed all demos.

**Related GitHub Issues:**
None.

PR #975, opened 2023-11-02
Interactive demo assets base path fix

ashishks0522:
## Changes
* Make asset past constant with respect to root of PennyLane Website
* This fixes assets not loading for [interactive demo](https://pennylane.ai/qml/demos/circuits_as_fourier_series/)

## Note 
- this might not work correctly with current QML preview, but will work on react site.
- this will work with https://github.com/PennyLaneAI/qml/pull/967


rashidnhm (approved):
The original workflow had to account for both PR preview (sphinx) and deployment (react). But now, since both of those scenarios will be react, we can revisit this later to see what we'd need to change.

For now, this workaround LGTM

PR #985, opened 2023-11-14
Demo RL with pulse programming

BorjaRequena:
**Title:** Reinforcement learning for gate calibration

**Summary:** In this demo we learn how quantum gates are implemented in superconducting quantum circuits in practice. We will understand why we need to calibrate every qubit individually and how to deal with this process in an experimentally-friendly way with reinforcement learning.

**Relevant references:** It implements a protocol similar to the one introduced in [this paper](https://journals.aps.org/prxquantum/abstract/10.1103/PRXQuantum.2.040324)

----

It is related to the project I developed during my summer residency. It's a cool application of the pulse module in combination with reinforcement learning that can be easily taken from simulations to actual real devices.

* AUDIENCE — Who is this for?
For anyone in the field of quantum computing looking for an excuse to learn about reinforcement learning.

reinforcement learning, pulse programming, ... (need to think about these)

I currently don't have access to the linked notion page. It's on the lengthy side.


BorjaRequena:
The code will probably not be able to run because I need libraries that are not automatically installed like [qutip](https://qutip.org/) (for visualization) and [flax](https://flax.readthedocs.io/en/latest/) (to build a neural network). What's the standard approach for this?

josh146:
@BorjaRequena I'll let @KetpuntoG provide more details here, but there are two options:

- We can add these dependencies to the repo so that your demo will build
- We can convert your demo to a non-executable one, with the outputs hardcoded.

The first option is always preferred, but it depends on whether the dependencies are big/easier to install etc.

KetpuntoG:
> @BorjaRequena I'll let @KetpuntoG provide more details here, but there are two options:
> 
> * We can add these dependencies to the repo so that your demo will build
> * We can convert your demo to a non-executable one, with the outputs hardcoded.
> 
> The first option is always preferred, but it depends on whether the dependencies are big/easier to install etc.

Hey @BorjaRequena , nice to see this demo 😊
I would try to include them in the repo, at least I know that `qutip` should not give any issue. [Here](https://github.com/PennyLaneAI/qml#adding--editing-dependencies) you can find instructions on how to do it. let me know if there are any problems :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1081 @@
...
+
+axs[2].axis("off")
+ax2 = fig.add_subplot(1, 3, 3, projection="3d")
+plot_rotation_axes(rot_axis, fig=fig, ax=ax2)
+ax2.set_title(f"Average gate fidelity {avg_gate_fidelity:.3f}", fontsize=14)
+plt.show()
+
+######################################################################
+# Beyond single-qubit quantum computers and gates
+# -----------------------------------------------
```
BorjaRequena, line 890: The code in this section is just to give the reader an idea of how to extend the demo to deal with 2-qubit gates and hint how to go to multi-qubit devices. It won't work off the bat by just copy/pasting the cells below in their respective places above. Let me know if it is understandable.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1081 @@
...
+    # Single qubit pulses
+    qml.evolve(H_int + H_sq)(params_sq, t, atol=1e-5)
+
+    return qml.state()
+
+
+######################################################################
+# The state of the environment would be the 2-qubit state, and we would treat the qubit pair as our
+# previous single qubit. We would need to decide how many segments we wish to split each pulse into,
+# and define the time windows within ``play_episodes`` accordingly.
```
BorjaRequena, line 1003: Note to self: mention that the entire negative CR can be skipped as a single segment in the evolution as there are no actions happening.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1081 @@
...
+# 2. We reset the qubit.
+# 3. We perform quantum tomography to determine the qubit's state.
+# 4. With this information, the agent fixes the parameters for the next pulse segment.
+# 5. We reset the qubit and execute the pulse up to the last segment with fixed parameters.
+# 6. We repeat steps 3-5 until we reach the end of the pulse and evaluate the average gate fidelity.
+#
+# 
+# In the image below, we see an example of the tomography-parameters-reset&execute loop.
+#
+# .. figure:: ../demonstrations/rl_pulse/sketch_protocol.png
```
BorjaRequena, line 244: Let me know if the protocol is clear. The image is clearly provisional and if you'd like to see something in the plot that'd helped you understand it all better, please let me know!

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1081 @@
...
+# task through the interaction with the environment.
+#
+# In order to complete the desired task, the agent can observe the environment and perform *actions*,
+# which can affect the environment and change its *state*. This way, the interaction between them is
+# cyclic, as depicted in the figure below. At a given point in time, the agent observes the
+# environment's state. With this information, it can choose to perform a certain action. Doing so
+# affects the environment, which provides the agent with the new state it is found in and a *reward*.
+# The reward is a measure of how well the agent is performing the task given the last interaction.
+#
+# .. figure:: ../demonstrations/rl_pulse/sketch_rl.png
```
BorjaRequena, line 193: Clearly provisional 😅

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1081 @@
...
+The most common way to represent quantum computations is through gate-based quantum circuits. These
+provide a level of abstraction that allows us to write quantum algorithms without considering the
+hardware that will execute them. However, every quantum platform offers a different set of
+interactions and controls that allow us to simulate the effect of a few quantum gates. These are
+commonly referred to as the *native gates* of the device, and they constitute the fundamental
+building blocks of any quantum algorithm that we run in it. Therefore, it is essential that we
+perform such operations as accurately as possible, which requires the careful tunning of the
+hardware's controls. In this demo, we will learn how to use reinforcement learning to find the
+optimal control parameters to execute quantum gates in superconducting quantum devices.
+
```
BorjaRequena, line 14: Would be cool to have an image here like in [this demo](https://pennylane.ai/qml/demos/tutorial_optimal_control/) or [this other one](https://pennylane.ai/qml/demos/tutorial_teleportation/) 😄

demonstrations/tutorial_rl_pulse.metadata.json
```diff
@@ -0,0 +1,115 @@
...
+    "dateOfLastModification": "2023-12-25T00:00:00+00:00",
+    "categories": [
+        "Optimization",
+        "Quantum Hardware"
+    ],
+    "tags": [],
+    "previewImages": [
+        {
+            "type": "thumbnail",
+            "uri": "/_images/thumbnail_tutorial_here_comes_the_sun.png"
```
BorjaRequena, line 18: I just put in some random thumbnails and publication dates. These will need to be updated 😄

demonstrations/tutorial_rl_pulse.metadata.json
```diff
@@ -0,0 +1,115 @@
+{
+    "title": "Gate calibration with reinforcement learning",
```
BorjaRequena, line 2: I'm not too sold on this title. If you have a better one, I'm all ears.

pyproject.toml
```diff
@@ -56,7 +56,9 @@ covalent = "0.202.0.post1"
 pyzx = "0.7.0"
 neural-tangents = "0.6.2"
 plotly = ">=4.5.0"
-optax = "0.1.3"
+optax = "0.1.7"
```
BorjaRequena, line 59: Introducing flax required a higher jax version which then required upgrading optax. I hope this does not break any other demo.

BorjaRequena (commented):
I managed to deploy the demo. I hope you like it! Some comments below 😀

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1081 @@
...
+# To simulate the time evolution of our system, we can use the function :func:`~pennylane.evolve`.
+# This also allows us to extract the unitary matrix associated to the time evolution and visualize the
+# resulting rotation axis in the Bloch sphere. Let's see the effect that the amplitude and phase have
+# in the resulting single-qubit rotation.
+#
+
+import jax.numpy as jnp
+import warnings
+
+warnings.simplefilter("ignore")  # Silence JAX's warnings (nothing to worry about)
```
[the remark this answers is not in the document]
Qottmann: I would put `jax.config.update("jax_enable_x64", True)`
On newer versions of jax there is also no warning about no GPU anymore, alternatively you can set the device to CPU, `jax.config.update("jax_platform_name", "cpu")`

pyproject.toml
Qottmann: _should_ be okay, AFIK there are no breaking changes in the updated versions here

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+    qml.evolve(H_int + H_sq)(params_sq, t, atol=1e-5)
+
+    # Echoed CR
+    qml.evolve(H_int + H_cr_pos)(params_cr, t, atol=1e-5)
+    qml.PauliX(0)  # Flip control qubit
+    qml.evolve(H_int - H_cr_neg)(params_cr, t, atol=1e-5)  # Negative CR
+    qml.PauliX(0)  # Recover control qubit
+
+    # Single qubit pulses
+    qml.evolve(H_int + H_sq)(params_sq, t, atol=1e-5)
```
Qottmann, line 995: Does this actually work? I would have thought that you need to provide the correct timespans here as well.
I would recommend just working with fixed time durations, so `t=time_duration_sq` and `t=time_duration_ce` in the evolve function as well as the `get_drive` definitions above.

So overall you just have a `H_sq` and a `H_cr` that you reuse in the qnode with different parameters (and reusing `-params_cr`)

Qottmann (commented):
This is seriously a m a z i n g @BorjaRequena 😍 ! This will be a great resource for anybody wanting to do anything with reinforcement learning and quantum - in general 👌 🥂 

I went over the draft once and will certainly need a second time to fully appreciate all details. In the meantime, I left some detail-focused suggestions in the code. Here are some high level observations:

**We need an intro paragraph about the problem setting.** In this you can elaborate the setting of doing RL in the style of the Q-CTRL paper that uses measurement to calibrate real hardware (and here we do it in simulation, but that basically you can excahnge the environment for a real black box device). You can also cite directly the paper (but make it clear that its not a direct implementation [becauase its closed source, dont mention that though] but just "in the style" or following the ideas of that paper). That is one point to make in the overview/intro paragraph, the other is focussing on reinforcement learning.

Since the demo is already long, I would suggest to scrap the entire first section"Quantum gates in superconducting quantum computers", or at least reducing it significantly. The contents are more or less exactly the same to the "Transmon physics" section in [this recent demo](https://pennylane.ai/qml/demos/oqc_pulse/) to which you can link instead. The focus should really be that this demo shows you the RL pipeline and we just use transmon qubits in simulation as _one_ possible example for an environment. The details of transmon qubits are not that important, providing the definitions of the Hamiltonian and linking to the demo should be sufficient.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+#
+# Naively, we could define the time-dependent pulse parameters beforehand, execute the pulse, evaluate
+# the results, and modify the parameters for the next try based on the outcome. However, this
+# black-box approach suffers from major limitations. For instance, we would need to perform a separate
+# optimization for every single qubit every time we need to calibrate the quantum computer.
+#
+# With this approach, we would be neglecting the interaction loop between the agent and the
+# environment, which is perhaps the most important aspect of the reinforcement learning pipeline. It
+# is best if the agent receives information about the qubit's evolution during the pulse execution.
+# This allows the agent to react to the qubit's peculiarities and adapt the pulse parameters
```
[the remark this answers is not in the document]
BorjaRequena: On an initial draft I had some black-box examples of optimization, and then this was a motivation to use RL. However, after some modifications, it's a bit awkward. I think we could scrap the first paragraph (?) and adapt the second one accordingly.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# accordingly to execute the target gate. This way, we can train a single agent to calibrate all the
+# qubits in our computer.
```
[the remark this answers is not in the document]
BorjaRequena: Hmm I'll read the paper again to double-check 🤔

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+Gate calibration with reinforcement learning
+============================================
+
+The most common way to represent quantum computations is through gate-based quantum circuits. These
+provide a level of abstraction that allows us to write quantum algorithms without considering the
+hardware that will execute them. However, every quantum platform offers a different set of
+interactions and controls that allow us to simulate the effect of a few quantum gates. These are
+commonly referred to as the *native gates* of the device, and they constitute the fundamental
+building blocks of any quantum algorithm that we run in it. Therefore, it is essential that we
+perform such operations as accurately as possible, which requires the careful tunning of the
```
trbromley, line 11:
```suggestion
perform such operations as accurately as possible, which requires the careful tuning of the
```

demonstrations/rl_pulse/sketch_protocol.png
trbromley, line 1: @BorjaRequena for the images, did you make them yourself for this demo?

trbromley:
Thanks @BorjaRequena! This will be an awesome resource for reinforcement learning :star_struck:

demonstrations/rl_pulse/sketch_protocol.png
BorjaRequena: Yes. I sketched them fast just to show an idea of what they would represent to get feedback before polishing 😄

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+hardware's controls. In this demo, we will learn how to use reinforcement learning to find the
+optimal control parameters to execute quantum gates in superconducting quantum devices.
```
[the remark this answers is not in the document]
BorjaRequena: I rewrote it a bit. Let me know what you think :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+
+Quantum gates in superconducting quantum computers
+--------------------------------------------------
+
+In superconducting quantum computers, the qubits are coupled to wave guides that allow us to target
+them with microwave pulses. These pulses change the state of the qubits depending on their
+properties, allowing us to perform operations like single qubit rotations. When we do so, we say
+that we *drive* the qubit. Furthermore, some qubits are connected to others with coupling buses that
```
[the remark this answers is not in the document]
BorjaRequena: I'm getting rid of this section entirely.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+
+import pennylane as qml
+
+# Define the single-qubit device
+qubit_freqs = [4.81]  # GHz
+connections = []  # No connections
+couplings = []  # No couplings
+wires = [0]
+
+H_int = qml.pulse.transmon_interaction(qubit_freqs, connections, couplings, wires)
```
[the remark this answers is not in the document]
BorjaRequena: Removing this section. I will add the Hamiltonian in the Environment section

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# is best if the agent receives information about the qubit's evolution during the pulse execution.
+# This allows the agent to react to the qubit's peculiarities and adapt the pulse parameters
+# accordingly to execute the target gate. This way, we can train a single agent to calibrate all the
+# qubits in our computer.
+#
+
+######################################################################
+# The hardest part of this approach is extracting the information about the qubit's evolution under
+# the pulse, since observing it destroys the quantum state. Here, we follow a similar
+# experimentally-friendly approach to the one introduced in [#BaumPRXQ21]_. The main idea is to
```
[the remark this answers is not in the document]
BorjaRequena: I have put it in the new intro section

demonstrations/tutorial_rl_pulse.py
BorjaRequena: Yep, I don't have access to the notion page 😢

KetpuntoG: Post a sketch or an idea here and I'll take care of the rest :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# environment. The environment contains the relevant information about the problem and defines the
+# “rules of the game”. The main goal of the agent is to find the optimal strategy to perform a given
+# task through the interaction with the environment.
+#
+# In order to complete the desired task, the agent can observe the environment and perform *actions*,
+# which can affect the environment and change its *state*. This way, the interaction between them is
+# cyclic, as depicted in the figure below. At a given point in time, the agent observes the
+# environment's state. With this information, it can choose to perform a certain action. Doing so
+# affects the environment, which provides the agent with the new state it is found in and a *reward*.
+# The reward is a measure of how well the agent is performing the task given the last interaction.
```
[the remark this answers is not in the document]
BorjaRequena: Since I had already used chess to illustrate the reward below, I've thrown in a chess example in every step of the cycle.

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
BorjaRequena: I removed it because I didn't find a clear answer

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# 
+# In the image below, we see an example of the tomography-parameters-reset&execute loop.
+#
+# .. figure:: ../demonstrations/rl_pulse/sketch_protocol.png
+#    :align: center
+#    :width: 90%
+#
+# With this protocol, the agent iteratively builds a PWC pulse according to the qubit's evolution.
+# Even though this involves multiple executions to perform the intermediate tomography steps, the
+# overall cost is rather low provided that it is only for the qubit(s) involved in the gate.
```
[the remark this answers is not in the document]
BorjaRequena: I think you may be mixing two things: on one hand, a single pulse is comprised of various segments between which we perform quantum state tomography on the qubit, which is needed for the agent to predict the next segment's parameters. These tomography steps involve multiple re-executions, `n_shots`, of the same pulse (up to the given segment). In total, we would have `n_segments * n_shots` cost for a single pulse. On the other hand, training the agent involves multiple repetitions of this process called episodes, yielding `n_episodes * n_segments * n_shots`.

The argument here is that `n_shots` is low because we're interested only in the qubits involved in the gate, typically 1 or 2. I will try to make it clearer in the text, because I can see where the confusion stems from in the sentence.

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
BorjaRequena: I got rid of this paragraph. I believe it's clearer now 😄

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# - The environment (hardware, actions, and rewards)
+# - The agent (the policy and how to act)
+# - The learning algorithm
+#
+# Then, we'll put all the pieces together and train our agent. Let's introduce these concepts one by
+# one and implement them from scratch with PennyLane and JAX.
+#
+
+######################################################################
+# The environment
```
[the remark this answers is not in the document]
BorjaRequena: I agree! I've enriched this section a bit to compensate for the intro to superconducting gates, and I've added the Hamiltonian for a single driven qubit.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+
+H_drive = qml.pulse.transmon_drive(amplitude, phase, freq, wires)
+
+# Full time-dependent parametrized Hamiltonian
+H = H_int + H_drive
+
+######################################################################
+# Now that we have the effective model of our system, we need to simulate its time evolution. Since we
+# are simulating the whole process, we will simplify the qubit reset, evolution and tomography steps.
+# We can simply stop the simulation, look at the qubit's state, and then continue after the agent
```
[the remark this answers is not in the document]
BorjaRequena: I rewrote this paragraph. Let me know what you think 😄

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# experimentally feasible range, and associate every action to a combination of amplitude and phase
+# values.
+#
+
+values_phase = jnp.linspace(-jnp.pi, jnp.pi, 9)[1:]
+values_ampl = jnp.linspace(0.0, 0.2, 11)
+ctrl_values = jnp.stack(
+    (jnp.repeat(values_ampl, len(values_phase)), jnp.tile(values_phase, len(values_ampl))), axis=1
+)
+n_actions = len(ctrl_values)
```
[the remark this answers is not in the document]
BorjaRequena: Done!

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# parameters. The main principle of REINFORCE is to directly modify the policy to favour series of
+# actions within the agent's experience that have lead to high rewards. This way, past beneficial
+# actions are more likely to happen again.
+#
+# We achieve this by maximizing the expected *return* of the policy. To keep it simple, we can
+# understand the return (typically :math:`G`) as the weighted sum of rewards along an *episode*, which
+# are full executions of our reinforcement learning “game”. In our case, an episode would be the full
+# execution of a pulse program, which is comprised by several interactions between the agent and the
+# environment. Since the reward will only be given at the end, we will take the return to be the final
+# reward. For those who are familiar with the reinforcement learning terminology, we're setting the
```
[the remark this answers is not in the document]
BorjaRequena: I removed the last sentence. It did not really bring any value given that those who are familiar will already know that it is gamma=1, and those who are not familiar will have no idea of what's going on.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1082 @@
...
+# the current policy :math:`\pi_{\mathbf{\theta}}`. Then, we can perform a gradient ascent update to
+# the policy parameters :math:`\mathbf{\theta}`, sample some episodes with the new parameters, and
+# repeat until we converge to the optimal policy.
+#
+# Let's implement these ideas one by one, starting by the episode sampling. The episodes start with
+# the qubit in the :math:`|0\rangle` state. Then, we start the observe-act loop, where the agent
+# observes the qubit's state and chooses the parameters for the next pulse segment according to its
+# policy. We can sample the action and compute the score function at the same time to make the code as
+# efficient as possible. Finally, when we reach the end of the pulse program, we compute its reward.
+# We like it when things go fast, so we'll sample all the episodes in parallel!
```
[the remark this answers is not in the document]
BorjaRequena: Not quite! Thanks for reading it to the detail though!! 😀

demonstrations/tutorial_rl_pulse.metadata.json
```diff
@@ -0,0 +1,115 @@
...
+    "authors": [
+        {
+            "id": "borja_requena"
+        }
+    ],
+    "dateOfPublication": "2023-12-25T00:00:00+00:00",
+    "dateOfLastModification": "2023-12-25T00:00:00+00:00",
+    "categories": [
```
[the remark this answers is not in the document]
BorjaRequena: I have added "Getting Started". I'm not sure about QML, depending on what you (Xanadu) consider as QML. If you count classical ML applied to quantum stuff as QML, I can add it, although I don't usually call it QML 😄

demonstrations/tutorial_rl_pulse.py
BorjaRequena: I've replaced it by 2 images and I have extended the description. I think it should be clearer, let me know what you think!

demonstrations/tutorial_rl_pulse.py
BorjaRequena: Done!

demonstrations/tutorial_rl_pulse.py
BorjaRequena: I had to split `H_sq` into `H_sq_ini` and `H_sq_end` for it to work properly. Defining the PWC parametrized Hamiltonians with a time span from the beginning guarantees that they are 0 outside of that time span. I did it this way because in `play_episodes` we perform the time-evolution segment by segment providing a time window for the current segment we are interested in: `evolve_states(..., time_window)`. This way, the control of the segments and time windows is outside of the evolution function. I am not sure this is the cleanest, my motivation here was to try to provide something that was both understandable and (somewhat) plug-and-play for the previous code.

I am not sure about the negative parameters in the CR. It would provide both a negative amplitude and phase. Is this what we want? In my head it was only a negative amplitude 🤔

BorjaRequena:
I have written a revised version of the demo taking all the feedback into account. The plots are still provisional, and I will need to find an example where there's a better convergence because the resulting gate in the example is quite bad 😅

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
BorjaRequena: Done! It's cleaner now

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+    # Echoed CR
+    qml.evolve(H_int + H_cr_pos)(params_cr, t, atol=1e-5)
+    qml.PauliX(0)  # Flip control qubit
+    qml.evolve(H_int - H_cr_neg)(params_cr, t, atol=1e-5)  # Negative CR
+    qml.PauliX(0)  # Recover control qubit
+
+    # Single qubit pulses
+    qml.evolve(H_int + H_sq_end)(params_sq, t, atol=1e-5)
+
+    return qml.state()
```
Qottmann, line 911: So just to understand, you have already run this and it works right? We just dont include the actual training because it would be taking too long to run here?
Would it be possible to make a code cell that _would_ run the training successfully (without the reader having to adapt anything) or would that become too messy with too many adjustments to be made?

Qottmann (commented):
This will be an incredible resource for anyone wanting to do anything with reinforcement learning in jax and pennylane! Really, really good demo @BorjaRequena !

I left mostly minor comments, some of which tagged as personal opinion (PO, so feel free to disregard)

I think the very last paragraphs are a bit confusing and I am not sure I understand the intention here. With fixing that, the demo is good to go from my side (there will likely a third review from the community team though)



- [ ] We need to start the thumbnail process. Could you provide @KetpuntoG with a sketch or an idea? Simple is king :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+Gate-based quantum circuits are the most common representation of quantum computations. These
+provide an abstraction layer that enables the development of quantum algorithms without considering
+the hardware in charge of the execution. However, every quantum platform offers a different set of
+interactions and controls that define the natural operations that can be performed in the hardware.
+These are known as the *native gates* of the device, and they constitute the fundamental
+building blocks of any quantum algorithm executed in it. Therefore, it is essential that such
+operations are performed as accurately as possible, which requires the careful tunning of the
+hardware's controls. In this demo, we will learn how to use reinforcement learning to find the
+optimal control parameters to accurately execute quantum gates using superconducting quantum
+devices as an example.
```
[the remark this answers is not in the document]
BorjaRequena: I agree! Done :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+calibrated for each individual qubit in the hardware. For further details about superconducting
+quantum computers and their control see `this demo <https://pennylane.ai/qml/demos/oqc_pulse/>`__.
+
+A common strategy to calibrate quantum gates involves the detailed modelling of the quantum
+computer, enabling the gate optimization through analytical and numerical techniques. Nevertheless,
+developing such models requires an exhaustive characterization of the hardware and, even then, it
+is very challenging to account for all the relevant interactions. Therefore, it is usually
+unfeasible to develop highly accurate models, in practice.
+
+An alternative promissing approach is through the direct interaction with the device, refraining
```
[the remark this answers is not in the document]
BorjaRequena: Yup, done!

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+experimental nuances associated with the process of executing the gates, such as the effect of the
+most relevant noise sources. This makes the resulting agent an excellent calibrator that is robust
+to these phenomena.
+
+The procedure that we present below is entirely agnostic to the quantum hardware.
+Among all the possibilities, we will ilustrate it using coupled-transmon superconducting quantum
+computers simulated with PennyLane. This will allow us to focus on the method itslef, skipping some
+of the nuances associated with the execution on real devices, while ensuring that the resulting
+code can be easily adapted to run in a quantum computer using the PennyLane plugins, as shown in
+`this demo <https://pennylane.ai/qml/demos/tutorial_optimal_control/>`__.
```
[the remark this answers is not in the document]
BorjaRequena: It looks like I messed up here with the url. I probably wanted to link the [OQC Lucy demo](https://pennylane.ai/qml/demos/oqc_pulse/) there to provide a reference of pulse-level control with real devices.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+the frequency and the connectivity to other qubits. These differences cause the same pulse programs
+to produce different operations for every qubit. Consequently, every gate must be carefully
+calibrated for each individual qubit in the hardware. For further details about superconducting
+quantum computers and their control see `this demo <https://pennylane.ai/qml/demos/oqc_pulse/>`__.
+
+A common strategy to calibrate quantum gates involves the detailed modelling of the quantum
+computer, enabling the gate optimization through analytical and numerical techniques. Nevertheless,
+developing such models requires an exhaustive characterization of the hardware and, even then, it
+is very challenging to account for all the relevant interactions. Therefore, it is usually
+unfeasible to develop highly accurate models, in practice.
```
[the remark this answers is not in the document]
BorjaRequena: I have tamed it down a bit. Let me know what you think or whether you would add anything else.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+# are rewards after every action is performed, as in the schematic depiction above. However, in some
+# cases, those rewards are zero until the task is finished. For example, if we are training an agent
+# to play chess, we cannot evaluate every single move on its own, and we need to wait until the game
+# is resolved in order to provide the agent with a reward.
+#
+# Our case is similar to the chess example. The intermediate states visited along the time evolution
+# do not necessarily provide a clear indication of how well the target gate is being executed.
+# Therefore, we need to evaluate the pulse program as a whole after we have gone through all the pulse
+# segments. The reward will be the average gate fidelity of our pulse program with respect to the
+# target gate.
```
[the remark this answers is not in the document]
BorjaRequena: I have rewritten a bit the paragraph. Although that now that I see your comment again, maybe you meant the last sentence? 🤔

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+
+
+policy_model = MLP(hidden_size=30, out_size=n_actions)
+
+# Initialize the parameters passing a mock sample
+key = jax.random.PRNGKey(3)
+key, subkey = jax.random.split(key)
+
+mock_state = jnp.empty((1, state_size))
+policy_params = policy_model.init(subkey, mock_state)
```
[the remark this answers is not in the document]
BorjaRequena: Hmm that's right. This is probably inherited from the original code where this was a function and the key was then passed on. Finding a key that works well is a but clunky because I get different outcomes between my local execution and the result in the demo. If this is not too critical I would rather leave it as is.

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
Qottmann: Yeah mainly the point that the reward in our case (and chess case) is only given at the end

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
Qottmann: Yeah thats fine 👍

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+# understand the return (typically :math:`G`) as the weighted sum of rewards along an *episode*, which
+# are full executions of our reinforcement learning “game”. In our case, an episode would be the full
+# execution of a pulse program, which is comprised by several interactions between the agent and the
+# environment. Since the reward will only be given at the end, we will take the return to be the final
+# reward.
+#
+# We perform the maximization of the expected return by gradient ascent over the policy parameters. We
+# can compute the gradient of the expected return as follows
+#
+# .. math:: \nabla_{\mathbf{\theta}} \mathbb{E}\left[G\right] = \mathbb{E}\left[\sum_{t=1}^{T}G_t\nabla_{\mathbf{\theta}}\log\pi_{\mathbf{\theta}}(a_t|s_t)\right],
```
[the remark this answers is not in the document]
BorjaRequena: This is a construction that allows us to compute the gradient without any knowledge about the dynamics of the underlying system. Otherwise, we would need to derive an explicit model for the transition probability between states given actions $p(s_{t+1}|s_t, a_t)$. This is partly what gives the name to the "model-free" RL algorithms that avoid attempting to learn anything about the system.

Since we're on a fully controlled simulation, I think it would be possible to derive an analytic form for the expression. However, that would defeat the point of the protocol that precisely tries to avoid building an explicit model of the quantum computer.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+#
+# where the expectation values are over episodes sampled following the policy
+# :math:`\pi_{\mathbf{\theta}}`. The sum goes over the episode time steps :math:`t`, where the agent
+# observes the state :math:`s_t` and chooses to perform the action :math:`a_t`. The term
+# :math:`\nabla_{\mathbf{\theta}}\log\pi_{\mathbf{\theta}}(a_t|s_t)` is known as the *score function*
+# and it is the gradient of the logarithm of the probability with which the action is taken. Finally,
+# :math:`G_t` is the return associated to the episode from time :math:`t` onwards, which is always the
```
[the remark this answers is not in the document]
Qottmann: I think already seeing that $G_t$ is a sum over past rewards helps (maybe this is already in and I just missed it). I think it makes sense to explain these but since we have the scenario of just a single reward at the end, I would put it in a note maybe? Like FYI/Good to know generally, but not super relevant here (and explain why not).

basically the first paragraph in your answer could make a good note 👍

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+
+def get_optimizer(params, learning_rate):
+    """Create and initialize an Adam optimizer for the parameters."""
+    optimizer = optax.adam(learning_rate)
+    opt_state = optimizer.init(params)
+    return optimizer, opt_state
+
+
+######################################################################
+# And we will define our parameter update function to perform the gradient ascent step. These
```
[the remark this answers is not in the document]
BorjaRequena: Hmm thinking about it it's not that straightforward because the gradient is a pytree of nested dicts. If I remember correctly the built in function to perform the update does the same `tree_map` with `p + u` instead of `p - u`, so maybe this is the jax version of "just a minus sign"? 🤔 However, the training loop looks nice with one line for each thing: play episodes, compute gradients, update parameters.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+# gates involving multiple qubits and larger devices.
+#
+# To continue learning about this topic, try implementing one of the two extensions we explain above.
+# To learn more about reinforcement learning, we recommend [#SuttonBarto18]_ for an introduction to
+# the topic, and [#Dawid22]_ for an introduction of machine
+# learning for physics (reinforcement learning in chapter 6). To learn more about how superconducting
+# quantum computers work, see [#KrantzAPR19]_ for an extensive review, and the related `PennyLane
+# documentation <https://docs.pennylane.ai/en/stable/code/qml_pulse.html>`__.
+#
+# Finally, check out these other demos for alternative ways to tune pulse programs.
```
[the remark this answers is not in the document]
BorjaRequena: Indeed! I have added them and extended the conclusions a bit to highlight again the strengths of the method that we introduce

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+# and define the time windows within ``play_episodes`` accordingly. Given that the negative CR
+# pulse uses the same parameters as the positive CR one, we can skip it as an entire segment that
+# does not involve any intermediate tomography steps.  
+#
+# Finally, in order to train a calibrator for a quantum computer with several qubits (or qubit pairs),
+# we need to define a separate drive Hamiltonian for each indidividual qubit. Then, every episode
+# samples one of these at random to perform the evolution. This way, if we are calibrating the
+# :math:`R_X(\pi/2)` gate for a device with multiple qubits, we would have a separate ``H_drive`` with
+# the control pulse for each qubit and every episode would carry the time evolution of
+# ``H_int + H_drive`` for a randomly selected qubit. This is a bit more involved to adapt. Mainly, we
```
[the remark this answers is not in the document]
BorjaRequena: Yeah, more or less. I am not too convinced with this paragraph actually. To calibrate devices with many qubits, we can either train a calibrator for all the qubits in the device, or train a calibrator for every single qubit or qubit pair. I honestly don't know how this is done in practice, which is why I was throwing both ideas in there.  I've rewritten the paragraph and it hopefully makes a bit more sense. Let me know what you think, even if you think it requires rethinking from scratch.

demonstrations/tutorial_rl_pulse.py
BorjaRequena: I have tested that this function provides the correct Hamiltonian at every point in time. However, I have not performed a gate optimization with it. Indeed, I never got to get a properly optimized 2-qubit gate in my scarce testing last summer (although it was with a different ansatz). This was mainly intended to provide an example of how to do it, but it does require further modifications to the code. In particular, the code above assumes segments of constant length, and here we would need to consider pulse segments of variable length, given that the CR pulses are way longer than the single qubit ones. This is something that is mentioned in the text, but there's no code example for it. I have extended the text explanations to make them more precise.

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1006 @@
...
+
+
+######################################################################
+# The state of the environment would be the 2-qubit state, and we would treat the qubit pair as our
+# previous single qubit. We would need to decide how many segments we wish to split each pulse into,
+# and define the time windows within ``play_episodes`` accordingly. Given that the negative CR
+# pulse uses the same parameters as the positive CR one, we can skip it as an entire segment that
+# does not involve any intermediate tomography steps.  
+#
+# Finally, in order to train a calibrator for a quantum computer with several qubits (or qubit pairs),
```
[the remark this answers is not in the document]
BorjaRequena: I'll close this to discuss it in the next comment, as they are quite related

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
BorjaRequena: I have added the note! Maybe we should add another note mentioning that reinforce (the algorithm) tends to converge to local minima and that changing the seed there may be different results?

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
BorjaRequena: Yep

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1049 @@
...
+refer to as a *pulse program*. However, each qubit in the device has distinct properties, such as
+the frequency and the connectivity to other qubits. These differences cause the same pulse programs
+to produce different operations for every qubit. Consequently, every gate must be carefully
+calibrated for each individual qubit in the hardware. For further details about superconducting
+quantum computers and their control see `this demo <https://pennylane.ai/qml/demos/oqc_pulse/>`__.
+
+A common strategy to calibrate quantum gates involves the detailed modelling of the quantum
+computer, enabling the gate optimization through analytical and numerical techniques. Nevertheless,
+developing such accurate models requires an exhaustive characterization of the hardware, and it
+can be challenging to account for all the relevant interactions in practice.
```
Qottmann, line 39: Nice! Thanks for the update :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1049 @@
...
+the hardware in charge of the execution. However, every quantum platform offers a different set of
+interactions and controls that define the natural operations that can be performed in the hardware.
+These are known as the *native gates* of the device, and they constitute the fundamental
+building blocks of any quantum algorithm executed in it. Therefore, it is essential that such
+operations are performed as accurately as possible, which requires the careful tunning of the
+hardware's controls. In this demo, we will learn how to use reinforcement learning to find the
+optimal control parameters to accurately execute quantum gates. We will implement an
+experimentally-friendly protocol based on the direct interaction with the hardware, following
+the main ideas in [#BaumPRXQ21]_, which we illustrate using superconducting quantum computers as
+an example.
```
Qottmann, line 16:
```suggestion
the main ideas in [#BaumPRXQ21]_, which we illustrate using superconducting qubits as
an example.
```

This may sound like we are doing stuff experimentally, while what you mean is we look at superconducting qubit platforms I think

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1049 @@
...
+#     where :math:`\gamma\in[0, 1]` is a *discount factor* that favours early rewards vs latter
+#     ones. For instance, :math:`\gamma\to0` only values immediate rewards, whereas
+#     :math:`\gamma\to1` accounts equally for all the rewards regardless of the time. The return
+#     from time :math:`t` weights the rewards relative to the given time:
+#
+#     .. math:: G_t=\sum_{k=0}^{T-1-t} \gamma^k r_{k+t+1},
+#
+#     where :math:`k` denotes the number of steps after :math:`t` (note that
+#     :math:`G\equiv G_{t=0}`). This presents a recursive definition that is exploited in some
+#     algorithms :math:`G_t = r_{t+1} + \gamma G_{t+1}`.
```
Qottmann, line 438: What is meant by $G \equiv G_{t=0}$ here?
The second sentence reads a bit awkwardly, is there a part missing?

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
Qottmann: sounds good 👍

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1049 @@
...
+        pulse_duration_sq,
+        pulse_duration_sq,
+        pulse_duration_cr,
+        pulse_duration_cr,
+        pulse_duration_sq,
+        pulse_duration_sq,
+    ]
+)
+change_times = jnp.cumsum(pulse_durations)
+timespans = [(t0.item(), t1.item()) for t0, t1 in zip(change_times[:-1], change_times[1:])]
```
Qottmann, line 898: Note that this is not really necessary as there is no concept of absolute time between different pulses. You might as well just pass the `pulse_duration_sq/cr` times as the timespans here to simplify the code. Not necessary though

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1049 @@
...
+# the topic, and [#Dawid22]_ for an introduction of machine learning for physics (reinforcement
+# learning in chapter 6). To learn more about how superconducting
+# quantum computers work, see [#KrantzAPR19]_ for an extensive review, and the related `PennyLane
+# documentation <https://docs.pennylane.ai/en/stable/code/qml_pulse.html>`__.
+#
+# Finally, check out the related demos for alternative ways to tune pulse programs. In particular,
+# `this demo <https://pennylane.ai/qml/demos/tutorial_optimal_control/>`__ for an optimal control
+# approach to gate calibration, and `this <https://pennylane.ai/qml/demos/tutorial_odegen/>`__ and 
+# `this other <https://pennylane.ai/qml/demos/tutorial_pulse_programming101/>`__ demos for examples
+# of variational quantum eigensolvers.
```
Qottmann, line 996:
```suggestion
# Finally, check out the related demos for alternative ways to tune pulse programs. In particular,
# `this demo <https://pennylane.ai/qml/demos/tutorial_optimal_control/>`__ for an optimal control
# approach to gate calibration, and `this demo on optimizing pulses using hardware compatible gradients <https://pennylane.ai/qml/demos/tutorial_odegen/>`__ and 
# `this more general intro to differentiable pulse programming <https://pennylane.ai/qml/demos/tutorial_pulse_programming101/>`__. 
```

Qottmann (approved):
Thanks for the updates @BorjaRequena 
Looks good to me now! Just some minor comments, but overall good 👍 

Do you have ideas for a thumbnail? If yes, can you please send it to @KetpuntoG ? We should start the process soon :)

demonstrations/tutorial_rl_pulse.py
BorjaRequena: Yep, I did it to simplify the qnode. This way, when the qnode receives a time interval inside `play_episodes`, there is no need for any control flow to know which terms of the Hamiltonian should be acting and so on because the inactive terms will be 0. There's probably a cleaner way to handle this because the code actually assumes that the block `qml.PauliX(0)`-negative CR-`qml.PauliX(0)` is always executed as a single segment (explained in the text below). I'm definitely open to suggestions for improvements 😄

demonstrations/tutorial_rl_pulse.py
BorjaRequena: Typically $G$ denotes the return of a full episode, and $G_t$ denotes the return on an episode from time $t$ onwards. By definition, $G_t$ is $G$ when we take the initial time $t=0$. I don't know if this is clear.

I have rephrased this part. Let me know if it is understandable. Either way, it does not play a role in this demo so it could be safely scratched if it ends up being more confusing than anything.

Qottmann: Thanks, much clearer now :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+[#BaumPRXQ21]_. In this setting, a reinforcement learning agent learns to calibrate the gates by
+tuning the control parameters and directly observing the response of the qubits. Through this
+process, the agent implicitly learns an effective model of the device, as it faces all the
+experimental nuances associated with the process of executing the gates, such as the effect of the
+most relevant noise sources. This makes the resulting agent an excellent calibrator that is robust
+to these phenomena.
+
+The procedure that we present below is entirely agnostic to the quantum hardware.
+Among all the possibilities, we will ilustrate it using coupled-transmon superconducting quantum
+computers simulated with PennyLane. This will allow us to focus on the method itslef, skipping some
```
KetpuntoG, line 57:
```suggestion
computers simulated with PennyLane. This will allow us to focus on the method itself, skipping some
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+# reward will be the average gate fidelity of our pulse program with respect to the target gate.
+#
+# In order to evaluate it, we sample several random initial states and apply the pulse program a few
+# consecutive times. Then, we compute the average fidelity between the resulting intermediate and
+# final states from our pulse, and the expected states from the target gate. Finally, we take the
+# weighted average of all the fidelities, giving more relevance to the initial gate applications. This
+# process of repeatedly applying the pulse programs accumulates the errors, making our reward function
+# more sensitive to them. This allows the agent to better refine the pulse programs.
+#
+# The number of initial states and gate repetitions are hyper parameters that will add up really
```
KetpuntoG, line 280:
```suggestion
# The number of initial states and gate repetitions are hyperparameters that will add up really
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+# according to the discrete probability distribution provided by the output layer. However, we will
+# see how to implement it below to act and compute the policy gradient at once.
+#
+
+######################################################################
+# The learning algorithm
+# ``````````````````````
+#
+# Among the plethora of reinforcement learning algorithms, we use the REINFORCE [#Williams1992]_
+# algorithm. It belongs to the policy gradient algorithms family, which propose a parametrization of
```
KetpuntoG, line 397:
```suggestion
# algorithm. It belongs to the policy gradient algorithms family, which proposes a parametrization of
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+# the policy (as we have conveniently done in the previous section) and directly optimize its
```
KetpuntoG, line 398:
```suggestion
# the policy (as we have conveniently done in the previous section) and directly optimizes its
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+# step further. We will subtract a baseline :math:`b` to the return such that
+#
+# .. math:: \nabla_{\mathbf{\theta}} \mathbb{E}\left[G\right] = \mathbb{E}\left[\sum_{t=0}^{T-1}(G_t - b(s_t))\nabla_{\mathbf{\theta}}\log\pi_{\mathbf{\theta}}(a_t|s_t)\right].
+#
+# Intuitively, the magnitude of the reward is arbitrary and depends on our function of choice. Hence,
+# it provides the same information if we shift it. For example, our reward based on the average gate
+# fidelity is bound between zero and one, but everything would conceptually be the same if we added
+# one. In a sense, we could consider the original expression to have a baseline of zero, and this new
+# one to be a generalization. Any baseline is valid provided that it does not depend on the action
+# :math:`a_t` because this ensures it has a null expectation value, leaving the gradient unafected.
```
KetpuntoG, line 565:
```suggestion
# :math:`a_t` because this ensures it has a null expectation value, leaving the gradient unaffected.
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+    """Create and initialize an Adam optimizer for the parameters."""
+    optimizer = optax.adam(learning_rate)
+    opt_state = optimizer.init(params)
+    return optimizer, opt_state
+
+
+######################################################################
+# And we will define our parameter update function to perform the gradient ascent step. These
+# optimizers default to gradient descent, which is the most common in machine learning problems.
+# Hence, we'll have to subtract the parameter udpate they provide to go in the opposite direction.
```
KetpuntoG, line 640:
```suggestion
# Hence, we'll have to subtract the parameter update they provide to go in the opposite direction.
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+
+######################################################################
+# The training
+# ------------
+#
+# We have all the building blocks that we need to train an :math:`R_X(\pi/2)` calibrator for our
+# superconducting quantum computer. We just need to ensemble the pieces together and choose the right
+# parameters for the task.
+#
+# Let's define the config object that will contain all the hyper parameters of the training process.
```
KetpuntoG, line 659:
```suggestion
# Let's define the config object that will contain all the hyperparameters of the training process.
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+# The CNOT gate is typically constituted by a series of single-qubit pulses and cross-resonant (CR)
+# pulses. Every single-qubit pulse is resonant with the qubit it targets, i.e., it has the same
+# frequency, whereas CR pulses are shone on the control qubit at the target's frequency. This way, the
+# target is indirectly driven through the coupling with the control qubit, which entangles them.
+#
+# We will make the CR pulse take the first qubit as control and the second as target. We will
+# implement what's known as an echoed CR pulse, which consists of flipping the state of the control
+# qubit (applying an :math:`X` gate) in the middle of the CR pulse. The second half of the pulse is
+# the negative of the previous one. This “echoes out” the rapid interactions between the qubits, while
+# preserving the entangling slower interaction that we are interestd in, as introduced in [#SheldonPRA16]_.
```
KetpuntoG, line 872:
```suggestion
# preserving the entangling slower interaction that we are interested in, as introduced in [#SheldonPRA16]_.
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+#
+# Our full pulse ansatz will consist of sandwitching the echoed CR gate between single qubit pulses.
```
KetpuntoG, line 874: could you double check "sandwitching" is correct?

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+# appropriate ``time_window`` within ``play_episodes``. This can be achieved by modifying the 
+# ``config.segment_duration`` to be an array that contains the time spans of every segment, such
+# that ``time_window = config.segment_duration[s]``, or similar. Given that the negative CR pulse
+# uses the same parameters as the positive CR one, we can skip it as an entire segment merged with
+# the last from the positive one that does not involve any intermediate tomography steps.
+#
+# Finally, when dealing with quantum computers with several qubits, we can opt for two strategies:
+# train specialized calibrators for every qubit (or qubit pair), or train a single general
+# calibrator for all the qubits. In these cases, we need to define a separate drive Hamiltonian for
+# each indidividual qubit.
```
KetpuntoG, line 961:
```suggestion
# each individual qubit.
```

demonstrations/tutorial_rl_pulse.metadata.json
```diff
@@ -0,0 +1,127 @@
+{
+    "title": "Gate calibration with reinforcement learning",
+    "authors": [
+        {
+            "id": "borja_requena"
+        }
+    ],
+    "dateOfPublication": "2023-12-25T00:00:00+00:00",
+    "dateOfLastModification": "2023-12-25T00:00:00+00:00",
```
KetpuntoG, line 9:
```suggestion
    "dateOfPublication": "2024-04-05T00:00:00+00:00",
    "dateOfLastModification": "2024-04-05T00:00:00+00:00",
```

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+#     P. Krantz, M. Kjaergaard, F. Yan, T. P. Orlando, S. Gustavsson and W. D. Oliver. (2019)
+#     "A quantum engineer's guide to superconducting qubits."
+#     `Applied physics reviews 6(2) <https://pubs.aip.org/aip/apr/article/6/2/021318/570326/A-quantum-engineer-s-guide-to-superconducting>`__.
+#
+
+######################################################################
+# About the author
+# ----------------
+#
+# .. include:: ../_static/authors/borja_requena.txt
```
KetpuntoG, line 1053:
```suggestion
```

Extracted from the metadata :)

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+
+import jax
+from functools import partial
+
+device = qml.device("default.qubit", wires=1)
+
+
+@partial(jax.jit, static_argnames="H")
+@partial(jax.vmap, in_axes=(0, None, 0, None))
+@qml.qnode(device=device, interface="jax")
```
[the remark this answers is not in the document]
Qottmann: We dont need to set `interface="jax"` explicitly anymore when doing pulses?

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1053 @@
...
+
+    # Single qubit pulses
+    qml.evolve(H_int + H_sq_end)(params_sq, t, atol=1e-5)
+
+    return qml.state()
+
+
+######################################################################
+# The state of the environment is now a 2-qubit state for which on the quantum computer we need to
+# perform :math:`\mathcal{O}(4^2)` measurements for tomography.
```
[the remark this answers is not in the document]
Qottmann: Yes tomography is $\mathcal{O}(4^n-1)$ measurements 🥲

demonstrations/tutorial_rl_pulse.py
Qottmann: I think it should be "sandwiching"

Qottmann:
> As a comment, I personally like to see that the code works as it is displayed or more frequently. It has been at the end of the demo when it has been executed for the first time. Anyway, a very interesting topic!

@KetpuntoG what do you mean by this?

KetpuntoG:
> > As a comment, I personally like to see that the code works as it is displayed or more frequently. It has been at the end of the demo when it has been executed for the first time. Anyway, a very interesting topic!
> 
> @KetpuntoG what do you mean by this?

For instance in [this demo](https://pennylane.ai/qml/demos/tutorial_liealgebra/) you can see that you print something after each cell. That helps me to know that everything is working well 😄 I know it doesn't always make sense, but I find it easier to read

demonstrations/tutorial_rl_pulse.py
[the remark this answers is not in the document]
KetpuntoG: Oh I didn't know that pulses worked differently. In general in the devices is something that we do not specify but if in pulses is different ignore this comment 👌

demonstrations/tutorial_rl_pulse.py
```diff
@@ -0,0 +1,1052 @@
...
+# frequency, whereas CR pulses are shone on the control qubit at the target's frequency. This way, the
+# target is indirectly driven through the coupling with the control qubit, which entangles them.
+#
+# We will make the CR pulse take the first qubit as control and the second as target. We will
+# implement what's known as an echoed CR pulse, which consists of flipping the state of the control
+# qubit (applying an :math:`X` gate) in the middle of the CR pulse. The second half of the pulse is
+# the negative of the previous one. This “echoes out” the rapid interactions between the qubits, while
+# preserving the entangling slower interaction that we are interested in, as introduced in [#SheldonPRA16]_.
+#
+# Our full pulse ansatz will consist of sandwitching the echoed CR gate between single qubit pulses.
```
KetpuntoG, line 874:
```suggestion
# Our full pulse ansatz will consist of sandwiching the echoed CR gate between single qubit pulses.
```

KetpuntoG (approved):
Awesome 😎

PR #997, opened 2023-12-11
Update variational classifier demo

dwierichs:
This PR updates the variational classifier demo with respect to the following:

- Some coding strategies are being updated, based on changed (mostly improved) behaviour of PennyLane.
- Some comments are outdated, probably because code has been updated in overhauls but the text has not.
- Changed the padding of the Iris data to be a tad less arbitrary (maybe?)
- Use broadcasting for the Iris dataset, moving the execution time on my local machine from a few minutes to ~15 seconds.

@mariaschuld This is the update I mentioned :)


demonstrations/tutorial_variational_classifier.py
```diff
@@ -222,80 +201,73 @@ def cost(weights, bias, X, Y):
...
-    )
+    current_cost = cost(weights, bias, X, Y)
+    acc = accuracy(Y, predictions)
 
+    print(f"Iter: {it+1:4d} | Cost: {current_cost:0.7f} | Accuracy: {acc:0.7f}")
 
 ##############################################################################
+# As we can see, the variational classifier learned to classify the bit strings correctly by
+# training the model on the cost function. Note that we used (batches of) the
+# same data for training the model and for testing it, which we would not do in practice.
```
mariaschuld, line 235: IIIerkh. We cannot do this in a demo, even if the comment warns users. Can't believe I coded this up. 

Shall we split training and test data?

mariaschuld:
@KetpuntoG , David and I overhauled this beginner demo, a lot of stuff was outdated. 

We wanted to invite you to have a look because the content changed a bit in the first example: we now use it to show what generalisation means. Does that work?

demonstrations/tutorial_variational_classifier.py
```diff
@@ -194,23 +174,22 @@ def cost(weights, bias, X, Y):
 # .. note::
 #
```
KetpuntoG, line 175: Could you include also a link to the paraity_test.txt?

demonstrations/tutorial_variational_classifier.py
```diff
@@ -219,83 +198,111 @@ def cost(weights, bias, X, Y):
...
-        np.sqrt(x[2] ** 2 + x[3] ** 2)
-        / np.sqrt(x[0] ** 2 + x[1] ** 2 + x[2] ** 2 + x[3] ** 2)
-    )
+    beta2 = 2 * np.arcsin(np.linalg.norm(x[2:]) / np.linalg.norm(x))
 
     return np.array([beta2, -beta1 / 2, beta1 / 2, -beta0 / 2, beta0 / 2])
 
 
-def statepreparation(a):
+def state_preparation(a):
```
KetpuntoG, line 305: Is this the qml.MottonenStatePreparation? 
If so, we can simplify this cell

demonstrations/tutorial_variational_classifier.py
```diff
@@ -219,83 +198,111 @@ def cost(weights, bias, X, Y):
...
+acc_test = accuracy(Y_test, predictions_test)
+print("Accuracy on unseen data:", acc_test)
 
 ##############################################################################
+# The quantum circuit has also learnt to predict all unseen examples perfectly well! 
+# This is actually remarkable, since the encoding strategy creates quantum states 
+# from the data that have zero overlap -- and hence the states created from the test 
+# set have no overlap with the states created from the training set. There are 
+# many functional relations the variational circuit could learn from this kind 
+# of representation, but the classifier choose to label bit strings according
```
KetpuntoG, line 266:
```suggestion
# of representation, but the classifier chooses to label bit strings according 
```

KetpuntoG (commented):
Thanks for the update 😊

demonstrations/tutorial_variational_classifier.py
mariaschuld: We could, yes...but this is a super simple beginners demo, so I wanted to unpack black boxes as much as possible. Unless it is distracting?

demonstrations/tutorial_variational_classifier.py
mariaschuld: Good point!

mariaschuld:
@KetpuntoG I hope it's ready now? Definitely a huge improvement to before

KetpuntoG (approved):
Awesome! 🚀 This is ready to merge :)
Feel free to mark as approved @mariaschuld  (we need two)

mariaschuld (approved):

mariaschuld:
Done, I'll leave it for you to merge though! :)

PR #1002, opened 2023-12-15
Renamed demo to make it distinct

BM7878:


doctorperceptron (approved):
:+1: 

Might be a good idea to add the "Getting Started" category to this one too, so it shows up with the other one in that section?

justinpickering:
Hey @Mandrenkov - would you be able to review this please? This is causing a P0 bug, and we will aim to release glass onion demos search Monday morning - thank you!

josh146:
Just for some context, this page not showing up in any of the demo homepage categories was deliberate, as this is simply 'appendix' information for another demo that links directly to it.

What was the P0 bug in this case?

doctorperceptron:
> Just for some context, this page not showing up in any of the demo homepage categories was deliberate, as this is simply 'appendix' information for another demo that links directly to it.
> 
> What was the P0 bug in this case?

This being closely related to the primary demo, it was being recommended along with it. Having the same name and thumbnail made it look like the system was recommending the same demo twice. Additionally, this demo was being recommended to follow after the main demo, but once again having the same name and thumbnail looked like the main demo was recommending itself.

josh146:
Yep let's remove the category!

Mandrenkov (approved):
@justinpickering Looks good to me if @doctorperceptron and @josh146 approve!

PR #1005, opened 2024-01-03
update auto-run demos before 0.34 release

timmysilv:
- learning2learn was failing because we call a QNode with a vanilla list of parameters at some point, and it wrongly auto-assumed that the interface was `autograd`. Not sure why this only came up now, but forcing it to `tf` fixes it
- the Barrier in the teleportation demo was just to make it look like the two measurements happen simultaneously, but that doesn't happen anymore regardless... so I removed it
- the second full circuit MPL drawing (with the QNode) is supposed to show that defer_measurements is applied, so I had to add `expansion_strategy="device"` to the QNode to make sure it is applied (it used to be applied at qnode-level, but now it's part of device pre-processing)
  - question: should I put the kwarg in the qnode construction, or in the invocation of `draw_mpl`? I think they both accept it
- unitary_designs was using the old `qfunc_transform` and a warning was being raised. Updated to the new `transform` API to avoid the warning

### DIFF
--- a/demonstrations/tutorial_teleportation.py
+++ b/demonstrations/tutorial_teleportation.py
@@ -313,13 +313,13 @@ _ = qml.draw_mpl(teleport, style="pennylane")(state)
 # PennyLane, when you bind a circuit to a device that does not support them,
 # it will automatically apply the principle of deferred measurement and update
 # your circuit to use controlled operations instead. Note that you need to
-# specify ``expansion_strategy="device"`` when creating the QNode for the
-# drawer to know to do device expansion before drawing.
+# specify ``expansion_strategy="device"`` when calling ``draw_mpl`` so it
+# runs the device pre-processing before drawing the circuit.
 
 dev = qml.device("default.qubit", wires=["S", "A", "B"])
 
 
-@qml.qnode(dev, expansion_strategy="device")
+@qml.qnode(dev)
 def teleport(state):
     state_preparation(state)
     entangle_qubits()
@@ -328,7 +328,7 @@ def teleport(state):
     return qml.density_matrix(wires=["B"])
 
 
-_ = qml.draw_mpl(teleport, style="pennylane")(state)
+_ = qml.draw_mpl(teleport, style="pennylane", expansion_strategy="device")(state)
 
 ##############################################################################
 #
--- a/demonstrations/tutorial_teleportation.py
+++ b/demonstrations/tutorial_teleportation.py
@@ -297,7 +297,6 @@ def teleport(state):
     state_preparation(state)
     entangle_qubits()
     basis_rotation()
-    qml.Barrier(["S", "A"], only_visual=True)
     measure_and_update()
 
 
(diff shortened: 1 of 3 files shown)


demonstrations/tutorial_teleportation.py
```diff
@@ -313,12 +312,14 @@ def teleport(state):
 # working in a system that does not support mid-circuit measurements. In
 # PennyLane, when you bind a circuit to a device that does not support them,
 # it will automatically apply the principle of deferred measurement and update
-# your circuit to use controlled operations instead.
+# your circuit to use controlled operations instead. Note that you need to
+# specify ``expansion_strategy="device"`` when creating the QNode for the
+# drawer to know to do device expansion before drawing.
```
albi3ro, line 317: I'd probably stick it in the drawer instead, since its for the drawer.

demonstrations/tutorial_unitary_designs.py
```diff
@@ -467,20 +467,21 @@ def noisy_operations(damp_factor, depo_factor, flip_prob):
...
+    # Apply the original operations, then the noisy ones
+    noisy_ops = tape.operations + noisy_tape.operations
+
+    def null_postprocessing_fn(results):
+        return results[0]
 
     # Apply the original measurements
-    for m in tape.measurements:
-        qml.apply(m)
+    return (type(tape)(noisy_ops, tape.measurements),), null_postprocessing_fn
```
albi3ro, line 484: Should we be adding shots as well?

demonstrations/tutorial_teleportation.py
timmysilv: done

demonstrations/tutorial_unitary_designs.py
timmysilv: there are no shots used in this tutorial, but we could to be a good example if you wanna

albi3ro: I think because this also showing other people how to user transforms to do things.  And someone might be messing around with the code and want to try finite shots.

timmysilv: updated!

demonstrations/tutorial_unitary_designs.py
```diff
@@ -491,24 +492,25 @@ def apply_noise(tape, damp_factor, depo_factor, flip_prob):
...
 
-@qml.qfunc_transform
+@qml.transform
 def conjugate_with_unitary(tape, matrix):
-    qml.QubitUnitary(matrix, wires=0)
-
-    for op in tape.operations:
-        qml.apply(op)
+    new_ops = [
+        qml.QubitUnitary(matrix, wires=0),
```
mudit2812, line 505: You might have to `qml.queuing.QueuingManager.stop_recording()` here. @albi3ro recently had to fix a bug that was causing new ops added in a transform to be queued twice. This might not be problematic in the context of the demo (iirc it was only a problem with transforming qfuncs), but it's probably good practice to make it public in a demo for people that might want to create their own transforms.

mudit2812 (commented):
One comment, otherwise ready to approve

demonstrations/tutorial_unitary_designs.py
timmysilv: I actually don't want to add `stop_recording` for exactly that reason. It's clumsy, so I _don't_ want to _have_ to tell users to do it. Luckily, Christina already realized that, so she changed all qfunc-transforms to stop recording internally. side-note for posterity: QNode and tape transforms don't depend on queuing, so they don't need it and won't be affected

mudit2812 (approved):

albi3ro (approved):

PR #1009, opened 2024-01-05
Update hybrid jobs demos

mudit2812:
The hybrid jobs were failing. Some of them were because of incorrect variable names. Others because functions defined outside the hybrid job context were being used inside it. This PR fixes both of those.

From discussions outside GitHub:

> * The `getting_started_with_hybrid_jobs` demo uses some pickling and is only expected to work with 0.33.1. Once changes in this PR are incorporated, it does work with 0.33.1. An empty dictionary for results is returned with the most recent PL, and it's not a bug, its built in on the Braket end.
> * Once the fixes for `braket_parallel_gradients` are added, everything runs, the results are just incorrect. It's a static demo, so it will continue to look like the results are correct unless someone actually tries it. We'll let AWS know.

There is also a deprecation warning about `Observable.return_types` in the `braket_parallel_gradients` demo, but it's originating from the braket plugin. Their release is incosistent with ours, so it will be fixed, but probably not in time for 0.34


demonstrations/getting_started_with_hybrid_jobs.py
```diff
@@ -232,7 +250,11 @@ def qubit_rotation_hybrid_job(num_steps=1, stepsize=0.5):
 import pandas as pd
 import matplotlib.pyplot as plt
 
-df = pd.DataFrame(job.metrics())
+metrics = job.metrics()
+print(metrics)
+
+df = pd.DataFrame(metrics)
+print(df)
```
timmysilv, line 257: should these `print` statements be here, and if so, should the outputs be added to the demo outputs somewhere?

demonstrations/getting_started_with_hybrid_jobs.py
```diff
@@ -169,7 +169,25 @@ def qubit_rotation(num_steps=10, stepsize=0.5):
 
 @hybrid_job(device="local:pennylane/lightning.qubit")
 def qubit_rotation_hybrid_job(num_steps=1, stepsize=0.5):
-    return qubit_rotation(num_steps=num_steps, stepsize=stepsize)
```
timmysilv, line 172: if we aren't using this anymore, we should update the end of the text blurb above

demonstrations/getting_started_with_hybrid_jobs.py
mudit2812: I thought I removed those 😅

timmysilv (approved):

KetpuntoG (commented):
Thanks! Should we update the warning on the demo to PL 0.34?

lillian542:
> Thanks! Should we update the warning on the demo to PL 0.34?

It's still compatible with 0.32, as claimed in the warning, and when we were testing, it also worked with 0.33 and 0.33.1. It is **not** compatible with 0.34, because of things defined on the Braket end. So I think we should leave the warning as-is, or maybe say "only guaranteed to be compatible with Python 3.10 and PennyLane 0.32"

KetpuntoG (approved):
LGTM thanks!

PR #1010, opened 2024-01-07
Demo about implementing Shor's algorithm with Catalyst

glassnotes:
**Title:** JIT compiling Shor's algorithm with PennyLane and Catalyst

**Summary:**  Over the past year, PennyLane has been integrated with Catalyst, allowing for quantum just-in-time compilation of classical and quantum code together. This demo leverages that integration to implement a version of Shor's factoring algorithm that is just-in-time compiled from end-to-end, classical control structure and all.

**Relevant references:** See within

**Possible Drawbacks:** As it stands, it's quite long (especially for the amount of actual output); I am not sure about the ending.

**Related GitHub Issues:** N/A

----

Highlight PennyLane and Catalyst integration, and expose challenges (and opportunities) in scaling up quantum compilation pipelines.

* AUDIENCE — Who is this for?

Primarily people researchers and developers focused on quantum compilation.

quantum compilation; quantum circuits; quantum algorithms; Shor's algorithm

- [ ] Tutorial
- [X] Demo
- [ ] How-to

[sc-75726]


demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+
+# TODO: insert code here
+
+######################################################################
+# The quantum part
+# ^^^^^^^^^^^^^^^^
+# 
+# TODO: explain how the modular exponentiation circuits work (not sure yet how to
+# best include these, because some are quite large)
```
KetpuntoG, line 103: If you don't want to go into too much detail here, there is an option to upload a .py and simply import the operator. However, I think it would be worth a small tutorial in the future. People would love it!

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,140 @@
...
+interact with quantum software is evolving. Writing and optimizing quantum
+circuits by hand for algorithms with hundreds or thousands of qubits is
+unsustainable, even for the most seasoned quantum programmers. To develop
+large-scale algorithms, we need frameworks that allow us to sit at a comfortable
+level of abstraction, and tools we can trust to do the heavy lifting under the
+hood. The integration of version 0.34 of PennyLane with `Catalyst
+<https://docs.pennylane.ai/projects/catalyst/en/latest/index.html>`_ represents
+a positive step in this direction. In this demonstration we will show how
+Catalyst enables an implementation of Shor's factoring algorithm that is
+just-in-time compiled from end-to-end, classical control structure and all.
```
KetpuntoG, line 28: I think we should recommend to the reader something about just-in-time since I think it is super important here and it is not a known term in quantum computing: "If you still don't know what this Jit thing is and the potential it has, I invite you to take a look to ..." or similar

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,206 @@
...
+#
+# Just-in-time compilation involves compiling code *during* execution, for instance,
+# while an interpreter is doing its job. 
+#
+
+######################################################################
+# The quantum part
+# ^^^^^^^^^^^^^^^^
+# 
+# TODO: explain how the modular exponentiation circuits work (not sure yet how to
```
KetpuntoG, line 168: Update: pennylane now has [qml.ModExp](https://docs.pennylane.ai/en/stable/code/api/pennylane.ModExp.html). So we can probably save the explanation here 😄

glassnotes: I saw! Are they `qjit` compatible?  There is a fair amount of customization in my implementation so I will probably still provide brief explanations, and the circuits, but only what's needed.

KetpuntoG: It should be, let me know if you have any problems :)

demonstrations/tutorial_haar_measure.py
```diff
@@ -809,4 +809,4 @@ def qr_haar_random_unitary():
 #     Z. Holmes, K. Sharma, M. Cerezo, and P. J. Coles (2021) "Connecting ansatz
 #     expressibility to gradient magnitudes and barren plateaus". (`arXiv
 #     <https://arxiv.org/abs/2101.02138>`__)
-#
\ No newline at end of file
+#
```
glassnotes, line 812:
```suggestion
#
```

demonstrations/tutorial_shor_catalyst.py
josh146: Curious if this ended up working!

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1036 @@
...
+
+def QFT(wires):
+    """The standard QFT, redefined because the PennyLane one uses terminal SWAPs."""
+    shifts = jnp.array([2 * jnp.pi * 2**-i for i in range(2, len(wires) + 1)])
+
+    for i in range(len(wires)):
+        qml.Hadamard(wires[i])
+
+        for j in range(len(shifts) - i):
+            qml.ControlledPhaseShift(shifts[j], wires=[wires[(i + 1) + j], wires[i]])
```
KetpuntoG, line 788: The demo is looking great! Just a comment to make this more visible
In Pennylane the number 4 is represented as:

|1>---
|0>---
|0>--

In qiskit as:

|0>---
|0>-
|1>--

I see you are going with the qiskit format. Perhaps it may cause confusion as the results will not match the Pennylane arithmetic operators

glassnotes: Ah, I am using the normal PennyLane ordering, but I written things to prevent a bunch of unnecessary SWAPs. It should match the order in the circuits here.

demonstrations/tutorial_shor_catalyst.py
glassnotes: I haven't tried - I stayed with my own implementations because there are  "higher-level" optimizations being applied (like removing entire QFTs, etc.).

glassnotes:
Marking as ready-for-review to test the build (it worked locally). I'll tag people once a full first draft of all the text and code is finished.

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1059 @@
...
+#
+# TODO: plots of performance
+
+# TODO: relevant code
+
+######################################################################
+# Conclusions
+# -----------
+#
+# TODO
```
josh146, line 1031:
```suggestion
# TODO!
```

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1105 @@
...
+# down into a couple distinct steps. Suppose we wish to decompose an integer
+# :math:`N` into its two constituent prime factors, :math:`p` and :math:`q`.
+#
+#  - First, we randomly select a candidate integer, :math:`a`, between 2 and
+#    :math:`N-1` (before proceeding, we double check that we did not get lucky and randomly select one of the true factors)
+#  - Using our chosen a, we proceed to the quantum part of the algorithm: order-finding.
+#    Quantum circuits are generated, and the circuit is executed on a device. The results
+#    are used to make a guess for a non-trivial square root.
+#  - If the square root is non-trivial, we test whether we found the factors. Otherwise, we try again
+#    with more shots. Eventually, we try with a different value of a.
```
josh146, line 91:
```suggestion
# - First, we randomly select a candidate integer, :math:`a`, between 2 and
#   :math:`N-1` (before proceeding, we double check that we did not get lucky and randomly select one of the true factors)
# - Using our chosen a, we proceed to the quantum part of the algorithm: order-finding.
#   Quantum circuits are generated, and the circuit is executed on a device. The results
#   are used to make a guess for a non-trivial square root.
# - If the square root is non-trivial, we test whether we found the factors. Otherwise, we try again
#   with more shots. Eventually, we try with a different value of a.
```

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+        end = time.time()
+        execution_times.append((N, a, end - start))
+
+labels = [f"{ex[0]}, {int(ex[1])}" for ex in execution_times][::2]
+times = [ex[2] for ex in execution_times]
+
+plt.scatter(range(len(times)), times, c=[ex[0] for ex in execution_times])
+plt.xticks(range(0, len(times), 2), labels=labels, rotation=80)
+plt.xlabel("N, a")
+plt.ylabel("Runtime (s)");
```
glassnotes, line 1098: I can't figure out how to suppress the text output of the plot. Adding a semicolon does not help. Also, would be great to get some feedback on clarity of the plot and whether it is effective in capturing the main points

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+# compile, and optimize very large circuits, even within the context of a larger
+# workflow. There is still much work to be done, however. For one, the generated
+# circuits are not optimized at the individual gate level, so the resource
+# counts will be large and impractical. One also observes that, even though
+# "higher-level" optimizations become possible, there is still a significant
+# amount of manual labour involved in implementing them. Compilation tools that
+# can automatically determine how subroutines can be simplified based on
+# classical information or on the input quantum superposition would be valuable
+# to develop, as they would enable co-optimization of the classical and quantum
+# parts of workflows.
```
glassnotes, line 1121: Suggestions welcome. Honestly this was so long in the making that I was at a loss for what to write here :sweat_smile:

glassnotes (commented):
@KetpuntoG @josh146 @dime10 At long last, here is a first draft for review. It's a long one, so please feel free to tag whoever has some bandwidth. I left some comments within. Two other top-level things:
 - Most of the 90000 lines of this PR is SVG. The actual demo is ~1100
 - The webpage does not seem to fully build. I can find the demo from the front, but most of the links give me XML access denied errors. I don't know if that's expected.

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+                if q != 1:
+                    p = N // q
+
+    return p, q
+
+
+######################################################################
+# To run the algorithm, we must randomly choose ``a``. In principle, we could
+# randomly generate ``a`` within the function above. However, this cannot
+# currently be QJITted. Passing ``n_bits`` as a static argument is also to work
```
dime10, line 1010: Oh does this not work? Generating random numbers is possible with JAX, the main difference to numpy is that the PRNG state is [explicit rather than implicit](https://docs.jax.dev/en/latest/random-numbers.html#explicit-random-state). We also use them in the MSD demo: https://docs.pennylane.ai/projects/catalyst/en/stable/demos/magic_distillation_demo.html#T-Type-Magic-States

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+# around some QJIT limitations. For now, let's see if our algorithm can
+# successfully factor.
+
+from jax import random
+
+
+def factor_with_shor(N, n_shots=100):
```
dime10, line 1017: I'm hesitant about the term shots here, since they could be confused with the concept of PennyLane/device shots. Should we just call these trials or similar?

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+plt.scatter(range(len(times)), times, c=[ex[0] for ex in execution_times])
+plt.xticks(range(0, len(times), 2), labels=labels, rotation=80)
+plt.xlabel("N, a")
+plt.ylabel("Runtime (s)");
+
+######################################################################
+# This plot demonstrates exactly what we suspect: changing :math:`N` and
+# :math:`a` does not lead to recompilation of the program! This will be
+# particularly valuable for large :math:`N`, where compilation time may become
+# long.
```
dime10, line 1104: Was this saying that the JIT compilation time is long for large N or that "traditional" (e.g. PennyLane) circuit construction and processing becomes very expensive?
I think when I measured it the increase in JIT compilation time was relatively flat now. 
```suggestion
# particularly valuable for large :math:`N`, where traditional circuit processing times can
# grow very large.
```

demonstrations/tutorial_shor_catalyst.py
dime10: This plot does show the lack of recompilation for different `a`s and even `N`s, but I think what might be missing is demonstrating why that is valuable to the user 🤔 
What about comparing to the non-jitted version where re-computing and processing new circuits for each `a` takes a lot of time?

demonstrations/tutorial_shor_catalyst.py
dime10: I remember you mentioning how great it was to be able to implement optimizations that are dependent on `a` without any extra cost, maybe that is something we could highlight in the conclusion? :) 
> The fact that you can make all these really specific, a-dependent optimizations in the circuit, and it just works after the first run because of jit, is seriously amazing

dime10 (commented):
First of all, this is an incredible demo!! I'm amazed how well you explain and show the different pieces going in to this, must have been a lot work 😍 The introduction is also really nice :)

demonstrations/tutorial_shor_catalyst.py
dime10: > I can't figure out how to suppress the text output of the plot. 

Have you tried this version?
```suggestion
plt.ylabel("Runtime (s)")
plt.show()
```

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+# programmer really need to know anything quantum is happening, if the software
+# library can generate and compile appropriate quantum code (though, they should
+# probably have at least some awareness, since the output of ``guess_order`` is
+# probabilistic!).  This raises the important question of what exactly gets
+# compiled, as well as where, when, and how that compilation happens.
+#
+# Over the past year, PennyLane has been integrated with `Catalyst
+# <https://docs.pennylane.ai/projects/catalyst/en/latest/index.html>`_, allowing
+# for quantum just-in-time compilation of classical and quantum code
+# together. This demo leverages that integration to implement a version of
```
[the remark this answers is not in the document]
glassnotes: Ahh, this was written as things had just changed from functions like `adjoint` and `ctrl` needing special Catalyst versions to being able to just use `qml.adjoint`. I will adjust the language.

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+# ^^^^^^^^^^^^^^^^^^^
+# Quantum compilation, like its classical counterpart, lowers an algorithm from
+# high-level instructions down to low-level ones. The bulk of
+# this process involves converting a circuit from a generic, abstract gate
+# set to a sequence of gates that satisfy the constraints of a particular
+# hardware device. Quantum compilation also involves multiple passes through
+# one or more intermediate representations, and both
+# machine-independent and dependent optimizations, as depicted below.
+#
+# .. figure:: ../_static/demonstration_assets/shor_catalyst/compilation-stack.svg
```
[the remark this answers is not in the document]
glassnotes: Yeah I'll change that! I usually use this diagram in talks, not text, so I'm explaining as I go :sweat_smile:

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+    return p, q
+
+######################################################################
+# We will find ways to leverage knowledge of :math:`a` to
+# construct more optimal quantum circuits within the QJITted function.
+#
+# QJIT compiling Shor's algorithm
+# -------------------------------
+#
+# Quantum subroutines
```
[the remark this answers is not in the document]
glassnotes: Alright so I've relegated three big chunks of circuit explanations to an appendix at the end - see what you think!

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+#
+# There is a tradeoff here, however. Each controlled operation is different, so
+# we must compile and optimize more than one circuit. However, additional
+# compilation time could be outweighed by the fact that we now run only
+# :math:`t` controlled operations, instead of :math:`1 + 2 + 4 + \cdots +
+# 2^{t-1} = 2^t - 1`.
+#
+# Next, let's zoom in on an arbitrary controlled-:math:`U_a`.
+#
+# .. figure:: ../_static/demonstration_assets/shor_catalyst/c-ua.svg
```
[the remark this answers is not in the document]
glassnotes: I see what you mean - check out the new version, I tried putting it inside a curly brace. If this makes it clearer, I can update the other diagrams too.

demonstrations/tutorial_shor_catalyst.py
glassnotes: Admittedly I haven't tried in a long time - the last time I did, I couldn't get it to work, but I don't remember what the issue was. I'll see if I can work it out!

demonstrations/tutorial_shor_catalyst.py
glassnotes: Sure, `trials` works!

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1244 @@
...
+# Finally, let's put it all together in the core portion of Shor's algorithm,
+# under the ``@qml.qjit`` decorator.
+
+from jax import random
+
+
+# ``n_bits`` is a static argument because ``jnp.arange`` does not currently
+# support dynamically-shaped arrays when jitted.
+@qml.qjit(autograph=True, static_argnums=(3))
+def shors_algorithm(N, key, a, n_bits, n_trials):
```
glassnotes, line 788: @dime10 I got the random number generation working within the qjitted algorithm. The reason it wasn't working before is that previously, I was generating a list of valid `a` using `jnp.arange` and then using `jnp.random.choice`. But if I simply use `randint` between a range, this avoids the problem. 

I've also made it so `a` can still be sent in manually (which bypasses the random generation, and allows us to time and do resource estimates for specific `a`).

Finally, I've made `n_trials` a parameter as well, so we really can qjit the whole thing! :tada:

```python
p, q, _, a, success_prob = shors_algorithm(N, key.astype(jnp.uint32), 0, n_bits, 100)
```

glassnotes:
The demo is now ready for re-review (requested from David, but please add anyone else who will be taking a look). The major updates include:
 - moving a lot of the circuit construction details and diagrams into appendices
 - adding a couple more plots to demonstrate results, including a comparison against a standard PennyLane implementation (which leads to a lot of code duplication, unfortunately)
 - expanding the scope of the qjitted function to include also random generation of `a`. So *everything* (except the static argument for bit width) is now qjitted  :tada:

demonstrations/tutorial_shor_catalyst.py
dime10: Sweet 🚀 🚀

dime10 (approved):
Love it 😍 The demo is more concise now, and I like the new results section which the comparison to PennyLane! High-quality demo :)

The below are all very minor:

demonstrations/tutorial_shor_catalyst.py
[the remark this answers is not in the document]
dime10: Much clearer now :)

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+# 1 \rangle`, which gets sent to :math:`\vert a \rangle`. This is effectively
+# just doing controlled addition of :math:`a - 1` to :math:`1`. Since :math:`a`
+# is selected from between :math:`2` and :math:`N-2` inclusive, the addition is
+# guaranteed to never overflow. This means we can simply do a controlled Fourier
+# addition, and save a significant number of resources!
+#
+# We can also make some optimizations to the end of the algorithm by keeping
+# track of the powers of :math:`a`. If at iteration :math:`k` we have
+# :math:`a^{2^k} = 1`, no further multiplication is necessary, and we can
+# effectively terminate the algorithm early.
```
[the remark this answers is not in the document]
dime10: 👌

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+
+
+def as_integer_ratio(f):
+    """QJIT compatible conversion of a floating point number to two 64-bit
+    integers such that their quotient equals the input to available precision.
+    """
+    mantissa, exponent = jnp.frexp(f)
+
+    i = 0
+    while jnp.logical_and(i < 300, mantissa != jnp.floor(mantissa)):
```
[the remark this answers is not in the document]
dime10: Omg 😆 Okay I went digging a bit and found the source again, this function is a conversion of the Python `float_as_integer_ratio` implementation to JAX:
https://github.com/python/cpython/blob/e0bc9d2a0c448cf46df233f8d84344c1f55a190f/Objects/floatobject.c#L1562-L1565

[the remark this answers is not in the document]
dime10: I just had another thought, I think the implementation is bit-width limited (besides memory restrictions on simulators), such that we can only have `n_bits < 64`. I think the reason is the use of integers to do bit-string manipulations in some places. Is it worth mentioning in the demo?

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1150 @@
...
+
+    if exponent > 0:
+        num_to_return, denom_to_return = numerator << abs_exponent, denominator
+    else:
+        num_to_return, denom_to_return = numerator, denominator << abs_exponent
+
+    return num_to_return, denom_to_return
+
+
+def phase_to_order(phase, max_denominator):
```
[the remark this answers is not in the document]
dime10: Hmm it might be kind of awkward to insert attributions in the middle of the demo code. So maybe at the bottom of the demo if it fits but I don't think it's a big deal either way, you did most of the hard work here :)

demonstrations/tutorial_shor_catalyst.py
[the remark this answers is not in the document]
dime10: Looks great!

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1424 @@
...
+#
+#    Initial circuit for order finding with quantum phase estimation (QPE).
+#
+# This high-level view hides the circuit's complexity; the implementation
+# details of :math:`U_a` are not shown, and auxiliary qubits are omitted. In
+# what follows, we'll leverage shortcuts afforded by Catalyst and the hybrid
+# nature of computation. Specifically, with mid-circuit measurement and reset we
+# can reduce the number of estimation wires to :math:`t=1`. Most arithmetic will
+# be performed in the Fourier basis. Moreover, since we know :math:`a` in
+# advance, we can vary circuit structure on the fly and save resources.
```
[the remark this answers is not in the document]
glassnotes: Yeah good point - I will adjust that!

demonstrations/tutorial_shor_catalyst.py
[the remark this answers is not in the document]
glassnotes: Oh good catch! I think we can just leave it - it's actually mentioned in the docstring.

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1424 @@
...
+def shors_algorithm(N, key, a, n_bits, n_trials):
+    # If no explicit a is passed (denoted by a = 0), randomly choose a
+    # non-trivial value of a that does not have a common factor with N.
+    if a == 0:
+        key, subkey = random.split(key)
+        a = random.randint(subkey, (1,), 2, N - 1)[0]
+
+        while jnp.gcd(a, N) != 1:
+            key, subkey = random.split(key)
+            a = random.randint(subkey, (1,), 2, N - 1)[0]
```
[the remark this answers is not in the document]
glassnotes: I like it! `jnp.gcd` is efficient anyways, and this looks much cleaner.

glassnotes:
@josh146 @dime10 All changes are now incorporated - I think we just need a check on the metadata, and a new thumbnail once available :+1: :tada:

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1434 @@
...
+#
+# .. [#Draper2000]
+#
+#     Thomas G. Draper (2000) *Addition on a Quantum Computer.* arXiv preprint,
+#     arXiv:quant-ph/0008033.
+#
+# About the author
+# ----------------
+# .. include:: ../_static/authors/olivia_di_matteo.txt
+#
```
justinpickering, line 1207: About the author section should go at the end. Like this demo: https://pennylane.ai/qml/demos/learning2learn

```suggestion
```

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1434 @@
...
+#     \mathbf{M}_{k} = \begin{pmatrix} 1 & 0 \\ 0 & e^{-2\pi i\sum_{\ell=0}^{k}  \frac{\theta_{\ell}}{2^{k + 2 - \ell}}} \end{pmatrix}.
+#
+# With a bit of index gymnastics, we obtain our final QPE algorithm with a single estimation qubit:
+#
+# .. figure:: ../_static/demonstration_assets/shor_catalyst/qpe_full_modified_power_with_qft-7.svg
+#    :width: 800
+#    :align: center
+#    :alt: QPE circuit with one estimation qubit.
+#
```
justinpickering, line 1434: Insert About the author section at the end.

```suggestion
#
# About the author
# ----------------
#
```

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1434 @@
...
+# .. figure:: ../_static/demonstration_assets/shor_catalyst/doubly-controlled-adder-with-control-not-on-qft.svg
+#    :width: 700
+#    :align: center
+#    :alt: Controlled addition of :math:`ax` using a series of double-controlled Fourier adders, controls on QFT removed.
+#
+#    The controls on the QFT can be removed without altering the effect of the circuit
+#    [#Beauregard2003]_.
+#
+# At first glance ,it not clear how this produces :math:`a x`. The qubits in
```
justinpickering, line 347: typo
```suggestion
# At first glance, it's not clear how this produces :math:`a x`. The qubits in
```

demonstrations/tutorial_shor_catalyst.metadata.json
```diff
@@ -0,0 +1,96 @@
+{
+    "title": "JIT compiling Shor's algorithm with PennyLane and Catalyst",
```
justinpickering, line 2: Typically, we avoid having acronyms in the title, unless they're very well known (like NASA :D).
See other demos that 'spell-out' JIT:
- [How to optimize a QML model using Catalyst and quantum just-in-time (QJIT) compilation](https://pennylane.ai/qml/demos/tutorial_How_to_optimize_QML_model_using_JAX_catalyst_and_Optax)
- [How to quantum just-in-time (QJIT) compile Grover’s algorithm with Catalyst](https://pennylane.ai/qml/demos/tutorial_qjit_compile_grovers_algorithm_with_catalyst)

Perhaps this can be:
- `Quantum just-in-time (QJIT) compiling Shor's algorithm with PennyLane and Catalyst`, or
- `Compiling Shor's algorithm with PennyLane and Catalyst`

We may also do without `PennyLane` in the title, since it mentions Catalyst, and this implies that PennyLane is used.

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1434 @@
...
+
+                    if candidate_q != 1:
+                        candidate_p = N // candidate_q
+
+                if candidate_p * candidate_q == N:
+                    p, q = candidate_p, candidate_q
+                    successful_trials += 1
+
+    return p, q, key, a, successful_trials / n_trials
+
```
alvaro-at-xanadu, line 889: Code blocks are so long and hard to parse 😓 Is it possible to add some comments to explain some of the key pieces (nothing too long)

alvaro-at-xanadu (changes requested):
Thank you! It looks great, just minor copy-edits. My major piece of feedback is that the code blocks are super hard to read. I was wondering if we can comment on them more?

demonstrations/tutorial_shor_catalyst.py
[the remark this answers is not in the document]
glassnotes: Sure - this conflicts with the suggested title change to remove PennyLane, though. Maybe `tutorial_shors_algorithm_catalyst` (`shors` with an s)?

demonstrations/tutorial_shor_catalyst.py
```diff
@@ -0,0 +1,1434 @@
...
+# the algorithm down into a few distinct steps to identify where quantum
+# compilation happens (for a full description of Shor's algorithm, the
+# interested reader is referred to the `PennyLane Codebook
+# <https://pennylane.ai/codebook/10-shors-algorithm/>`_).
+#
+#  - Randomly select an integer, :math:`a`, between 2 and
+#    :math:`N-1` (double check we didn't get lucky and :math:`a` has a common factor with :math:`N`)
+#  - Using :math:`N` and :math:`a`, generate a circuit for *order finding* on a quantum computer. Execute it, and use the measurement results to obtain a candidate non-trivial square root
+#  - If the square root is non-trivial, test for valid factors. Otherwise, take more measurement shots, or try a different :math:`a`.
+#
```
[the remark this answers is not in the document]
glassnotes: Do you mean from a rendering standpoint? (I wanted to have the description visually separate, whether it's in a box or not makes no difference on my side)

demonstrations/tutorial_shor_catalyst.metadata.json
glassnotes: I'm a fan of "Quantum just-in-time compiling Shor's algorithm with Catalyst"

josh146: I like @glassnotes's suggestion!

glassnotes:
@alvaro-at-xanadu @justinpickering I've incorporated your feedback - the comments didn't transfer over when I changed the filename, but I adjusted the language about Fourier adders as requested, and added a couple comments to the code (though I'm not sure how useful they are :sweat_smile: ). 

I did a quick check of the rendering locally, but won't have time to check on the build output until later today. Let me know if you'd like more updates, or if anything looks amiss!

demonstrations/tutorial_shor_catalyst.py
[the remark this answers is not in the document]
justinpickering: That sounds good, thanks!

alvaro-at-xanadu (approved):

justinpickering (approved):

PR #1036, opened 2024-02-09
Add new note from `qml.eigvals` to the tutorial on qubit tapering

Jonas-Jaeger:
**Title:** Add new note from `qml.eigvals` to the tutorial on qubit tapering

**Summary:** In order to clear some previous confusion regarding the eigensprectrum for the charged molecule HeH+ in the qubit tapering tutorial a note was added to the `qml.eigvals` (by @obliviateandsurrender). Since the confusion initially originated from reading the tutorial, it would be a great benefit, in my opinion, to add this note to the tutorial as well. 
Hence, I copied it and added a brief connection to the specific charged molecule. Adjustments to the text are welcome! 

**Relevant references:** N/A

**Possible Drawbacks:** N/A

**Related GitHub Issues:** #5095 and #5016


demonstrations/tutorial_qubit_tapering.py
```diff
@@ -185,6 +185,14 @@
 print("\nEigenvalues of H_tapered:\n", qml.eigvals(H_tapered_sparse, k=4))
 
 ##############################################################################
+# Note that a second-quantized Hamiltonian is independent of the number of electrons and its
+# eigenspectrum contains the energies of the neutral and charged molecules. Therefore, the
+# smallest eigenvalue returned by :func:`~.pennylane.qml.eigvals` for a molecular Hamiltonian
+# might not always correspond to the neutral molecule. Qubit tapering facilitates the conservation
```
soranjh, line 191: I suggest this:

"... might correspond to the neutral or charged molecules"

demonstrations/tutorial_qubit_tapering.py
```diff
@@ -185,6 +185,14 @@
 print("\nEigenvalues of H_tapered:\n", qml.eigvals(H_tapered_sparse, k=4))
 
 ##############################################################################
+# Note that a second-quantized Hamiltonian is independent of the number of electrons and its
+# eigenspectrum contains the energies of the neutral and charged molecules. Therefore, the
+# smallest eigenvalue returned by :func:`~.pennylane.qml.eigvals` for a molecular Hamiltonian
+# might not always correspond to the neutral molecule. Qubit tapering facilitates the conservation
+# of the number of electrons and the spin symmetry of the Hartree-Fock state when choosing the
+# optimal sector. Therefore, the ground-state energy of the :math:`\textrm{HeH}^+` cation is
```
soranjh, line 193: I suggest rephrasing the third sentence as this (or something similar):

"Qubit tapering allows specifying the optimal sector of the eigenvalues that correspond to the correct number of electrons."

demonstrations/tutorial_qubit_tapering.py
```diff
@@ -185,6 +185,14 @@
 print("\nEigenvalues of H_tapered:\n", qml.eigvals(H_tapered_sparse, k=4))
 
 ##############################################################################
+# Note that a second-quantized Hamiltonian is independent of the number of electrons and its
+# eigenspectrum contains the energies of the neutral and charged molecules. Therefore, the
+# smallest eigenvalue returned by :func:`~.pennylane.qml.eigvals` for a molecular Hamiltonian
+# might not always correspond to the neutral molecule. Qubit tapering facilitates the conservation
+# of the number of electrons and the spin symmetry of the Hartree-Fock state when choosing the
+# optimal sector. Therefore, the ground-state energy of the :math:`\textrm{HeH}^+` cation is
+# obtained by the smallest eigenvalue of the tapered Hamiltonian.
```
soranjh, line 194:
```suggestion
# the smallest eigenvalue of the tapered Hamiltonian.

```

soranjh (changes requested):
Thank you very much @Jonas-Jaeger for adding the note. 🙏 Left some minor suggestions.

demonstrations/tutorial_qubit_tapering.py
```diff
@@ -185,6 +185,14 @@
 print("\nEigenvalues of H_tapered:\n", qml.eigvals(H_tapered_sparse, k=4))
 
 ##############################################################################
+# Note that a second-quantized Hamiltonian is independent of the number of electrons and its
+# eigenspectrum contains the energies of the neutral and charged molecules. Therefore, the
+# smallest eigenvalue returned by :func:`~.pennylane.qml.eigvals` for a molecular Hamiltonian
```
soranjh, line 190:
```suggestion
# smallest eigenvalue returned by :func:`~.pennylane.eigvals` for a molecular Hamiltonian

```

demonstrations/tutorial_qubit_tapering.py
Jonas-Jaeger: I am a bit hesitant to formulate it in such a way that the number of electrons is specified, because I believe that only the parity is conserved (i.e., odd vs. even number of electrons). Thus, if the optimal sector is determined with respect to an HF reference state with an even (odd) number of electrons $N$, then the tapered Hamiltonian or eigenspace consists of states with an even (odd) number of electrons, but not necessarily only states with $N$ electrons. I do not consider myself an expert on this topic, so I may be missing something here.

soranjh: You can check it for `HeH+` with 1, 2, 3, 4 electrons (-1, 0, +1, +2 charge) and see that the correct eigenvalue sectors are returned.

Jonas-Jaeger: Yes it is true that it holds for `HeH+` (the eigenstates in the optimal sector all evaluate the number operator to 2 matching the HF state), however, this is not the case in general. In the following example using `H4` you can see an example where eigenstates with electron numbers show up in the optimal sector that are different from the number of electrons in the HF state. As I mentioned above, at least the parity is conserved, i.e., since the HF state has an even number of electrons, only eigenstates with an even electron number show up in this sector: 
```
import pennylane as qml
from pennylane import numpy as np

symbols = ["H", "H", "H", "H"]
geometry = np.array([[0.00000000, 0.00000000, 1*0.88],
                     [0.00000000, 0.00000000, 2*0.88],
                     [0.00000000, 0.00000000, 3*0.88],
                     [0.00000000, 0.00000000, 4*0.88],],)

H, qubits = qml.qchem.molecular_hamiltonian(symbols, geometry, charge=0)
num_particles = qml.qchem.Molecule(symbols, geometry, charge=0).n_electrons
print(f"Number of Electrons: {num_particles}")

eig_vals, eig_vecs = np.linalg.eigh(qml.matrix(H))

# Building particle number operator
num_op = qml.qchem.particle_number(qubits)

print("The following particle numbers are present in the Hamiltonians eigenstates:")
print(sorted(set(int(np.round(state.T @ qml.matrix(num_op) @ state).real) for state in eig_vecs.T)))

## output: ##
# The following particle numbers are present in the Hamiltonians eigenstates:
# [0, 1, 2, 3, 4, 5, 6, 7, 8]

# information needed for tapering
generators = qml.symmetry_generators(H)
paulixops = qml.paulix_ops(generators, qubits)
paulix_sector = qml.qchem.optimal_sector(H, generators, num_particles)

# tapering the Hamiltonian and getting the eigen-spectrum
H_tapered = qml.taper(H, generators, paulixops, paulix_sector)
eig_vals, eig_vecs = np.linalg.eigh(qml.matrix(H_tapered))

# taper the operators for making it compatible with tapered subspace
num_op_tapered = qml.taper(num_op, generators, paulixops, paulix_sector) 

print("Tapered Hamiltonian within the optimal sector: ")
print("# \t Eigenvalue \t\t #elec")
for idx, vals in enumerate(eig_vals):
    state = eig_vecs[:, idx]
    print(idx, 
          "\t", vals, 
          "\t", int(np.round(state.T @ qml.matrix(num_op_tapered) @ state).real))

## output: ##
# Tapered Hamiltonian within the optimal sector: 
# #        Eigenvalue              #elec
# 0        -1.4865952309456287     4
# 1        -0.050488808376015534   4
# 2        0.08984345463238812     4
# 3        0.5182365331484711      4
# 4        1.405562470995986       2
# 5        1.4055624709959875      2
# 6        1.6081032873890968      4
# 7        1.6779320481239317      4
# 8        1.776714813698555       4
# 9        2.700343198439295       4
# 10       2.700343198439297       4
# 11       2.700343198439298       4
# 12       3.1169267120584734      4
# 13       3.201253203779686       4
# 14       3.261058187974911       4
# 15       3.3603750676567765      4
# 16       3.507155980637822       4
# 17       4.057532336145187       2
# 18       4.057532336145194       2
# 19       4.474392851702405       6
# 20       4.474392851702412       6
# 21       4.621659114272911       4
# 22       4.82030618157768        4
# 23       4.924242424242418       0
# 24       5.144971267033627       4
# 25       6.111002072778478       4
# 26       6.445782438578079       4
# 27       6.7311088369906855      4
# 28       7.441149046148575       6
# 29       7.4411490461485785      6
# 30       8.03076359988483        4
# 31       11.376689514952242      8
```

Overall, this is why I think we should not write explicitly that tapering helps us to fix/determine the particle number. What do you think?

soranjh: You can make the statement specific for HeH+, e.g., by saying " In the case of ...".

I found the original sentence saying "Qubit tapering facilitates the conservation of the number of electrons ..." a bit general and hard to digest. That is why I suggested something simple and more clear.

Jonas-Jaeger: Thank you for the suggestion @soranjh 
I incorporated the following change:
_"While in the case of HeH+, qubit tapering allows specifying the optimal sector of the eigenvectors corresponding only to the correct number of electrons, it is generally guaranteed that the optimal sector covers all eigenvectors with the correct number of electrons, but may contain additional eigenvectors of different charge."_
Please let me know what you think and if we can finalize the PR.
Cheers,
Jonas

soranjh (approved):
Thanks @Jonas-Jaeger for adding this!

obliviateandsurrender (approved):

PR #1038, opened 2024-02-21
raise PL deprecations as errors

timmysilv:
**Context:**
we'd like to catch PL deprecation warnings as early as possible, so we can fix them well before feature freeze. We already do this with PL and plugin tests, so it's become rather standard

**Changes:**
- Raise `qml.PennyLaneDeprecationWarning`'s as errors when building demos
- update uses of `qml.matrix()` that hit the deprecated behaviour, either by adding a wire order or providing an object that has wires
- Unrelated but blocking: add `disable_new_opmath` to the end of the `tutorial_liealgebra.py`, as it enabled it at the start and this caused other demos to fail

Side note: I know we're moving away from the term "ancilla". I decided to not change it in the encounters I had in this PR because there are 162 instances of the string in the repo, but we might want to swap it out in another PR.

[sc-57573]


mudit2812 (approved):

KetpuntoG (approved):
Thanks for the update! I looks good

PR #1042, opened 2024-02-29
Update Rigetti demos

lillian542:
A few updates to the Rigetti demos:

- Pytorch Noise
   - Update Aspen-M-2 to Aspen-M-3 for the hardware version (M-2 is decommissioned)
   - Add the info about connecting to Rigetti with Docker (included in the other Rigetti demos)
- Ensemble Multi QPU
   - Move the information about connecting to the Rigetti device to _before_ the section where we try to connect to the Rigetti device
   - Change how we cast to torch during prediction phase


Mandrenkov (approved):
Looks reasonable to me! 👍

jay-selby (approved):

PR #1047, opened 2024-03-05
update dateOfLastModification for updated demos

lillian542:
We didn't know we needed to update dateOfLastModification in all the metadata files to make the demos update on the website, so we didn't do it. This PR does it for all the demos updated this release. We will need to go back and see how much this covers of past updates, but in the interest of time, here are the ones for the 0.35 release.

Here are all the demos modified for the 0.35 release: https://github.com/PennyLaneAI/qml/pull/1044


astralcai (approved):
🎸

Issue #1045, opened 2024-03-05
[BUG] QML JAX and Optax demo does not use the correct definition of optimizer


Gopal-Dahale:
#### Issue description

In the demo "How to optimize a QML model using JAX and Optax", the `update_step_jit` does not use the correct definition of the optimizer that is defined in `optimization_jit` ([Line 170](https://github.com/PennyLaneAI/qml/blob/master/demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py#L170)) as it is not passed as an argument to the `update_step_jit` function. It possibly uses the earlier definition of the optimizer ([Line 123](https://github.com/PennyLaneAI/qml/blob/master/demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py#L123)).

* *Expected behavior:* In principle, the optimizer should get passed to the `update_step_jit` function ([Line 165-189](https://github.com/PennyLaneAI/qml/blob/master/demonstrations/tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py#L165-L189))

* *Actual behavior:* The code works fine but uses the earlier definition of the optimizer.

* *System information:* Using standard google colab.

KetpuntoG:
Thanks for opening the issue, I look at it!

KetpuntoG:
Solved 😄

KetpuntoG:
The changed will be deployed in a few hours

PR #1052, opened 2024-03-08
Update tutorial_How_to_optimize_QML_model_using_JAX_and_Optax.py

KetpuntoG:
Solving issue #1045 

As indicated in the issue, the optimizer being used within the `update_step_jit` function is not the one defined in the `optimization_jit` function. I tried to pass `opt` but it gave problems so I simply defined the optimizer outside.

In the case where jit is not being applied, I was able to pass the optimizer, so I updated those functions.
Tagging @josh146 in case there is a better solution in this case


josh146 (approved):
thanks @KetpuntoG

PR #1055, opened 2024-03-13
Update items_QUBO.png

KetpuntoG:
updating one image


tarik-elkhateeb (approved):
Thanks @KetpuntoG this is the correct image :)

PR #1057, opened 2024-03-13
Fix canonical URL of force fields demo

Mandrenkov:
**Title:**

Fix canonical URL of **Symmetry-invariant quantum machine learning force fields** demo.

**Summary:**

The canonical URL of a demo must agree with its file name. In this case, `-` must be replaced with `_`.

**Relevant references:**

_N/A_

**Possible Drawbacks:**

_N/A_

**Related GitHub Issues:**

_N/A_


KetpuntoG (approved):
Thanks for the fix!

ashishks0522 (approved):

PR #1066, opened 2024-04-09
[MCM 1/3] How-to: Statistics of mid-circuit measurements

dwierichs:
**Title:**
How to collect statistics of mid-circuit measurements

**Summary:**
PennyLane supports mid-circuit measurements not only to create dynamic circuits with control flow using the measurement results, but also supports a wide range of postprocessing options of the measured values.
This how-to explains the usage of `qml.measure` and briefly compares `qml.defer_measurements` and `qml.dynamic_one_shot`. Then it shows how (not) to measure qubit correlations with MCMs, and afterwards walks through the supported postprocessing of MCM values and the allowed return functions such as `qml.expval` or `qml.counts`.
Finally, ____ is shown as an application for obtaining statistics of MCMs.

**Relevant references:**
N/A

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**
N/A

----

This how-to promotes mid-circuit measurement functionality in PennyLane, which has been extended in recent releases. In particular, it contains a comparison of the brand-new native MCM support via `dynamic_one_shot` with `defer_measurements` and can serve to advertise this new feature.

In addition, this how-to is part of the broader goal of emphasizing MCMs and dynamic circuits in our contents.

* AUDIENCE — Who is this for?

Typical how-to audience: Fundamental concepts of measurements and quantum computing are known, but the audience may not yet be familiar with PennyLane's syntax for MCMs and returning their outcomes, or they may want to consult this how-to as a reference of how to realize their circuit with MCMs.
Anyone really with the goal of coding up a circuit that returns MCM values can be in the audience.

  - Mid-circuit measurements
  - Statistics
  - [optional] postprocessing

- [ ] Tutorial
- [ ] Demo
- [x] How-to

[sc-60836]


demonstrations/tutorial_how_to_collect_mcm_stats.metadata.json
```diff
@@ -0,0 +1,43 @@
+{
+    "title": "How to collect statistics about mid-circuit measurements",
+    "authors": [
+        {
+            "id": "david_wierichs"
+        }
+    ],
+    "dateOfPublication": "2024-01-18T00:00:00+00:00",
+    "dateOfLastModification": "2024-03-08T00:00:00+00:00",
```
dwierichs, line 9: Placeholder comment as reminder to update.

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,252 @@
...
+#
+#     The bitwise operators ``&`` and ``|`` do not necessarily raise an error when used with
+#     ``float``\ s or ``np.ndarray``\ s, even if they return incorrect results!
+#
+# .. note::
+#
+#     The bitwise "xor" operator ``^`` currently is not supported but can be obtained by using
+#     ``(a | b) - (a & b)``. The Python operators ``and`` and ``or`` are not supported.
+#     They usually can be obtained using implicit conversion to integers when applying
+#     arithmetics. E.g. ``mcm0 and mcm1`` often is equivalent to ``mcm0 * mcm1``.
```
josh146, line 220: @dwierichs sorry I find this section very hard to follow!

It definitely comes a bit out of left field, and with the lack of examples I'm tempted to just skip it (as a reader).

My suggestions:

- this is something that should definitely live in the reference documentation if it does not already, and we can just link out to it.
- in the examples above, when we _do_ use MCM combination operators, we can briefly discuss it there, but it does not need this level of depth.

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,252 @@
...
+#     itself are returned.
+#   - ``qml.expval`` and ``qml.var`` are not supported for sequences of MCMs.
+#     ``qml.probs``, ``qml.sample``, and ``qml.counts`` are supported for sequences but
+#     only if they do not contain arithmetic expressions of these MCMs. That is,
+#     ``qml.sample([mcm0, mcm1, mcm2])`` is supported, ``qml.sample([mcm0 + mcm1, mcm2])``
+#     is not. You can use multiple return values instead, i.e.
+#     ``qml.sample(mcm0 + mcm1), qml.sample(mcm2)``.
+#
+# As we saw in the ``QNode`` above, MCM statistics can be returned alongside
+# standard terminal measurements.
```
josh146, line 251: This section does fit much more with the how-to! If you could slim it down slightly that would be great, otherwise this one is really helpful.

josh146 (changes requested):
Thanks @dwierichs! First half of the how-to is really great, but I struggled to follow it a bit once the MCM processing started

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,252 @@
...
+the syntax and features around mid-circuit measurements (MCMs). In this how-to, we will focus on
+extracting statistics about measurements that are performed while a quantum circuit is up and
+running --- mid-circuit measurement statistics!
+
+.. figure:: ../_static/demonstration_assets/how_to_collect_mcm_stats/socialthumbnail_large_how_to_collect_mcm_stats.png
+    :align: center
+    :width: 50%
+
+If you are interested in how to use mid-circuit measurements to create dynamic circuits
+in PennyLane, stay tuned for the related how-to on that topic appearing soon!
```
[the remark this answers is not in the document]
dwierichs: Fair enough! I'd think to edit this part in the next how-to PR, but we can also just skip it.

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+    # Measure first qubit
+    mcms.append(qml.measure(first_wire))
+
+    # Entangle all qubits with first qubit
+    for w in other_wires:
+        qml.CNOT([first_wire, w])
+
+    # Measure and reset all qubits but the first
+    for w in other_wires:
+        mcms.append(qml.measure(w))
```
josh146, line 61: Do you need to set `reset=True` here?

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+    # Rotate all qubits
+    for w, x_ in enumerate(x):
+        qml.RX(x_, w)
+
+    # Measure first qubit
+    mcms.append(qml.measure(first_wire))
+
+    # Entangle all qubits with first qubit
+    for w in other_wires:
+        qml.CNOT([first_wire, w])
```
josh146, line 57: optional suggestion: it might be easier to simply have

```python
def ansatz(x):
    mcms = []

    # Rotate all qubits
    for w, x_ in enumerate(x):
        qml.RX(x_, w)

    # Measure first qubit
    mcms.append(qml.measure(0))

    # Entangle all qubits with first qubit
    for w in range(1, 4):
        qml.CNOT([0, w])

    # Measure and reset all qubits but the first
    for w in range(1, 4):
        mcms.append(qml.measure(w))
```

That is, rather than abstract away the wires behind variable names, I think it would be fine to just be explicit (especially since you already have the comment that specifies 'first wire', etc.)

I say optional here, because I haven't read further yet and don't know if you have another reason for abstracting away wires :)

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+# bit strings sampled from the other three MCMs. To implement the ``QNode``, we also define a
+# shot-based qubit device.
+#
+
+dev = qml.device("default.qubit", shots=100)
+
+
+@qml.qnode(dev)
+def simple_node(x):
+    mcm1, *mcms2 = ansatz(x)
```
josh146, line 82:
```suggestion
    # apply the ansatz, and collect mid-circuit measurements.
    # here, mcm1 is the measurement of wire 0, and mcm2 is
    # a list of measurements of the remaining wires.
    mcm1, *mcms2 = ansatz(x)
```
optional suggestion, wasn't entirely clear to me

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+######################################################################
+# A quantum circuit with basic MCM statistics
+# -------------------------------------------
+#
+# Before we post-process the mid-circuit measurements in this ansatz or expand the ansatz itself,
+# let's construct a simple :class:`~.pennylane.QNode` and look at the statistics of the four
+# performed MCMs; We compute the probability vector for the MCM on the first qubit and count the
+# bit strings sampled from the other three MCMs. To implement the ``QNode``, we also define a
```
josh146, line 73:
```suggestion
# let's construct a simple :class:`~.pennylane.QNode` and look at the statistics of the four
# performed MCMs:
#
# 1. We compute the probability vector for the MCM on the first qubit, and
#
# 2. count the bit strings sampled from the other three MCMs.
#
# To implement the ``QNode``, we also define a
```

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+######################################################################
+# Before executing the circuit, let's draw it! For this, we sample some random  parameters, one
+# for each qubit, and call the Matplotlib drawer :func:`~.pennylane.draw_mpl`.
+#
+
+x = np.random.random(num_wires)
+fig, ax = qml.draw_mpl(simple_node)(x)
+import matplotlib.pyplot as plt
+
+plt.show()
```
josh146, line 95: Nice 😎 But shouldn't the import come before the `qml.draw()`?

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+print(f"Bit string counts on other qubits: {counts}")
+
+######################################################################
+# We see that the first qubit has a probability of about :math:`23.3\%` to be in the state
+# :math:`|1\rangle` after the rotation. We also observe that we only sampled bit strings from
+# the other three qubits for which the second and third bit are identical. (Quiz question:
+# Is this an analytic probability or just because we did not sample enough?)
+#
+# Note that we applied the ``defer_measurements`` transform above. This is for reproducibility
+# reasons and is not required in general.
```
josh146, line 112: What do you mean by this @dwierichs?

josh146: I would move this comment to right _before_ you use `defer_measurements`, so that a user understands clearly that it is optional.

It could also be worth (briefly) explaining why you use it.

---

Or preferentially, I would swap! Show the example _without_ `defer_measurements`, and then briefly introduce it, explain why it is useful, and then run it again with it.

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+
+probs, counts = qml.defer_measurements(simple_node)(x)
+print(f"Probability vector of first qubit MCM: {np.round(probs, 5)}")
+print(f"Bit string counts on other qubits: {counts}")
+
+######################################################################
+# We see that the first qubit has a probability of about :math:`23.3\%` to be in the state
+# :math:`|1\rangle` after the rotation. We also observe that we only sampled bit strings from
+# the other three qubits for which the second and third bit are identical. (Quiz question:
+# Is this an analytic probability or just because we did not sample enough?)
```
josh146, line 109: I appreciate the quiz question @dwierichs but might not be fully appropriate for this type of content 😆 

I'm trying to work out how we could keep it here. I think at the very least, we would need to provide the answer to the reader

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+#
+# Note that we applied the ``defer_measurements`` transform above. This is for reproducibility
+# reasons and is not required in general.
+#
+# Post-processing mid-circuit measurements
+# ----------------------------------------
+# We now set up a more interesting ``QNode``. It executes the ``ansatz`` from above twice and
+# compares the obtained MCMs (note that we did not define ``comparing_function`` yet, we will
+# get to that shortly):
```
josh146, line 118: I skim read past this and was definitely confused for a second 😆

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+# ``comparing_function``. We ask the following question: Is the measurement on the first qubit
+# equal between the two sets of MCMs, and do the other three measured values have the same parity?
+#
+# In contrast to quantum measurements at the end of a :class:`~.pennylane.QNode`,
+# PennyLane supports a number of unary and binary operators for MCMs even *within*
+# ``QNode``\ s. This enables us to phrase the question above as a boolean function.
+# Consider the
+# `introduction on measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html#mid-circuit-measurements-and-conditional-operations>`_
+# and the documentation if you want to learn more about the supported operations.
+#
```
josh146, line 141: 💯

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def comparing_function(first_mcms, second_mcms):
+    """A function that compares two sets of MCM outcomes."""
+    equal_first = first_mcms[0] == second_mcms[0]
+    # Computing the parity can be done with bitwise "and" with the number 1
+    first_parity = sum(first_mcms[1:]) & 1
+    second_parity = sum(second_mcms[1:]) & 1
+    equal_parity = first_parity == second_parity
+    return equal_first & equal_parity
```
josh146, line 151: This is great!

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+
+num_shots = 10000
+counts = qml.defer_measurements(interesting_qnode)(x, shots=num_shots)
+p_yes = counts[True] / num_shots
+p_no = counts[False] / num_shots
+print(f'The probability to answer with "yes" / "no" is {p_yes:.5f} / {p_no:.5f}')
+
+######################################################################
+# **This concludes our how-to on statistics and post-processing of
+# mid-circuit measurements.**
```
josh146, line 184:
```suggestion
# This concludes our how-to on statistics and post-processing of
# mid-circuit measurements.
```

josh146: This is also a good place for some CTAs (calls to action). E.g.,

```rest
This concludes our how-to on statistics and post-processing of
mid-circuit measurements. If you would like to explore mid-circuit measurement
applications, be sure to check out :doc:`demo1` and :doc:`demo2`. Or,
see all available functionality in our :doc:`quickstart <etc>`.
```

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+# Additional information
+# ----------------------
+#
+# Below, you can find some complementary information
+# on the supported return types with (post-processed) MCMs.
+# For more details consider the
+# `introduction on measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html#mid-circuit-measurements-and-conditional-operations>`_
+# and the documentation of :func:`~.pennylane.measure`.
+# For performance considerations, take a look at
+# :func:`~.pennylane.defer_measurements` and :func:`~.pennylane.dynamic_one_shot`.
```
josh146, line 195: Something that has always struck me: I don't know what `dynamic_one_shot` refers to (I see a lot of people talking about it) 😅

Maybe more a problem that's out of scope here, but here we could briefly explain these two functions before linking out to them.

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+p_yes = counts[True] / num_shots
+p_no = counts[False] / num_shots
+print(f'The probability to answer with "yes" / "no" is {p_yes:.5f} / {p_no:.5f}')
+
+######################################################################
+# **This concludes our how-to on statistics and post-processing of
+# mid-circuit measurements.**
+#
+# Additional information
+# ----------------------
```
josh146, line 187: We could make this section heading 'Appendix: supported MCM return types' (or something similar) to avoid the double nesting.

Not sure I am sold on this idea though --- happy to leave it as is :)

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+# :func:`~.pennylane.probs`,
+# :func:`~.pennylane.sample`, **and**
+# :func:`~.pennylane.counts`,
+# **are supported** with the following exceptions:
+#
+#   - ``qml.sample`` and ``qml.counts`` are not supported for ``shots=None``.
+#   - ``qml.probs`` is not supported for MCMs collected in arithmetic expressions.
+#   - ``qml.expval`` and ``qml.var`` are not supported for sequences of MCMs.
+#     ``qml.probs``, ``qml.sample``, and ``qml.counts`` are supported for sequences but
+#     only if they do not contain arithmetic expressions of these MCMs.
```
josh146, line 217:
```suggestion
# - ``qml.sample`` and ``qml.counts`` are not supported for ``shots=None``.
# - ``qml.probs`` is not supported for MCMs collected in arithmetic expressions.
# - ``qml.expval`` and ``qml.var`` are not supported for sequences of MCMs.
#   ``qml.probs``, ``qml.sample``, and ``qml.counts`` are supported for sequences but
#   only if they do not contain arithmetic expressions of these MCMs.
```

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+# For performance considerations, take a look at
+# :func:`~.pennylane.defer_measurements` and :func:`~.pennylane.dynamic_one_shot`.
+#
+# Supported MCM return types
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Depending on the processing applied to the MCM results, not all return types are supported.
+# For example, ``qml.probs(2 * mcm0)`` is not a valid return value, because it is not clear
+# which probabilities are being requested.
+# Furthermore, available return types depend on whether or not the device is
```
josh146, line 203:
```suggestion
# which probabilities are being requested.
#
# Furthermore, available return types depend on whether or not the device is
```

Just adding a line break for clarity

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,218 @@
...
+# shot-based (``qml.sample`` can not be returned if the device is not sampling).
+# Overall, **all combinations of post-processing and all of**
+# :func:`~.pennylane.expval`,
+# :func:`~.pennylane.var`,
+# :func:`~.pennylane.probs`,
+# :func:`~.pennylane.sample`, **and**
+# :func:`~.pennylane.counts`,
+# **are supported** with the following exceptions:
```
josh146, line 211:
```suggestion
# **are supported** for mid-circuit measurements with the following exceptions:
```
maybe good to be explicit here

josh146 (approved):
Really nice work @dwierichs! I think it's almost there. I've 'pre-approved' but left some additional comments and suggestions (some marked optional)

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Ah yes! Thanks!

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: I thought that in this case it might be nice to have an variable assignment: This way, the wires _are_ the integer values we give them. For `range`, I feel like more thinking has to happen to understand what exactly is going on.

I'm fine with doing it either way.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Included locally.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: We actually don't need this in the repository, this is left over from running locally. But funnily enough, we don't need to import before `draw_mpl`, because PennyLane also imports internally.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: I argued now that it is nice to use here because we can pick a larger number of shots without paying computational cost.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: I added the answer to the end of the how-to. I think it's a cute sidetrack that people may or may not want to think about. Also, it could be seen as a reminder that sampling behaviour is not necessarily representative for small shot counts :D

dwierichs: If it _is_ in the way, I can just take it out.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Should I change something here?

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: I abstractly describe the two:
>For performance considerations, take a look at
:func:`~.pennylane.defer_measurements` and :func:`~.pennylane.dynamic_one_shot`,
two simulation techniques that PennyLane uses under the hood to run circuits
like the ones in this how-to.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: No, I like the reduction! It was more useful before with the two appendix sections, but now we can just have one :)

dwierichs: It made me notice that there was some redundancy in the text as well, so now it's even more compact :)

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Ohhh, that makes so much sense :facepalm:

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+
+np.random.seed(511)
+
+first_wire = 0
+other_wires = [1, 2, 3]
+
+######################################################################
+# Now we create a quantum circuit ansatz that switches between a layer of simple rotation gates
+# (:class:`~.pennylane.RX`), mid-circuit measurements(:func:`~.pennylane.measure`), and a layer
+# of entangling two-qubit gates (:class:`~.pennylane.CNOT`) between the first and all other qubits.
```
KetpuntoG, line 40: What is important here is the definition of the ansatz rather than the libraries. I would say that the first thing is to define the circuit we are going to work with, explain it, and then merge the two code cells into one.

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+######################################################################
+# A quantum circuit with basic MCM statistics
+# -------------------------------------------
+#
+# Before we post-process the mid-circuit measurements in this ansatz or expand the ansatz itself,
+# let's construct a simple :class:`~.pennylane.QNode` and look at the statistics of the four
+# performed MCMs:
+
```
KetpuntoG, line 72:
```suggestion
#
```

demonstrations/tutorial_how_to_collect_mcm_stats.py
KetpuntoG: I am not used to this notation but it simplifies things a lot 👌

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+probs, counts = qml.defer_measurements(simple_node)(x)
+print(f"Probability vector of first qubit MCM: {np.round(probs, 5)}")
+print(f"Bit string counts on other qubits: {counts}")
+
+######################################################################
+# We see that the first qubit has a probability of about :math:`20\%` to be in the state
+# :math:`|1\rangle` after the rotation. We also observe that we only sampled bit strings from
+# the other three qubits for which the second and third bit are identical.
+# (Quiz question: Is this an analytic probability or just because we did not sample enough?
```
KetpuntoG, line 111: I was a bit confused about the question, what do you think about:

(Quiz question: Is this an expected behavior or maybe we did not sample enough?

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+    first_mcms = ansatz(x)
+    second_mcms = ansatz(-x)
+    output = comparing_function(first_mcms, second_mcms)
+    return qml.counts(output)
+
+
+######################################################################
+# Before we can run this more interesting ``QNode``, we need to actually specify the
+# ``comparing_function``. We ask the following question: Is the measurement on the first qubit
+# equal between the two sets of MCMs, and do the other three measured values have the same parity?
```
KetpuntoG, line 133: I don't think parity in this context is a well-known concept. Could you clarify?

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+#
+
+
+def comparing_function(first_mcms, second_mcms):
+    """A function that compares two sets of MCM outcomes."""
+    equal_first = first_mcms[0] == second_mcms[0]
+    # Computing the parity can be done with bitwise "and" with the number 1
+    first_parity = sum(first_mcms[1:]) & 1
+    second_parity = sum(second_mcms[1:]) & 1
+    equal_parity = first_parity == second_parity
```
KetpuntoG, line 150: Would this work?

`equal_parity = sum(first_mcms[1:]) == sum(second_mcms[1:])`

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+# Consider the
+# `introduction on measurements <https://docs.pennylane.ai/en/stable/introduction/measurements.html#mid-circuit-measurements-and-conditional-operations>`_
+# and the documentation if you want to learn more about the supported operations.
+#
+
+
+def comparing_function(first_mcms, second_mcms):
+    """A function that compares two sets of MCM outcomes."""
+    equal_first = first_mcms[0] == second_mcms[0]
+    # Computing the parity can be done with bitwise "and" with the number 1
```
KetpuntoG, line 147: A common mistake for users is to use "and" instead of "&", maybe you could put a note?

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+    second_parity = sum(second_mcms[1:]) & 1
+    equal_parity = first_parity == second_parity
+    return equal_first & equal_parity
+
+
+######################################################################
+# We can again inspect this ``QNode`` by drawing it:
+#
+
+fig, ax = qml.draw_mpl(interesting_qnode)(x)
```
KetpuntoG, line 158: I really like see this circuits 😄

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+
+######################################################################
+# This concludes our how-to on statistics and post-processing of mid-circuit measurements.
+# If you would like to explore mid-circuit measurement applications, be sure to check out
+# our :doc:`MBQC demo </demos/tutorial_mbqc>` and the
+# :doc:`demo on quantum teleportation </demos/tutorial_teleportation>`. Or, see all available functionality in our
+# `measurements quickstart page <https://docs.pennylane.ai/en/stable/introduction/measurements.html#mid-circuit-measurements-and-conditional-operations>`_.
+#
+# Appendix: Supported MCM return types
+# ------------------------------------
```
KetpuntoG, line 187: To follow the structure of other demos, I would put this part under the title "Conclusion", (or at the end of the how-to). 
Also, in this title I would remove the word "Appendix:" and start the paragraph by saying. "Before finishing, it would be good to show some final clarifications" or similar :)

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+# :func:`~.pennylane.defer_measurements` and :func:`~.pennylane.dynamic_one_shot`,
+# two simulation techniques that PennyLane uses under the hood to run circuits
+# like the ones in this how-to.
+#
+# And finally, the answer to our quiz question above: It's not an analytic property!
+# Sampling more shots eventually also gives us bit strings with differing second and third
+# bits, but they remain rare:
+
+probs, counts = qml.defer_measurements(simple_node)(x, shots=10000)
+print(f"Bit string counts on last three qubits: {counts}")
```
KetpuntoG, line 222: Here is missed:
```
About the author
----------------
```

no need the #include......., but yes the title

demonstrations/tutorial_how_to_collect_mcm_stats.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+"""
+
+######################################################################
+# Defining the circuit ansatz
+# ---------------------------
+#
+# We start with standard imports, setting a randomness seed, and specifying wires. As we
+# will treat the first wire different than all other wires, we define it as separate variable.
```
KetpuntoG, line 26:
```suggestion
# will treat the first wire differently than all other wires, we define it as separate variable.
```

KetpuntoG (commented):
Cool how-to! Thanks for working on this 🚀
Here I left some comments

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs:
```suggestion
# (Quiz question: Is this expected behaviour or did we just not sample often enough?
```

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Not quite, because we want even and odd numbers to be "equal", i.e. the condition should return `True` on `2` and `4`, for example.

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Sounds good! 

>(or at the end of the how-to)

You mean behind the quiz question's answer?

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: Good call! Modified the ordering/structure and merged the code cells. How is it now?

demonstrations/tutorial_how_to_collect_mcm_stats.py
josh146: Sorry @KetpuntoG but I would disagree here! The appendix isn't a conclusion, in the sense that a user does not need to read this to finish the how-to.

Since a how-to is focused more on readers who are _working_, the natural end point is when the last step required to achieve the objective is complete. Anything beyond that is not necessary.

Also I would add we don't need to finish every demo with a conclusion section :)

demonstrations/tutorial_how_to_collect_mcm_stats.py
josh146: Oh, more so that I had the same thought: 'wait, `comparing_function` has not yet been defined!' and only then did I notice you also said this

demonstrations/tutorial_how_to_collect_mcm_stats.py
dwierichs: I did not go for the conclusion but really moved everything else that is not part of this "appendix" to the section before. I think this should underline the structure, let me know what you think :)

demonstrations/tutorial_how_to_collect_mcm_stats.py
josh146: That's perfect, and all that's needed here 😎

demonstrations/tutorial_how_to_collect_mcm_stats.py
josh146: @dwierichs I would say the way you have it now is perfect:

image

KetpuntoG: Sounds good, let's keep it this way then

demonstrations/tutorial_how_to_collect_mcm_stats.py
KetpuntoG: Now it's great! 💪

KetpuntoG (approved):
Good job 🚀🚀🚀

PR #1067, opened 2024-04-09
Update demos to use new opmath

Qottmann:
Remove explicit `qml.operation.Tensor` instances, as well as `qml.ops.Hamiltonian` usage

branching against `dev` as new opmath will be default only in `v0.36` (scheduled for 07. May)

- [x] IMPORTANT change back final cell in `tutorial_qubit_tapering` when https://github.com/PennyLaneAI/pennylane/issues/5532 is fixed


trbromley (approved):

vincentmr (approved):
LGTM, cheers @Qottmann .

Qottmann:
Because deprecation warnings are treated as failures, I just pushed another change removing all occurences of explicit `H.ops` or `H.coeffs` calls and replaced them with `H_coeffs, H_ops = H.terms()`.

Qottmann:
[sc-53506]

Qottmann:
Currently blocked by `tutorial_qubit_tapering` due to https://github.com/PennyLaneAI/pennylane/issues/5532

Qottmann:
Just noticed a bug in another demo with the latest updates to pennylane. Should be fixed now 🤞

demonstrations/tutorial_diffable_shadows.py
```diff
@@ -421,6 +420,8 @@ def circuit():
 
         # execute qwc measurements
         dev_finite = qml.device("default.qubit", wires=range(n_wires), shots=int(shots))
+
+        @qml.transforms.hamiltonian_expand
```
Qottmann, line 424: necessary for the I() term in the Hamiltonian

demonstrations/tutorial_qubit_tapering.py
```diff
@@ -129,7 +129,7 @@
                      [0.00000000, 0.00000000,  0.87818362]])
 
 H, qubits = qml.qchem.molecular_hamiltonian(symbols, geometry, charge=1)
-print(H)
+H
```
vincentmr, line 132: This won't show anymore if one just types `python demonstrations/tutorial_qubit_tapering.py`. Is that what we want?

Qottmann:
> Looks good to me! To be sure, have we checked that all of these still run without error?

Yes, this PR is in fact building all demos due to the tag

demonstrations/tutorial_qubit_tapering.py
Qottmann: I'd say that is fine! Just fyi, the reason for this change is that the print output doesnt look as nice with new opmath because that is printing the `__str__` and not the `__repr__`

In the demo it looks as intended:

image

_static/demonstration_assets/barren_gadgets/barren_gadgets.py
```diff
@@ -28,12 +30,13 @@ def gadgetize(self, Hamiltonian, target_locality=3):
         # checking for unaccounted for situations
         self.run_checks(Hamiltonian, target_locality)
         computational_qubits, computational_locality, computational_terms = self.get_params(Hamiltonian)
+        Hamiltonian_coeffs, Hamiltonian_ops = Hamiltonian.terms()
```
trbromley, line 33: Just want to double check, is `Hamiltonian.coeffs` deprecated? I thought it wasn't.

demonstrations/tutorial_diffable_shadows.py
trbromley: Shouldn't this happen automatically? :thinking:

_static/demonstration_assets/barren_gadgets/barren_gadgets.py
Qottmann: `ops.Hamiltonian.coeffs` and `ops.LinearCombination.coeffs` are not deprecated no, afaik mainly due to technical reasons.
In many places of these demos, code that used to generate a `ops.Hamiltonian` now generates a `Sum` instance, for which it is indeed deprecated.

Additionally, my understanding was to avoid using `op.coeffs` calls in docs/demos and promote `op.terms()` for consistency and avoid confusion among users what to use (`.terms()` is always defined and not deprecated)

demonstrations/tutorial_diffable_shadows.py
trbromley: Yes, CC @Alex-Preciado @albi3ro @astralcai

_static/demonstration_assets/barren_gadgets/barren_gadgets.py
trbromley: > Additionally, my understanding was to avoid using op.coeffs calls in docs/demos and promote op.terms() for consistency and avoid confusion among users what to use (.terms() is always defined and not deprecated). Did I get that wrong?

I think that makes sense in the demos. I'm just concerned about things breaking for users in the wild.

> In many places of these demos, code that used to generate a ops.Hamiltonian now generates a Sum instance, for which it is indeed deprecated.

Oh I was expecting most things to be a `LinearCombination` :thinking: What are the common workflows that lead to `Sum`?

Is the typehint for the function in this particular file wrong then? `Hamiltonian (qml.Hamiltonian)   : target Hamiltonian to decompose`?

Qottmann: > Oh I was expecting most things to be a LinearCombination 🤔 What are the common workflows that lead to Sum?

Generators and some qchem functions. E.g. molecular_hamiltonian is creating a `Sum` now

Qottmann: > Is the typehint for the function in this particular file wrong then? Hamiltonian (qml.Hamiltonian)   : target Hamiltonian to decompose?

As far as I can see, in the demo they are constructing the "Hamiltonians" that are passed to that method via python operators, see e.g. https://github.com/PennyLaneAI/qml/blob/master/demonstrations/tutorial_barren_gadgets.py#L229

demonstrations/tutorial_diffable_shadows.py
albi3ro: Should be fixed soon: https://github.com/PennyLaneAI/pennylane/pull/5570

Qottmann: Is this PR currently a blocker for other PRs? I see some other PRs with the build-all-demos tag. If yes, we could merge this and remove the hamiltonian expands in a follow-up to unblock the others. If not, we can wait for https://github.com/PennyLaneAI/pennylane/pull/5570 to be merged

astralcai (approved):

_static/demonstration_assets/barren_gadgets/barren_gadgets.py
Qottmann: Removed the explicit hamiltonian_expand transforms now that https://github.com/PennyLaneAI/pennylane/pull/5570 is merged 🥳

demonstrations/tutorial_diffable_shadows.py
Qottmann: Removed the explicit hamiltonian_expand transforms now that https://github.com/PennyLaneAI/pennylane/pull/5570 is merged 🥳

PR #1080, opened 2024-04-24
Update block encoding demo

soranjh:
Update the demo on block encoding with matrix access oracles to:

1. Add the newly implemented `qml.FABLE` to the demo.
2. Correct an equation to allow block encoding `2 x 2` matrices.

### DIFF
--- a/demonstrations/tutorial_block_encoding.py
+++ b/demonstrations/tutorial_block_encoding.py
@@ -79,7 +79,10 @@ The angles :math:`\alpha` are obtained from the matrix elements of the matrix :m
 be obtained with the :func:`~.pennylane.templates.state_preparations.mottonen.compute_theta`
 function of PennyLane.
 
-Let's now construct the FABLE block encoding circuit for a structured matrix.
+The :class:`~.pennylane.templates.subroutines.fable.FABLE` circuit is implemented in PennyLane and
+can be easily used to block encode matrices of any given shape. Here, we manually construct the
+circuit for a structured :math:`4 \times 4` matrix.
+.
 """
 
 import pennylane as qml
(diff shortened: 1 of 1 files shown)


austingmhuang (approved):
Looks good @soranjh! Do you want to refactor this demo cause there are some lines 

e.g. `wire_order = ancilla_wires + wires_i[::-1] + wires_j[::-1]`
`M = len(A) * qml.matrix(circuit,wire_order=wire_order)().real[0:len(A),0:len(A)]`

in the code that don't work quite the same when using the template instead. With the template, we don't need to set the wire_order this way, just need
`M = len(A) * qml.matrix(circuit)().real[0 : len(A), 0 : len(A)]`

There's also code like 
`wire_map = {control_index : wire for control_index, wire in enumerate(wires_j + wires_i)}` that can be replaced with what's in the template
`wire_map = dict(enumerate(wires_j + wires_i))`

Not a super big deal and the demo works as is though.

soranjh:
> Looks good @soranjh! Do you want to refactor this demo cause there are some lines
> 
> e.g. `wire_order = ancilla_wires + wires_i[::-1] + wires_j[::-1]` `M = len(A) * qml.matrix(circuit,wire_order=wire_order)().real[0:len(A),0:len(A)]`
> 
> in the code that don't work quite the same when using the template instead. With the template, we don't need to set the `wire_order` this way, just need `M = len(A) * qml.matrix(circuit)().real[0 : len(A), 0 : len(A)]`
> 
> There's also code like `wire_map = {control_index : wire for control_index, wire in enumerate(wires_j + wires_i)}` that can be replaced with what's in the template `wire_map = dict(enumerate(wires_j + wires_i))`
> 
> Not a super big deal and the demo works as is though.

Thanks @austingmhuang, the demo codes are a bit verbose for educational purposes, so no need to change them if they run correctly.

austingmhuang (approved):

soranjh:
> * We could optionally show a circuit using `qml.Fable` and then draw the circuit using `my_circuit.tape.expand().draw()` to show the decomposed operation

Checked and confirmed locally but adding it here breaks the current nice workflow.

> * Just make sure that the link to the Fable template works well once the CI builds

Confirmed with @KetpuntoG that it will work when merged.

PR #1106, opened 2024-05-16
[WIP] [g-sim 2] (g + P)-sim

Qottmann:
This is a follow-up for https://github.com/PennyLaneAI/qml/pull/1085, our new demo on g-sim.
Here, I show some original extensions of g-sim for product operators in an efficient fashion.

[sc-62859]


Qottmann:
Merged both approaches in https://github.com/PennyLaneAI/qml/pull/1118

PR #1110, opened 2024-05-16
[TEST] Add copy button

doctorperceptron:


doctorperceptron:
Turns out this is more complex than expected. Will need to add this button on the REACT side rather than here.

PR #1111, opened 2024-05-16
Update actions using old Node version

doctorperceptron:
Many of our old actions use Node 16, which is deprecated in GitHub Actions. Bumping:
- `actions/checkout` to v4
- `actions/setup-python` to v5
- `actions/changed-files` to v44
- `actions/cache` to v3.2.2 (latest)

I have not bumped `actions/upload-artifact` or `actions/download-artifact` as v4 [has breaking API changes](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) and I will definitely mess that up.


ashishks0522 (approved):
🚀

rashidnhm (dismissed):
Thanks for taking the initiative on this!

If the goal is to remove all the warnings, we should also:
- Bump `actions/setup-python` to using `v5`
- Bump `tj-actions/changed-files` to `v44`

We should also build the demos as part of this PR just to be safe

doctorperceptron:
@rashidnhm this is ready for another look. See the note above about the upload/download artifact actions. May need you to tackle those.

PR #1122, opened 2024-05-31
Update tutorial_lcu_blockencoding.py

KetpuntoG:
In the block encoding example from [this demo](https://pennylane.ai/qml/demos/tutorial_lcu_blockencoding/), the resulting matrix does not appear correctly. 
This PR solves the problem.


DSGuala (approved):
Nice quick change :+1:

ikurecic (approved):
Seems good to me.

PR #1123, opened 2024-06-06
QROM demo

KetpuntoG:
### Before submitting

**Title:**

**Summary:**

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

----

* AUDIENCE — Who is this for?

- [ ] Tutorial
- [ ] Demo
- [ ] How-to


KetpuntoG:
[sc-63852]

demonstrations/tutorial_intro_qrom.metadata.json
```diff
@@ -0,0 +1,69 @@
...
+    "title": "Intro to QSVT",
+    "authors": [
+        {
+            "id": "guillermo_alonso"
+        }
+    ],
+    "dateOfPublication": "2023-05-23T00:00:00+00:00",
+    "dateOfLastModification": "2024-03-04T00:00:00+00:00",
+    "categories": [
+        "Algorithms",
```
KetpuntoG, line 11: The metadata I usually updated at the end when we already have the thumbnails

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+import pennylane as qml
+from functools import partial
+
+control_wires = [0,1,2]
+target_wires = [3,4]
+
+bitstrings = ["01", "11", "11", "00", "01", "11", "11", "00"]
+Ui = [qml.BasisEmbedding(int(bitstring, 2), target_wires) for bitstring in bitstrings]
+
+dev = qml.device("default.qubit", shots = 1)
```
ixfoduap, line 50: Why shots=1? Isn't this unnecessary?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+
+# This line is included for drawing purposes only.
+@partial(qml.devices.preprocess.decompose,
```
ixfoduap, line 53: What does this do to affect the drawing? I've never seen this before and as a reader it leaves me puzzled

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+##############################################################################
+# Once we have defined the circuit, we can draw it and check that the outputs are as expected.
+
+import matplotlib.pyplot as plt
+
+qml.draw_mpl(circuit, style = "pennylane")(0)
+plt.show()
+
+for i in range(8):
+    print(f"The bitstring stored in the {i}-index is: {circuit(i)}")
```
ixfoduap, line 72:
```suggestion
    print(f"The bitstring stored in index{i} is: {circuit(i)}")
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# This line is included for drawing purposes only.
+@partial(qml.devices.preprocess.decompose,
+         stopping_condition = lambda obj: False,
+         max_expansion=1)
+
+@qml.qnode(dev)
+def circuit(index):
+    qml.BasisEmbedding(index, wires=control_wires)
+    qml.Select(Ui, control=control_wires)
+    return qml.sample(wires=target_wires)
```
ixfoduap, line 61: Oh you're getting the output state by sampling? This seems very hacky. Isn't there a more elegant way to do this? :thinking:

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+
+##############################################################################
+# Once we have defined the circuit, we can draw it and check that the outputs are as expected.
+
+import matplotlib.pyplot as plt
+
+qml.draw_mpl(circuit, style = "pennylane")(0)
```
ixfoduap, line 68: Looking at the plot, I find the `BasisEmbedding` uninsightful, since it seems to indicate we do the same thing every time, even though we don't. I think it's better to explicitly construct the unitary that prepares each basis state. This is simply as it is only a sequence of X gates

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+plt.show()
+
+for i in range(8):
+    print(f"The bitstring stored in the {i}-index is: {circuit(i)}")
+
+
+##############################################################################
+# Nice, you can see that the outputs match the elements of our initial data list.
```
ixfoduap, line 76: Maybe good to write down the list here again fir easy reference

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+#
+# Although the algorithm works correctly, we can see that the number of multicontrol gates is high.
+# The decomposition of these gates is expensive and there are numerous works attempt to simplify this.
+# We can highlight the work [#unary]_ which introduces an efficient technique using measurements in the middle
+# of the circuit. Another clever approach was introduced in [#selectSwap]_ , with a smart structure known as SelectSwap.
```
ixfoduap, line 81:
```suggestion
# Although the algorithm works correctly, the number of multicontrol gates is high.
# The decomposition of these gates is expensive and there are numerous works that attempt to simplify this.
# We highlight the work [#unary]_ which introduces an efficient technique using measurements in the middle
# of the circuit. Another clever approach was introduced in [#selectSwap]_ , with a smart structure known as SelectSwap, which we describe below.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
+r"""Intro to QROM
+=============================================================
+
+Managing data is a crucial task on any computer, and quantum computers are no exception. Efficient data management is vital in Quantum Machine Learning (QML), search algorithms or state preparation.
+In this demonstration, we will introduce the concept of Quantum Read-Only Memory (QROM), a data structure designed to facilitate these tasks.
```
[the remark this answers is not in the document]
ixfoduap: In particular, we should probably mention that we have a specific QROM operation that implements it directly, and link to it already at the beginning for easy access

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+#
+# Although the algorithm works correctly, we can see that the number of multicontrol gates is high.
+# The decomposition of these gates is expensive and there are numerous works attempt to simplify this.
+# We can highlight the work [#unary]_ which introduces an efficient technique using measurements in the middle
+# of the circuit. Another clever approach was introduced in [#selectSwap]_ , with a smart structure known as SelectSwap.
+#
+# SelectSwap
+# ~~~~~~~~~~
+# The SelectSwap goal is to trade depth for the width of the circuit. That is, using auxiliary qubits,
+# reduce the number of gates required to build the QROM. We can detail the algorithm in two steps.
```
ixfoduap, line 86:
```suggestion
# The goal of the SelectSwap construction is to trade depth for width. That is, using multiple auxiliary qubits,
# we reduce the circuit depth required to build the QROM. We can detail the algorithm in two steps.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+#
+# .. figure:: ../_static/demonstration_assets/qrom/select_swap.jpeg
+#    :align: center
+#    :width: 70%
+#    :target: javascript:void(0)
+#
+# First we use the auxiliary qubits to store more than one bitstring per column.
```
ixfoduap, line 93:
```suggestion
# First, we use the auxiliary qubits to store more than one bitstring per column.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# In this way, we reduce the number of operators that Select requires.
+# The control qubits of the Select block determine in which column is the bitstring we want to load.
+# Secondly, the swap block detects the row where the searched bitstring is located and
+# swap it to the target wires.
```
ixfoduap, line 97:
```suggestion
# In this way, we reduce the number of operators that Select requires. The main idea is to organize the data in two dimensions, with each bitstring labelled by a column index and a row index. 
# The control qubits of the Select block determine the column index, while a
# swap block is used to detect the row index and swap the data to the
# the target wires.
```

ixfoduap: I like this way of presenting it; let me know what you think

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# The decomposition of these gates is expensive and there are numerous works attempt to simplify this.
+# We can highlight the work [#unary]_ which introduces an efficient technique using measurements in the middle
+# of the circuit. Another clever approach was introduced in [#selectSwap]_ , with a smart structure known as SelectSwap.
+#
+# SelectSwap
+# ~~~~~~~~~~
+# The SelectSwap goal is to trade depth for the width of the circuit. That is, using auxiliary qubits,
+# reduce the number of gates required to build the QROM. We can detail the algorithm in two steps.
+#
+# .. figure:: ../_static/demonstration_assets/qrom/select_swap.jpeg
```
ixfoduap, line 88: Move the figure to be below the text that explains it, not before

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+#    :target: javascript:void(0)
+#
+# First we use the auxiliary qubits to store more than one bitstring per column.
+# In this way, we reduce the number of operators that Select requires.
+# The control qubits of the Select block determine in which column is the bitstring we want to load.
+# Secondly, the swap block detects the row where the searched bitstring is located and
+# swap it to the target wires.
+#
+# Note that with more auxiliary qubits we could make larger groupings of bitstrings reducing the workload of the
+# Select operator.
```
ixfoduap, line 100:
```suggestion
# Select operator. Below we show an example with two columns and four rows:
```

demonstrations/tutorial_intro_qrom.py
ixfoduap: I think this could also use an step-by-step example for a specific bitstring showing how this circuits successfully implements QROM

demonstrations/tutorial_intro_qrom.py
ixfoduap: Could also use an explicit example here

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# been returned to state :math:`|0\rangle`. This can cause unwanted behaviors, so we will present the technique shown
+# in [#cleanQROM]_ to solve this.
+#
+# .. figure:: ../_static/demonstration_assets/qrom/clean_version_2.jpeg
+#    :align: center
+#    :width: 90%
+#    :target: javascript:void(0)
+#
+# Following the same idea as before, the control wires of the select block will choose the column :math:`c` where the
+# bitstring is located and those of the swap block the row :math:`r`.
```
ixfoduap, line 120:
```suggestion
# Following the same idea as before, the control wires of the select block will choose the column :math:`c` where the
# bitstring is located, and the control wires of the swap block determine the row :math:`r`.
```

ixfoduap: Should introduce the labels c and r earlier in the text

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# We can summarize the idea of why the circuit works in a few simple steps:
+#
+# 1. **We start by generating the uniform superposition on the r-th register**.
```
ixfoduap, line 123: The figure doesn't use r-register. Please use control, target, and work wires instead, or label the figure with r-wires etc

demonstrations/tutorial_intro_qrom.py
ixfoduap: I think this is missing one or two sentences explaining the intuition behind the construction. Right now it appears a bit too much out of the blue for my taste

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+#
+# 3. **We apply the Hadamard's in r-th register.**
+#
+# .. math::
+#       |c\rangle |r\rangle |b_{c0}\rangle \dots |0\rangle_r \dots |b_{cR}\rangle
+#
+# 4. **We apply Select again to the state.** Note that applying Select twice on a state is equivalent to the identity.
+#
+# .. math::
+#       |c\rangle |r\rangle |0\rangle \dots |b_{cr}\rangle_r \dots |0\rangle
```
ixfoduap, line 142: I find this description confusing. Let's chat in person so you can guide me through it

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+#
+# That's it! With a last swap we have managed to load the bitstring of column :math:`c` and row :math:`r` in the target wires.
+#
+# QROM in PennyLane
+# -----------------
+# Coding these ideas from scratch can be painful but with the help of PennyLane you can use it in just one line.
```
ixfoduap, line 148:
```suggestion
# Coding a QROM circuit from scratch can be painful, but with the help of PennyLane you can do it in just one line.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# We are going to encode the same bitstrings, using the first SelectSwap approach. In this example we use two work wires
```
ixfoduap, line 149:
```suggestion
# We are going to encode the same bitstrings, using the SelectSwap approach. In this example we use two work wires
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+target_wires = [3,4]
+work_wires = [5,6]
+
+# This function is included for drawing purposes only.
+def my_stop(obj):
+  if obj.name in ["CSWAP", "PauliX", "Hadamard"]:
+    return True
+  return False
+
+@partial(qml.devices.preprocess.decompose, stopping_condition = my_stop, max_expansion=2)
```
ixfoduap, line 164: What's the purpose of this @partial?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+@qml.qnode(qml.device("default.qubit"))
+def circuit():
+
+  qml.PauliX(control_wires[-3])
+  qml.Hadamard(control_wires[-1])
+  qml.Hadamard(control_wires[-2])
```
ixfoduap, line 170: I find this confusing, like you're trying to do too much. I suggest instead creating a more complicated QROM circuit and drawing it to show that it's correct and easy to implement with PL

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,248 @@
...
+# As a curiosity, this template works with work wires that are not initialized to zero.
+#
+#
+#
+# Conclusion
+# ----------
+#
+# By implementing various versions of the QROM operator, such as Select and SelectSwap, we optimize quantum circuits
+# for enhanced performance and scalability. Numerous studies demonstrate the efficacy of these methods in improving
+# State Preparation [#StatePrep]_ techniques by reducing the number of required gates, which I recommend you explore.
```
ixfoduap, line 216:
```suggestion
# state preparation [#StatePrep]_ techniques by reducing the number of required gates, which we recommend you explore.
```
Use "we", not "I"

ixfoduap (changes requested):
The overall structure is great, but I think there is a lot of room for improvement in the presentation

demonstrations/tutorial_intro_qrom.py
KetpuntoG: otherwise it would be a box that says "QROM".  In demos like QFT I also use it. It's not ideal but it's the only way to do it with PennyLane right now

demonstrations/tutorial_intro_qrom.py
KetpuntoG: I clarified the text

demonstrations/tutorial_intro_qrom.py
KetpuntoG: I added the example in the previous case. I think it is clear with that, so I add here a line inviting the reader to try it

demonstrations/tutorial_intro_qrom.py
KetpuntoG: I have added clarifications. Happy to talk in person if still needed

demonstrations/tutorial_intro_qrom.py
KetpuntoG: so that a box saying "QROM" does not appear

demonstrations/tutorial_intro_qrom.py
KetpuntoG: This returns the bitstring directly so it is easy to check that it is the wanted element. If we want to go by probabilities (or state) we would have to see in which position we have a 1 in the output vector and pass that index to binary. I think it might be a bit more confusing

demonstrations/tutorial_intro_qrom.py
KetpuntoG: https://github.com/PennyLaneAI/qml/pull/1123#discussion_r1663057198

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+
+.. math::
+     \begin{align}
+     \text{QROM}|00\rangle|000\rangle &= |00\rangle|011\rangle \\
+     \text{QROM}|01\rangle|000\rangle &= |01\rangle|101\rangle \\
+     \text{QROM}|10\rangle|000\rangle &= |10\rangle|111\rangle \\
+     \text{QROM}|11\rangle|000\rangle &= |11\rangle|100\rangle
+     \end{align}
+
+We will now explain three different implementations of a QROM: Select, SelectSwap, and an extension of SelectSwap.
```
soranjh, line 30:
```suggestion
We will now explain three different implementations of QROM: Select, SelectSwap, and an extension of SelectSwap.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+
+Select
+~~~~~~~
+
+More generally, Select is an operator that prepares quantum states associated with indices. It is defined as:
```
soranjh, line 35:
```suggestion
Select is an operator that prepares quantum states associated with indices. It is defined as:
```

soranjh: You might want to link to the Select operator in PennyLane.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+
+.. math::
+
+    \text{Sel}|i\rangle|0\rangle = |i\rangle U_i|0\rangle =|i\rangle|\phi_i\rangle,
```
soranjh, line 39:
```suggestion
    \text{Select}|i\rangle|0\rangle = |i\rangle U_i|0\rangle =|i\rangle|\phi_i\rangle,
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+
+where :math:`|\phi_i\rangle` is the i-th state we want to encode generated by a known-gate :math:`U_i`.
```
soranjh, line 41:
```suggestion
where :math:`|\phi_i\rangle` is the :math:`i`-th state we want to encode, generated by a known unitary :math:`U_i`.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+@partial(qml.devices.preprocess.decompose,
+         stopping_condition = lambda obj: False,
+         max_expansion=1)
+
+@qml.qnode(dev)
+def circuit(index):
+    qml.BasisEmbedding(index, wires=control_wires)
+    qml.Select(Ui, control=control_wires)
+    return qml.sample(wires=target_wires)
+
```
soranjh, line 71: The drawing is better to come right after the circuit.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+##############################################################################
+# Once we have defined the circuit, we can check that the outputs are as expected.
+# We will also draw the circuit for the particular case in which we want to know the output of index :math:`3`.
+# Note that in this case we generate in the control qubits the state :math:`|011\rangle`, representation of the
```
soranjh, line 75: This sentence is not very clear.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+~~~~~~~
+
+More generally, Select is an operator that prepares quantum states associated with indices. It is defined as:
+
+.. math::
+
+    \text{Sel}|i\rangle|0\rangle = |i\rangle U_i|0\rangle =|i\rangle|\phi_i\rangle,
+
+where :math:`|\phi_i\rangle` is the i-th state we want to encode generated by a known-gate :math:`U_i`.
+Since the bitstrings are encoded in computational basis states, we can view QROM as a special case of a Select operator.
```
soranjh, line 42: This sentence is a bit vague. You might reorder it as: QROM can be considered a special case of the Select operator where the encoded states are computational basis states. Or something similar to this. Then continue with: Then the unitaries Ui can be simply X gates ...

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+The gates :math:`U_i` would simply be :math:`X` gates that determine whether each bit is a :math:`0` or a :math:`1`.
+We use :class:`~.pennylane.BasisEmbedding` as a useful template where given the bitstrings, it places the :math:`X` gates
```
soranjh, line 44: "... as a useful template for preparing bitstrings, it places ...." seems better to me.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+in the right position. Let's use a longer string for the following example:
+
+"""
+
+import pennylane as qml
+from functools import partial
+
+bitstrings = ["01", "11", "11", "00", "01", "11", "11", "00"]
```
soranjh, line 52: Why you change the example?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+         max_expansion=1)
+
+@qml.qnode(dev)
+def circuit(index):
+    qml.BasisEmbedding(index, wires=control_wires)
+    qml.Select(Ui, control=control_wires)
+    return qml.sample(wires=target_wires)
+
+##############################################################################
+# Once we have defined the circuit, we can check that the outputs are as expected.
```
soranjh, line 73: Better if you use 000 as example, I got confused for a moment when I saw you add X, and you just mentioned using BasisEmbedding instead of adding X.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+for i in range(8):
+    print(f"The bitstring stored in index {i} is: {circuit(i)}")
+
+
+##############################################################################
+# Nice, you can see that the outputs match the elements of our initial data list: :math:`[01, 11, 11, 00, 01, 11, 11, 00]`.
+#
+# Although the algorithm works correctly, the number of multicontrol gates is high.
+# The decomposition of these gates is expensive and there are numerous works that attempt to simplify this.
+# We highlight the work [#unary]_ which introduces an efficient technique using measurements in the middle
```
soranjh, line 92:
```suggestion
# We highlight reference [#unary]_ which introduces an efficient technique using measurements in the middle
```

soranjh: It shows as reference 4, but it should be 1.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+# with each bitstring labelled by a column index :math:`c` and a row index :math:`r`.
+# The control qubits of the Select block determine the column :math:`c`, while the
+# control qubits of the swap block are used to specify the row index :math:`r`.
+#
+# .. figure:: ../_static/demonstration_assets/qrom/select_swap.jpeg
+#    :align: center
+#    :width: 70%
+#    :target: javascript:void(0)
+#
+# Let's look at an example by assuming we want to load in the target wires the bitstring with
```
soranjh, line 108: This paragraph is not easy to follow. What are b0, b1 ...? Also, seems better to stick with one example instead of using different examples in different sections.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+#    :align: center
+#    :width: 70%
+#    :target: javascript:void(0)
+#
+# Using the same example, we have that :math:`c = |1\rangle` and :math:`r = |01\rangle`. In this case, the columns are
+# determined by a single index but we need two indexes for the rows. We invite you to check that :math:`b_5` is actually
+# loaded in the target wires.
+#
+#
+# Reusable qubits
```
soranjh, line 130: This section is not easy to follow.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+# 4. **We apply Select again to the state.** Note that loading the bitstring twice in the same register leaves the state as :math:`|0\rangle`. (:math:`X^2 = \mathbb{I}`)
+#
+# .. math::
+#       |c\rangle |r\rangle |0\rangle \dots |b_{cr}\rangle_r \dots |0\rangle
+#
+# That's it! With a last swap we have managed to load the bitstring of column :math:`c` and row :math:`r` in the target wires.
+#
+# QROM in PennyLane
+# -----------------
+# Coding a QROM circuit from scratch can be painful, but with the help of PennyLane you can do it in just one line.
```
soranjh, line 170:
```suggestion
# Coding a QROM circuit from scratch can be painful, but with the help of PennyLane you can do it in just one line of code.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+# We are going to encode longer bitstrings and we will use enough work wires to group four bitstrings per column:
```
soranjh, line 171:
```suggestion
# Let's see an example where we encode longer bitstrings and we will use enough work wires to group four bitstrings per column:
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+
+bitstrings = ["01", "11", "11", "10", "01", "11", "11", "00",
+              "01", "11", "11", "11", "01", "00", "11", "00"]
+
+control_wires = [0, 1, 2, 3]
+target_wires = [4, 5]
+work_wires = [6, 7, 8, 9, 10, 11]
+
+# This function is included for drawing purposes only.
```
soranjh, line 180: You can skip the drawing, or just show QROM in the circuit to make the code more compact and easier to understand.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+# Beautiful! The circuit is more complex, but the work wires are clean.
+# As a curiosity, this template works with work wires that are not initialized to zero.
+#
+#
+#
+# Conclusion
+# ----------
+#
+# By implementing various versions of the QROM operator, such as Select and SelectSwap, we optimize quantum circuits
+# for enhanced performance and scalability. Numerous studies demonstrate the efficacy of these methods in improving
```
soranjh, line 233: You mention several studies but cite only one. You can simply say that these methods improve state prep and cite the ref.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,266 @@
...
+#       "Qubitization of Arbitrary Basis Quantum Chemistry Leveraging Sparsity and Low Rank Factorization",
+#       `Quantum 3, 208 <http://dx.doi.org/10.22331/q-2019-12-02-208>`__, 2019
+#
+# .. [#StatePrep]
+#
+#       Lov Grover and Terry Rudolph,
+#       "Creating superpositions that correspond to efficiently integrable probability distributions",
+#       `arXiv:quant-ph/0208112 <https://arxiv.org/abs/quant-ph/0208112>`__, 2002
+#
+# .. [#unary]
```
soranjh, line 259: Seems like a duplication of the first one.

demonstrations/tutorial_intro_qrom.py
KetpuntoG: Based on [this comment](https://github.com/PennyLaneAI/qml/pull/1123#discussion_r1649006413)

In text it would be a very large example (with 8 items) and in code it is too small (4 items) to visualize it

demonstrations/tutorial_intro_qrom.py
KetpuntoG: These are the operators that generate the bitstrings. I can change it to U_i for consistency. Would it be better?

demonstrations/tutorial_intro_qrom.py
KetpuntoG: Since Soran found it confusing as well, I have removed the drawing part

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+#
+# 1. **We start by generating the uniform superposition on the r-th register**. To do this, we put the Hadamard in the target wires and moved it to the :math:`r` -row with the swap block.
+#
+# .. math::
+#       |c\rangle |r\rangle |0\rangle \dots |+\rangle_r \dots |0\rangle
+#
+# 2. **We apply the Select block.** Note that in the :math:`r`-th position, the Select has no effect since the state :math:`|+\rangle` is not modified by :math:`X` gates.
+#
+# .. math::
+#       |c\rangle |r\rangle |b_{c0}\rangle \dots |+\rangle_r \dots |b_{cR}\rangle
```
EmilianoG-byte, line 149: Were we introduced to what big `R` is?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+#
+# .. figure:: ../_static/demonstration_assets/qrom/clean_version_2.jpeg
+#    :align: center
+#    :width: 90%
+#    :target: javascript:void(0)
+#
+# To see how this circuit works, let's suppose we want to load the bitstring :math:`b_{cr}`, in the target wires.
+# We can summarize the idea in a few simple steps:
+#
+# 1. **We start by generating the uniform superposition on the r-th register**. To do this, we put the Hadamard in the target wires and moved it to the :math:`r` -row with the swap block.
```
EmilianoG-byte, line 141: I think what is confusing for me is that the register `r` is called the same as the `r` row in the work wires 🤔

EmilianoG-byte: I think I get the idea that the value of the `r` register will determine the specific row on which the superposition is created, but then maybe we can differentiate the name of the row-register and its value

EmilianoG-byte, line 141:
```suggestion
# 1. **We start by generating the uniform superposition on the r-th register of the work wires**. To do this, we put the Hadamard in the target wires and move it to the :math:`r` -row with the swap block.
```

EmilianoG-byte: I think is worth emphasizing here that the specific row :math:`r` is determined by the binary representation of the row-register in the control wires. As mentioned in my comment above, I think it might be useful to make a distinction between the name of these two registers (the one controlling the row, and the actual row in the work wires)

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+# .. math::
+#       |c\rangle |r\rangle |0\rangle \dots |+\rangle_r \dots |0\rangle
+#
+# 2. **We apply the Select block.** Note that in the :math:`r`-th position, the Select has no effect since the state :math:`|+\rangle` is not modified by :math:`X` gates.
+#
+# .. math::
+#       |c\rangle |r\rangle |b_{c0}\rangle \dots |+\rangle_r \dots |b_{cR}\rangle
+#
+#
+# 3. **We apply the Hadamard's in r-th register.** The two swap blocks and the Hadamard gate in target wires achieve this.
```
EmilianoG-byte, line 152:
```suggestion
# 3. **We apply the Hadamard's in r-th register of the work wires.** The two swap blocks and the Hadamard gate in target wires achieve this.
```

EmilianoG-byte (commented):
Great demo! 🤩  Just left some comments here and there regarding clarity of the text, and particularly on the qubit reuse section.

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+# .. math::
+#       |c\rangle |r\rangle |0\rangle_0 |0\rangle_1 \dots |+\rangle_r \dots |0\rangle_{R-1}
+#
+# 2. **We apply the Select block.** Note that in the :math:`r`-th position, the Select has no effect since the state :math:`|+\rangle` is not modified by :math:`X` gates.
+#
+# .. math::
+#       |c\rangle |r\rangle |b_{c0}\rangle_0 |b_{c1}\rangle_1 \dots |+\rangle_r \dots |b_{c(R-1)}\rangle_{R-1}
+#
+#
+# 3. **We apply the Hadamard's in r-th register of the work wires.** The two swap blocks and the Hadamard gate in target wires achieve this.
```
EmilianoG-byte, line 152:
```suggestion
# 3. **We apply the Hadamard's to the  r-th register of the work wires.** The two swap blocks and the Hadamard gate applied to the target wires achieve this.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+# We highlight reference [#unary]_ which introduces an efficient technique using measurements in the middle
+# of the circuit. Another clever approach was introduced in [#selectSwap]_ , with a smart structure known as SelectSwap, which we describe below.
+#
+# SelectSwap
+# ~~~~~~~~~~
+# The goal of the SelectSwap construction is to trade depth for width. That is, using multiple auxiliary qubits,
+# we reduce the circuit depth required to build the QROM. The main idea is to organize the data in two dimensions,
+# with each bitstring labelled by a column index :math:`c` and a row index :math:`r`.
+# The control qubits of the Select block determine the column :math:`c`, while the
+# control qubits of the swap block are used to specify the row index :math:`r`.
```
EmilianoG-byte, line 97: Do you think it's worth showcasing the relation:

$2^{len(r)} \times 2^{len(c)} = N$ 

to emphasize that together with the columns and dimensions we want to be able to reach the $N$ different $b_i$'s?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+#    :width: 70%
+#    :target: javascript:void(0)
+#
+# Let's look at an example by assuming we want to load in the target wires the bitstring with
+# the index :math:`5`, i.e. :math:`b_5 = 11 `.
+# For it, we put as input in the control wires the state :math:`|101\rangle` (5 in binary), where the first two bits refer to the
+# index :math:`c = |10\rangle` and the last one to the index :math:`r = |1\rangle`.  After applying the Select block, we
+# obtain :math:`|101\rangle|01\rangle|11\rangle`, loading the bitstrings :math:`b_4` and :math:`b_5` respectively.
+# Since the third
+# control qubit (i.e., :math:`r`) is a :math:`|1\rangle`, it will activate the swap block, generating the state :math:`|101\rangle|11\rangle|01\rangle`
```
EmilianoG-byte, line 110:
```suggestion
# control qubit (i.e., :math:`|r\rangle`) is a :math:`|1\rangle`, it will activate the swap block, generating the state :math:`|101\rangle|11\rangle|01\rangle`
```

to make it consistent with the labels in the diagram

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+#
+# 1. **We start by generating the uniform superposition on the r-th register of the work wires**. To do this, we put the Hadamard in the target wires and move it to the :math:`r` -row with the swap block. We denote by :math:`R` the number of rows.
+#
+# .. math::
+#       |c\rangle |r\rangle |0\rangle_0 |0\rangle_1 \dots |+\rangle_r \dots |0\rangle_{R-1}
+#
+# 2. **We apply the Select block.** Note that in the :math:`r`-th position, the Select has no effect since the state :math:`|+\rangle` is not modified by :math:`X` gates.
+#
+# .. math::
+#       |c\rangle |r\rangle |b_{c0}\rangle_0 |b_{c1}\rangle_1 \dots |+\rangle_r \dots |b_{c(R-1)}\rangle_{R-1}
```
EmilianoG-byte, line 149: Initially this notation (with the additional sub indices) seems a bit loaded, but IMO it makes it easier to follow the transformation of the registers after each step

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,253 @@
...
+# That's it! With a last swap we have managed to load the bitstring of column :math:`c` and row :math:`r` in the target wires.
+#
+# QROM in PennyLane
+# -----------------
+# Coding a QROM circuit from scratch can be painful, but with the help of PennyLane you can do it in just one line of code.
+# Let's see an example where we encode longer bitstrings and we will use enough work wires to group four bitstrings per column:
+
+bitstrings = ["01", "11", "11", "00", "01", "11", "11", "00"]
+
+control_wires = [0, 1, 2]
```
EmilianoG-byte, line 171: To make a parallel with the explanation in the text, one can also divide the control wires as:

```suggestion
column_wires = [0,1]
row_wires = [2]
control_wires = column_wires + row_wires
```

KetpuntoG: Good idea! However I'll leave it as it is because I hope the demo will help people to know how to use the template (and that would not be the expected use)

demonstrations/tutorial_intro_qrom.py
KetpuntoG: Let me see if I can fit it in somewhere

demonstrations/tutorial_intro_qrom.py
soranjh: This is what the comment suggests:
QROM |00> |00> = |00>|11>
QROM |01> |00> = |01>|01>
QROM |10> |00> = |10>|11>
QROM |11> |00> = |11>|00>

But the demo has:

QROM |00> |000> = |00>|011>

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+QROM
+-----
+
+The QROM is an operator that allows us to load classical data into a quantum computer. Data is represented as a collection of bitstrings (list of 0s and 1s) that we denote by :math:`b_0, b_2, \ldots, b_{N-1}`. The QROM operator is then defined as:
+
+.. math::
+
+    \text{QROM}|i\rangle|0^{\otimes m}\rangle = |i\rangle|b_i\rangle,
+
+where :math:`|b_i\rangle` is the bitstring associated with the :math:`i`-th computational basis and :math:`m` is the length of the bitstrings. We have assumed all the bistrings are of equal length.
```
austingmhuang, line 23:
```suggestion
where :math:`|b_i\rangle` is the bitstring associated with the :math:`i`-th computational basis and :math:`m` is the length of the bitstrings. We have assumed all the bitstrings are of equal length.
```
Typo

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
+r"""Intro to QROM
+=============================================================
+
+Managing data is a crucial task on any computer, and quantum computers are no exception. Efficient data management is vital in quantum machine learning, search algorithms, and state preparation.
+In this demonstration, we will discuss the concept of a Quantum Read-Only Memory (QROM), a data structure designed to load classical data on a quantum computer.
+You will also see how easy it is to use this operator in PennyLane through the :class:`~.pennylane.QROM` template.
```
ixfoduap, line 6:
```suggestion
You will also see how easy it is to use this operator in PennyLane using the :class:`~.pennylane.QROM` template.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+
+.. figure:: ../_static/demonstration_assets/qrom/qrom_opengraph.png
+    :align: center
+    :width: 50%
+    :target: javascript:void(0)
+
+QROM
+-----
+
+The QROM is an operator that allows us to load classical data into a quantum computer. Data is represented as a collection of bitstrings (list of 0s and 1s) that we denote by :math:`b_0, b_2, \ldots, b_{N-1}`. The QROM operator is then defined as:
```
ixfoduap, line 17:
```suggestion
The QROM is an operator that allows us to load classical data into a quantum computer. Data is represented as a collection of bitstrings (list of 0s and 1s) that we denote by :math:`b_0, b_1, \ldots, b_{N-1}`. The QROM operator is then defined as:
```
ATTENTION TO DETAILS! How did multiple reviewers miss this?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+
+.. math::
+
+    \text{QROM}|i\rangle|0^{\otimes m}\rangle = |i\rangle|b_i\rangle,
+
+where :math:`|b_i\rangle` is the bitstring associated with the :math:`i`-th computational basis and :math:`m` is the length of the bitstrings. We have assumed all the bistrings are of equal length.
+
+For example, suppose our data consists of eight bitstrings, each with two bits: :math:`[01, 11, 11, 00, 01, 11, 11, 00]`. Then, the index register will consist of three
+qubits (:math:`3 = \log_2 8`) and the target register of two qubits (e.g., :math:`m = 2`). For instance, for the
```
ixfoduap, line 26:
```suggestion
qubits (:math:`3 = \log_2 8`) and the target register of two qubits (:math:`m = 2`). For instance, for the
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+first four indices, the QROM operator acts as:
+
+.. math::
+     \begin{align}
+     \text{QROM}|000\rangle|00\rangle &= |000\rangle|01\rangle \\
+     \text{QROM}|001\rangle|00\rangle &= |001\rangle|11\rangle \\
+     \text{QROM}|010\rangle|00\rangle &= |010\rangle|11\rangle \\
+     \text{QROM}|011\rangle|00\rangle &= |011\rangle|00\rangle
```
ixfoduap, line 34:
```suggestion
     \text{QROM}|011\rangle|00\rangle &= |011\rangle|00\rangle.
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+
+:class:`~.pennylane.Select` is an operator that prepares quantum states associated with indices. It is defined as:
+
+.. math::
+
+    \text{Select}|i\rangle|0\rangle = |i\rangle U_i|0\rangle =|i\rangle|\phi_i\rangle,
+
+where :math:`|\phi_i\rangle` is the :math:`i`-th state we want to encode, generated by a known unitary :math:`U_i`.
+QROM can be considered a special case of the Select operator where the encoded states are computational basis states.
+Then the unitaries :math:`U_i` can be simply :math:`X` gates that determine whether each bit is a :math:`0` or a :math:`1` satisfying:
```
ixfoduap, line 50:
```suggestion
Then the unitaries :math:`U_i` can be simply products of :math:`X` gates satisfying:
```
Your previous wording made it sound like the U_i are themselves X gates, which is not the case

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+
+.. math::
+
+    U_i|0\rangle =|b_i\rangle.
+
+We use :class:`~.pennylane.BasisEmbedding` as a useful template for preparing bitstrings, it places the :math:`X` gates
```
ixfoduap, line 56: In light of the upcoming changes to state preparation, do we still want to use basis embedding here?

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+in the right position. Let's see how it could be implemented:
```
ixfoduap, line 57:
```suggestion
We use :class:`~.pennylane.BasisEmbedding` as a useful template for implementing the gates :math:`U_i`. Let's see how it could be written in code:
```

demonstrations/tutorial_intro_qrom.py
ixfoduap: Any way we can expand the BasesEmbedding blocks to show the X gates instead? Because that would be better

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+
+##############################################################################
+# Now we can check that all the outputs are as expected:
+
+for i in range(8):
+    print(f"The bitstring stored in index {i} is: {circuit(i)}")
+
+
+##############################################################################
+# Nice, the outputs match the elements of our initial data list: :math:`[01, 11, 11, 00, 01, 11, 11, 00]`.
```
ixfoduap, line 97:
```suggestion
# The outputs match the elements of our initial data list: :math:`[01, 11, 11, 00, 01, 11, 11, 00]`. Nice!
```
Just a suggestion

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+control_wires = [0,1,2]
+target_wires = [3,4]
+work_wires = [5,6]
+
+@qml.qnode(dev)
+def circuit(index):
+    qml.BasisEmbedding(index, wires=control_wires)
+    qml.QROM(bitstrings, control_wires, target_wires, work_wires, clean = False)
+    return qml.sample(wires=control_wires + target_wires + work_wires)
+
```
ixfoduap, line 138: I find this a bit random here at the beginning without any explanation. Consider moving it to a later part after the main explanation

demonstrations/tutorial_intro_qrom.py
ixfoduap: This explanation is indeed not very clear. @KetpuntoG You can do better! Take another look and figure out how to explain this more clearly

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+#
+# Note that with more auxiliary qubits we could make larger groupings of bitstrings reducing the depth of the
+# Select operator. Below we show an example with two columns and four rows:
+#
+# .. figure:: ../_static/demonstration_assets/qrom/select_swap_4.jpeg
+#    :align: center
+#    :width: 70%
+#    :target: javascript:void(0)
+#
+# The QROM template will put as many rows as possible using the ``work_wires`` we pass.
```
ixfoduap, line 183: Good place for another code block showing how to implement it. For this code, it would be good to see a circuit diagram from the code that matches the schematic one in the drawing

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+#
+#
+# Reusable qubits
+# ~~~~~~~~~~~~~~~~ 
+#
+# The above approach has a drawback. The work wires have been altered, i.e., after applying the operator they have not
+# been returned to state :math:`|00\rangle`. This can cause unwanted behaviors but in PennyLane can be easily solved
```
ixfoduap, line 190:
```suggestion
# been returned to state :math:`|00\rangle`. This can cause unwanted behaviors, but in PennyLane can be easily solved
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+    qml.QROM(bitstrings, control_wires, target_wires, work_wires, clean=True)
+    return qml.sample(wires=target_wires + work_wires)
+
+for i in range(8):
+    print(f"The bitstring stored in index {i} is: {circuit(i)[:2]}")
+    print(f"The work wires for that index are in the state: {circuit(i)[2:4]}\n")
+
+
+##############################################################################
+# Great! To achieve this, we have followed the technique shown in [#cleanQROM]_ where the proposed circuit is as follows:
```
ixfoduap, line 213:
```suggestion
# Great! We confirm that all the work wires have been reset. To achieve this, we have followed the technique shown in [#cleanQROM]_ where the proposed circuit is as follows:
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+#
+# .. figure:: ../_static/demonstration_assets/qrom/clean_version_2.jpeg
+#    :align: center
+#    :width: 90%
+#    :target: javascript:void(0)
+#
+# where :math:`R` the number of rows. To see how this circuit works, let's suppose we want to load the bitstring :math:`b_{cr}` in the target wires, where :math:`b_{cr}`
```
ixfoduap, line 220:
```suggestion
# where :math:`R` is the number of rows. To see how this circuit works, let's suppose we want to load the bitstring :math:`b_{cr}` in the target wires, where :math:`b_{cr}`
```

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,289 @@
...
+#
+#
+# Conclusion
+# ----------
+#
+# By implementing various versions of the QROM operator, such as Select and SelectSwap, we optimize quantum circuits
+# for enhanced performance and scalability. These methods improve the efficiency of
+# state preparation [#StatePrep]_ techniques by reducing the number of required gates, which we recommend you explore.
+# As the availability of qubits increases, the relevance of these methods will grow making this operator an
+# indispensable tool for developing new algorithms and an interesting field for further study.
```
ixfoduap, line 259:
```suggestion
# state preparation [#StatePrep]_ techniques. After all, state preparation is a special case of data encoding, where the data are the coefficients that define the state.
# As the availability of qubits increases, QROM methods will become more attractive as a practical due to their superior asymptotic efficiency. 
# This makes them an indispensable tool for developing new algorithms.
```

ixfoduap (changes requested):
Overall this looks great, but the explanation of the SelectSwap construction, arguably the most important part of the demo, needs to be improved as it is currently not very clear

demonstrations/tutorial_intro_qrom.py
KetpuntoG: I will be able to upgrade it to `BasisState`, although `BasisEmbedding` will still exist as an alias for `BasisState`

demonstrations/tutorial_intro_qrom.py
KetpuntoG: The objective of this change is that the user who simply wants to learn how to use it, does not get lost with the theory. Let me see how I can readjust it

demonstrations/tutorial_intro_qrom.py
KetpuntoG: Ok, I will update it with an example so the number 3 is encoded with two PauliX gates

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,288 @@
...
+#
+# 4. **Select block is applied.** Note that loading the bitstring twice in the same register leaves the state as :math:`|0\rangle`. (:math:`X^2 = \mathbb{I}`)
+#
+# .. math::
+#       |c\rangle |r\rangle |0\rangle |0\rangle \dots |b_{cr}\rangle_r \dots |0\rangle.
+#
+# 5. **Swap block is applied.** With this, we move :math:`b_{cr}` that is encoded in the :math:`r`-row to the target wires.
+#
+# .. math::
+#       |c\rangle |r\rangle |b_{cr}\rangle |0\rangle \dots |0\rangle_r \dots |0\rangle.
```
gmlejarza, line 248: I get the general idea of applying Hadamard gates and then a combination of controlled SWAPs and SELECT blocks to make sure the work_wires are cleaned. However, I really think an example after this could be great to fully grasp everything that is going on, because the math is not trivial. Maybe you can even use the same example as in the previous section, so everything is more self-contained. Usually, what I like about Pennylane demos is that after a very technical explanation, from which I understand let's say 50-70%, there's a particular example, and after going through the example, I can go back to the technical explanation and understand up to 90-100%.

KetpuntoG: I hope that with the clarifications in the previous section now it is easier to follow

demonstrations/tutorial_intro_qrom.py
```diff
@@ -0,0 +1,288 @@
...
+##############################################################################
+# Great! We confirm that all the work wires have been reset. To achieve this, we have followed the technique shown in [#cleanQROM]_ where the proposed circuit is as follows:
+#
+# .. figure:: ../_static/demonstration_assets/qrom/clean_version_2.jpeg
+#    :align: center
+#    :width: 90%
+#    :target: javascript:void(0)
+#
+# where :math:`R` is the number of rows. To see how this circuit works, let's suppose we want to load the bitstring :math:`b_{cr}` in the target wires, where :math:`b_{cr}`
+# is the bitstring whose operator :math:`U` is placed in the c-th column and r-th row in the two dimensional representation shown in the Select block.
```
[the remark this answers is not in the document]
KetpuntoG: Added an extra image

josh146:
@KetpuntoG I think this can be changed to merge into the `master` branch now

KetpuntoG:
> @KetpuntoG I think this can be changed to merge into the `master` branch now

This includes a lot of changed files  and can complicate the review. 
I will wait for Juan Miguel's review and then I will make the change to master to avoid problems
Do you know of a way to rebase and not modify extra files @josh146 ?

KetpuntoG:
Reopened only to continue the review here

mudit2812:
> > @KetpuntoG I think this can be changed to merge into the `master` branch now
> 
> This includes a lot of changed files and can complicate the review. I will wait for Juan Miguel's review and then I will make the change to master to avoid problems Do you know of a way to rebase and not modify extra files @josh146 ?

This might require some fancy git commands. I don't have any suggestions other than using stack overflow or other forums/tools of the sort.

PR #1125, opened 2024-06-07
Fix for installation of torch and torchvision on macOS

austingmhuang:
**Title:**
Fix for installation of `torch` and `torchvision` on macOS

**Summary:**
After installing poetry (version 1.8.3) on macOS and running `make environment`, you get the following error. Fixing this error will lead to a similar one for `torchvision`.
```
  - Installing torch (1.13.1+cpu): Failed

  RuntimeError

  Unable to find installation candidates for torch (1.13.1+cpu)

  at ~/.local/pipx/venvs/poetry/lib/python3.12/site-packages/poetry/installation/chooser.py:74 in choose_for
       70│ 
       71│             links.append(link)
       72│ 
       73│         if not links:
    →  74│             raise RuntimeError(f"Unable to find installation candidates for {package}")
       75│ 
       76│         # Get the best link
       77│         chosen = max(links, key=lambda link: self._sort_key(package, link))
       78│ 
 ```

This is because the installer is using the `+cpu` version (despite the fact that the non `+cpu` version is in `pyproject.toml`). Specifying the source seems to fix the issue and the installer proceeds without error.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### CODE REVIEW

File: pyproject.toml [hunk 1]
```diff
@@ -70,11 +70,11 @@ stim = "*"
 
 # Install a difference version of torch from PyPI as the one from PyTorch repo is not compatible with MacOS
 torch = [
-    { version = "1.13.1", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" },
+    { version = "1.13.1", source="PyPI", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" },
```
rashidnhm, line 73:
```suggestion
    { version = "1.13.1", source="pypi", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" },
```

pyproject.toml
```diff
@@ -70,11 +70,11 @@ stim = "*"
...
     { version = "1.13.1+cpu", source = "pytorch-cpu", markers = "platform_machine == 'x86_64'" }
 ]
 torchvision = [
-    { version = "0.14.1", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" },
+    { version = "0.14.1", source="PyPI", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" },
```
rashidnhm, line 77:
```suggestion
    { version = "0.14.1", source="pypi", markers = "sys_platform == 'darwin' and platform_machine == 'arm64'" },
```

rashidnhm (approved):
Thanks for taking a look into this. After some investigation this does seem to be the solution [as outlined by a poetry contributor here](https://github.com/python-poetry/poetry/issues/7748#issuecomment-1491799076).

To me this seems to be a limitation (dare I even say bug) of poetry since pypi is supposed to be the default search location anyways for packages. But alas, glad it got sorted

ikurecic (approved):
Okay, this makes sense to me, let's merge it in.
Thanks a lot, @austingmhuang !

PR #1133, opened 2024-06-17
How to use the new `default.tensor` quantum device

PietropaoloFrisoni:
### Before submitting

- [x] Ensure that your tutorial executes correctly, and conforms to the

- [x] Remember to do a grammar check of the content you include.
- [x] All tutorials conform to

**Title:** How to simulate quantum circuits with tensor networks using DefaultTensor

**Summary:** This how-to guide shows the first steps to use the new ``default.tensor`` device that will be available to users in PennyLane 0.37

**Relevant references:** To be discussed with reviewer(s)

**Possible Drawbacks:** None that I can think of, except for package dependencies

**Related Shortcut Stories:** [sc-65345]

----

* AUDIENCE — Who is this for?

- [ ] Tutorial
- [ ] Demo
- [x] How-to


demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,181 @@
...
+# Since we did not specify the number of qubits when instantiating the device,
+# the number of tensors in the tensor network is inferred from the last execution of the quantum circuit.
+# Let's visualize the MPS representation of the quantum circuit with 12 qubits.
+#
+
+circuit(theta, phi, num_qubits=12)
+dev.draw(color="auto", show_inds=True, figsize=(8, 6))
+
+##############################################################################
+# .. figure:: ../_static/demonstrations_assets/how_to_simulate_quantum_circuits_with_tensor_networks/MPS_circuit.png
```
ikurecic, line 105:
```suggestion
# .. figure:: ../_static/demonstration_assets/how_to_simulate_quantum_circuits_with_tensor_networks/MPS_circuit.png
```

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,181 @@
...
+# As before, let's visualize the tensor network representation of the circuit with 15 qubits.
+# We can see that the Tensor Network (TN) method generates a more complex tensor network than the Matrix Product State (MPS) method.
+# In this case, the representation depends on the structure of the quantum circuit and the specified keyword arguments.
+#
+
+circuit(theta, depth, num_qubits=15)
+dev.draw(color="auto", show_inds=True, figsize=(7, 7))
+
+##############################################################################
+# .. figure:: ../_static/demonstrations_assets/how_to_simulate_quantum_circuits_with_tensor_networks/TN_circuit.png
```
ikurecic, line 173:
```suggestion
# .. figure:: ../_static/demonstration_assets/how_to_simulate_quantum_circuits_with_tensor_networks/TN_circuit.png
```

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+# Choosing the method to simulate quantum circuits
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The ``default.tensor`` device can simulate quantum circuits using two different computational methods.
+# We only need to specify the ``method`` keyword argument when instantiating the device.
+#
+# The first is the Matrix Product State (MPS) representation, and the second is the Tensor Network (TN) approach.
+# The MPS representation can be seen as a particular case of the TN method, where the tensor network has a one-dimensional structure.
+# If not specified, the default method is the MPS representation.
+#
```
Qottmann, line 28: I know we are going to do more in-depth demos on these two simulations techniques, but I think it would still be good to explain with another paragraph and ideally a figure to highlight the difference between the two methods

You could, for example, mention that MPS is much closer to a state vector simulator with the difference that the underlying state description is in form on an MPS, and that a full contraction schemes always contracts the full psi0 - U - H - U - psi0 expectation value network

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+"""
+
+######################################################################
+# Choosing the method to simulate quantum circuits
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The ``default.tensor`` device can simulate quantum circuits using two different computational methods.
+# We only need to specify the ``method`` keyword argument when instantiating the device.
+#
+# The first is the Matrix Product State (MPS) representation, and the second is the Tensor Network (TN) approach.
```
Qottmann, line 25:
```suggestion
# The first is the Matrix Product State (MPS) representation, and the second is the full contraction approach using a Tensor Network (TN).
```

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+Tensor networks are a powerful computational tool for simulating quantum circuits.
+They provide a way to represent quantum states and operations in a compact form. 
+Unlike the state vector approach, tensor networks are particularly useful for large-scale simulations of quantum circuits.
+
+Here, we demonstrate how to simulate quantum circuits using the ``default.tensor`` device in PennyLane.
+This simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations. 
+The ``default.tensor`` device is convenient for simulations with tens, hundreds, or even thousands of qubits.
+Other devices based on the state vector approach may be more suitable for small circuits 
+since the overhead of tensor network contractions can be significant.
+
```
Qottmann, line 13: Personal opinion: we can already put a disclaimer that this is work-in-progress development and list capabilities and limitations

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+# We can also visualize the tensor network representation of the quantum circuit with the ``draw`` method.
+# This method generates a graphical representation of the tensor network using the ``quimb``'s plotting functionalities.
+#
+# Since we did not specify the number of qubits when instantiating the device,
+# the number of tensors in the tensor network is inferred from the last execution of the quantum circuit.
+# Let's visualize the MPS representation of the quantum circuit with 12 qubits.
+#
+
+circuit(theta, phi, num_qubits=12)
+dev.draw(color="auto", show_inds=True, figsize=(7, 5))
```
Qottmann, line 100: I am bit puzzled by this functionality now. How can the draw method be a device method? Is it plotting the last execution? I find this a bit confusing, I would assume that I can draw a circuit, not the device.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+#
+
+######################################################################
+# Simulating a quantum circuit with the MPS method
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Let's start by showing how to simulate a quantum circuit using the Matrix Product State (MPS) method.
+# If the number of wires is not specified when instantiating the device, it is inferred from the circuit at runtime.
+#
+# We choose the maximum bond dimension as ``None``, meaning the bond dimension is not restricted.
```
Qottmann, line 37: I think showing using `None` is not a good idea because we dont want to teach people to do that, they should use a bond dimension and be able to learn from this demo how to set it and get some examples.

Note that because your example circuit has only one layer the maximum required bond dimension should be <= 10 (a CNOT is x2 but not sure what the two multi-qubit gates add)

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+#
+
+import pennylane as qml
+import numpy as np
+
+# Define the keyword arguments for the MPS method
+kwargs_mps = {
+    "max_bond_dim": None,
+    "cutoff": np.finfo(np.complex128).eps,
+    "contract": "auto-mps",
```
Qottmann, line 48: would be good to comment on this choice here

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+#
+
+######################################################################
+# Simulating a quantum circuit with the TN method
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The Tensor Network (TN) method is a more general approach than the Matrix Product State (MPS) method.
+# It can be more efficient for simulating quantum circuits when the MPS method
+# may require a large bond dimension to accurately represent the state, leading to increased computational and memory costs.
+# For example, this method can be helpful in simulating circuits with a higher degree of entanglement.
```
Qottmann, line 120: This can be a bit misleading because it sounds like highly entangled circuits are no problem for full contraction schemes. They can also invoke high computational and memory costs. I would rephrase that a bit to make those subtleties more clear.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+#
+
+import pennylane as qml
+import numpy as np
+
+# Define the keyword arguments for the TN method
+kwargs_tn = {
+    "contract": "auto-split-gate",
```
Qottmann, line 128: needs comment

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+    "local_simplify": "ADCRS",
```
Qottmann, line 129: needs comment

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+    "contraction_optimizer": "auto-hq",
```
Qottmann, line 130: also

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+dev.draw(color="auto", show_inds=True, figsize=(6, 6))
+
+##############################################################################
+# .. figure:: ../_static/demonstration_assets/how_to_simulate_quantum_circuits_with_tensor_networks/TN_circuit.png
+#    :align: center
+#    :width: 90%
+
+######################################################################
+# We can see that the Tensor Network (TN) method generates a more complex tensor network than the Matrix Product State (MPS) method.
+# In this case, the representation depends on the structure of the quantum circuit and the specified keyword arguments.
```
Qottmann, line 181: Isnt that because in the MPS case we just see the final MPS after all the contractions? I dont get this point 🤔

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+#
```
Qottmann, line 182: It would be good to equip the user with some tips and tricks on how to set the kwargs of this device. What do the different settings do, when is it useful?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+
+######################################################################
```
Qottmann, line 184: conclusion section highlighting in which scenarios this device can be useful, where a user needs to be cautious

Qottmann (changes requested):
Thanks for the quick turnaround @PietropaoloFrisoni !

Currently, this reads more like a documentation "usage details" section. Here in a How-To demo, we can expand much more and equip the user with all the useful tipps and tricks to successfully use this device. I left some comments and suggestions on what to expand.

Further, it would be crucial to show a bond dimension scaling in the MPS section. That is, take some circuit expectation value and scale the bond dimension from something small to something large. The expectation value should then converge to the true value once a suitably large circuit is found. With this you can well describe the function of the bond dimension and how to set it.

For that, I suggest a reasonably deep circuit with 19/20 qubits to compare to the exact result. You can plot max bond dim (x) vs expectation value (y) on a (loglog) plot (not sure, but I think it should yield a power law scaling, so straight line in loglog).

(if you are very eager you can also plot versus 1/bond_dim and extrapolate to the bond_dim=inf limit (at 0).)

PietropaoloFrisoni:
Thanks @Qottmann. Before addressing each point, I see that you asked to comment on the specified kwargs. I actually did it at first, but then I removed the explanations since this was one of the guidelines:

_It shouldn't teach or explain, it should assume the reader already understands what X, Y, and Z are. They just want to see the code to remember how to do it in PL._ 

I can insert an explanation for each of the keyword arguments I used, but I am not sure if this is the right thing to do in such a short guide

Qottmann:
> _It shouldn't teach or explain, it should assume the reader already understands what X, Y, and Z are. They just want to see the code to remember how to do it in PL._
> 
> I can insert an explanation for each of the keyword arguments I used, but I am not sure if this is the right thing to do in such a short guide

I think the teaching and explaining refers to the underlying theory. I agree that in-depth explanations of tensor network methods are better suited for the upcoming demos on that, but it would still be crucial and valuable to explain to the user what those settings that you set do.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
trbromley: :thinking: What capabilities/limitations are you thinking of?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+# Simulating a quantum circuit with the MPS method
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Let's start by showing how to simulate a quantum circuit using the Matrix Product State (MPS) method.
+# If the number of wires is not specified when instantiating the device, it is inferred from the circuit at runtime.
+#
+# We choose the maximum bond dimension as ``None``, meaning the bond dimension is not restricted.
+# The ``cutoff`` parameter is set to the machine epsilon of the ``numpy.complex128`` data type.
+#
```
trbromley, line 39: We could point to the device docstrings to understand more about what these arguments do. Eventually we could point to the upcoming tutorial content.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
trbromley: Are these just the defaults?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+import time
+
+# Simulate the circuit for different numbers of qubits
+for num_qubits in range(50, 201, 50):
+    print(f"Number of qubits: {num_qubits}")
+    start_time = time.time()
+    result = circuit(theta, phi, num_qubits)
+    end_time = time.time()
+    print(f"Result: {result}")
+    print(f"Execution time: {end_time - start_time:.4f} seconds")
```
trbromley, line 87:
```
Execution time: 7.2229 seconds
Execution time: 0.9611 seconds
Execution time: 1.6116 seconds
Execution time: 2.5679 seconds
```
I'm confused. Why is the first execution slower?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
trbromley, line 100: Why do we need to use a file for the drawing rather than an automatically generated plot?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,188 @@
...
+#    :align: center
+#    :width: 90%
+
+######################################################################
+# We can see that the Tensor Network (TN) method generates a more complex tensor network than the Matrix Product State (MPS) method.
+# In this case, the representation depends on the structure of the quantum circuit and the specified keyword arguments.
+#
+
+######################################################################
+# About the author
```
trbromley, line 185: Syntax is not quite right here.
image

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
trbromley: This could be useful, but please can we keep it quite brief, e.g., just a few sentences? We're aiming for the how-to's to be short recipes.

trbromley (commented):
Thanks @PietropaoloFrisoni

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: > You could, for example, mention that MPS is much closer to a state vector simulator with the difference that the underlying state description is in form on an MPS, and that a full contraction schemes always contracts the full psi0 - U - H - U - psi0 expectation value network

Based on the guidelines and suggestions that @trbromley and @josh146 provided, I think that this explanation of the underlying theory might be optimal in a different context, for example [this demo](https://app.shortcut.com/xanaduai/epic/64218?group_by=none&ct_workflow=all&cf_workflow=500000005). When the latter has been completed, we can always add a link/reference to this how-to 👍

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I inserted a disclaimer (specifying that this is a work in progress) in a final remark section

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I agree, I changed the `max_bond_dim` to be 50 specifying that this is more than enough for this circuit

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Except for `"contract": "auto-mps"`, all the kwargs shown in the latest version of this how-to are different from the default values.

Based on the conversations we had on Slack, I mark this conversation and the remaining ones regarding keyword argument descriptions as resolved (we can chat about them separately in a Slack thread)

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Yes, exactly. I added a comment about this after the graphical representation of the MPS. Hopefully this is more clear now, but please tell me if this is not the case

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I added some of these comments as a few sentences in a final remark 👍

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Because I didn't know it was possible XD.

I removed the plot files, now figures are automatically generated. Thanks!

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: This is most likely due to the initial overhead in `quimb`. I added a sentence to clarify that

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Good point. I added a few sentences to comment on those choices and linked the `default.tensor` documentation.  I agree that we can add a link to the tutorial comment

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Same as in the above conversation, I added a few comments to each kwarg and the final remark section. Based on recommendations received on Slack, I think we should not go deeper in the description in such a context

PietropaoloFrisoni:
Thanks for the suggestions @Qottmann, @trbromley, and @josh146 (I tagged you all here even though some of you provided feedback on Slack).

I did my best to find a compromise between all the suggestions you provided. I have addressed each entry individually, so I summarize the general points here.

Regarding the keyword arguments provided, I changed some of them to slightly improve the performance of the circuits (this lowers the total execution time of this guide by about 10 seconds). I have inserted one/two sentences to comment on each choice, referring to the device documentation for further details. I also added a final section of a few sentences and general comments on performance and future developments.

> Further, it would be crucial to show a bond dimension scaling in the MPS section. That is, take some circuit expectation value and scale the bond dimension from something small to something large. The expectation value should then converge to the true value once a suitably large circuit is found. With this you can well describe the function of the bond dimension and how to set it.

@Qottmann I really like the idea, but based on all the feedback and guidelines I received it seems to me that this might not be the context. This is definitely something we should insert in [the demo on Matrix Product States](https://app.shortcut.com/xanaduai/epic/64229?group_by=none&ct_workflow=all&cf_workflow=500000005), and then we can link the demo to this short how-to guide. As an additional comment, to scale up the maximum bond dimension it would be nice to be able to change execution configuration at runtime (this requirement was eventually removed from this project) so that we don't have to re-instantiate the device every time.

As a final comment, probably the "usage details" section of `default.tensor` is too detailed compared to all the other devices in PennyLane (most of them don't even have one) and makes this how-to seem almost like a repetition, but it seems to me that this how-to overall reflects the requirements provided. We could think about reducing the "usage details" section of this device, but at this point, I don't think it makes sense (besides, more information is better than less).

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Yep, that's a long story and I agree that it might seem confusing. 

In a nutshell, this functionality was added to exploit `quimb`'s drawing functionality. That is, we are drawing the _wavefunction associated with the circuit_, which is handled internally by `quimb`, rather than the circuit itself. Therefore, to achieve this, the `draw` function is a method of the device class. 

If we instantiate the device with a certain number of wires, the internal number of tensorial indices handled by `quimb` is fixed and is reflected in the `draw` method. It does not matter if we run the circuit with fewer wires, the `draw` function will always display the wires we provided. On the other hand, if we do not instantiate the device by providing the number of wires, the latter is selected at runtime for each execution to match those required by the circuit. In this last case, the `draw` method reflects the number of wires provided during the last execution.

To summarize, the current behavior allows us to select the number of wires at runtime and to exploit `quimb`'s drawing functionality. I hope I clarified

Qottmann (changes requested):
I agree that linking to keyword argument descriptions is best, great addition 👍 

My main concern is that the demo how-to conveys that you can simulate arbitrarily large circuits with this device. I made some suggestions that I think help add more nuance.

If overall length of the demo is the problem, I would suggest scrapping the drawing paragraph in the MPS section. Drawing is then still covered in the TN section where this is more relevant.

In its stead, I really suggest to show some code playing with the bond dimension. To me this is the central thing to convey in a mps/tn simulator demo. It requires little to no explanation, just code. If you want, you can use this example:

```python
dataset = qml.data.load("qchem", molname="H6", bondlength=1.3, basis="STO-3G")[0]

H = dataset.hamiltonian
n_wires = len(H.wires)

def circuit():
    qml.BasisState(dataset.hf_state, wires=H.wires)
    for op in dataset.vqe_gates:
        qml.apply(op)
    return qml.expval(H)

bond_dims = 2**np.arange(2, (n_wires//2)+1) # maximum required bond dinemsion is 2**(n_wires//2) = 64
ress = []

for bond_dim in bond_dims:
    mps = qml.device("default.tensor", wires=n_wires, method="mps", max_bond_dim=bond_dim, contract="nonlocal")
    res = qml.QNode(circuit, mps)()
    ress.append(res)


plt.plot(bond_dims, ress, "x:", label="mps sim")
plt.hlines(dataset.vqe_energy, bond_dims[0], bond_dims[-1], label="exact result")
plt.xscale("log")
plt.xlabel("bond dim")
plt.xticks(bond_dims, bond_dims)
plt.legend()
plt.show()
```

image

The text could be something along the lines of 
"The bond dimension lets the user control the degree of approximation of the circuit.
Depending on how entangled the circuit is, we require a higher bond dimension. The highest possible bond dimension is $2^{n/2}$. In the example below, we see that we can already obtain a decent approximation before that."

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,203 @@
...
+######################################################################
+# Choosing the method to simulate quantum circuits
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The :class:`default.tensor <pennylane.devices.DefaultTensor>` device can simulate quantum circuits using two different computational methods.
+# We only need to specify the ``method`` keyword argument when instantiating the device.
+#
+# The first is the Matrix Product State (MPS) representation, and the second is the full contraction approach using a Tensor Network (TN).
+# The MPS representation can be seen as a particular case of the TN method, where the tensor network has a one-dimensional structure.
+# If not specified, the default method is the MPS representation.
```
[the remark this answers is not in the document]
PietropaoloFrisoni: I re-formulated these statements and the ones above in a different way :+1:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,203 @@
+r"""How to simulate quantum circuits with tensor networks
+====================================================================
+
+Tensor networks are a powerful computational tool for simulating quantum circuits.
+They provide a way to represent quantum states and operations in a compact form. 
+Unlike the state vector approach, tensor networks are particularly useful for large-scale simulations of quantum circuits.
+
+Here, we demonstrate how to simulate quantum circuits using the :class:`default.tensor <pennylane.devices.DefaultTensor>` device in PennyLane.
+This simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations. 
+This device is convenient for simulations with tens, hundreds, or even thousands of qubits.
```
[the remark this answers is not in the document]
PietropaoloFrisoni: This is definitely true for MPS, but for TN, this might not always be the case, right? I added a statement to clarify this in the section after this :+1:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,203 @@
...
+for num_qubits in range(25, 101, 25):
+    print(f"Number of qubits: {num_qubits}")
+    start_time = time.time()
+    result = circuit(theta, depth, num_qubits)
+    end_time = time.time()
+    print(f"Result: {result}")
+    print(f"Execution time: {end_time - start_time:.4f} seconds")
+
+######################################################################
+# As before, let's visualize the tensor network representation of the circuit.
```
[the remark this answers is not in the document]
PietropaoloFrisoni: Done :+1:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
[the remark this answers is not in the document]
Qottmann: Also for tn you cannot simulate arbitrarily entangled circuits, otherwise there would be no point to building a quantum computer :)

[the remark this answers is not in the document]
PietropaoloFrisoni: I agree, but it seems to me that the TN method can be helpful precisely when the entanglement is such that MPS cannot be applied (obviously with some limitations). If you think right now this is not clear enough in the way this is explained, I'll add a few more statements to clarify it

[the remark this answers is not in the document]
PietropaoloFrisoni: I specified this even in the above introduction. Hopefully, this should make things clear enough :+1:

Qottmann (approved):

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+They provide a way to represent quantum states and operations in a compact form. 
+Unlike the state vector approach, tensor networks are particularly useful for large-scale simulations of quantum circuits.
+
+Here, we demonstrate how to simulate quantum circuits using the :class:`default.tensor <pennylane.devices.DefaultTensor>` device in PennyLane.
+This simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations. 
+This device is convenient for simulations of circuits with tens, hundreds, or even thousands of qubits, 
+provided that the circuit structure is not too entangled. Other devices based on the state vector approach may be more suitable for small circuits 
+since the overhead of tensor network contractions can be significant.
+
+TODO: Insert figure
```
trbromley, line 14: What is the plan for the figure here? @tarik-elkhateeb is drawing an illustration for our release blog post, and we could also use that image here.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+# On the other hand, the TN method always yields an exact result but can demand higher computational and memory costs depending
+# on the underlying circuit structure and contraction path.
+#
+
+######################################################################
+# Simulating a quantum circuit with the MPS method
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Let's start by showing how to simulate a quantum circuit using the Matrix Product State (MPS) method.
+# If the number of wires is not specified when instantiating the device, it is inferred from the circuit at runtime.
```
trbromley, line 40: Not sure we need this sentence.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# We set the maximum bond dimension to 50 and the ``cutoff`` parameter is set to the machine epsilon of the ``numpy.complex128`` data type.
+# For this circuit, retaining a maximum of 50 singular values in the SVD decomposition is more than enough to represent the quantum state accurately.
+# Finally, the contraction strategy is set to ``auto-mps``. For an explanation of these parameters, refer to
+# the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`__ of
+# the :class:`default.tensor <pennylane.devices.DefaultTensor>` device.
```
trbromley, line 46: We could shift this paragraph below the code.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+    return qml.expval(
+        qml.X(num_qubits - 1) @ qml.Y(num_qubits - 2) @ qml.Z(num_qubits - 3)
+    )
+
+
+######################################################################
+# We can now simulate the quantum circuit for different numbers of qubits.
+#
+# The execution time will generally increase as the number of qubits grows.
+# The first execution is typically slower due to the initial setup and compilation processes of ``quimb``.
```
trbromley, line 85: Given that the circuits are quite different in qubit number, it's not clear to me what quimb is compiling/setting up that can be used for subsequent runs.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+# We can also visualize the tensor network representation of the quantum circuit with the ``draw`` method.
+# This method generates a graphical representation of the tensor network using ``quimb``'s plotting functionalities.
+#
+# Since we did not specify the number of qubits when instantiating the device,
+# the number of tensors in the tensor network is inferred from the last execution of the quantum circuit.
+# Let's visualize the MPS representation of the quantum circuit with 10 qubits.
+#
+
+circuit(theta, phi, num_qubits=10)
+dev.draw(color="auto", show_inds=True, figsize=(9, 5), return_fig=True)
```
trbromley, line 110: Where did we centre on @Qottmann's suggestion of cutting this? I'm :+1: with removing this, as the drawing is really just a linear graph.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+# The wave function associated with the quantum circuit is inferred from the last execution of the circuit on the device.
+#
+
+circuit(theta, depth, num_qubits=15)
+dev.draw(color="auto", show_inds=True, return_fig=True)
+
+
+######################################################################
+# We can see that the Tensor Network (TN) method generates a more complex tensor network than the Matrix Product State (MPS) method.
+# In this case, the representation depends on the structure of the quantum circuit and the specified keyword arguments.
```
trbromley, line 190: I'm on board with visualizing this :+1:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+#
+
+######################################################################
+# Final remarks
+# ~~~~~~~~~~~~~
+#
+# We have shown the basic usage of the :class:`default.tensor <pennylane.devices.DefaultTensor>` device in PennyLane to simulate quantum circuits using tensor networks.
```
trbromley, line 197: This link doesn't work, I'm not sure why :thinking:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+# As a general rule, choosing the appropriate method and setting the optimal keyword arguments is essential to achieve the best performance for a given quantum circuit.
+#
+# As mentioned in the introduction, devices based on the state vector approach, such as :class:`default.qubit <pennylane.devices.DefaultQubit>`
+# or :class:`lightning.qubit <pennylane.devices.LightningQubit>` may be more suitable for circuits with a few qubits.
```
trbromley, line 201: This link also doesn't work. Here note that `LightningQubit` is in a separate package.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,209 @@
...
+#
+# Finally, this device is still under development. Further improvements and new features are expected in future releases.
+#
```
trbromley, line 204: Not sure we need the final remarks section. It mostly feels like a repeat of other things, and the document isn't long enough in my opinion to warrant this.

trbromley (commented):
Thanks @PietropaoloFrisoni, I've left a few comments.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I'll remove it

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Did we suggest removing this? If that's the case I missed it. Ok, I'll remove it

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Is it because this device is in the latest doc version and not in the stable one?

PietropaoloFrisoni: Edit: I think this happens because this device is not imported in `pennylane.devices` to avoid warnings
from `quimb` if the latter is installed on the system. I'll try to fix this (otherwise, we can always use the ``  `` syntax)

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Ok, I'll remove it and move some of these statements

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I had to switch to ``default.tensor``

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I moved the explanations below the code sections. I am not sure this clarifies the workflow, but I'm pretty agnostic about this choice

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+
+######################################################################
+# We set the maximum bond dimension to 50 and the ``cutoff`` parameter is set to the machine epsilon of the ``numpy.complex128`` data type.
+# For this circuit, retaining a maximum of 50 singular values in the SVD decomposition is more than enough to represent the quantum state accurately.
+# Finally, the contraction strategy is set to ``auto-mps``. For an explanation of these parameters, we refer to
+# the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`__ of
+# the ``default.tensor`` device.
+#
+# As a general rule, choosing the appropriate method and setting the optimal keyword arguments is essential
+# to achieve the best performance for a given quantum circuit.
```
ikurecic, line 84: Could you add that it depends on each specific problem/setup/purpose/system/circuit/(pick your favorite)? I'm asking because I can already predict user questions about this sentence.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+######################################################################
+# Selecting the MPS method, each gate is immediately contracted into the MPS representation of the wavefunction.
+# Therefore, the structure of the MPS is maintained after each gate application.
+#
+
+######################################################################
+# Simulating a quantum circuit with the TN method
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The Tensor Network (TN) method is a more general approach than the Matrix Product State (MPS) method.
```
ikurecic, line 111:
```suggestion
# The tensor network (TN) method is a more general approach than the matrix product state (MPS) method.
```

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+
+import pennylane as qml
+import numpy as np
+
+# Define the keyword arguments for the TN method
+kwargs_tn = {
+    "contract": False,
+    "local_simplify": "DCRS",
+    "contraction_optimizer": None,
+}
```
ikurecic, line 127: The same suggestion here to include concise comments for the arguments in each line. You're presenting a device, so it would be great to point out its components.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+
+# The parameters of the quantum circuit
```
ikurecic, line 129:
```suggestion
# Parameters of the quantum circuit
```

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+    for _ in range(1, depth - 1):
+        for i in range(0, num_qubits, 2):
+            qml.CNOT(wires=[i, i + 1])
+        for i in range(num_qubits % 5):
+            qml.RZ(theta, wires=i)
+        for i in range(1, num_qubits - 1, 2):
+            qml.CZ(wires=[i, i + 1])
+    for i in range(num_qubits):
+        qml.CNOT(wires=[i, (i + 1)])
+    return qml.var(qml.X(num_qubits - 1))
```
ikurecic, line 151: I'm wondering if you want to (for both the TN and the MPS case) add a very short sentence that says _why_ the circuit is built this way? Or, honestly, maybe two simple figures? Tensor networks are so visual... I'm missing figures.
Feel free to text me about this on Slack.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+
+######################################################################
+# Here, we lazily add each gate to the tensor network without contracting by setting the ``contract`` keyword argument to ``False``.
+# The contraction optimizer is set to ``None``, and the simplification sequence is set to ``DCRS``.
+# As for the MPS method, we refer to the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`__ for
+# a list and explanation of the keyword arguments available for the TN method.
+#
+# We can also visualize the tensor network representation of the quantum circuit with the ``draw`` method.
+# This method produces a graphical representation of the tensor network using ``quimb``'s plotting functionalities.
+# The Tensor Network (TN) method usually generates a more complex tensor network than the MPS method.
```
ikurecic, line 171:
```suggestion
# The tensor network method usually generates a more complex tensor network than the MPS method.
```

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+#
+# Since we did not specify the number of qubits when instantiating the device,
+# the number of tensors in the tensor network is inferred from the last execution of the quantum circuit.
+# Let's visualize the tensor network for 15 qubits.
```
ikurecic, line 175:
```suggestion
# the number of tensors in the tensor network is inferred from the last execution of the quantum circuit,
# and we can visualize the tensor network for 15 qubits.
```
"Let's" is usually followed by an explanation, and this is the last sentence of the demo, so I'm suggestion this change.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+#
+
+circuit(theta, depth, num_qubits=15)
+dev.draw(color="auto", show_inds=True, return_fig=True)
+
+##############################################################################
+# About the author
+# ----------------
```
ikurecic, line 183: You're using the wrong heading level here. I suggest you change all section headings to have `-` instead of `~`, that's the standard approach.

ikurecic (commented):
Thanks for sharing, @PietropaoloFrisoni !
I've left some comments and suggestions for you.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
ikurecic: Just to confirm, based on the explanation via Slack, it would be great to add a sentence explaining the circuit choice (for both the MPS and the TN case) before getting into the code.

Qottmann: Could do a dataset vqe circuit, that requires little to no explanation and are super relevant :)

```python
dataset = qml.data.load("qchem", molname="H6", bondlength=1.3, basis="STO-3G")[0]

H = dataset.hamiltonian
n_wires = len(H.wires)

def circuit():
    qml.BasisState(dataset.hf_state, wires=H.wires)
    for op in dataset.vqe_gates:
        qml.apply(op)
    return qml.expval(H)
```

ikurecic: I know what you mean, but I think it would be a good to keep this as simple as possible, so it's useful as a step-by-step guide for everyone (instead of a great showcase of available features/resources). There will be space for this in future demos for sure.
Probably the best thing that could be done here is for Pietropaolo to write that sentence explaining the circuit choice _and_ possibly also say: "for example, this could be a very useful approach when analyzing systems from PennyLane Datasets(insert_link)" or "in a real-world example this circuit might be an XYZVQEABCDE(insert_link) circuit instead," or something like that.

I wouldn't push it past that in this case, so it can be kept clean and accessible for beginners.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,184 @@
...
+====================================================================
+
+Tensor networks are a powerful computational tool for simulating quantum circuits.
+They provide a way to represent quantum states and operations in a compact form. 
+Unlike the state vector approach, tensor networks are particularly useful for large-scale simulations of quantum circuits.
+
+Here, we demonstrate how to simulate quantum circuits using the ``default.tensor`` device in PennyLane.
+This simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations. 
+This device is convenient for simulations of circuits with tens, hundreds, or even thousands of qubits, 
+provided that the circuit structure is not too entangled. Other devices based on the state vector approach may be more suitable for small circuits
```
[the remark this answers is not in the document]
PietropaoloFrisoni: I am re-formulating the sentence differently, so that it's more clear

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I'll try. Let's see the deployed version

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Done :+1:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I added a link to an existing paper specifically on MPS, although the scientist in me is disgusted by this way of citing references/papers. I understand that this is just a how-to guide and not a demo, so I'll leave it as it is.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: I am re-opening this discussion @trbromley.

I agree with @ikurecic that we might lack visual representation in the current status of the how-to guide. I think the graphical representation of tensor networks is a compelling feature. It is certainly true that the MPS representation is always the same for every circuit, but perhaps showing it explicitly is very effective for users to understand what's going on under the hood (that is, the user can see that the state is always kept in a matrix product form). 

What do you think?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
PietropaoloFrisoni: Thanks, @ikurecic, I saw the Notion page as soon as you created it. Please, let me know if there is any action required from me at this stage. 

> Does this in any way clash with the other brief related to this topic? 

I am not sure I understand. Just to be clear, I inserted the `thumbnail_FALQON` image only to deploy the current version, but obviously, this will not be the final preview. My understanding is that the image used for the preview will be the same as shown here. I have no special preferences for the image to display.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.metadata.json
```diff
@@ -0,0 +1,29 @@
...
+            "uri": "/_static/demonstration_assets/regular_demo_thumbnails/thumbnail_FALQON.png"
+        }
+    ],
+    "seoDescription": "Learn how to simulate quantum circuits with tensor networks using DefaultTensor.",
+    "doi": "",
+    "canonicalURL": "/qml/demos/tutorial_how_simulate_quantum_circuits_with_tensor_networks_using_DefaultTensor",
+    "references": [],
+    "basedOnPapers": [],
+    "referencedByPapers": [],
+    "relatedContent": []
```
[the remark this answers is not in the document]
PietropaoloFrisoni: I inserted it, thanks for letting me know. However, there's something strange in the preview. Is it because this is still a deployed and not published version, or is there something wrong?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
ikurecic: @trbromley , is there a conflict here with the illustrations and the design brief I created for this Demo (linked above)?

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
trbromley: So the suggestion is that this how-to, in its current form, doesn't have enough visual content? I.e., there is just the TN visualization at the bottom?

With the addition of an illustration at the top of the page, I'm not sure I'm convinced about this worry :thinking:

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.metadata.json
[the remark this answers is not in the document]
ikurecic: This looks fine in the preview, it seems the error is on your end — all good.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,203 @@
+r"""How to simulate quantum circuits with tensor networks
+====================================================================
+
+`Tensor networks <https://pennylane.ai/qml/demos/tutorial_tn_circuits/>`__ are a powerful computational tool for simulating quantum circuits.
+They provide a way to represent quantum states and operations in a compact form. 
+Unlike the state-vector approach, tensor networks are particularly useful for large-scale simulations of quantum circuits.
+
+Here, we demonstrate how to simulate quantum circuits using the ``default.tensor`` device in PennyLane.
+We refer to the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`__ for more details.
+This simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations.
```
ikurecic, line 10:
```suggestion
Here, we demonstrate how to simulate quantum circuits using the ``default.tensor`` device in PennyLane.
The used simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations, and we refer to the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`__ for more details.
```

ikurecic (commented):
I left minor copy edit suggestions, @PietropaoloFrisoni 
The only thing still missing are the visuals, and I will update you when those are ready. Thank you.

demonstrations/tutorial_How_to_simulate_quantum_circuits_with_tensor_networks.py
```diff
@@ -0,0 +1,203 @@
...
+Here, we demonstrate how to simulate quantum circuits using the ``default.tensor`` device in PennyLane.
+We refer to the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`__ for more details.
+This simulator is based on `quimb <https://quimb.readthedocs.io/en/latest/>`__, a Python library for tensor network manipulations. 
+This device is well-suited for simulations of circuits with tens, hundreds, or even thousands of qubits as long as the degree of entanglement 
+within the circuit remains manageable. In general, the effectiveness of this device 
+depends on the specific circuit structure and the provided keyword arguments.
+Other simulators based on the state-vector approach may be more suitable for small circuits 
+since the overhead of tensor network contractions can be significant.
+
+This device has just been released and is still under development.
```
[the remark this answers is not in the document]
PietropaoloFrisoni: There is no problem with everything else, but the link for sphynx directive `_:class:~pennylane.devices.default_tensor.DefaultTensor_` does not work (although rendered correctly). I am not sure why. I have already tried several times but I had to give up. Should I insert it anyway, although the link does not work? Thanks!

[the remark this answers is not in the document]
ikurecic: That should be because this links to the stable version (and not the latest version) — it should be live with the new PennyLane release. Actually, maybe the other links to that docs page should be updated to lead to the stable URL, as well. Could you do that?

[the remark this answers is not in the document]
ikurecic: This demo is merging to `dev`, which means that it will only get published _after_ the new PennyLane version is released (when `dev` will be merged to `master`). This means the documentation will already have been updated, so this should work fine.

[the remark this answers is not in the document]
PietropaoloFrisoni: I'll replace all links to the documentation so that they point to `stable` and not `latest`, except for the one in the statement:

_check the latest functionality in the `documentation <https://docs.pennylane.ai/en/latest/code/api/pennylane.devices.default_tensor.DefaultTensor.html>`___

Since here I think it makes sense to point to the latest version of the doc :+1:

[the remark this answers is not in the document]
josh146: Hey @PietropaoloFrisoni @ikurecic, we should try and avoid any links in the demos that use absolute URLs, since the docs URLs change with code structure and this will be hard to maintain.

Instead we can use:

```rest
- :doc:`link text <path/to/docs/page>`
- :func:`~.path.to.func`
- :class:`~.path.to.class`
```

etc. This will also (usually) respect a users preference to which version of the docs to link to.

In terms of working out what to cross reference, you can use `sphobjinv` https://github.com/bskinn/sphobjinv:

```bash
$ wget https://docs.pennylane.ai/en/latest/objects.inv
$ sphobjinv suggest objects.inv SimplifiedTwoDesign -st 58
```

This will show the expected cross-ref path.

For doc pages, they will usually have the form

```rest
:doc:`circuit quickstart <introduction/circuits>`
```

conf.py
```diff
@@ -198,6 +198,6 @@
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
-    "pennylane": ("https://docs.pennylane.ai/en/stable/", None),
+    "pennylane": ("https://docs.pennylane.ai/en/latest/", None),
     "catalyst": ("https://docs.pennylane.ai/projects/catalyst/en/stable", None)
```
josh146, line 202: @ikurecic @PietropaoloFrisoni this line in the `conf.py` file determines _which version of the docs_ the demos will try and cross reference against. I updated it to `latest` to allow you to test the cross-referenced links in the rendered version of the PR.

However, please remember to revert it to `stable` either:
- before this PR is merged, _or_
- before `dev` is merged into `master` on the 10th!

PietropaoloFrisoni: Thanks @josh146 , I added a reminder in my calendar to make sure we don't forget 👍

ikurecic (approved):

PR #1136, opened 2024-06-19
updating image

KetpuntoG:
Changing to a better image


_static/demonstration_assets/mapping/long_image.png
soranjh, line 1: Thanks @KetpuntoG, the thumbnail image should be also replaced. The current one only shows BK.

KetpuntoG: Ah, this design does not exist. Shall I request it to Tarik?
(we cannot crop the image ourselves)

soranjh: Yes, please. The original image with only BK is misleading.

soranjh (approved):

ddhawan11 (approved):

PR #1140, opened 2024-07-03
exclude `demonstrations_categories.metadata.json` from `upload-text` workflow

michaelvu97:
It's not useful metadata and breaks the schema a little bit


doctorperceptron (approved):
Nice!

KetpuntoG (approved):

PR #1146, opened 2024-07-04
Quantum volume updates

albi3ro:
The lima device used in the quantum volume demo no longer exists.  This update switches to make it use the fake lima device instead.

Since when running the demo, I got data that no longer agreed with the conclusions of the data present in the demo, I added a warning that device details may have changed over time.  

Since we now use a fake device instead of a real device, this demo could theoretically be automatically run. But the difference in the data output means that the automatically run results would not tell the same story.


demonstrations/quantum_volume.py
```diff
@@ -14,6 +14,11 @@
 
 *Author: Olivia Di Matteo — Posted: 15 December 2020. Last updated: 15 April 2021.*
 
+.. warning::
+    The data is this demo was generated using the lima device that no longer exists.
+    While the code now uses a fake version of the lima device, the noise and calibration
+    may have shifted since the data was generated.
```
glassnotes, line 20:
```suggestion
    The data in this demo was originally generated using system calibration data from the ``ibmq_lima`` device, 
    which has since been retired. While the code now uses a fake version of the lima device, the noise and calibration
    may have shifted since the data was generated.
```

**suggestion**: Further to our Slack discussion, if the demo gets rendered live, I would provide an additional warning that the conclusions in text may not match what is shown in the plots. But if I recall correctly, it is one of the "static" demos anyways, so this isn't an issue.

glassnotes (approved):
Thanks so much, @albi3ro ! :tada:

demonstrations/quantum_volume.py
```diff
@@ -586,13 +572,11 @@ def heavy_output_set(m, probs):
...
-from qiskit.providers.aer import noise
+from qiskit_aer.noise import NoiseModel
 
-noise_model = noise.NoiseModel.from_backend(dev_lima.backend)
+noise_model = NoiseModel.from_backend(FakeLimaV2())
 
-dev_noisy = qml.device(
-    "qiskit.aer", wires=dev_lima.num_wires, shots=1000, noise_model=noise_model
-)
+dev_noisy = qml.device("qiskit.aer", wires=5, shots=1000, noise_model=noise_model)
```
lillian542, line 579: Maybe not worth spending time on, but are the results more similar to the original results if we just use the `FakeLima` device directly, instead of using its noise model on `"qiskit.aer"`? Something like `dev_noisy = qml.device("qiskit.remote", backend=FakeLimaV2())`. 

Even if it doesn't change the results though, if it runs the same, it might be nice to switch it out - it just seems a little funny to use a simulator to simulate another simulator, and if we do it that way, we would actually be demonstrating code that is compatible with directly running on real hardware (by just switching out the backend), which seems more useful.

demonstrations/quantum_volume.py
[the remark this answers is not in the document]
albi3ro: capitalized.

demonstrations/quantum_volume.py
albi3ro: So that actually fixes the problem with different results!

lillian542 (approved):
Looks good to me! Nice to see this updated after so many conversations about it 😅

PR #1148, opened 2024-07-04
Update braket static demos; Update torch dependency

mudit2812:
**Summary:**
* Update `ahs_aquila` and `braket_parallel_gradients` demos to be up to date with PennyLane 0.37.
* Update `pyproject.toml` to install torch 2.1.2. The old version (1.13) was quite old. I tried changing the version to 2.3.1 which requires `typing-extensions>=4.8`, but that caused a dependency conflict with tensorflow which required `typing-extensions<4.6`

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**


astralcai:
I've added the "build-all-qml-demos" tag, can you retrigger CI to check that the change of dependency didn't break anything else? Just to be safe.

astralcai (approved):

PietropaoloFrisoni (approved):

PR #1149, opened 2024-07-05
In-equation syntax in parshift Demo fix

ikurecic:
To address this
image
as described [here](https://xanaduhq.slack.com/archives/C03FRM3RCCV/p1720170995178649).


dwierichs (approved):
Thanks for the quick fix, @ikurecic !

josh146 (approved):

PR #1150, opened 2024-07-05
Demo checks for v0.37

Shiro-Raven:


demonstrations/getting_started_with_hybrid_jobs.py
```diff
@@ -373,7 +373,7 @@ def circuit(params):
 # is not as smooth as the simulator, but the minimum still is detected correctly!
 #
 
-df = pd.DataFrame(qpu_job.metrics())
+df = pd.DataFrame(job.metrics(metric_type=MetricType.ITERATION_NUMBER))
```
lillian542, line 376: why are we updating this to access the previous job and not the `qpu_job`?

lillian542 (commented):
One question about the changes, and then of course the messaging about Lucy needs a quick update. Lmk when its ready for another look! :)

demonstrations/getting_started_with_hybrid_jobs.py
Shiro-Raven: That's a copy-paste mistake, you are right

astralcai (approved):

lillian542 (approved):
Looks good, thanks! 🎉

PR #1162, opened 2024-07-17
Update error mitigation demo to use `split_non_commuting`

Shiro-Raven:
`hamiltonian_expand` is deprecated and so the demo is being updated to use the unified `split_non_commuting` interface instead.

[[sc-64018](https://app.shortcut.com/xanaduai/story/64018)]


EmilianoG-byte (approved):
Looks good to me 🥳 !

PR #1170, opened 2024-07-24
sc-66381-demos-page-optimize-imagess

ashishks0522:
## Changes
- [sc-66381-demos-homepage-individual-demo-page-compress](https://app.shortcut.com/xanaduai/story/66381/demos-homepage-individual-demo-page-compress-images-or-convert-them-to-webp-or-avif)
- Compress images to improve page performance


doctorperceptron (approved):
Thanks @ashishks0522! A couple of the author images look pretty small (e.g. Alain and Diksha), but if we're moving to using their profiles soon this may not matter?

ashishks0522:
> Thanks @ashishks0522! A couple of the author images look pretty small (e.g. Alain and Diksha), but if we're moving to using their profiles soon this may not matter?

Correct we are moving away from this soon, but also the max size of author image on the website is 85px by 85px so compared to that all images are at least 2x that size. So they should look good on the website.

AndrewGardhouse (approved):
Looks like all the images have the same file names and the dimensions are the same 🚀

PR #1176, opened 2024-07-31
Replacement of legacy devices with `default.qubit`

Shiro-Raven:
**Summary:**
The legacy qubit devices `default.qubit.{tf,torch,jax,autograd,legacy}` are now deprecated and need to be replaced with `default.qubit`.

[[sc-70260](https://app.shortcut.com/xanaduai/story/70260)]

### DIFF
--- a/demonstrations/ahs_aquila.py
+++ b/demonstrations/ahs_aquila.py
@@ -512,7 +512,7 @@ amplitude_params = [max_amplitude, displacement, sigma]
 params = [amplitude_params]
 ts = [0.0, 1.75]
 
-default_qubit = qml.device("default.qubit.jax", wires=3, shots=1000)
+default_qubit = qml.device("default.qubit", wires=3, shots=1000)
 
 
 @qml.qnode(default_qubit, interface="jax")
--- a/demonstrations/learning2learn.py
+++ b/demonstrations/learning2learn.py
@@ -239,8 +239,7 @@ def qaoa_from_graph(graph, n_layers=1):
     def hamiltonian(params, **kwargs):
         """Evaluate the cost Hamiltonian, given the angles and the graph."""
 
-        # We set the default.qubit.tf device for seamless integration with TensorFlow
-        dev = qml.device("default.qubit.tf", wires=len(graph.nodes))
+        dev = qml.device("default.qubit", wires=len(graph.nodes))
 
         # This qnode evaluates the expectation value of the cost hamiltonian operator
         cost = qml.QNode(circuit, dev, diff_method="backprop", interface="tf")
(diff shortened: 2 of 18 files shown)


demonstrations/learning2learn.py
```diff
@@ -373,9 +372,7 @@ def recurrent_loop(graph_cost, n_layers=1, intermediate_steps=False):
     # We perform five consecutive calls to 'rnn_iteration', thus creating the
     # recurrent loop. More iterations lead to better results, at the cost of
     # more computationally intensive simulations.
-    out0 = rnn_iteration(
-        [initial_cost, initial_params, initial_h, initial_c], graph_cost
-    )
+    out0 = rnn_iteration([initial_cost, initial_params, initial_h, initial_c], graph_cost)
```
albi3ro, line 375: just line length formatting?

Shiro-Raven: Yup

albi3ro (commented):
I'm double checking the output of the `learning2learn` demo by getting a crash. But it may be due to me not setting up Keras properly.  Think I've made mistakes in Keras versioning in the past. Stay tuned...

Shiro-Raven:
> I'm double checking the output of the `learning2learn` demo by getting a crash. But it may be due to me not setting up Keras properly. Think I've made mistakes in Keras versioning in the past. Stay tuned...

Maybe this link is what you are looking for? 
https://qml-build-previews.pennylane.ai/pull_request_build_preview/1176/qml/demos/learning2learn/

PietropaoloFrisoni (approved):
Thank you @Shiro-Raven , looks good to me!

@albi3ro the `learning2learn.py` demo fails with the latest version of `tensorflow` and `keras`, but on my laptop works with `tensorflow-2.14.1` and `keras-2.14.0` (the same used in the `poetry.lock` file)

demonstrations/ahs_aquila.py
```diff
@@ -512,7 +512,7 @@ def gaussian_fn(p, t):
 params = [amplitude_params]
 ts = [0.0, 1.75]
 
-default_qubit = qml.device("default.qubit.jax", wires=3, shots=1000)
+default_qubit = qml.device("default.qubit", wires=3, shots=1000)
```
albi3ro, line 515: Currently issues with the probability distribution not being normalized, but we can fix that independently afterward.

demonstrations/tutorial_jax_transformations.py
```diff
@@ -253,8 +259,8 @@ def circuit(param):
 # To learn more about how JAX handles randomness, visit their
 # `documentation site. <https://jax.readthedocs.io/en/latest/jax.random.html>`__
 #
-# .. note:: 
-#     This example only applies if you are using ``jax.jit``. Otherwise, PennyLane 
+# .. note::
```
albi3ro, line 262: Trying to figure out what was meant by this statement. 

I'm tempted to say we just get rid of it, because I find it confusing.  What about this example only works with jax jit?

Maybe we can add a comment that seed can also accept any integer.

albi3ro (approved):
We should independently resolve the probability normalization issue, but I think we can consider that a different task.

For the note in `tutorial_jax_transformations.py`, I'll think about that note and come back to it. But for now, let's just resolve that conversation and move on.

PR #1185, opened 2024-08-09
[WIP] Demo on "Constant-depth preparation of matrix product states with adaptive quantum circuits"

dwierichs:
**Title:**
Constant-depth preparation of matrix product states with dynamic circuits

**Summary:**
Matrix product states (MPS) form an important class of quantum many-body states. Preparing them is an important subroutine for any algorithm that want to use MPS as an initial state, for example. A recent paper, on which this demo is based, and which it is following closely, generalized a method that reduces the depth of such a preparation circuit to constant scaling, by using auxiliary qubits and dynamic quantum circuits (mid-circuit measurements together with classical feedforward).
We here focus on the simplest scenario in the main text of the paper, and implement a simple example in code.
The implemented MPS has a parameter controlling the correlation length of the prepared state, which we verify explicitly.

**Relevant references:**
https://arxiv.org/pdf/2404.16083
https://arxiv.org/pdf/quant-ph/0501096
https://journals.aps.org/prxquantum/pdf/10.1103/PRXQuantum.4.020315

**Possible Drawbacks:**

**Related GitHub Issues:**

----

  Mid-circuit measurements recently are/were a focus point in PennyLane. MPS states are used in some quantum chemistry algorithms as initial state.

* AUDIENCE — Who is this for?

  Researchers in quantum algorithms

TBD

- [ ] Tutorial
- [x] Demo
- [ ] How-to


demonstrations/tutorial_constant_depth_mps_prep.metadata.json
```diff
@@ -0,0 +1,53 @@
+{
+    "title": "Constant-depth preparation of matrix product states with dynamic circuits",
+    "authors": [
+        {
+            "id": "david_wierichs"
+        }
+    ],
+    "dateOfPublication": "2024-09-10T00:00:00+00:00",
+    "dateOfLastModification": "2024-09-10T00:00:00+00:00",
```
dwierichs, line 9: To do

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+#    .. math::
+#
+#        |\psi_1\rangle
+#        &= \prod_{i=N}^{1} U^{(i)} |0\rangle^{\otimes N} \frac{1}{\sqrt{D}} \sum_j|jj\rangle _D\\
+#        &= \frac{1}{\sqrt{D}} \sum_j \prod_{i=N-1}^{1} U^{(i)}
+#        \sum_{m_N=0}^{d-1} |0\rangle^{\otimes N-1}|m_N\rangle A^{m_N}|jj\rangle _D\\
+#        &= \frac{1}{\sqrt{D}} \sum_j
+#        \sum_{\vec{m}} |\vec{m}\rangle A^{m_1} A^{m_2}\cdots A^{m_N}|jj\rangle_D\\
+#        &= \frac{1}{\sqrt{D}} \sum_{i,j}
+#        \sum_{\vec{m}} |\vec{m}\rangle \langle i|A^{m_1} A^{m_2}\cdots A^{m_N}|j\rangle |ij\rangle_D
```
Qottmann, line 190: (opinion) is there a way to illustrate this? I found it a bit hard to follow

Qottmann: Also, is this a block encoding in some way? Or is this going beyond that?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+E_00 = np.array([[1, 0], [0, 0]])  # |0><0|
+E_10 = np.array([[0, 0], [1, 0]])  # |1><0|
+U_first_term = np.kron(A_0, E_00) + np.kron(A_1, E_10)
+
+print(np.round(U_first_term, 5))
+print(np.linalg.norm(U_first_term, axis=0))
+
+######################################################################
+# We see that this fixed part has two norm-1 columns already, and we are
+# able to complement this by :math:`C_\perp` using the same columns, but with shifted rows:
```
Qottmann, line 221: I dont think I get this 🤔 
Is there a more constructive / general way to complement the matrix?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+# We will verify the prepared state below, when comparing to the constant-depth
+# preparation circuit. Let's just draw the circuit for now.
+
+N = 7
+_ = qml.draw_mpl(sequential_circuit)(N, g)
+
+######################################################################
+# As we can see, the sequential preparation circuit already uses mid-circuit
+# measurements. However, there is no feed forward control that modifies the circuit
+# dynamically based on measured values.
```
Qottmann, line 321: The way it is drawn this is not obvious to everyone, a small comment on how that resembles mid-circuit measurements would be helpful I think

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+#
+# Fusion of MPS states with mid-circuit measurements
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# The next ingredient for the constant-depth MPS preparation circuit is to fuse
+# two MPS states together into one MPS. Under the hood, this is
```
Qottmann, line 327: what is meant by fusing two MPS states together? Are we talking about a superposition of two states? A tensor product?
Even after going over the following section it is still not clear to me 😓

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+# tensor :math:`A`. We will not go into detail about these conditions but assume here that
+# we work with compatible MPS and measurements.
+#
+# Operator pushing then allows us to push a defect operator :math:`B^k` from one bond
+# axis through the tensor :math:`A` of the MPS to the other bond axis and to the physical axis.
+# While doing so, the operator in general will change, i.e., we end up with different operators
+# :math:`C^k` and :math:`D^k` on the bond and physical axes, respectively.
+# Visually, we find:
+#
+# .. image:: ../_static/demonstration_assets/constant_depth_mps_prep/operator_pushing.png
```
Qottmann, line 457: is this the correct image? I dont see a $B^k$, $C^k$ or $D^k$ anywhere

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+# beyond the scope of the demo.
+#
+# Block size :math:`q`
+# ~~~~~~~~~~~~~~~~~~~~
+#
+# The size of the blocks that have to be prepared in the first step depends on multiple
+# considerations. First, the block must be such that operator pushing relations are
+# available. Often this will determine a minimal block size, and multiples of that
+# size are valid choices as well. Second, the depth of the (parallelized) sequential
+# preparation circuits is proportional to :math:`q`, determining a key contribution
```
Qottmann, line 566: what is q?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+# :math:`|\Psi(g)\rangle`.
+# Before we evaluate the states it produces, let's see how the circuit looks.
+# We'll add some boxes for the individual subroutines, and recommend opening
+# the figure in a separate tab.
+#
+
+N = 9
+q = 3
+g = -0.8
+fig, ax = qml.draw_mpl(constant_depth_circuit)(N, g, q)
```
Qottmann, line 660: super pretty plot! please tell me you didnt do this by hand 😬 ?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+ax.legend()
+plt.show()
+
+######################################################################
+# Note that due to the periodic boundary conditions, we have :math:`C(N-j)=C(j)`.
+# This is why the correlations increase again after :math:`j=\frac{N}{2}`.
+# To obtain the correlation length, we fit an exponential to the measured correlators
+# up until :math:`j=\frac{N}{2}` and invert the decay rate.
+# We can also compare this result to the analytic result for the correlation
+# length :math:`xi` mentioned in the beginning:
```
Qottmann, line 798:
```suggestion
# length :math:`\xi` mentioned in the beginning:
```

Qottmann (commented):
Very impressive work @dwierichs 
To be honest I found it quite hard to follow, despite the excellent code examples, probably simply because this is just a very advanced topic.
Any chance to dumb this down a bit? E.g. is it necessary to talk about qudits, or could we just use qubits and concrete d=2 for simplicity?
I think the demo could also benefit from additional illustrations accompanying the math whereever possible (the illustrations that are already there are great!)

TBH I got lost at the fusion part as it was not really clear to me what this fusion is doing. It would be good to clarify the general idea better to not have people lost, but I'll also try to re-read another time.

Most comments marked with (opinion) are optional and can be disregarded at your discretion :)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+Matrix product states
+~~~~~~~~~~~~~~~~~~~~~
+
+Matrix product states (MPS) are an important class of quantum states.
+They can be described and manipulated conveniently on classical computers and
+are able to approximate relevant states in quantum many-body systems.
+In particular, MPS can efficiently describe ground states of (gapped local)
+one-dimensional Hamiltonians, which in addition can be found efficiently using
+density matrix renormalization group (DMRG) algorithms.
+For a review of MPS see [#cirac]_.
```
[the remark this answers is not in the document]
dwierichs: Perfect, thanks! It is indeed the "next best one" I found, coming from the main paper. Was hoping to get your opinion on this :wink:

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+This MPS also is discussed in Sec. III C 1. of [#smith]_.
+In general, the correlation length of :math:`|\Psi(g)\rangle` is given by
+
+.. math::
+
+    \xi(g) = \left|\ln\left(\frac{1+g}{1-g}\right)\right|^{-1}.
+
+Long-range correlated states require linear-depth unitary circuits in general.
+Therefore, the constant-depth circuit for :math:`|\Psi(0)\rangle` will demonstrate
+that dynamic quantum circuits are more powerful than unitary operations alone.
```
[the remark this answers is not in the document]
dwierichs: I meant that if we restrict to unitary circuits, we need linear-depth. If we allow (non-unitary) dynamic circuit components, we can get away with constant depth :)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+
+    A^0 = \eta \left(\begin{matrix} 1 & 0 \\ \sqrt{-g} & 0 \end{matrix}\right)
+    \quad
+    A^1 = \eta \left(\begin{matrix} 0 & -\sqrt{-g} \\ 0 & 1 \end{matrix}\right)
+
+where :math:`\eta = \frac{1}{\sqrt{1-g}}` and :math:`g\in[-1, 0]` is a freely chosen parameter.
+
+This MPS is a simple yet important example because it can be tuned from the long-range correlated
+GHZ state :math:`|\Psi(0)\rangle=\frac{1}{\sqrt{2}}(|0\rangle^{\otimes N} + |1\rangle^{\otimes N})`
+to the state :math:`|\Psi(-1)\rangle=|+\rangle^{\otimes N}` with vanishing correlation length.
```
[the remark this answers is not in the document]
dwierichs: I'll think about it :)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+    return (A_0, A_1)
+
+
+g = -0.6
+A_0, A_1 = A(g)
+is_left_canonical = np.allclose(A_0.conj().T @ A_0 + A_1.conj().T @ A_1, np.eye(2))
+print(f"The matrices A^m are in left-canonical form: {is_left_canonical}")
+
+xi = 1 / abs(np.log((1 + g) / (1 - g)))
+print(f"For {g=}, the theoretical correlation length is {xi=:.4f}")
```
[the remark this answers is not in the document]
dwierichs: Hahaha credit to @timmysilv for introducing this to me :heart:

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: I'll try to make it more visual/intuitive. Basically we chain up the physical sites like beads on a string, where the string is the bond site that is being passed on, is my intuition...

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Phew. To be honest I was happy that I caught this option somewhat quickly :D Let me see whether there is a generic method (for 4x4 matrices, maybe).
Would rephrasing the argument above help, maybe? Something like
"The states |00> and |10> already are mapped to normalized, orthogonal states. Therefore we just need to map |01> and |11> to normalized states that are mutually orthogonal with all other output states."
?

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Aight!

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Thanks, this is a very valuable feedback point!

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Nope, it's not :D Sorry, the pictures are tbd. I'll upload the temporary sketches !

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: The block size from step 1 in the first part of this section. I'll define it more prominently!

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: :eyes:

dwierichs: Thanks! :)

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
Qottmann: can you say it more like that 😄 ? That I immediately understood

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: certainly would help the procedure here, yeah, but I guess this is not what one would do in practice?

dwierichs (commented):
Thanks @Qottmann for the first review!
Let me know what you think about the individual points, I hope to have addressed all of them.

Regarding the overall complexity and maybe a less general scope with qubits only, I'll think some more about how to do this best :)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+# Great! This means that we already found the unitary :math:`U` for this MPS.
+# Combining the terms above on paper, we see that
+#
+# .. math::
+#
+#     U &= A^0 \otimes (|0\rangle\!\langle 0| + |1\rangle\!\langle 1|)
+#     + A^1\otimes (|1\rangle\!\langle 0| + |0\rangle\!\langle 1|)\\
+#     &= A^0 \otimes \mathbb{I} + A^1\otimes X
+#
+# This looks a lot like a :class:`~.pennylane.CNOT` gate.
```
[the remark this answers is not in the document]
Qottmann: Could alternatively just define 

```
def U_template(g, bond_idx, phys_idx):
    qml.QubitUnitary(A(g), wires=(bond_idx, phys_idx))
```

and leave the gate implementation for later/out of scope

(opinion)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+# distribution as the sequential circuit.
+#
+
+p_seq = sequential_circuit(N, g)
+p_const = constant_depth_circuit(N, g, q)
+print(f"The two probabilities agree for {g=:.1f}: {np.allclose(p_seq, p_const)}")
+
+
+######################################################################
+# Nice, we have a constant-depth circuit that reproduces the existing sequential
```
Qottmann, line 744: What is the success probability of preparing the desired state? I guess this decreases with the number of conditionals in the preparation? Is is true that we trade off depth with success rate?

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Wait, why not? :)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,865 @@
...
+dimension :math:`d`. We will further restrict to periodic boundary conditions,
+corresponding to translational invariance of the MPS on a closed loop.
+A general MPS with this properties is given by
+
+.. math::
+
+    |\Psi\rangle = \sum_{\vec{m}} \operatorname{tr}[A^{m_1}A^{m_2}\cdots A^{m_N}]|\vec{m}\rangle,
+
+where :math:`\vec{m}` is a multi-index of :math:`N` indices ranging over :math:`d`,
+i.e., :math:`\vec{m}\in\{0, 1 \dots, d-1\}^N`, and :math:`\{A^{m}\}_{m}` are :math:`d`
```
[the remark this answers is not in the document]
dwierichs: Ah, so there are $d$ different matrices, so (as an example) if $m_1 \neq m_2$, the first and second factor are different.
However, if $m_1 = m_2$, we get the same factor twice, because we prepare a translationally-invariant MPS. This is probably what you mean? I'll try and make this more explicit!

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+# to a physical site in :math:`|0\rangle` and a bond site, we obtain the same result as
+# applying :math:`A` to the bond site alone.
+#
+# We leave the action of :math:`U` on other physical input states undefined, and only require
+# it to make :math:`U` unitary overall.
+# In short, our construction can be written as
+#
+# .. math::
+#
+#     U = \sum_m A^m \otimes |m\rangle\!\langle 0| + C_\perp,
```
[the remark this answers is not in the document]
dwierichs: Yeah, I agree!

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: As we want to shorten, I'm down to doing the `QubitUnitary`. I'll leave it as an exercise left to the readers :D

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+# the figure in a separate tab.
+#
+
+N = 9
+q = 3
+g = -0.8
+fig, ax = qml.draw_mpl(constant_depth_circuit)(N, g, q)
+
+# Cosmetics
+options = {
```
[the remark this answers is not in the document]
dwierichs: Nope, there's only an `wire_options` dict for all wires. Nice idea for an addition, though! I'd suggest to recommend this internally (should be an about-one-line patch).

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: >What is the success probability of preparing the desired state? I guess this decreases with the number of conditionals in the preparation? 

Note that there really is only the final projective measurement that is probabilistically successful! On average (over bond site states), the probability of success will be $\mathcal{O}(D^{-2})$, I think.

> Is is true that we trade off depth with success rate?

No, because the sequential preparation circuit also has this probabilistically successful step! It's a "problem" of the overall MPS preparation strategy introduced by the sequential circuit, inherited by the constant-depth algorithm.

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+
+
+######################################################################
+# Nice, we have a constant-depth circuit that reproduces the existing sequential
+# preparation circuit, which has linear depth!
+# Not that while the dynamic operations of step 3 seem to have a linear depth,
+# they can be applied in parallel because they are only controlled by classical
+# registers.
+#
+# Correlation length
```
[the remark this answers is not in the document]
dwierichs: Mhm, yeah, this is to "use" the prepared state for something and demo that the example covers interesting scenarios. 
But now that you question it, I agree that the demo is fine without this. And I'd like to shorten it anyways...

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+
+
+mid_point = N // 2 + 1
+distances = list(range(1, mid_point))
+print(f"Correlation lengths xi\n======================")
+for g, correls in zip(gs, all_correls):
+    popt, pcov = curve_fit(exp_fn, distances, correls[1:mid_point])
+    xi_numerical = -1 / popt[1]
+    xi_predicted = 1 / abs(np.log((1 + g) / (1 - g)))
+    print(f"{g=:.1f}")
```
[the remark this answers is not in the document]
dwierichs: Ah, I accidentally pushed the change to fewer qubits (`N=12` above). I think for this shorter chain length, finite-size effects are kicking in. They are stronger for longer-ranged correlated states, hence the dependence of the deviation on $g$.

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: This is resolved: I removed the trailing section as you suggested, and included a tiny computation of correlators here, showcasing the difference between GHZ and $|+>$ state.

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,896 @@
...
+#
+# An MPS like the one above can be prepared in linear depth using an established technique
+# by Schön et al. [#schoen]_. We introduce this technique here because the new
+# constant-depth construction uses it as a (parallelized) subroutine, and because we will
+# later compare the two approaches.
+#
+# The MPS :math:`|\Psi\rangle` above is given by the rank-3 tensor :math:`A`.
+# We can think of this as an operator that acts on a bond site and simultaneously *creates*
+# a physical site. In order to prepare the state in a circuit, which usually is not allowed
+# to create qudits out of nowhere, we want to construct a unitary operation
```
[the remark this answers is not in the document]
dwierichs: Better?

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: Is there no constructive/automatic way to do this? Like can I not write a subroutine `A_to_U(mps)` for any input $A$?

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
Qottmann: What I meant was that typically there is a different 3-legged tensor A for each site, only implicitly differentiated by the label index, e.g. 
image

(eq. (37) in [schollwöck](https://arxiv.org/pdf/1008.3477))

All tensors are called $A$ but they are different on each site

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: > Note that there really is only the final projective measurement that is probabilistically successful! On average (over bond site states), the probability of success will be $O(D^{−2})$ , I think.

Oooooh right because its just a measurement-dependent "fix"! Nice! Could be worth mentioning (if it fits somewhere)

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
Qottmann: Ah right is the formula for infinite size systems?

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
Qottmann: Did you push the update already? I still see the onld version

image

I trust you fix the sentence, no need to reconfirm 🙂

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: tbh, I didn't check , but I'd assume so, yes :)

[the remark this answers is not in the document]
dwierichs: As this just left the demo, I won't investigate further :D

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: Sorry, forgot to push!

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: Ah, yes, okay!
I mention explicitly now that we reuse the same $A$ because it is a translation-invariant MPS

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
Qottmann: works for me 😆

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: cool! I always prefer the algorithmic way where I dont have to think and "see"

dwierichs: I get that, but I think I'd rather keep this as short as possible, and not introduce another algorithmic step, to make it less intimidating (although I get that for generalization it is nicer to have an algorithm)

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: Removed this part.

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: Yeah its a good point!

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,843 @@
...
+
+We will start by briefly (no, really!) introducing the building blocks for
+the algorithm from a mathematical perspective. One of these blocks
+is the sequential MPS preparation circuit to which we will compare later.
+Alongside this introduction we describe and code up each building block for a
+specific MPS. Then we combine the building blocks into the constant-depth
+algorithm by Smith et al. and run it for the example MPS.
+As a rough preview, this is what the algorithm will look like schematically:
+
+.. image:: ../_static/demonstration_assets/constant_depth_mps_prep/algorithm_preview.png
```
Qottmann, line 49: Would be great to enumerate each step and write a short sentence what we are seeing in the text, currently it is a bit disjoint

Qottmann: and potentially already call these steps by their name, mps prepare, fusion and operator push? (are those the three steps displayed here?)

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,843 @@
...
+# physical qubit, and a final basis change to measure the Bell basis. Overall, this amounts to a linear
+# operation count and circuit depth.
+#
+# These steps are visualized in the following sketch, with the bond qudits starting at the top.
+# Note that the sketch deviates from standard circuit diagrams in that it passes one bond qudit from
+# the top to the bottom, instead of keeping the position of the qudits fixed.
+#
+# .. image:: ../_static/demonstration_assets/constant_depth_mps_prep/sequential.png
+#     :width: 75%
+#     :align: center
```
Qottmann: How about moving this to the beginning of the section? It's a matter of personal preference, but I prefer to first _see_ and then get into the details of what is going on, rather than the other way around

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,843 @@
...
+#    its inverse.
+#
+# #. Perform the same projection step as in the sequential preparation algorithm on
+#    the two remaining bond sites.
+#
+# We summarize the algorithm in the following sketch, which again follows [#smith]_.
+#
+# .. image:: ../_static/demonstration_assets/constant_depth_mps_prep/algorithm.png
+#     :width: 90%
+#     :align: center
```
Qottmann, line 618: I honestly wouldnt mind having an image with that level of detail already at the beginning 😬 Like I can look at this with no prior knowledge and already extract the basic ideas and building blocks of the algorithm on a high level. When it's buried down here I might never see it because I dont make it all the way through the demo 😅

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,843 @@
...
+# considerations. First, the block must be such that operator pushing relations are
+# available. Often this will determine a minimal block size, and multiples of that
+# size are valid choices as well. Second, the depth of the (parallelized) sequential
+# preparation circuits is proportional to :math:`q`, determining a key contribution
+# to the overall depth of the algorithm. Third, the sequential algorithm requires
+# two bond qudits for each block, leading to :math:`\frac{2N}{q}` auxiliary qudits
+# overall. Note how the product of the depth and the number of auxiliary qubits is
+# linear in :math:`N`.
+#
+# **Example**
```
Qottmann, line 641: Shouldnt this be a subsection itself? Otherwise it feels like we are still discussing the block size q and then suddently there is this nice big example

Qottmann (commented):
This has turned out super nice, really well done @dwierichs 

Just have some further, rather nitpicky but I think important points about the (very nice) illustrations and their positioning 🙃

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: >Would be great to enumerate each step and write a short sentence what we are seeing in the text, currently it is a bit disjoint

So a bit more than what is written below?

dwierichs: >and potentially already call these steps by their name, mps prepare, fusion and operator push? (are those the three steps displayed here?)

Hm, yeah, they are subsummed a bit. Is it fine if I make the mapping to the names in the text or should it happen in the image?

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Mhm, oh dear :sweat_smile: So should I just put this one in the beginning as well?

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Will do! Did you have this impression in other places as well?

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: Ideally in the image but for three I guess it's okay to say first, second, and third

> So a bit more than what is written below?

Oh yeah, I read it, but didnt make the connection 🙈

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: I personally think it would be a good idea, but I leave it up to you to decide 😬

demonstrations/tutorial_constant_depth_mps_prep.py
Qottmann: no just here

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: That says enough :D Will improve the description then and make the reference to the steps explicit! :+1:

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: I think I'll include it. Then the steps also have a name right away and there is only one type of visual queue per step.

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: In order to be consistent I kept "Example" as a bold line only, but I demoted the "block size q" part to a bold line (instead of subsection) as well.

Qottmann: that works 👍

Qottmann (approved):
👨‍🍳 👌 
image

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+# following effect:
+#
+# - If it is a Pauli-:math:`Y`, all physical sites need to be corrected by a Pauli-:math:`X`, and
+#   a Pauli-:math:`Y` is pushed to the opposite bond site.
+#
+# - If it is a Pauli-:math:`X`, a Pauli-:math:`Y` is applied to the first physical qubit, and afterwards
+#   the case above is recovered, leading to applying Pauli-:math:`X` and pushing through a Pauli-:math:`Y`.
+#
+# - If it is a Pauli-:math:`Z`, a Pauli-:math:`Z` is applied to the first physical qubit and we are done.
+#   No operator is pushed through on the bond axis.
```
ikurecic, line 561: On pennylane.ai we normally write these as "Pauli X," without a hyphen. But if you prefer this, we can keep this.

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+#
+# We can recast this condition into the following dynamic circuit instructions:
+#
+# - If the first (second) bit of the measurement outcome bitstring is active, apply a Pauli-:math:`Z`
+#   (Pauli-:math:`Y`) to the first physical qubit
```
ikurecic, line 566:
```suggestion
#   (Pauli-:math:`Y`) to the first physical qubit.
```

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+    qml.cond(op_id[1], qml.Y)(w)
+    # Apply X on other physical sites if input is X or Y
+    for i in phys_wires[1:]:
+        qml.cond(op_id[1], qml.X)(i)
+    # Push through Y if input is X or Y
+    return np.array([op_id[1], op_id[1]])
+
+
+######################################################################
+# Piecing everything together
```
ikurecic, line 593: I see what you're trying to say, but it's a little vague. Could you choose something more specific and easily understandable at a glance? I can help with some ideas if you're not sure.

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+# ---------------------------
+#
+# Now that we discussed the sequential preparation algorithm, fusion with mid-circuit
+# measurements, and operator pushing, we have all components for the constant-depth
+# preparation algorithm. As inputs it requires the tensor :math:`A` (or the unitary
+# :math:`U` to reproduce :math:`A`), the total number of physical sites :math:`N`,
+# and a block size :math:`q`.
```
ikurecic, line 600:
```suggestion
# and the block size :math:`q`.
```
Unless there is a reason it would be "a."

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+#    the two remaining bond sites.
+#
+# We summarize the algorithm in the following sketch, which again follows [#smith]_.
+#
+# .. image:: ../_static/demonstration_assets/constant_depth_mps_prep/algorithm.png
+#     :width: 90%
+#     :align: center
+#
+# It is important to remember that showing the existence of suitable operator pushing
+# relations (and finding them explicitly) is a crucial step in general, which goes
```
ikurecic, line 625: What do you mean by it being a "crucial step in general?" I'm not quite following your thoughts here.

ikurecic (commented):
Thank you, @dwierichs . I've left some comments and suggestions. Please let me know if there's something I can clarify or help with. Thanks.

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+.. math::
+
+    \sum_m {A^m}^\dagger A^m = \mathbb{I}.
+
+We will not discuss additional redundancies here but refer to [#smith]_ and the
+mentioned reviews for more details.
+
+**Example**
+
+As an example, consider a chain of :math:`N` qubits (:math:`d=2`) and an MPS
```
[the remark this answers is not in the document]
dwierichs: Is there no "no-line-break" character in RST at all? :weary:

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+#
+# #. Start in the state :math:`|\psi_0\rangle = |0\rangle^{\otimes N}\otimes |00\rangle_D`.
+#    The last two sites are non-physical bond sites, encoded by :math:`D`-dimensional qudits.
+#
+# #. Entangle the bond qudits into the state
+#    :math:`|I\rangle = \frac{1}{\sqrt{D}}\sum_j |jj\rangle_D`.
+#
+# #. Apply the unitary :math:`U` to each of the :math:`N` physical sites, with the
+#    :math:`D`-dimensional tensor factor always acting on the first bond qudit.
+#    Denoting :math:`U` acting on the :math:`i`\ th physical site and the first bond site
```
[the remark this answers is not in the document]
dwierichs: This one actually sounds a bit funky in my ears :thinking: I'll rephrase the sentence

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+    """Run the preparation circuit and projectively measure the bond qubits."""
+    wires = list(range(N + 2))
+    sequential_preparation(g, wires)
+    project_measure(0, N + 1)
+    return qml.probs(wires=wires[1 : N + 1])
+
+
+######################################################################
+# We will verify the prepared state below, when comparing to the constant-depth
+# preparation circuit. Let's just draw the circuit for now.
```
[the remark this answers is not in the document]
dwierichs: I meant that we check what kind of state we're actually preparing later on. But as we use the linear-depth preparation as source of truth later on, this is a circular statement. I'll just remove this.

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: Can just skip the "in general", does it work then?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+Then, all that is left to do is to 4) correct the defect at the boundary and
+finally 5) to measure two remaining bond qudits projectively.
+
+Building blocks
+---------------
+
+We briefly introduce MPS and a sequential circuit with linear
+depth to prepare them, as well as two tools from quantum information theory
+that we will use below: fusion of MPS with mid-circuit measurements
+and *operator pushing*.
```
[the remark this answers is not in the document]
dwierichs: The first two are "briefly introduce MPS" and "a sequential circuit with linear depth to prepare them".
Should I make it more explicit to mirror the section structure?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+They can be described and manipulated conveniently on classical computers and
+are able to approximate relevant states in quantum many-body systems.
+In particular, MPS can efficiently describe ground states of (gapped local)
+one-dimensional Hamiltonians, which in addition can be found efficiently using
+density matrix renormalization group (DMRG) algorithms.
+For a review of MPS see [#schollwoeck]_.
+
+Following [#smith]_, we will look at translation-invariant MPS
+of a quantum :math:`N`-body system where each body, or site, has local (physical)
+dimension :math:`d`. We will further restrict to periodic boundary conditions,
```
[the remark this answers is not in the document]
dwierichs: How does it work with ":math:\`N\`-body"? Should it be ":math:\`N-\`\ body? :sweat_smile:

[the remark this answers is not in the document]
Qottmann: could do `:math:'N-\text{body}'` though I am not sure if the font will be latex default or from the document

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: Mhm, actually rephrasing it does not really evade the problem.
Is it not fine to say "U acts on ...", does one say "The U acts on..." ?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+#
+# The last building block we need is so-called operator pushing. In fact, whether or not an MPS
+# can be prepared with the constant-depth circuit depends on whether certain operator pushing
+# relations are satisfied between the matrices :math:`B^k` from the fusion step and the MPS
+# tensor :math:`A`. We will not go into detail about these conditions but assume here that
+# we work with compatible MPS and measurements.
+#
+# Operator pushing then allows us to push a defect operator :math:`B^k` from one bond
+# axis through the tensor :math:`A` of the MPS to the other bond axis and to the physical axis.
+# While doing so, the operator in general will change, i.e., we end up with different operators
```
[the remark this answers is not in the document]
dwierichs: Went for "In general, while doing so will change the operator"

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+#
+# .. image:: ../_static/demonstration_assets/constant_depth_mps_prep/operator_pushing_example.png
+#     :width: 90%
+#     :align: center
+#
+# That is, in the notation above we have
+# :math:`X\mapsto (Y, Y)`,
+# :math:`Y\mapsto (Y, X)`, and
+# :math:`Z\mapsto (I, Z)` as well as the trivial
+# :math:`I\mapsto (I, I)`.
```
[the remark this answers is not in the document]
dwierichs: I think I'd rather delete the mention of `I` here, because it is really trivial, and it's not worth it to call in Tarik again to redo the image, no?

demonstrations/tutorial_constant_depth_mps_prep.py
```diff
@@ -0,0 +1,848 @@
...
+# In the fusion step above we picked the Bell basis as measurement basis, without
+# further specifying why. As we saw, this basis leads to Pauli matrices as impurities
+# :math:`B^k`. It turns out that those matrices satisfy simple pushing relations together with
+# the tensor :math:`A` of our example MPS :math:`|\Psi(g)\rangle`, explaining this choice of
+# measurement basis. In particular, the relations are
+#
+# .. image:: ../_static/demonstration_assets/constant_depth_mps_prep/operator_pushing_example.png
+#     :width: 90%
+#     :align: center
+#
```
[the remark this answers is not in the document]
dwierichs: Like so?

demonstrations/tutorial_constant_depth_mps_prep.py
dwierichs: something like this?
"Piecing the algorithm together"
"Building the algorithm from the building blocks"
"Putting the building blocks together"

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
dwierichs: Oh dear :sweat_smile:

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
ikurecic: Actually, I think this is me being confused. I have no idea how I didn't clock this. Please disregard.

demonstrations/tutorial_constant_depth_mps_prep.py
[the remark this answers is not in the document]
ikurecic: I leave that to your artistic license. Some of these are complicated and neither option looks good.

ikurecic (approved):

PR #1195, opened 2024-08-20
Fixed Qrack demo. State cannot work with finite shots

mehrdad2m:
**Title:**
Modified qrack simulator demo.

**Summary:**
State measurements can not work with finite shots. However, there is such a case in qrack simulator demo which can cause failure with the latest catalyst which verifies this.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**
[sc-70792]


dime10 (approved):
Looks good 👍  @josh146 second review? :)

josh146 (approved):
Nice catch!

PR #1196, opened 2024-08-21
Fix Failing Demo after Facade Merge

Shiro-Raven:
[[sc-71847](https://app.shortcut.com/xanaduai/story/71847)]

### DIFF
--- a/demonstrations/tutorial_error_mitigation.py
+++ b/demonstrations/tutorial_error_mitigation.py
@@ -153,9 +153,7 @@ from pennylane.transforms import mitigate_with_zne
 extrapolate = RichardsonFactory.extrapolate
 scale_factors = [1, 2, 3]
 
-mitigated_qnode = mitigate_with_zne(
-    noisy_qnode, scale_factors, fold_global, extrapolate
-)
+mitigated_qnode = mitigate_with_zne(noisy_qnode, scale_factors, fold_global, extrapolate)
 mitigated_qnode(w1, w2)
 
 ##############################################################################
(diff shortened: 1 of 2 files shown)


demonstrations/tutorial_error_mitigation.py
```diff
@@ -525,7 +516,7 @@ def qchem_circuit(phi):
     def executor(circuit):
 
         # Add Hamiltonian measurement to circuit
-        circuit_with_meas = qml.tape.QuantumTape(circuit.operations, [qml.expval(H)])
```
josh146, line 528: Hey @Shiro-Raven, how come specifying shots on the device doesn't work any more? 

Just asking because this is quite a common user pattern (creating multiple devices with different settings) they I see, so there might be some impact beyond this demo

EmilianoG-byte (approved):
Looks good to me! 👏

demonstrations/tutorial_error_mitigation.py
EmilianoG-byte: Was this actually failing @Shiro-Raven or is it just because we are trying to move to add shots on tapes instead of devices?

Shiro-Raven: > Hey @Shiro-Raven, how come specifying shots on the device doesn't work any more? 
> 
> Just asking because this is quite a common user pattern (creating multiple devices with different settings) they I see, so there might be some impact beyond this demo

This only needs to be taken care of for Legacy devices, otherwise nothing's changed

Shiro-Raven: > Was this actually failing @Shiro-Raven or is it just because we are trying to move to add shots on tapes instead of devices? 

It was falling

josh146: > This only needs to be taken care of for Legacy devices, otherwise nothing's changed

Is this something we could fix in the legacy device fade? As there are still quite a large number of legacy devices out there, so this would be considered a breaking change in the UI 🤔

josh146: Nevermind, ignore me! It's because this demo is using `tape` and not a `QNode` 🤦

mudit2812 (approved):

PR #1199, opened 2024-08-27
Merge master branch into dev

josh146:


josh146:
hey everyone, quite a significant PR here, as I noticed that `master` has not been merged into `dev` in a long time.

- In the `master` branch, image locations were refactored. I am not sure though why so many images are appearing as new/changed' files though 🤔 
- There are a few new demos in the `master` branch.

@mudit2812 @Alex-Preciado this PR should be merged in _as soon as possible_ this week, as the `dev` CI we have been using will not be accurate until it is merged in.

@mudit2812 I imagine that when we merge `dev` -> `master` next week, we may get a lot of merge conflicts with the images :(

mudit2812 (commented):
I don't think I have the context about `qml`'s architecture to meaningfully review this PR. I'll leave it to others to do so, but definitely agree @josh146 that this is extremely important to merge as soon as possible.

josh146:
At the moment, a couple of things are happening:

- `tutorial_how_to_quantum_just_in_time_compile_vqe_catalyst` is failing, but this will be fixed shortly by https://github.com/PennyLaneAI/catalyst/pull/1062

- `tutorial_qubit_rotation` is failing:

  ```
  WARNING: /home/runner/work/qml/qml/demonstrations/tutorial_qubit_rotation.py failed to execute correctly: Traceback (most recent call last):
  File "/home/runner/work/qml/qml/demonstrations/tutorial_qubit_rotation.py", line 271, in <module>
    print(dcircuit(phi1, phi2))
  TypeError: Gradient only defined for scalar-output functions. Output had shape: (1,).
  ```

  @mudit2812 is this to be expected? 🤔 

- We are also seeing issues with pip installing the nightly PL/Lightning/Catalyst wheels! In particular, see this part of the logs: https://github.com/PennyLaneAI/qml/actions/runs/10586577302/job/29335696184?pr=1199#step:9:551

  - It seems that catalyst 0.9-dev0 is being correctly installed (the rc), but it is failing to find lightning/kokkos 0.39-dev0 rc wheels (only 0.38-dev wheels).
  - As a result, catalyst is being downgraded to 0.8-dev1!!

mudit2812:
> * `tutorial_qubit_rotation` is failing:
>   ```
>   WARNING: /home/runner/work/qml/qml/demonstrations/tutorial_qubit_rotation.py failed to execute correctly: Traceback (most recent call last):
>   File "/home/runner/work/qml/qml/demonstrations/tutorial_qubit_rotation.py", line 271, in <module>
>     print(dcircuit(phi1, phi2))
>   TypeError: Gradient only defined for scalar-output functions. Output had shape: (1,).
>   ```
>   @mudit2812 is this to be expected? 🤔

Definitely not expected. I'm going to investigate this, thanks for pointing it out.

> * We are also seeing issues with pip installing the nightly PL/Lightning/Catalyst wheels! In particular, see this part of the logs: https://github.com/PennyLaneAI/qml/actions/runs/10586577302/job/29335696184?pr=1199#step:9:551
>   
>   * It seems that catalyst 0.9-dev0 is being correctly installed (the rc), but it is failing to find lightning/kokkos 0.39-dev0 rc wheels (only 0.38-dev wheels).
>   * As a result, catalyst is being downgraded to 0.8-dev1!!

@josh146  Who do you think is the best person to look at this?

josh146:
> Definitely not expected. I'm going to investigate this, thanks for pointing it out.

I just ran the qubit rotation tutorial locally and everything worked fine 😅 So maybe it's a false alarm, let's see how the CI goes.

josh146:
@mudit2812 ah never mind! When I pull the latest PL RC I do see qubit rotation failing locally. There must have been a change in the last few hours. Have fixed it here: [6a3e20b](https://github.com/PennyLaneAI/qml/pull/1199/commits/6a3e20bfc795f67ab010daf5b9212b18d2e6ad5f)

Makefile
```diff
@@ -77,7 +77,7 @@ environment:
 			$$PYTHON_VENV_PATH/bin/python -m pip install --upgrade git+https://github.com/PennyLaneAI/pennylane-cirq.git#egg=pennylane-cirq;\
 			$$PYTHON_VENV_PATH/bin/python -m pip install --upgrade git+https://github.com/PennyLaneAI/pennylane-qiskit.git#egg=pennylane-qiskit;\
 			$$PYTHON_VENV_PATH/bin/python -m pip install --upgrade git+https://github.com/PennyLaneAI/pennylane-qulacs.git#egg=pennylane-qulacs;\
-			$$PYTHON_VENV_PATH/bin/python -m pip install -i https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade;\
-			$$PYTHON_VENV_PATH/bin/python -m pip install -i https://test.pypi.org/simple/ PennyLane-Catalyst --pre --upgrade;\
+			$$PYTHON_VENV_PATH/bin/python -m pip install --extra-index-url https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade;\
+			$$PYTHON_VENV_PATH/bin/python -m pip install --extra-index-url https://test.pypi.org/simple/ PennyLane-Catalyst --pre --upgrade;\
```
josh146, line 81: Updating from `-i` to `--extra-index-url` solves the issue with downgrades of the catalyst and lightning packages.

pyproject.toml
```diff
@@ -42,13 +42,13 @@ pennylane = "0.37.0"
 pennylane-cirq = "0.36.0"
 pennylane-qiskit = "0.37.0"
 pennylane-qulacs = "0.36.0"
-pennylane-catalyst = "0.7.0"
+# pennylane-catalyst = "0.7.0"
```
josh146, line 45: This will need to be uncommented and updated (to `0.8.0`) when we merge `dev` into `master`.

pyproject.toml
```diff
@@ -61,17 +61,21 @@ neural-tangents = "0.6.2"
 plotly = ">=4.5.0"
 pydantic = "^2.8.2"
 ply = "3.11"
-optax = "0.1.7"
+optax = "0.2.3"
```
josh146, line 64: optax needs to be updated to work with jax 0.4.28

pyproject.toml
```diff
@@ -42,13 +42,13 @@ pennylane = "0.37.0"
...
-pennylane-catalyst = "0.7.0"
+# pennylane-catalyst = "0.7.0"
 ##########################################################
 
 
 matplotlib = "3.7.2"
-jax = "0.4.23"
-jaxlib = "0.4.23"
+jax = "0.4.28"
+jaxlib = "0.4.28"
```
josh146, line 51: jax needs to be updated to work with catalyst 0.8

pyproject.toml
```diff
@@ -27,7 +27,7 @@ sphinx = ">=5.0.2"
 sphinx_gallery = "0.12.2"
 Jinja2 = "3.0.3"
 markupsafe = "2.1.1"
-numpy = "~1.23"
+numpy = "~1.24"
```
josh146, line 30: optax update requires a newer version of numpy

demonstrations/tutorial_how_to_quantum_just_in_time_compile_vqe_catalyst.py
```diff
@@ -116,15 +116,14 @@ def cost(params):
...
 from jax import numpy as jnp
 
-hf = jnp.array(dataset.hf_state)
+hf = np.array(dataset.hf_state)
 
 @qml.qjit
 @qml.qnode(dev)
 def cost(params):
-    qml.BasisState(hf, wires=range(qubits))
+    qml.BasisState.compute_decomposition(hf, wires=range(qubits))
```
josh146, line 126: This is because of this bug here: https://github.com/PennyLaneAI/catalyst/issues/1065

If this bug is resolved before the release, we should revert this change.

demonstrations/tutorial_jax_transformations.py
```diff
@@ -43,15 +43,13 @@
 # set up a two-qubit device for computations. We'll be using the ``default.qubit`` device
 # for the first part of this tutorial.
 
-# Added to silence some warnings.
-from jax.config import config
-
-config.update("jax_enable_x64", True)
```
josh146, line 49: Importing the configuration directly was deprecated in jax v0.4.25

demonstrations/tutorial_tn_circuits.py
```diff
@@ -233,7 +233,7 @@ def circuit(template_weights):
 
 shapes = qml.SimplifiedTwoDesign.shape(n_layers=1, n_wires=4)
 weights = [np.random.random(size=shape) for shape in shapes]
-template_weights = [weights] * 3
+template_weights = qml._np.array([weights] * 3, dtype="object")
```
josh146, line 236: Unfortunately, ragged arrays were deprecated in NumPy v1.24. Here, `qml._np` is a hacky way to get vanilla numpy (since this fix doesn't work with autograd).

albi3ro:
> @mudit2812 ah never mind! When I pull the latest PL RC I do see qubit rotation failing locally. There must have been a change in the last few hours. Have fixed it here: [6a3e20b](https://github.com/PennyLaneAI/qml/pull/1199/commits/6a3e20bfc795f67ab010daf5b9212b18d2e6ad5f)

That was probably the "treat batch size one as batch size one" change that went in yesterday.  So yes, it did cause an issue here :( , but easy to remove the batch dimension from the input parameters.

demonstrations/tutorial_tn_circuits.py
mudit2812: Why don't we just change the imports in the tutorial to the following:

```python
import pennylane.numpy as pnp
import numpy as np
```

albi3ro: Because `template_weights = qml.numpy.array([weights] * 3, dtype="object")` still raises an error.

mudit2812: But we can just use vanilla numpy. I just want to avoid using `qml._np`

module_resets.py
```diff
@@ -0,0 +1,8 @@
+import importlib
+
+
+def reset_jax(gallery_conf, fname):
+	import jax
+    importlib.reload(jax)
```
mudit2812, line 6:
```suggestion
def reset_jax(gallery_conf, fname):
    import jax
    importlib.reload(jax)
```

demonstrations/tutorial_tn_circuits.py
josh146: yeah this was a quick hack by me just to get the demo building asap so we can merge this PR :) btw @mudit2812 even using vanilla numpy here fails in numpy 1.25+, so we probably need a change in PL for the long term fix here

PR #1202, opened 2024-08-28
Update failing demos

mudit2812:
As name says. Lots of demos are failing after #1199 was merged. This PR aims to fix all those failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,11 +73,11 @@ environment:
 		if [ '$(UPGRADE_PL)' = 'true' ]; then\
 			echo "Updating PennyLane and plugins to latest ... ";\
 			PYTHON_VENV_PATH=`$(POETRY_BIN) env info --path`;\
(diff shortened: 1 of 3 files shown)


mudit2812:
The failing demo has been fixed. The trace back does not reflect the changes that have been merged into the RC branch in PennyLane (see [here](https://github.com/PennyLaneAI/qml/actions/runs/10607618982/job/29400379522?pr=1202#step:19:1936)), so my guess is that pennylane is not being installed from the RC branch. This shouldn't be the case since the `Makefile` in `dev` has been updated to install pennylane from the RC branch (see [here](https://github.com/PennyLaneAI/qml/blob/1714f2a4d322adee4fdf24ef3ebd241616188362/Makefile#L76))

josh146 (approved):

Alex-Preciado (approved):

PR #1220, opened 2024-09-18
Remove code being deprecated in v0.39

mudit2812:
[sc-73935]

### DIFF
--- a/demonstrations/tutorial_backprop.py
+++ b/demonstrations/tutorial_backprop.py
@@ -69,19 +69,28 @@ key = jax.random.PRNGKey(42)
 # create a device to execute the circuit on
 dev = qml.device("default.qubit", wires=3)
 
+
+def CNOT_ring(wires):
+    """Apply CNOTs in a ring pattern"""
+    n_wires = len(wires)
+
+    for w in wires:
+        qml.CNOT([w % n_wires, (w + 1) % n_wires])
+
+
 @qml.qnode(dev, diff_method="parameter-shift")
 def circuit(params):
     qml.RX(params[0], wires=0)
     qml.RY(params[1], wires=1)
     qml.RZ(params[2], wires=2)
 
-    qml.broadcast(qml.CNOT, wires=[0, 1, 2], pattern="ring")
+    CNOT_ring(wires=[0, 1, 2])
 
     qml.RX(params[3], wires=0)
     qml.RY(params[4], wires=1)
     qml.RZ(params[5], wires=2)
 
-    qml.broadcast(qml.CNOT, wires=[0, 1, 2], pattern="ring")
+    CNOT_ring(wires=[0, 1, 2])
     return qml.expval(qml.PauliY(0) @ qml.PauliZ(2))
 
 
--- a/demonstrations/tutorial_kernels_module.py
+++ b/demonstrations/tutorial_kernels_module.py
@@ -250,7 +250,9 @@ def layer(x, params, wires, i0=0, inc=1):
         i += inc
         qml.RY(params[0, j], wires=[wire])
 
-    qml.broadcast(unitary=qml.CRZ, pattern="ring", wires=wires, parameters=params[1])
+    n_wires = len(wires)
+    for p, w in zip(params[1], wires):
+        qml.CRZ(p, wires=[w % n_wires, (w + 1) % n_wires])
 
 
 ##############################################################################
--- a/demonstrations/tutorial_local_cost_functions.py
+++ b/demonstrations/tutorial_local_cost_functions.py
@@ -241,14 +241,16 @@ def global_cost_simple(rotations):
     for i in range(wires):
         qml.RX(rotations[0][i], wires=i)
         qml.RY(rotations[1][i], wires=i)
-    qml.broadcast(qml.CNOT, wires=range(wires), pattern="chain")
+    for i in range(wires - 1):
+        qml.CNOT([i, i + 1])
     return qml.probs(wires=range(wires))
 
 def local_cost_simple(rotations):
     for i in range(wires):
         qml.RX(rotations[0][i], wires=i)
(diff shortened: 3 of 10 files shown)


demonstrations/tutorial_classical_kernels.metadata.json
```diff
@@ -6,7 +6,7 @@
         }
     ],
     "dateOfPublication": "2022-03-01T00:00:00+00:00",
-    "dateOfLastModification": "2024-09-04T00:00:00+00:00",
```
astralcai, line 9: The demo updated was `tutorial_kernels_module`, not `tutorial_classical_kernels`.

demonstrations/tutorial_kernels_module.py
```diff
@@ -347,7 +349,7 @@ def kernel(x1, x2, params):
 init_kernel = lambda x1, x2: kernel(x1, x2, init_params)
 K_init = qml.kernels.square_kernel_matrix(X, init_kernel, assume_normalized_kernel=True)
 
-with np.printoptions(precision=3, suppress=True):
+with pnp.printoptions(precision=3, suppress=True):
```
astralcai, line 352: A non-blocking comment, more of a personal preference: I typically prefer to keep both `pnp` and `np` and only use `pnp` to create parameters that are meant to be trainable. For everything else I still use vanilla `numpy`

demonstrations/tutorial_local_cost_functions.py
```diff
@@ -371,7 +373,8 @@ def tunable_cost_simple(rotations):
     for i in range(wires):
         qml.RX(rotations[0][i], wires=i)
         qml.RY(rotations[1][i], wires=i)
-    qml.broadcast(qml.CNOT, wires=range(wires), pattern="chain")
```
astralcai, line 374: Is `qml.broadcast` deprecated now?

mudit2812: Yep, the recommended alternative is to use loops

demonstrations/tutorial_classical_kernels.metadata.json
mudit2812: Thanks for catching this.

demonstrations/tutorial_kernels_module.py
mudit2812: Actually I thought more about it, and from a reader's perspective, I don't think it makes sense to import both `numpy` and `pennylane.numpy` unless absolutely necessary. That was the case in `tutorial_tn_circuits`, because there's a point where we create a ragged numpy array (it's possible with `dtype="object"`). This doesn't work with the version of autograd that gets installed in `qml`, and the `qml.MPS` template doesn't work if I don't wrap the ragged list inside an array.

Makefile
```diff
@@ -76,8 +76,8 @@ environment:
 			$$PYTHON_VENV_PATH/bin/python -m pip install --upgrade git+https://github.com/PennyLaneAI/pennylane-cirq.git#egg=pennylane-cirq;\
 			$$PYTHON_VENV_PATH/bin/python -m pip install --upgrade git+https://github.com/PennyLaneAI/pennylane-qiskit.git#egg=pennylane-qiskit;\
 			$$PYTHON_VENV_PATH/bin/python -m pip install --upgrade git+https://github.com/PennyLaneAI/pennylane-qulacs.git#egg=pennylane-qulacs;\
-			$$PYTHON_VENV_PATH/bin/python -m pip install --extra-index-url https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade;\
 			$$PYTHON_VENV_PATH/bin/python -m pip install --extra-index-url https://test.pypi.org/simple/ PennyLane-Catalyst --pre --upgrade;\
+			$$PYTHON_VENV_PATH/bin/python -m pip install --extra-index-url https://test.pypi.org/simple/ PennyLane-Lightning --pre --upgrade;\
```
mudit2812, line 80: I reordered the installations such that Catalyst now gets installed before Lightning. The reason is that after Lightning got installed, Catalyst would downgrade it when it was getting installed.

mudit2812:
The build failure [here](https://github.com/PennyLaneAI/qml/actions/runs/10961656107/job/30439178406?pr=1220) seems to come from `sphinx_gallery`, and I don't know how to debug it, since there is no traceback. I've verified that the versions of pennylane and lightning that are installed account for all deprecations. Catalyst _could_ be causing the issue, but `qml` has no references to `qml.Device`, which is the deprecated class that is causing the failure.

josh146:
Hey @mudit2812, how come the demos were updated to use `pnp` instead of `np`?

While we have been using `pnp` internally in the source code, we have deliberately kept demos/tutorials using `from pennylane import numpy as np` to stay consistent with all of our public documentation, papers, slides, codebook, etc.

josh146:
> What is the recommendation for demos where both vanilla and pennylane numpy are needed?

I don't think we've been in a position where we've had to do this in a demo before 🤔 

> Though for the `tutorial_tn_circuits` demo, I can't get the demo to work unless we use vanilla numpy for the template_weights given to `qml.MPS`, and we need `pennylane.numpy` for optimization. 

Interesting! do we know why MPS doesn't work with PL numpy (sounds like a bug?)

In the meantime, I think it's okay to import vanilla numpy here (since original numpy is only being used to patch a bug/issue, we could import it as `import numpy as onp`)

demonstrations/tutorial_backprop.py
```diff
@@ -55,7 +55,7 @@
 Let's have a go implementing the parameter-shift rule manually in PennyLane.
 """
 import pennylane as qml
-from jax import numpy as np
+from jax import numpy as jnp
```
josh146, line 58: nice catch!

josh146 (approved):

astralcai (approved):

PR #1224, opened 2024-09-25
Periods, commas etc. inside inline math command

ikurecic:
As requested [here](https://xanaduhq.slack.com/archives/C03FRM3RCCV/p1727273358129059).


dwierichs (approved):
Looked at about 100 corrections and couldn't spot any issues, so I'm gonna assume that the regex was correct :D :100:

Thanks @ikurecic !

justinpickering (approved):

PR #1211, opened 2024-09-05
Update qchem demos to be JAX or JAX/JIT compatible

austingmhuang:
Title:
Update qchem demos to be JAX or JAX/JIT compatible

Summary:
As we phase out autograd, we need to ensure that our current qchem demos can work with either vanilla numpy or jax. Demos that are JIT compatible were made to use JIT as well. For demos that don't have any differentiation, we opt to use vanilla numpy.

Notes: the qubit tapering demo is not JIT compatible due to 2 issues: 1) Wire ordering is not ascending (Not a bug, but might need attention) 2) qml.taper_operation() uses the Exp operator, which has a conditional statement (https://github.com/PennyLaneAI/pennylane/issues/5993)

Relevant references:

Possible Drawbacks:

Related GitHub Issues:
[[sc-69776](https://app.shortcut.com/xanaduai/story/69776)]

PR #1226, opened 2024-10-01
Fix demos to use jax that require grads

austingmhuang:
Doesn't build due to the fact that [this PR](https://github.com/PennyLaneAI/pennylane/pull/6096) is not yet merged. JSON files need to be updated as well.

**Summary:**
Fixes remaining demos that checked for requires_grad. Won't build for now due to missing a fix in [this PR](https://github.com/PennyLaneAI/pennylane/pull/6324)

[[sc-69776](https://app.shortcut.com/xanaduai/story/69776)] [[sc-69778](https://app.shortcut.com/xanaduai/story/69778)]


demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -191,12 +194,12 @@ def H(x):
 # The ``hf`` array is used by the :class:`~.pennylane.BasisState` operation to initialize
 # the qubit register. Then, the :class:`~.pennylane.DoubleExcitation` operations are applied
 # First, we define the quantum device used to compute the expectation value.
-# In this example, we use the ``lightning.qubit`` simulator:
+# In this example, we use the ``default.qubit`` simulator:
 num_wires = 6
-dev = qml.device("lightning.qubit", wires=num_wires)
+dev = qml.device("default.qubit", wires=num_wires)
```
austingmhuang, line 199: For some reason lightning.qubit doesn't work with JAX here. Not sure why but I had to change to using default to get it working.

demonstrations/tutorial_resource_estimation.py
```diff
@@ -65,12 +65,16 @@
 `6-31g basis set <https://en.wikipedia.org/wiki/Basis_set_(chemistry)>`_ as an example.
 """
 import pennylane as qml
-from pennylane import numpy as np
+import numpy as np
+from jax import numpy as jnp
+import jax
```
soranjh, line 70: Let's just use vanilla numpy in this demo.

demonstrations/tutorial_mapping.py
```diff
@@ -202,11 +202,11 @@
...
 
 from pennylane import qchem
-from pennylane import numpy as np
+from jax import numpy as jnp
 
 symbols  = ['H', 'H']
-geometry = np.array([[0.0, 0.0, -0.69434785],
-                     [0.0, 0.0,  0.69434785]], requires_grad = False)
+geometry = jnp.array([[0.0, 0.0, -0.69434785],
+                     [0.0, 0.0,  0.69434785]])
```
soranjh, line 209: This demo is already updated in another [PR](https://github.com/PennyLaneAI/qml/pull/1211), correct?

demonstrations/tutorial_fermionic_operators.py
```diff
@@ -108,7 +108,7 @@
 # The matrix representation of the qubit Hamiltonian in the computational basis can be diagonalized
 # to get its eigenpairs.
 
-from pennylane import numpy as np
+import numpy as np
```
soranjh, line 111: Can we update this demo in this [PR](https://github.com/PennyLaneAI/qml/pull/1211) as well?

soranjh (changes requested):
Thanks @austingmhuang. befor doing a full review, could you please make sure that this PR targets dev instead of master? Also, if there is any demo that does not require features in dev, their upgrade should be done in this [PR](https://github.com/PennyLaneAI/qml/pull/1211).

demonstrations/tutorial_mapping.py
austingmhuang: No, it's not updated in that PR. Well that PR had a one line change in this demo by accident.

austingmhuang: This demo doesn't work without the requires_grad changes because it calls fermionic_hamiltonian directly

demonstrations/tutorial_fermionic_operators.py
austingmhuang: No, it cannot be done. This demo calls `electronic_integrals()` directly, which checks for requires_grad, leading to an error.

demonstrations/tutorial_differentiable_HF.py
```diff
@@ -228,13 +230,13 @@
 # plane.
 
 n = 30 # number of grid points along each axis
-
+qml.qchem.hf_energy(mol)()
```
KetpuntoG, line 233: What is this line doing?

demonstrations/tutorial_differentiable_HF.py
```diff
@@ -313,44 +317,41 @@ def circuit(*args):
...
+args = [circuit_param, geometry, coeff, alpha]
 
-    args = [circuit_param, geometry, alpha, coeff]
+for n in range(36):
+    args = [circuit_param, geometry, coeff, alpha]
     mol = qml.qchem.Molecule(symbols, geometry, alpha=alpha, coeff=coeff)
 
     # gradient for circuit parameters
-    g_param = grad(energy(mol), argnum=0)(*args)
+    g_param = jax.grad(energy(), argnums=[0, 1, 2, 3])(*args)[0]
```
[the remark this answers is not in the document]
austingmhuang: Yes 👍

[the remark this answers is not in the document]
austingmhuang: We were doing that beforehand too (geom, alpha, coeff, and circuit_param were all labelled with `requires_grad=True`. We just update the UI to use jax UI instead.

demonstrations/tutorial_differentiable_HF.py
austingmhuang: It is a bit of a hack to reset the information in `mol` to use numpy arrays instead of jax arrays. When we did `jax.grad(qml.qchem.hf_energy(mol), argnums=0)(geometry, mol.coeff, mol.alpha)`, internally some of the information of the molecule automatically uses jax arrays. This is great, but not for this graphing part, where we want to stick to using numpy.

demonstrations/tutorial_differentiable_HF.py
```diff
@@ -101,16 +101,18 @@
...
 
 symbols = ["H", "H"]
 # optimized geometry at the Hartree-Fock level
-geometry = np.array([[-0.672943567415407, 0.0, 0.0],
-                     [ 0.672943567415407, 0.0, 0.0]], requires_grad=True)
+geometry = jnp.array([[-0.672943567415407, 0.0, 0.0],
+                     [ 0.672943567415407, 0.0, 0.0]])
 
 ##############################################################################
 # The use of ``requires_grad=True`` specifies that the nuclear coordinates are differentiable
```
austingmhuang, line 118: We should delete this line.

demonstrations/tutorial_mol_geo_opt.py
```diff
@@ -312,44 +310,39 @@ def grad_x(params, x):
...
+    theta = theta - 0.8 * g_param
 
-    energy.append(cost(theta, x))
-    bond_length.append(np.linalg.norm(x[0:3] - x[3:6]) * bohr_angs)
+    # gradient for coordinates
+    value, _ = jax.value_and_grad(cost, argnums=1)(theta, x)
+    grad = grad_x(theta, x)
+    x = x - 0.8 * grad
+    energies.append(value)
+    bond_length.append(jnp.linalg.norm(x[0] - x[1]) * bohr_angs)
```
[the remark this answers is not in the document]
austingmhuang: `x` is still the nuclear gradients. Just now we are manually updating its value instead of letting opt_x.step() take care of that.

KetpuntoG (approved):
Look good! please update the metadata before merging

PR #1233, opened 2024-10-08
Update Deprecated Github Actions

brownj85:
**Summary:**
- Updates `actions/checkout` to v4
- Updates `actions/setup-node` to v4 and uses node version 20

**Story**
https://app.shortcut.com/xanaduai/story/60888/update-actions-checkout-version-in-qml-workflows


Mandrenkov (commented):
Looks reasonable to me. Is there a way to easily verify that the workflows still work?

brownj85:
> Looks reasonable to me. Is there a way to easily verify that the workflows still work?
 
With `actions/setup-node`, we can see it working here: https://github.com/PennyLaneAI/qml/actions/runs/11257971811

With `actions/checkout`, really hard to see anything breaking - the [interface](https://github.com/actions/checkout) is unchanged and we don't use any fancy features. If anything does break it'll be a quick fix.

Mandrenkov (approved):

doctorperceptron (approved):
:partying_face:

PR #1237, opened 2024-10-10
Fix a bug in dense_to_mps

Qottmann:
Bug reported by @hsim13372 , thanks a lot!

Quick fix. This was missed because the function is just used to show the dimensionality.


dwierichs (approved):
lgtm!

ikurecic (approved):

PR #1240, opened 2024-10-15
Update qulacs to latest version

mudit2812:
Failures happening in both `dev` and `master` builds due to `ModuleNotFoundError: No module named 'qulacs'`. This PR updates the `qulacs` version in `pyproject.toml` to attempt resolving this issue.

[sc-75989]


josh146 (approved):
Thanks @mudit2812! Could you check that the tutorial(s) using qulacs are still returning the same results as on main?

mudit2812:
Confirmed @josh146 . The only demo that uses qulacs (Intro to QAOA) has the same output as on the public website.

astralcai (approved):

PR #1242, opened 2024-10-17
Update `update-dev.yml` to use `git` instead of third-party action

mudit2812:
The `update-dev` action keeps failing due to unrelated histories between the `dev` and `master` branches, and the third party action used for the nightly merge does not include any options to allow merging with unrelated histories. So, this PR updates the action to use `git` instead.

This will also give us more flexibility in the future.

Couple of updates:
* I have confirmed that configuring `git` using `secrets.NIGHTLY_TOKEN` does not expose the token in the workflow logs. See [here](https://github.com/PennyLaneAI/qml/actions/runs/11408914907/job/31748090296#step:3:4) for an example.

[sc-76254]


.github/workflows/update-dev.yml
```diff
@@ -27,12 +27,31 @@ jobs:
...
-        with:
-          stable_branch: 'master'
-          development_branch: 'dev'
-          allow_ff: false
         env:
-          GITHUB_TOKEN: ${{ secrets.NIGHTLY_TOKEN }}
+          MERGE_ARGS: --verbose --no-ff --allow-unrelated-histories
+        run: |
+          git config --global user.email '${{ secrets.AUTO_UPDATE_VERSION_RINGO_EMAIL }}'
+          git config --global user.name "ringo-but-quantum"
```
mudit2812, line 46: Using ringo so that we can merge without worrying about branch protection. I just copied this from another workflow that uses ringo.

.github/workflows/update-dev.yml
```diff
@@ -27,12 +27,31 @@ jobs:
...
-        uses: robotology/gh-action-nightly-merge@v1.3.3
-        with:
-          stable_branch: 'master'
-          development_branch: 'dev'
-          allow_ff: false
         env:
-          GITHUB_TOKEN: ${{ secrets.NIGHTLY_TOKEN }}
+          MERGE_ARGS: --verbose --no-ff --allow-unrelated-histories
+        run: |
+          git config --global user.email '${{ secrets.AUTO_UPDATE_VERSION_RINGO_EMAIL }}'
```
rashidnhm, line 45: Does this secret exist in QML repo?

.github/workflows/update-dev.yml
rashidnhm, line 1: Is the merge being pushed up anywhere?

mudit2812: Damn, good catch 😆

.github/workflows/update-dev.yml
mudit2812: I have no idea. If not, what would be the strategy? Adding ringo to `qml`? Or is there some other way to merge directly into `dev`, which has branch protection?

rashidnhm: There does exist a nightly token that we use now. 


But using ringo might be a more future proof idea

.github/workflows/update-dev.yml
```diff
@@ -27,12 +27,28 @@ jobs:
...
+          # https://github.com/robotology/gh-action-nightly-merge/blob/master/entrypoint.sh
+
+          git config --global user.name "$CONFIG_USERNAME"
+          git config --global user.email "$CONFIG_EMAIL"
+
+          git fetch origin $MERGE_HEAD
+          (git checkout $MERGE_HEAD && git pull)
+
+          git fetch origin $MERGE_BASE
+          (git checkout $MERGE_BASE && git pull)
```
mudit2812, line 51:
```suggestion
          git checkout $MERGE_HEAD && git pull

          git fetch origin $MERGE_BASE
          git checkout $MERGE_BASE && git pull
```

rashidnhm (approved):

.github/workflows/update-dev.yml
mudit2812: Ringo is not available on this repo. Using the already available NIGHTLY_TOKEN

mlxd (approved):

PR #1246, opened 2024-10-22
Example data fits Sphinx Gallery's image dimensions

doctorperceptron:
**Title:**
Adapt example data in demo to fit Sphinx Gallery's image dimensions

**Summary:**
Sphinx Gallery renders single images at 640px by 480px. Adjusting the layout of the example data to better fit these dimensions and reduce white space. 

**Relevant references:**
https://app.shortcut.com/xanaduai/story/76532

### Before:
image

### After:
image

### DIFF
--- a/demonstrations/tutorial_post-variational_quantum_neural_networks.py
+++ b/demonstrations/tutorial_post-variational_quantum_neural_networks.py
@@ -130,9 +130,10 @@ y_test = (y_test - 4) / 2
 #
 
 fig, axes = plt.subplots(nrows=2, ncols=3, layout="constrained")
-for i in range(6):
-    axes[i].matshow(X_train[2*i])
-    axes[i].axis('off')
+for i in range(2):
+    for j in range(3):
+      axes[i][j].matshow(X_train[2*(i+j)])
+      axes[i][j].axis('off')
 fig.subplots_adjust(hspace=0.0)
 fig.tight_layout()
 plt.show()
--- a/demonstrations/tutorial_post-variational_quantum_neural_networks.py
+++ b/demonstrations/tutorial_post-variational_quantum_neural_networks.py
@@ -4,20 +4,20 @@ Post-variational quantum neural networks
 """
 
 ######################################################################
-# You're sitting in front of your quantum computer, excitement buzzing through your veins as your 
-# carefully crafted Ansatz for a variational algorithm is finally ready. But oh buttersticks — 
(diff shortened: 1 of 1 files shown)


demonstrations/tutorial_post-variational_quantum_neural_networks.py
```diff
@@ -127,62 +127,63 @@
...
 ######################################################################
 # A visualization of a few data points is shown below.
-# 
+#
 
-fig, axes = plt.subplots(nrows=1, ncols=5, layout="constrained")
-for i in range(5):
-    axes[i].matshow(X_train[2*i])
-    axes[i].axis('off')
+fig, axes = plt.subplots(nrows=2, ncols=3, layout="constrained")
```
DanielNino27, line 132: Would specifying` figsize `inside of subplots help or would that be superseded by the display settings in sphinx? It looks better but seems disproportionately big

DanielNino27 (changes requested):
Thanks @doctorperceptron ! Just one small question. If the settings on sphix negate specifying `figsize` than your fix looks good to me.

demonstrations/tutorial_post-variational_quantum_neural_networks.py
doctorperceptron: It might! Let's find out...

doctorperceptron: Hmm small problem, because we use `matplotlib inline` the figure gets cropped by Jupyter and makes the actual size unpredictable. I think this is the best we can do without disabling matplotlib inline (currently set globally for all demos).

https://matplotlib.org/stable/gallery/subplots_axes_and_figures/figure_size_units.html#id1

DanielNino27: Ok - thanks for checking it out. In that case, the fix is fine with me.

DanielNino27 (approved):
lgtm

justinpickering (approved):
LGTM!
Just a note, this new figure layout introduces a 6th figure. Not sure if that's an issue or not given the context of this demo. cc @ikurecic

doctorperceptron:
> LGTM! Just a note, this new figure layout introduces a 6th figure. Not sure if that's an issue or not given the context of this demo. cc @ikurecic

These figures are a representative sample of the data. Afaik the initial choice of 5 was arbitrary, so changing it to 6 shouldn't material affect the demo.

PR #1251, opened 2024-10-30
Update QAOA-Maxcut demo with a few good practices

dwierichs:
**Title:**

**Summary:**

Just updating the MaxCut demo with a few tiny changes that incorporate good practices in Python/PennyLane, which have changed since the creation back in 2019.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

----

* AUDIENCE — Who is this for?

- [ ] Tutorial
- [ ] Demo
- [ ] How-to


demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -3,16 +3,6 @@
...
-
-.. meta::
-    :property="og:description": Implementing the quantum approximate optimization algorithm using PennyLane to solve the MaxCut problem.
-    :property="og:image": https://pennylane.ai/qml/_static/demonstration_assets/qaoa_maxcut_partition.png
-
-.. related::
-   tutorial_qaoa_intro Intro to QAOA
-
-*Author: Angus Lowe — Posted: 11 October 2019. Last updated: 13 April 2021.*
-
```
mudit2812, line 15: @ikurecic could you confirm that this should be removed? I'm not particularly familiar with how `qml` metadata is stored. I know about the JSON files, but I'm not aware if any other metadata is used or not.

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -258,27 +242,26 @@ def objective(params):
...
+    sampled_ints = [bitstring_to_int(string) for string in bitstrings]
 
     # print optimal parameters and most frequently sampled bitstring
-    counts = np.bincount(np.array(bit_strings))
+    counts = np.bincount(np.array(sampled_ints))
     most_freq_bit_string = np.argmax(counts)
-    print("Optimized (gamma, beta) vectors:\n{}".format(params[:, :n_layers]))
-    print("Most frequently sampled bit string is: {:04b}".format(most_freq_bit_string))
+    print(f"Optimized (gamma, beta) vectors:\n{params[:, :n_layers]}")
+    print(f"Most frequently sampled bit string is: {most_freq_bit_string:04b}")
```
mudit2812, line 256: You can tell how old this demo is based on the fact that f-strings are not used anywhere (until now) 😆

demonstrations/tutorial_qaoa_maxcut.py
```diff
@@ -168,20 +158,17 @@ def U_B(beta):
...
 def U_C(gamma):
     for edge in graph:
-        wire1 = edge[0]
-        wire2 = edge[1]
-        qml.CNOT(wires=[wire1, wire2])
-        qml.RZ(gamma, wires=wire2)
-        qml.CNOT(wires=[wire1, wire2])
+        qml.CNOT(wires=edge)
+        qml.RZ(gamma, wires=edge[1])
+        qml.CNOT(wires=edge)
```
mudit2812, line 163: Isn't this just an `IsingZZ`?

mudit2812 (changes requested):
The graphs with the distribution for the bit-strings looks great! However, the actual numerical values in the demo output are dramatically different:

Published version:
```
p=1
Objective after step     5:  2.0000000
Objective after step    10:  2.0000000
Objective after step    15:  4.0000000
Objective after step    20:  3.0000000
Objective after step    25:  1.0000000
Objective after step    30:  2.0000000
Optimized (gamma, beta) vectors:
[[-0.82383311]
 [-1.13584295]]
Most frequently sampled bit string is: 0101

p=2
Objective after step     5:  4.0000000
Objective after step    10:  3.0000000
Objective after step    15:  4.0000000
Objective after step    20:  3.0000000
Objective after step    25:  4.0000000
Objective after step    30:  3.0000000
Optimized (gamma, beta) vectors:
[[ 0.87696449  1.31508883]
 [ 0.90138291 -0.34955715]]
Most frequently sampled bit string is: 0101
```

After this PR:
```
p=1
Objective after step     5:  0.5000000
Objective after step    10:  0.5000000
Objective after step    15:  0.5000000
Objective after step    20:  2.5000000
Objective after step    25:  0.5000000
Objective after step    30:  2.5000000
Optimized (gamma, beta) vectors:
[[-0.90759251]
 [-1.2352846 ]]
Most frequently sampled bit string is: 1010

p=2
Objective after step     5:  0.5000000
Objective after step    10:  2.5000000
Objective after step    15:  2.5000000
Objective after step    20:  2.5000000
Objective after step    25:  2.5000000
Objective after step    30:  2.5000000
Optimized (gamma, beta) vectors:
[[ 1.54174446 -0.7153231 ]
 [ 0.35679571 -0.72652338]]
Most frequently sampled bit string is: 0101
```

I think this is happening because the device only has one shot. Before, that would give 4 samples due to 4 executions, but now there's only one sample. It's surprising to see that the distribution still looks as good as it does 😆. I'm guessing changing [this line](https://github.com/PennyLaneAI/qml/blob/fa47d365020a5e83a91ef202021d532e41c43200/demonstrations/tutorial_qaoa_maxcut.py#L179) to the following will fix it:
```python
dev = qml.device("lightning.qubit", wires=n_wires, shots=len(graph))
```

demonstrations/tutorial_qaoa_maxcut.py
dwierichs: Yes, up to a global phase. I was considering to change it but the text explicitly describes this and has a circuit diagram, so I thought it could be fine without change.

But I'm all for reduction if we don't consider that too invasive

dwierichs:
Thanks for the analysis @mudit2812 , that's a good catch!
- Overall, this demo is stochastic and seeds don't seem to be fixed. So with single-shot expectation values, even the original version varies _wildly_ (The optimized parameters, for example, look almost unrelated between runs :sweat_smile: )
- The variance for the previous approach is $\sum_{(a,b)\in\mathcal{E}}1-\langleZ_aZ_b\rangle^2$ where $\mathcal{E}$ is the set of edges. For the new approach it is $|\mathcal{E}|+\sum_{(a,b)\in\mathcal{E}_{\text{nn}}} \langleZ_aZ_b\rangle-\left(\sum_{(a,b)\in\mathcal{E}}\langleZ_aZ_b\rangle\right)^2$, where $\mathcal{E}_{\text{nn}}$ are pairs of next-nearest neighbours. So you are right that the variances differ a lot, and this actually makes me concerned about the "simplification" I'm proposing here. It is more of a switch in measurement strategy, and one that is highly non-trivial to see through (in particular, while allocating `len(graph)` shots makes sense coming from the current version, I'm not sure whether it can be motivated from the measurement problem itself? :thinking: ) 

Mhm.

mudit2812:
Thanks @dwierichs . It sounds like a non-trivial change. Maybe it makes sense to go over the demo body as well just to make sure that the text is not inconsistent with the new implementation. Otherwise, happy to approve.

demonstrations/tutorial_qaoa_maxcut.py
dwierichs: So should I change it ? :D

dwierichs: I think I won't because I'm not the original author, and Angus put effort into making that figure etc. I'll just add a comment in a `note` environment.

dwierichs:
@mudit2812 The variance discussion stands, but I actually noticed a small bug I encoded with the change:
I basically changed the objective to $C=\frac{1}{2}(1-\sum_{(j,k)} \langle Z_j Z_k\rangle)$ but the correct thing of course is
$C=\frac{1}{2}\sum_{(j,k)} (1-\langle Z_j Z_k\rangle) = \frac{1}{2}( |\mathcal{E}| - \sum_{(j,k)} \langle Z_j Z_k\rangle)$. I fixed that and everything works as expected as as before now :) In particular, perfect bitstrings are being sampled in the end, as anticipated by the main text.
I also went through the main text and adapted where required for this update.

demonstrations/tutorial_qaoa_maxcut.py
mudit2812: Sorry for not responding, I think it's good to leave it as is based on the context you gave. Thanks :)

mudit2812 (approved):
LGTM!

Qottmann (changes requested):
Author is not shown at the end and start of the demo 😬 

Other than that lgtm!

ikurecic:
> Didnt realize the preview behavior changed (or has always been like this 🫨 ?)
> 
> All good then!

Yep — since the author bios began to be pulled from the live website's profile database instead of hardcoded into the Demos, we currently cannot preview them here before merging to production. However, I know the cloud team will have an improvement planned for this in the future.

angusjlowe:
Looks good to me too! Thanks for implementing these much-needed updates.

PR #1258, opened 2024-11-06
Add author profiles

doctorperceptron:
Apologies for the large PR, but this touches a large number of demos.

This PR replaces legacy demo author information with their new pennylane Profiles. It also introduces a new version of the `demo.metadata.schema` which requires the `username` field (PennyLane Profile), and will not accept the legacy `id` field for demo authors. The expectation is that all demo authors create PennyLane Profiles to store their avatar and bio going forward.

*Notes:*
The new metadata schema file is exactly the same as the previous version (copy paste) except for the `authors` field:
```
      "authors": {
        "description": "The author(s) of the demo. This array must contain at least one item.",
        "type": "array",
        "items": {
          "$ref": "file:./objects/author.schema.0.2.0.json"
        },
        "minItems": 1
      },
```


ikurecic (approved):
Thanks, that looks good to me.

ashishks0522 (approved):

PR #1273, opened 2024-12-04
Update QSVT on hardware how-to to use poly_to_angles

soranjh:
**Summary:**
The PR updates QSVT on hardware how-to to use PennyLane `poly_to_angles` function following [sc-79728].


demonstrations/tutorial_qsvt_hardware.py
```diff
@@ -73,11 +84,12 @@ def convert_angles(angles):
     return angles + update_vals
 
 
-angles = convert_angles(ang_seq)
-print(angles)
+angles_pg = convert_angles(ang_seq)
+print(angles_pg)
```
KetpuntoG, line 88: What do you mean by "pg"?

KetpuntoG (changes requested):
Great to see the functionality in action!
Could you update the last modification date in the metadata?

demonstrations/tutorial_qsvt_hardware.py
soranjh, line 88:
```suggestion
angles_pq = convert_angles(ang_seq)
print(angles_pq)
```

soranjh: Corrected to pq for pyqsp.

soranjh: @KetpuntoG Why the angles_pl and angles_pq are not the same?

KetpuntoG: They follow different strategies. In fact, if you run pyqsp again it will probably give you a different set. This is because an optimization is being done and there may be more than one global minimum

demonstrations/tutorial_quantum_analytic_descent.metadata.json
```diff
@@ -9,7 +9,7 @@
         }
     ],
     "dateOfPublication": "2021-06-30T00:00:00+00:00",
-    "dateOfLastModification": "2024-11-06T00:00:00+00:00",
+    "dateOfLastModification": "2024-12-04T00:00:00+00:00",
```
KetpuntoG, line 12: It is not the correct file

soranjh: Corrected.

KetpuntoG (approved):
🚀

KetpuntoG:
I updated one cell to use the new `qml.transform_angles` functionality. Now it is perfect :)

demonstrations/tutorial_qsvt_hardware.py
```diff
@@ -53,31 +62,16 @@
...
-    update_vals[1:-1] = np.pi / 2
-    update_vals[-1] = -np.pi / 4
-
-    return angles + update_vals
-
-
-angles = convert_angles(ang_seq)
-print(angles)
+angles_pyqsp = qml.transform_angles(ang_seq, "QSP", "QSVT")
+print(angles_pyqsp)
```
Jaybsoni, line 70: These angles don't match the ones generated by PL even after we transformed them. Maybe you should add a comment explaining that even though the angles are different, they will both produce the same polynomial transformation.

Jaybsoni (approved):
Looks good, left one comment. Otherwise good to go from my side 👍🏼

PR #1279, opened 2024-12-11
Update QJIT VQE demo to use `value_and_grad` and Catalyst printing

josh146:
When this demo was originally written, these features weren't available. Adding these features to the demo makes it more efficient (we aren't computing the cost function twice), and also shows users how to print from inside qjit.

### DIFF
--- a/demonstrations/tutorial_how_to_quantum_just_in_time_compile_vqe_catalyst.metadata.json
+++ b/demonstrations/tutorial_how_to_quantum_just_in_time_compile_vqe_catalyst.metadata.json
@@ -9,7 +9,7 @@
         }
     ],
     "dateOfPublication": "2024-04-26T00:00:00+00:00",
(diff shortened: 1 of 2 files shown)


justinpickering:
>Instead, we can use `Optax <https://github.com/google-deepmind/optax>`__, a library designed for optimization using JAX, as well as the :func:`~.catalyst.grad` function, which allows us to differentiate through quantum just-in-time compiled workflows.

Should we point to the [value_and_grad](https://docs.pennylane.ai/projects/catalyst/en/latest/code/api/catalyst.value_and_grad.html#:~:text=value_and_grad,-value_and_grad(fn%3DNone&text=A%20qjit()%20%2Dcompatible%20transformation,computed%20within%20the%20compiled%20program.) function here instead?

josh146:
fixed, thanks @justinpickering!

justinpickering (approved):
LGTM @josh146 👍

PR #1289, opened 2025-01-06
Updates to quantum volume and how to use qiskit

albi3ro:
Gets the Quantum Volume and how to use qiskit demos working


lillian542 (commented):
I'm not sure why the Qiskit 1.0 demo is static, looks to me like it could be running automatically, but maybe I'm missing something and either way that's not really for now.

Looks like the `dateOfLastModification` needs to be updated for both of these in the metadata.

PietropaoloFrisoni (approved):
They both work for me :rocket:

astralcai (approved):

lillian542 (approved):

PR #1291, opened 2025-01-07
QA demo process for PL 0.40 during feature freeze

PietropaoloFrisoni:
As the title says.


JerryChen97:
@isaacdevlugt I made use of the original author's notes inside the qnspsa notebook and left a warning before the very last block. Given the fact that the primary point of this demo is tutorial of qnspsa, only one warning there to indicate the issue with codes uploading to braket should be already enough.

demonstrations/vqe_parallel.py
```diff
@@ -15,6 +15,9 @@
 
 *Author: Tom Bromley — Posted: 14 February 2020. Last updated: 29 August 2023.*
 
+.. warning::
+    This demo is only compatible with PennyLane v0.40 or below. To use Rigetti hardware with newer versions of PennyLane please use the `PennyLane-Braket plugin <https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/stable/index.html>`__ instead.
```
lillian542, line 19: This demo uses Rigetti **simulators**, and I feel like the text regarding optimization, program overhead vs execution time, etc, are very simulator based. Because of that, I think we should consider directing people to a different simulator instead of to Rigetti hardware; the reference to accessing Rigetti hardware feels out of place. 

Though I haven't looked at it sufficiently in detail to be completely confident that you could just replace it with DefaultQubit or Braket's BasicSimulator and have everything else be the same.

demonstrations/vqe_parallel.py
```diff
@@ -15,6 +15,9 @@
 
 *Author: Tom Bromley — Posted: 14 February 2020. Last updated: 29 August 2023.*
 
+.. warning::
+    This demo is only compatible with PennyLane v0.40 or below. To use Rigetti hardware with newer versions of PennyLane please use the `PennyLane-Braket plugin <https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/stable/index.html>`__ instead.
+
 This tutorial showcases how using asynchronously-evaluated parallel QPUs can speed up the
 calculation of the potential energy surface of molecular hydrogen (:math:`H_2`).
```
lillian542, line 22: Do we also want to include a warning on this one that it requires Python <= 3.10, or is that overkill for a demo we are basically deprecating?

demonstrations/pytorch_noise.py
```diff
@@ -15,6 +15,10 @@
 
 *Author: Josh Izaac — Posted: 11 October 2019. Last updated: 9 November 2022.*
 
+.. warning::
+    This demo is only compatible with PennyLane v0.40 or below. To use Rigetti hardware with newer versions of PennyLane please use the `PennyLane-Braket plugin <https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/stable/index.html>`__ instead.
+
```
lillian542, line 20: It seems like this one also can't exist functionally in an environment with the latest version of Braket, because of dependency conflicts (Rigetti is pinned to very old versions of a couple of things). Is that something we want to address in the warning as well? Do we want to include anything about the Python version restriction?

lillian542 (commented):
We also need similar warnings about Rigetti in [this demo](https://pennylane.ai/qml/demos/ensemble_multi_qpu)

demonstrations/pytorch_noise.py
isaacdevlugt: Oh good point 🤔... what's the upper bound on the braket version for which this would still work? If that's a harder dependancy then I think that that should be stated over the PL restriction.

demonstrations/vqe_parallel.py
isaacdevlugt:
```suggestion
    This demo uses the PennyLane-Rigetti plugin, which is only compatible with PennyLane v0.40 or below. To run this demo with newer versions of PennyLane, you will need to use `a different simulator device <https://pennylane.ai/plugins>`__.
```

How about this?

isaacdevlugt: We could also do this (based on your comment here https://github.com/PennyLaneAI/qml/pull/1291/files#r1907628473)

```suggestion
    This demo requires Python <=3.10 and uses the PennyLane-Rigetti plugin, which is only compatible with PennyLane v0.40 or below. To run this demo with newer versions of PennyLane, you will need to use `a different simulator device <https://pennylane.ai/plugins>`__.
```

How about this?

```

File: demonstrations/vqe_parallel.py (hunk 2 above)
Comment (isaacdevlugt) [reply demonstrations/vqe_parallel.py:22]: See comment above

File: demonstrations/qnspsa.py [hunk 4]
```diff
@@ -913,11 +913,14 @@ def __apply_blocking(self, cost, params_curr, params_next):
...
-# with its dependencies in the `source_scripts`
-# `folder <https://github.com/aws/amazon-braket-examples/blob/main/examples/hybrid_jobs/6_QNSPSA_optimizer_with_embedded_simulator/source_scripts/>`__.
+# QN-SPSA in Fig. 1(b) of reference [#Gacon2021]_ with the following job. 
+#
+# .. warning:: 
+#   The following code is incomplete and might require additional attention. 
+#   You can find a more detailed version of the example in this
+#   `notebook <https://github.com/aws/amazon-braket-examples/blob/main/examples/hybrid_jobs/6_QNSPSA_optimizer_with_embedded_simulator/qnspsa_with_embedded_simulator.ipynb>`__,
+#   with its dependencies in the `source_scripts`
+#   `folder <https://github.com/aws/amazon-braket-examples/blob/main/examples/hybrid_jobs/6_QNSPSA_optimizer_with_embedded_simulator/source_scripts/>`__.
```
isaacdevlugt, line 923: @JerryChen97 thanks! Do we know why this outputs a failure message?

demonstrations/vqe_parallel.py
lillian542: I think the second one fits in all the information quite concisely :)

demonstrations/pytorch_noise.py
lillian542: I'm not actually sure yet - I just know it used to work and doesn't now, because of dependency version incompatibilities. Once I'm done with the QA for the PRs, I'll play around with this and see if I can get it working and let you know!

demonstrations/qnspsa.py
JerryChen97: @isaacdevlugt From what I've explored, the [script](https://github.com/amazon-braket/amazon-braket-examples/blob/main/examples/hybrid_jobs/6_QNSPSA_optimizer_with_embedded_simulator/source_scripts/benchmark_ref_paper_converge_speed.py) given by AWS is incomplete and shall fail with an Exception saying 'optimizer'. It should be some hyperparameters missing there, nothing fundamentally wrong but it's still far from getting the final demonstration figure in the end.

JerryChen97: I think it will be better if we have an issue to record this case. Completely fixing this demo might take some effort.

demonstrations/ensemble_multi_qpu.py
```diff
@@ -18,10 +18,10 @@
...
 
 .. warning::
-    This demo does not work with the latest version of Qiskit or the Pennylane-Qiskit plugin.
-    It is compatible with ``qiskit==0.46`` and ``pennylane-qiskit==0.35.1``. Older versions of
-    Qiskit and the Pennylane-Qiskit plugin should not be installed in environments with an
-    existing installation of Qiskit 1.0 or above.
+    This demo does not work with the latest version of Qiskit or the Pennylane-Qiskit plugin and
+    is compatible with Python versions 3.10 and lower. It is compatible with ``qiskit==0.46`` and 
+    ``pennylane-qiskit==0.35.1``. Older versions of Qiskit and the Pennylane-Qiskit plugin should 
+    not be installed in environments with an existing installation of Qiskit 1.0 or above.
```
isaacdevlugt, line 24: cc @lillian542

JerryChen97:
@isaacdevlugt Also, the oqc pulse demo is down for device retired as well. Do we plan to add warning similar here or have different treatment for it?

demonstrations/oqc_pulse.py
```diff
@@ -11,6 +11,10 @@
 
 *Author: Korbinian Kottmann — Posted: 30 October 2023.*
 
+.. warning::
+    The OQC Lucy device has been retired on Amazon Braket and is no longer accessible. As there is no alternative 
+    at this time, this demo should be used for educational purposes only.
```
isaacdevlugt, line 16: cc @CatalinaAlbornoz @lillian542

demonstrations/qnspsa.py
JerryChen97: Adjusted the sentences in a recent commit https://github.com/PennyLaneAI/qml/pull/1291/commits/4f8cf83d69e942719b35daf2ebaf09de64cc910b according to @Alex-Preciado suggestion.

demonstrations/oqc_pulse.py
lillian542: Seems clear to me!

demonstrations/pytorch_noise.py
lillian542: Turns out the problem I was seeing was actually an incompatibility between Rigetti and JupyterLabs, not Rigetti and Braket (I usually test the demos in a Jupyter notebook).

If you use most recent Braket and Rigetti and PennyLane, then pip complains about dependencies, but the demo runs fine (just not in a Jupyter notebook). So we can probably just put the basic "Rigetti won't work after PL 0.40 and it needs Python 3.10" warning, and _possibly_ also mention the last tested braket version (1.31.0), since it doesn't follow the PennyLane version like most of the plugins.

demonstrations/pytorch_noise.py
```diff
@@ -15,6 +15,10 @@
 
 *Author: Josh Izaac — Posted: 11 October 2019. Last updated: 9 November 2022.*
 
+.. warning::
+    This demo is only compatible with PennyLane v0.40 or below. To use Rigetti hardware with newer versions of PennyLane please use the `PennyLane-Braket plugin <https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/stable/index.html>`__ instead.
```
isaacdevlugt, line 19:
```suggestion
    This demo is only compatible with PennyLane v0.40 or below and Braket v1.31.0. To use Rigetti hardware with newer versions of PennyLane please use the `PennyLane-Braket plugin <https://amazon-braket-pennylane-plugin-python.readthedocs.io/en/stable/index.html>`__ instead.
```

isaacdevlugt: cc @lillian542

demonstrations/pytorch_noise.py
isaacdevlugt: See suggestion above 🙏

demonstrations/qnspsa.py
isaacdevlugt: Thanks @JerryChen97!

PietropaoloFrisoni:
Thanks @isaacdevlugt @lillian542 @JerryChen97!

Is there any other change that you want to push in this PR?

lillian542 (approved):
LGTM! Thanks for your patience while we sorted out the version info 😅

demonstrations/oqc_pulse.py
Alex-Preciado:
```suggestion
    The OQC Lucy device is no longer available on Amazon Braket. As there is no alternative at this time, this
    demo is now intended for educational purposes only.
```

isaacdevlugt (approved):
Thanks everyone! 🙌

JerryChen97 (approved):

PR #1298, opened 2025-01-13
[QA v0.40.0] Fix liesim_extension demo

Qottmann:
The order of the operators in the lie closure changed with https://github.com/PennyLaneAI/pennylane/pull/6724 such that the hard-coded indices are not correct anymore.

I changed the latter part where we pick the gate in the moment space according to the $p = O_1 O_2$ that is not in the moment space to automatically pick it from the list (instead of hard-coding the index). However, the first choice of $p$ is hard to automate, the only thing that matters is that $p = O_1 O_2$ is not in the DLA

Note for reviewers:
The important thing to look at is that the outputs are now again `True`

### DIFF
--- a/demonstrations/tutorial_liesim_extension.py
+++ b/demonstrations/tutorial_liesim_extension.py
@@ -133,7 +133,7 @@ U = expm(theta * gate)
 #
 # For that, let us construct a concrete example. First we pick two elements from :math:`\mathfrak{g}` such that their product is not in :math:`\mathfrak{g}.`
 
-p = dla[-5] @ dla[-2]
+p = dla[-5] @ dla[-1]
 p = next(iter(p)) # strip any scalar coefficients
 dla_vspace = qml.pauli.PauliVSpace(dla, dtype=complex)
 dla_vspace.is_independent(p.pauli_rep)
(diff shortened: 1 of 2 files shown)


dwierichs (approved):
LGTM :+1:

demonstrations/tutorial_liesim_extension.py
```diff
@@ -414,9 +414,11 @@ def Moment_step(ops, dla):
 #
 # We can now choose arbitrary DLA gates and a maximum of `one` :math:`P` gate to evolve the expectation value vector.
 
+P_index = Moment[comp_moment].index(1.*p) # pick the gate in the Moment space that p corrresponds to
```
PietropaoloFrisoni, line 417: Sorry for noticing just now, but there's a small typo here (3 `r`s).

I'll leave it up to you @Qottmann if you want to open another PR to address this

Qottmann: 🙈 oh no, I'll add a quick fix, thanks for noticing

Qottmann: fixed in https://github.com/PennyLaneAI/qml/pull/1299 ✅

PR #1317, opened 2025-02-14
Fix `tutorial_pulse_programming101.py` to type cast `expval`

andrijapau:
**Context:**

In https://github.com/PennyLaneAI/pennylane/pull/6939, a fix was made to `qml.expval` so that it no longer silently converts the result to a real number. However, in `tutorial_pulse_programming101.py` we use a Hamiltonian with complex-typed coefficients. This resulted in an imaginary expectation value (`X+0j`) and broke the `jax.value_and_grad` function.

**Description of change:**

Wrap the `qnode` function to get the expectation value and convert it to a real number. This allows `jax.value_and_grad` to be used and fixes the demo.

### DIFF
--- a/demonstrations/tutorial_pulse_programming101.py
+++ b/demonstrations/tutorial_pulse_programming101.py
@@ -327,7 +327,8 @@ dev = qml.device("default.qubit", wires=range(n_wires))
 
 
 def qnode(theta, t=duration):
-    @qml.qnode(dev, interface="jax")
+
+    @qml.qnode(dev)
     def _qnode_inner(theta, t=duration):
         qml.BasisState(jnp.array(data.tapered_hf_state), wires=H_obj.wires)
         qml.evolve(H_pulse)(params=(*theta, *theta), t=t)
--- a/demonstrations/tutorial_pulse_programming101.py
+++ b/demonstrations/tutorial_pulse_programming101.py
@@ -322,9 +322,10 @@ H_pulse = H_D + H_C
 
 ##############################################################################
 # Now we define the ``qnode`` that computes the expectation value of the molecular Hamiltonian.
+# We need to wrap the ``qnode`` in a function so that we can convert the expectation value to a real number.
+# This will enable use to make use of gradient descent methods that require real-valued loss functions.
 
 dev = qml.device("default.qubit", wires=range(n_wires))
-
 
 def qnode(theta, t=duration):
 
@@ -335,8 +336,7 @@ def qnode(theta, t=duration):
         return qml.expval(H_obj)
 
     expectation_value = _qnode_inner(theta, t)  # Execute the qnode
-    return jnp.real(expectation_value)  # Extract real part
-
+    return jnp.real(expectation_value)  # Typecast to real number
 
 value_and_grad = jax.jit(jax.value_and_grad(qnode))
 
--- a/demonstrations/tutorial_pulse_programming101.py
+++ b/demonstrations/tutorial_pulse_programming101.py
@@ -325,11 +325,16 @@ H_pulse = H_D + H_C
 
 dev = qml.device("default.qubit", wires=range(n_wires))
 
-@qml.qnode(dev, interface="jax")
+
 def qnode(theta, t=duration):
-    qml.BasisState(jnp.array(data.tapered_hf_state), wires=H_obj.wires)
-    qml.evolve(H_pulse)(params=(*theta, *theta), t=t)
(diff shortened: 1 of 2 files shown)


demonstrations/tutorial_pulse_programming101.py
```diff
@@ -325,11 +325,16 @@ def wrapped(p, t):
 
 dev = qml.device("default.qubit", wires=range(n_wires))
 
-@qml.qnode(dev, interface="jax")
+
 def qnode(theta, t=duration):
-    qml.BasisState(jnp.array(data.tapered_hf_state), wires=H_obj.wires)
-    qml.evolve(H_pulse)(params=(*theta, *theta), t=t)
-    return qml.expval(H_obj)
+    @qml.qnode(dev, interface="jax")
```
isaacdevlugt, line 330: Do we need `interface="jax"`? 🤔

demonstrations/tutorial_pulse_programming101.py
```diff
@@ -325,11 +325,16 @@ def wrapped(p, t):
...
+    def _qnode_inner(theta, t=duration):
+        qml.BasisState(jnp.array(data.tapered_hf_state), wires=H_obj.wires)
+        qml.evolve(H_pulse)(params=(*theta, *theta), t=t)
+        return qml.expval(H_obj)
+
+    expectation_value = _qnode_inner(theta, t)  # Execute the qnode
+    return jnp.real(expectation_value)  # Extract real part
```
isaacdevlugt, line 337: Does this work?

```pythyon
@qml.qnode(dev)
def qnode(theta, t=duration):
    ...

value_and_grad = jax.jit(jax.value_and_grad(jnp.real(qnode)))
```

isaacdevlugt: Just might be nicer to not have to formally wrap the QNode like what you're doing

andrijapau: Let me give it a try! 🤞🏼

demonstrations/tutorial_pulse_programming101.py
andrijapau: Good question, I don't think so since by default we have `"auto"`. Going to try it out.

demonstrations/tutorial_pulse_programming101.py
andrijapau: Unfortunately not ... 😿 
```
TypeError: Cannot interpret value of type <class 'pennylane.workflow.qnode.QNode'> as an abstract array; it does not have a dtype attribute
```

demonstrations/tutorial_pulse_programming101.py
andrijapau: Works! Addressed in https://github.com/PennyLaneAI/qml/pull/1317/commits/8ce2231f2bf7b2a787ab2bfdf0d71423a58724a9 😄

demonstrations/tutorial_pulse_programming101.py
isaacdevlugt: 🍅 booo 🍅 

Okay then 😁

demonstrations/tutorial_pulse_programming101.py
```diff
@@ -325,11 +325,17 @@ def wrapped(p, t):
 
 dev = qml.device("default.qubit", wires=range(n_wires))
 
-@qml.qnode(dev, interface="jax")
+
 def qnode(theta, t=duration):
-    qml.BasisState(jnp.array(data.tapered_hf_state), wires=H_obj.wires)
-    qml.evolve(H_pulse)(params=(*theta, *theta), t=t)
-    return qml.expval(H_obj)
```
PietropaoloFrisoni, line 332: If it is necessary to wrap the `qnode`in this way just to extract the real part of the result, would it make sense to specify it explicitly with a comment above the circuit in the demo?

PietropaoloFrisoni (approved):
Thanks a lot @andrijapau!

JerryChen97 (approved):
lgtm

demonstrations/tutorial_pulse_programming101.py
andrijapau: Sorry, what do you mean? There's no straight forward way to type cast the output of the qfunc (`qml.expval(...)`) as it returns a measurement operator to then be processed by the device.

andrijapau: Addressed in latest commit.

PR #1336, opened 2025-04-01
Updated title on ExpVal on Paramterized Circuits article.

cognigami:
Updated title.

### DIFF
--- a/demonstrations/tutorial_classical_expval_estimation.metadata.json
+++ b/demonstrations/tutorial_classical_expval_estimation.metadata.json
@@ -6,7 +6,7 @@
         }
     ],
     "dateOfPublication": "2024-09-10T00:00:00+00:00",
-    "dateOfLastModification": "2024-10-07T00:00:00+00:00",
+    "dateOfLastModification": "2025-04-01T00:00:00+00:00",
     "categories": [
         "Quantum Computing"
     ],
--- a/demonstrations/tutorial_classical_expval_estimation.metadata.json
+++ b/demonstrations/tutorial_classical_expval_estimation.metadata.json
@@ -6,7 +6,7 @@
         }
     ],
     "dateOfPublication": "2024-09-10T00:00:00+00:00",
-    "dateOfLastModification": "2025-04-01T00:00:00+00:00",
+    "dateOfLastModification": "2025-04-03T00:00:00+00:00",
     "categories": [
         "Quantum Computing"
     ],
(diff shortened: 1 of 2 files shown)


demonstrations/tutorial_classical_expval_estimation.metadata.json
```diff
@@ -1,12 +1,12 @@
 {
-    "title": "Classically estimating expectation values from parametrized quantum circuits",
+    "title": "Pauli Propogation: Classically estimating expectation values from parametrized quantum circuits",
```
dwierichs, line 2:
```suggestion
    "title": "Pauli Propagation: Classically estimating expectation values from parametrized quantum circuits",
```

demonstrations/tutorial_classical_expval_estimation.py
```diff
@@ -1,6 +1,6 @@
 r"""
-Classically estimating expectation values from parametrized quantum circuits
-============================================================================
+Pauli Propagation: Classically estimating expectation values from parametrized quantum circuits
+============================================================================================
```
dwierichs, line 3: Two things about this:
1. sorry, my suggestion still seems to have had the wrong length (the line of equality signs is supposed to have the same length as the title length, as far as I know).
2. it doesn't seem to matter and renders fine, as far as I can tell :shrug: :laughing:

demonstrations/tutorial_classical_expval_estimation.metadata.json
```diff
@@ -1,12 +1,12 @@
...
-    "title": "Classically estimating expectation values from parametrized quantum circuits",
+    "title": "Pauli Propagation: Classically estimating expectation values from parametrized quantum circuits",
     "authors": [
         {
             "username": "dwierichs"
         }
     ],
     "dateOfPublication": "2024-09-10T00:00:00+00:00",
-    "dateOfLastModification": "2024-10-07T00:00:00+00:00",
+    "dateOfLastModification": "2025-04-01T00:00:00+00:00",
```
dwierichs, line 9: Will have to be updated to date of merging.

dwierichs (approved):
LGTM!

demonstrations/tutorial_classical_expval_estimation.py
josh146:
```suggestion
==============================================================================================
```

that makes it match :) Note that while it will render fine, it will generate warnings in our build system

josh146 (approved):

PR #1348, opened 2025-04-10
Trigger CI

JerryChen97:
This PR is only for triggering the build process of catalyst related demos and shall be closed after we collected the runtime today.


JerryChen97:
Closed since catalyst runtime data collected

PR #1351, opened 2025-04-14
Merge `dev` into `master` (Release PL 0.41)

PietropaoloFrisoni:
Merging the `dev` branch into the `master` branch for the release of PL 0.41


JerryChen97:
Just a note that the new figure is caused by the change of keras demo from `tutorial_` to non tutorial (not automatically run by CI but only build static web) so it is necessarily required.

PietropaoloFrisoni:
> Just a note that the new figure is caused by the change of keras demo from tutorial_ to non tutorial (not automatically run by CI but only build static web) so it is necessarily required.

@JerryChen97 Sorry I am not sure I am following. Are you referring to the `sphx_glr_qnn_module_tf_001.png` figure?

JerryChen97:
> > Just a note that the new figure is caused by the change of keras demo from tutorial_ to non tutorial (not automatically run by CI but only build static web) so it is necessarily required.
> 
> @JerryChen97 Sorry I am not sure I am following. Are you referring to the `sphx_glr_qnn_module_tf_001.png` figure?

Yes. I was just clarifying in case any future reviewers feel suspicious about it.

PR #1356, opened 2025-04-24
SC-89340 | Add ability to deploy demos manually

Alan-eMartin:
**Summary:**
Allow internal users to build and deploy **all** or a select list of demos.

**Related GitHub Issues:**
Internal


.github/workflows/v2-build-and-deploy
```diff
@@ -0,0 +1,53 @@
...
+      branch:
+        default: master
+        description: Branch to build demos against `master` (stable) or `dev` (latest)
+        options:
+          - master
+          - dev
+        required: true
+        type: choice
+      demos:
+        description: Demos to build and deploy, if all demos are to be built leave blank
```
doctorperceptron, line 23: What does this string look like if it's more than one demo? A comma separated list?

.github/workflows/v2-build-and-deploy
```diff
@@ -0,0 +1,53 @@
...
+        required: false
+        type: string
+      as-previews:
+        default: false
+        description: Whether to deploy demos as preview
+        required: true
```
doctorperceptron, line 29: Usually, if a step has a default then it isn't required (same for branch above).
```suggestion
        required: false
```

.github/workflows/v2-build-and-deploy
```diff
@@ -0,0 +1,53 @@
...
+        type: choice
+      branch:
+        default: master
+        description: Branch to build demos against `master` (stable) or `dev` (latest)
+        options:
+          - master
+          - dev
+        required: true
+        type: choice
```
doctorperceptron, line 21: Per @rashidnhm's comment in the ADR, it may be clearer to refer to this as target rather than branch:
```suggestion
      target:
        default: stable
        description: PennyLane version to build the demos. Either 'latest' or the most recent 'stable' release.
        options:
          - stable
          - latest
        required: false
        type: choice
```

doctorperceptron (approved):
Functionally it looks good. Feel free to ignore my suggestions as we can always tweak this later.

rashidnhm (approved):

.github/workflows/v2-build-and-deploy
Alan-eMartin: This will be a _space-separated_ list e.g `"demo1 demo2 demo3"` I will update this now

.github/workflows/v2-build-and-deploy
Alan-eMartin: Sounds good, thanks paul!

.github/workflows/v2-build-and-deploy
Alan-eMartin: Will do, I am adding some input parsing steps now! I will tag you when they are ready for review

.github/workflows/v2-build-and-deploy.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Set branch
+        id: set-branch
+        run: |
+          if [[ "${{ github.event.inputs.target }}" == "stable" ]]; then
+            echo "branch=master" >> $GITHUB_OUTPUT
+          elif [[ "${{ github.event.inputs.target }}" == "latest" ]]; then
+            echo "branch=dev" >> $GITHUB_OUTPUT
+          else 
+            echo "branch=" >> $GITHUB_OUTPUT
+          fi
```
Alan-eMartin, line 51: I have added the following step to process the new clarified target input.

@doctorperceptron — please let me know if this works for you.

.github/workflows/v2-build-and-deploy.yml
```diff
@@ -0,0 +1,97 @@
...
+            "${{ github.event.inputs.as-previews }}" == "false" 
+          ]]; then
+            echo "========================="
+            echo "🚫 Invalid input detected:"
+            echo "Demos built with the latest version cannot be published to 'swc-staging' or 'swc-prod'."
+            echo "They can only be deployed as previews."
+            echo "Please set the 'as-previews' input to 'true' in your workflow configuration."
+            echo "========================="
+            exit 1
+          fi
```
Alan-eMartin, line 69: @rashidnhm and @doctorperceptron  — After further self-review, I discovered that I should probably add a test that prevents a user from deploying "published" (AKA non-preview) demos to the `swc-prod` and `swc-staging` when using the `latest` requirements and only allow them to be published to dev for testing. 

Let me know if this makes sense?

.github/workflows/v2-build-and-deploy.yml
```diff
@@ -0,0 +1,97 @@
...
+          - swc-dev
+        required: true
+        type: choice
+      target:
+        default: stable
+        description: PennyLane version to build the demos. Either 'latest' or the most recent 'stable' release.
+        options:
+          - latest
+          - stable
+        required: true
```
doctorperceptron, line 20: This shouldn't be required if there's a default value.
```suggestion
        required: false
```

.github/workflows/v2-build-and-deploy.yml
doctorperceptron: Yup this makes sense to me 👍 . You've now got me wondering as well: What is the use-case to deploy non-preview demos to production? Since this runs on workflow_dispatch and is in a public repo, in principle anyone could deploy a set of demos straight to prod? Is that ok?

Alan-eMartin: I don't believe people not part of PennyLane can do this, on internally people are allowed to dispatch this workflow. 

One use case I can think of is — we have made internal changes on how we process Sphinx-generated content and it needs a refresh on the server.

Alan-eMartin: Only previews from dev would be allowed to prod, and master would already be in prod. With the limited options for targets, we prevent anything that isn't already approved as well.

PR #1357, opened 2025-04-29
Minor corrections to new period finding demo

mariaschuld:
Fixes some minor issues raised by readers.


mariaschuld:
@alvaro-at-xanadu do we update the modification date if it's just a few typos?

dwierichs (approved):
Nice, thanks @mariaschuld :pray: :100: 
Now I actually understood why there are seven qubits, I got it wrong before! 
Up to the question about modification date, it looks good to me :+1:

alvaro-at-xanadu:
> @alvaro-at-xanadu do we update the modification date if it's just a few typos?

Hi @mariaschuld Yes, this is the best practice even if corrections are minor

PR #1363, opened 2025-05-06
FIX: unused import in `tutorial_annni` causing v2 build failure, and not used in demo

Alan-eMartin:
**Title:**
Delete unused import from `tutorial_annni`

**Summary:**
There was an unused import causing the v2 pipeline failure. The new pipeline will require that modules be installed, since this package is not required to run this demo, it will be deleted.

**Related GitHub Issues:**
_N/A_


doctorperceptron (approved):
# 🚀

mudit2812 (approved):

PR #1366, opened 2025-05-07
Bump poetry version in CI to v2.1.3

rashidnhm:
**Title:**
Bump poetry version and update lockfile

**Summary:**
Bumps the poetry version to the current latest one available.

The current CI was using an outdated version of poetry, making it more difficult to update the lockfile.

**Relevant references:**

**Possible Drawbacks:**
Contributors that want to add dependencies would need poetry v2 installed.

**Related GitHub Issues:**


Alan-eMartin (approved):
✅ 📖

mudit2812 (commented):
Looks good. Do we also need to update [this file](https://github.com/PennyLaneAI/qml/blob/a5081f225277a96b33138611898a49a5e3938acf/pyproject.toml#L30)?

rashidnhm:
> Looks good. Do we also need to update [pyproject.toml](https://github.com/PennyLaneAI/qml/blob/a5081f225277a96b33138611898a49a5e3938acf/pyproject.toml#L30)?

We should be updating pyproject to use the `[project]` header instead of `[tool.poetry]`. But I opted to do that as a separate PR as this is mainly to unblock Alan and v2 deployment work.

For now, the poetry will raise warning on the pyproject but it does not fail or functionally affect anything

mudit2812 (approved):
Thanks, approving.

PR #1369, opened 2025-05-12
Fix Jax bump upgrading numpy as well

rashidnhm:
**Title:**
Temp workaround for pinning numpy outside of poetry

**Summary:**
Currently, catalyst is hard pinning JAX 0.4.28. But upgrading it also upgrades numpy to 2.x which is breaking matplotlib.

**Relevant references:**
[sc-91141]

[Previous Failure](https://github.com/PennyLaneAI/qml/actions/runs/14910534756/job/41883635938?pr=1355)

**Possible Drawbacks:**
We need to roll this back prior to next release.

**Related GitHub Issues:**
None.


doctorperceptron:
FYI @rashidnhm I'm 99% sure we don't need this anymore, particularly as we are now switching to the V2 pipeline, so I'm closing this PR.

PR #1375, opened 2025-05-21
More v2 fixes

doctorperceptron:
More fixes for the V2 pipeline:
- Reverts the addition from #1374 . This was erroneously identified as the cause for the failures.
- Adds `workflow-run-id` and `github-token` to the `actions/download-artifact@v4` Action as these are needed to pull artifacts from external workflow runs (true cause of the failures).
- Separates the downloading artifact step into two: One for PRs (separate workflow) and one for same-workflow deployments.
- Adds a few more early checks to the workflows if:
  - There are no changed demos.
  - There are no artifacts to download.

### DIFF
--- a/.github/workflows/v2-build-pr.yml
+++ b/.github/workflows/v2-build-pr.yml
@@ -32,12 +32,6 @@ jobs:
           echo "Updated Demos: ${{ steps.get-changed-demos.outputs.updated }}"
           echo "Deleted Demos: ${{ steps.get-changed-demos.outputs.deleted }}"
 
-      - name: Exit if no changes
-        if: steps.get-changed-demos.outputs.updated == ''
-        run: |
-          echo "No changes found in demos. Exiting workflow."
-          exit 1
-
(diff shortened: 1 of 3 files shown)


.github/workflows/v2-build-pr.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         if: steps.get-changed-demos.outputs.updated == ''
         run: |
           echo "No changes found in demos. Exiting workflow."
+          exit 1
```
rashidnhm, line 39: Shouldn't we exit early vs exit with error?

As in, if we have PRs that don't update demos (which happen frequently enough), and we mark v2 as required workflow, wouldn't those PRs get blocked?

doctorperceptron: Yup, reverted this 👍

rashidnhm (approved):

ashishks0522 (approved):
Looks good!

PR #1376, opened 2025-05-22
Fix v2 actions

doctorperceptron:
The download artifact Action needs a token to pull an artifact from a separate workflow. Passing the default token doesn't seem to work, so trying to use it directly as it should be available for all workflows.


rashidnhm (approved):
Wait so github.token wasn't working before? It should work cause that's what v1 is using to download artifact from another workflow

ashishks0522 (approved):

PR #1379, opened 2025-05-23
Require at least one `category` for demos

doctorperceptron:
Currently demos are technically allowed to define an empty `categories` array in their metadata. In practice every demo defines at least one category, except for the `adjoint_diff_benchmarking` demo. This was a hack to avoid surfacing it in search because it is supplementary material. 

The new V2 pipeline requires demos to have a category, and we have better ways to omit this demo from search, so this PR adds a category (same as the parent demo) to this demo and updates the metadata schema (the relevant change [is here](https://github.com/PennyLaneAI/qml/pull/1379/files#diff-e8a88baa22c87205641827678200ebd8b54e8f640baad11e2b828783aef0f2d4R56).) and Action to require a category for demos. It also fixes issues (trailing .html) with the links in these demos.

### DIFF
--- a/.github/workflows/validate-demo-metadata.yml
+++ b/.github/workflows/validate-demo-metadata.yml
@@ -101,7 +101,7 @@ jobs:
           METADATA_FILE_LIST: ${{ needs.generate-metadata-file-list.outputs.metadata_files }}
         run: |
           cd metadata_schemas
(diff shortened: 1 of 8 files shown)


jzaia18 (approved):
Thanks, looks good. This fixes an issue I stumbled upon in the docs in the wild :sweat_smile:

PR #1383, opened 2025-05-27
iqpopt demo fix

josephbowles:
bug fix for the iqp circuit optimization demo. We updated the external package (IQPopt) and this caused an issue because we didn't specify a kwarg in one of the functions. Should be back now.


PietropaoloFrisoni (commented):
Thanks for this! 

- Except for the fact that right now `dev` is failing, is there a reason why this PR is targeting `master`?

- Should we also update the corresponding `metadata.json` file and modify the `dateOfLastModification` field?

josephbowles:
Have updated the json. And indeed yes should have been targeting dev.

PietropaoloFrisoni (approved):
🚀

PR #1390, opened 2025-05-30
Demo: X-ray Absorption Spectroscopy

colinjdale:
**Title:**

X-ray absorption spectroscopy simulation in the time domain

**Summary:**

This demo shows how to implement a time-domain XAS simulation algorithm for a small instance of an $N_2$ molecule. The algorithm is based on Fomichev et al. 2025, omitting some of the optimizations in that paper. XAS is introduced as well as why simulating XAS could be a promising use for early fault-tolerant quantum computers. The algorithm is described, and then implemented in PennyLane, with the time-domain and spectrum output plotted. Additional optimizations are discussed in an appendix after the conclusion.

**Relevant references:**

"Fast simulations of X-ray absorption spectroscopy for battery materials on a quantum computer" by Stepan Fomichev, Pablo A. M. Casares, Jay Soni, Utkarsh Azad, Alexander Kunitsa, Arne-Christian Voigt, Jonathan E. Mueller, Juan Miguel Arrazola, (2025) https://arxiv.org/abs/2506.15784

"Simulating X-ray absorption spectroscopy of battery materials on a quantum computer", by Stepan Fomichev, Kasra Hejazi, Ignacio Loaiza, Modjtaba Shokrian Zini, Alain Delgado, Arne-Christian Voigt, Jonathan E. Mueller, Juan Miguel Arrazola (2024)  https://arxiv.org/abs/2405.11015

**Possible Drawbacks:**

*Large knowledge overhead*
The demo makes extensive reference to previous quantum chemistry demos: [Using PennyLane with PySCF and OpenFermion](https://pennylane.ai/qml/demos/tutorial_qchem_external), [Initial state preparation for quantum chemistry](https://pennylane.ai/qml/demos/tutorial_initial_state_preparation), and [How to build compressed double-factorized Hamiltonians](https://pennylane.ai/qml/demos/tutorial_how_to_build_compressed_double_factorized_hamiltonians). This increases the overhead for the reader to understand the demo (IMO).

*Example too simple/not relevant*
$N_2$ is not a battery material, but calculating the ground state for some small cluster of a relevant material would take a long time, and is infeasible for a demo. 

*PennyLane factorization requires* `jax.config.update("enable_x64", True)`
`qml.qchem.factorize` does not seem to reliably produce a factorized Hamiltonian that works well in the algorithm unless I increase the precision used in Jax.

*Molecular orbital coefficient sign flip-flop/inconsistency between implementations*
The molecular orbital coefficients (MO coeffs) that are calculated in PySCF, PennyLane and OpenFermion end up slightly different from each other. Because CASCI methods are used to find the ground state in the active space, I had to overwrite the PySCF MO coeffs with the PennyLane calculated MO coeffs to match the ones used in the Hamiltonian and the dipole operator. I believe the MO coeffs have to be the same in all calculations for the result to be correct. However, `qml.qchem.scf` does not consistently produce the same MO coeffs, and I believe the same is true for PySCF. But, the beyond Hartree-Fock calculation methods in PySCF pull from the initial Hartree Fock instance, so the MO coefficients remain the same. However in PennyLane, the MO coeffs are recalculated when `dipole_moment` and `electron_integrals` are called, so I'm not sure if they will always give the same values. When trying an $N_2$ active space, I ended up changing the bond length to 1.0 Angstrom from 1.077 because that seemed to produce more stable results. 

**Related GitHub Issues:**

None.

----

To show off a recent Xanadu paper (https://arxiv.org/abs/2506.15784). This paper is of interest because it describes a quantum algorithm that could have industrial utility (battery material simulation). Xanadu's recent proposals for applications of quantum computers are focused on small but highly correlated systems, and this demo drives that point home. The implementation of the proposed algorithm in PennyLane is relatively simple, due to the functions in the quantum chemistry module of PennyLane. Showing off the utility of these functions is a main focus of the demo.

* AUDIENCE — Who is this for?

Chemistry researchers, anyone interested in Hamiltonian simulation algorithms.

X-ray absorption spectroscopy, Hamiltonian simulation, battery research, compressed double-factorized Hamiltonians, time-domain simulation, optimization, product formulas, commercial application.

- [ ] Tutorial
- [x] Demo
- [ ] How-to


demonstrations/tutorial_xas.py
```diff
@@ -0,0 +1,923 @@
...
+######################################################################
+# Note we can only obtain both real and imaginary expectation values in the simulated circuit. An
+# actual implementation would have to select real or imaginary by inserting a phase gate, like in the
+# circuit below.
+# 
+
+######################################################################
+# .. figure:: ../_static/demonstration_assets/xas/hadamard_test_circuit.png
+#    :alt: alt text
+#    :width: 70.0%
```
alvaro-at-xanadu, line 627: tk: add alt text for all figures

demonstrations/tutorial_xas.py
```diff
@@ -0,0 +1,923 @@
...
+# Figure X: *Hadamard test circuit to measure expectation value of time-evolution operator*. With the
+# phase gate :math:`S^\dagger` present (absent), this gives the real (imaginary) part of the
+# time-domain Green’s function :math:`\tilde G(\tau j)`.
+# 
+
+######################################################################
+# Run Simulation
+# --------------
+# 
+# Let’s define the simulation parameters we’re are going to use. This includes: - The Lorentzian width
```
alvaro-at-xanadu, line 638: List would probably be better here as well

daniela-angulo (approved):

PR #1372, opened 2025-05-19
[Demo] Add: Unitary synthesis with recursive KAK decompositions

dwierichs:
**Title:**
Unitary synthesis with recursive KAK decompositions

**Summary:**
Unitary synthesis, the process of compiling a unitary matrix into a quantum circuit, has been a topic of study for over
25 years. Three interesting techniques for unitary synthesis, namely the first, the (probably) most widely known, and the minimal-CNOT-count ("the best"), all use the same recursive KAK decomposition under the hood, as we recently
elaborated [in a paper](https://arxiv.org/abs/2503.19014).

This demo builds on our demo on KAK decompositions (added in #1227) and explains how such decompositions can be applied recursively to construct a many-qubit circuit and  how the three mentioned techniques all are powered by the same recursion.
It defers some optimization details to the literature but provides a brief gate count overview and showcases a simplified numerical implementation of all three techniques.

**Relevant references:**
Quite a few, see metadata file.

**Possible Drawbacks:**
N/A

**Related GitHub Issues:**

[sc-87548]

To do

- [x] Redo circuit diagrams
- [x] Add .tex files to repo
- [x] add examples/code

----
If you are writing a demonstration, please answer these questions to facilitate the marketing process.

* GOALS — Why are we working on this now?

- Promote `qml.liealg` features
- [promote a paper we recently did (see link above)]
- Follow up on started line of demos (#1227, #1261)

* AUDIENCE — Who is this for?
- Researchers in quantum compilation and quantum information
- Practicioners that want to learn where the complexity of their operations comes from

* KEYWORDS — What words should be included in the marketing post?
- (Recursive) KAK decomposition
- Quantum Shannon decomposition
- Block-ZXZ decomposition
- Unitary synthesis
- Compilation

* Which of the following types of documentation is most similar to your file? 
(more details [here](https://www.notion.so/xanaduai/Different-kinds-of-documentation-69200645fe59442991c71f9e7d8a77f8))
    
- [ ] Tutorial
- [x] Demo
- [ ] How-to

PR #1395, opened 2025-06-03
Fix reference formatting in KAK demo

dwierichs:
I used nested formatting in reference links in #1372.
ReST does not support this.
Here we remove the nested formatting (and adjust two small wordings that struck me as suboptimal upon re-reading)


Qottmann (changes requested):
to be consistent, we should also add links to the other journal papers

the second reference to QSD I think it would be more consistent to have the whole journal reference as the link, not just Vol and No

In other demos we often link arxiv as [arXiv:1234.1234](some_link) which I personaly prefer and think looks cleaner than writing "arXiv preprint" before each identifier-link

cognigami (approved):

PR #1396, opened 2025-06-03
Update concurrency group and workflow ref

doctorperceptron:
This PR updates the workflow files for building the V2 demos from dev and master to pul them from the respective branches rather than relying on the local versions that get checked out (see [the V1 workflows](https://github.com/PennyLaneAI/qml/blob/master/.github/workflows/build-branch-master.yml#L17) for reference.)

This also updates the concurrency groups, which were named the same as the V1 groups and were getting cancelled by those runs.

### DIFF
--- a/.github/workflows/v2-build-branch-dev.yml
+++ b/.github/workflows/v2-build-branch-dev.yml
@@ -5,12 +5,12 @@ on:
     - cron: '0 0 * * 0,2,4,6' # At 00:00 on Sunday, Tuesday, Thursday, and Saturday.
 
 concurrency:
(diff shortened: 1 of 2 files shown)


ashishks0522 (approved):
Looks good!

doctorperceptron:
> Is there a v2 build Branch dev workflow?

Sure is!

https://github.com/PennyLaneAI/qml/pull/1396/files#diff-a7089c8b0db3e46f54708ff72c7fb0c2579efcc28c0a9f5d5ec0071ca99df091

PR #1397, opened 2025-06-03
update attribution in GoSC demo

Qottmann:
see slack discussion [here](https://xanaduhq.slack.com/archives/C03FRM3RCCV/p1748901192865109)


Qottmann:
I went for a hybrid: Noting "Image source [1]." in the caption and leaving the attribution with the creative commons license explicitly mentioned @alvaro-at-xanadu @dwierichs

demonstrations/tutorial_game_of_surface_codes.py
```diff
@@ -57,7 +57,7 @@
 
     Qubits are defined as patches of tiles on the board. 
     A single qubit can occupy one tile (a) or multiple tiles (b), where dotted lines correspond to X and solid lines to Z operators.
-    Attribution see **
+    Image source [#Litinski]_.
```
dwierichs, line 60: As I understood the Slack thread, we want a colon there in order to make it grammatically correct?

demonstrations/tutorial_game_of_surface_codes.py
```diff
@@ -410,7 +410,7 @@
 # Attributions
 # ------------
 #
-# **: Images from `Game of Surface Codes <https://quantum-journal.org/papers/q-2019-03-05-128/>`__ by Daniel Litinski, `CC BY 4.0 <https://creativecommons.org/licenses/by/4.0/>`__
+# Images from `Game of Surface Codes <https://quantum-journal.org/papers/q-2019-03-05-128/>`__ by Daniel Litinski [#Litinski]_, `CC BY 4.0 <https://creativecommons.org/licenses/by/4.0/>`__
```
dwierichs, line 413: Not sure, but if this is about tracking a specific attribution, maybe it is actually better to just have one link (the Quantum paper)? I don't know enough about this to judge.

demonstrations/tutorial_game_of_surface_codes.py
Qottmann: oh right!

dwierichs (approved):
Looks good, sorry for the delay!

PR #1398, opened 2025-06-04
Test generate objects dot inv

doctorperceptron:
This PR updates the V2 `qml` demo tool to produce a master `objects.inv` file when using the html build option. This file is used to enable referencing demos from the docs using a short-form syntax and not needing to know the URL of the hosted demo files.

The reason this functionality is needed is because the new tool builds demos individually, so we need to explicitly merge the outputs from the individual `objects.inv` files together at the end of the build.

Note that this PR does not introduce a new workflow for generating and persisting this file. That will come in a separate PR.


ashishks0522 (approved):
Thanks Paul!

rashidnhm (approved):
Very creative solution! Great stuff :tada: :tada:

PR #1414, opened 2025-06-18
Update deprecated parameters in pyproject and bump numpy

rashidnhm:
**Title:** Update deprecated parameters in pyproject and pin numpyro

**Summary:**
The current master builds of QML are failing, from inspection of the logs the main culprit seems to be the version of numpyro being bumped.

Upon further inspection of the lockfile, it seems that the lockfile was also re-generated using an older version of poetry than what QML is using in CI. We use 2.x within CI but the lockfile was generated with 1.x. 

Though poetry 2.x tries it's best to be compatible with 1.x lockfile and pyproject, it is not in our best interest to support both, as it can cause problems.

As such, this PR updates the pyproject to address all deprecation warnings from poetry 2.x, including adding a minimum poetry version. This pyproject is now incompatible with poetry 1.x, and using poetry 1.x locally with QML repo will yield an error now. 

**Relevant references:**
sc-93662

**Possible Drawbacks:**
Users using poetry 1.x locally will need to use 2.x when updating lockfile for QML.

**Related GitHub Issues:**
None.


alvaro-at-xanadu (approved):

pyproject.toml
```diff
@@ -1,13 +1,35 @@
...
-repository = "https://github.com/PennyLaneAI/qml"
+license = "Apache-2.0"
+requires-python = "~3.10.0"
+authors = [
+    { name = "PennyLaneAI", email = "info@pennylane.ai" }
+]
 keywords = ["demo", "qml", "tensorflow", "automatic-differentiation", "tutorials", "pytorch", "autograd", "quantum-computing", "neural-networks", "quantum-chemistry", "key-concepts", "quantum-machine-learning"]
+dependencies = [
+    "typer>=0.15.1,<0.16.0",
+    "poetry>=1.8.5,<2.0.0",
```
andrijapau, line 14: Excuse my naivety but why is poetry restricted here but then down below it says `requires-poetry = ">=2.0"`?

rashidnhm: Oh my, this restriction should NOT be listed, here, tis' an error.

rashidnhm: Fixed

andrijapau:
Won't we have to do similar changes to the `dev` branch as well? Or else we will face merge conflicts come release time when we try to merge `dev` into `master`?

rashidnhm:
> Won't we have to do similar changes to the `dev` branch as well? Or else we will face merge conflicts come release time when we try to merge `dev` into `master`?

master gets merged to dev nightly so I think it should propagate on its own. 

If the nightly merge fails due to merge conflicts I'll open a new PR to dev with the same changes

andrijapau:
Sorry, one more question:

Remind me again why `dev` doesn't need this change? Is it because we are already using poetry 2.0 there?

rashidnhm:
> Sorry, one more question:
> 
> Remind me again why `dev` doesn't need this change? Is it because we are already using poetry 2.0 there?

Dev needs all these changes, which I'm hoping our nightly sync from master to dev can handle

Otherwise, as stated, this PR would need to be opened again for dev

PR #1431, opened 2025-07-09
Retain build for 3 days

doctorperceptron:
The V2 pipeline build artifacts are currently only retained for 1 day, but we only run the master and dev branch builds every 2 to 3 days, so the artifact goes stale between builds and can't be pulled. This PR bumps the retention period to 3 days so the artifact stays fresh and available until the next build.

### DIFF
--- a/.github/workflows/v2-build-demos.yml
+++ b/.github/workflows/v2-build-demos.yml
@@ -208,7 +208,7 @@ jobs:
         with:
           name: ${{ needs.generate-build-variables.outputs.artifact-name }}
           path: _build/pack/*.zip
-          retention-days: 1
+          retention-days: 3
           overwrite: true
       
       - name: Delete partial artifacts


andrijapau:
Why are builds only happening every 2-3 days?

rashidnhm:
Artifacts are only needed for a blip as once deploy workflow runs they are obsolete.

doctorperceptron:
> Artifacts are only needed for a blip as once deploy workflow runs they are obsolete.

... but these artifacts aren't used for deployments anymore. We deploy each demo whenever it's merged to master. Going forward the only use case I see for the "build master branch" and "build dev branch" workflows are for the [qml diff checker](https://github.com/PennyLaneAI/qml-diff-checker). Since this is run manually it could pull a stale build in between cron runs (this is what just happened to me, hence this PR).

@rashidnhm what are the [build branch dev](https://github.com/PennyLaneAI/qml/blob/master/.github/workflows/build-branch-dev.yml) and [build branch master](https://github.com/PennyLaneAI/qml/blob/master/.github/workflows/build-branch-master.yml) workflows for? These are the ones that run on this schedule.

rashidnhm:
It makes sense to parametrize the retention duration. For PR builds, anything over 1 is wastage

PR #1435, opened 2025-07-10
Add V2 objects.inv workflow

doctorperceptron:
The `objects.inv` file is generated by Sphinx and is used to enable easy syntax for referencing demos from the PennyLane docs.

This PR adds a workflow that builds all demos in a specified branch and uploads the generated `objects.in` file to s3. The `v2-build-demos-swc-env` workflow calls this workflow after merges to `master` or `dev` (currently using testing branches), so that the `objects.inv` files are created and uploaded whenever new demos are deployed to those environments. 

This new workflow is set to run on merges to the two testing branches, `test-v2-master` and `test-v2-dev`. We don't have testing buckets though, so merges to the `test-v2-dev` branch will push the `objects.inv` file to the `dev` bucket, while merges to the `test-v2-master` branch will push the `objects.inv` file to the `staging` bucket. Because this will push to the dev environment, we should hold off on merging this until after PennyLane v0.42.0 is released next week, or merge before then but refrain from testing the dev version of this workflow until then.

Note: I've been executing a lot of demos locally and am getting annoyed by the generated `sg_execution_times.rst` file tainting my branch, so I've also added this to `.gitignore`.


.gitignore
```diff
@@ -13,6 +13,7 @@ demos_community.rst
 .venv/
 .venv-build/
 **/*.egg-info/
+sg_execution_times.rst
```
Alan-eMartin, line 16: 🙌🏻

Alan-eMartin (approved):
This looks good to me. Do we want to wait until we're ready to transition pipelines before merging?

doctorperceptron:
> This looks good to me. Do we want to wait until we're ready to transition pipelines before merging?

I used the testing branch names so we can merge now and verify that the objects.inv file is being correctly uploaded during QA. Let's merge now and then make a PR to switch to the correct branches when we do the same for the other workflows.

ashishks0522 (approved):

PR #1444, opened 2025-07-16
Sync constraints post release

Alan-eMartin:
**Title:**
Bumping `deps` for `qml-v2` post PL release.

**Summary:**
Updates dependencies specific to QML V2 builds

### DIFF
--- a/.github/workflows/v2-build-branch-dev.yml
+++ b/.github/workflows/v2-build-branch-dev.yml
@@ -10,11 +10,11 @@ concurrency:
 
 jobs:
   build_dev:
-    uses: PennyLaneAI/qml/.github/workflows/v2-build-demos.yml@sync-constraints-post-release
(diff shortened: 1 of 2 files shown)


lib/qml/lib/demo.py
```diff
@@ -278,77 +278,77 @@ def _build_demo(
...
-        )
-        # We need to bump flax here, after Jax has been bumped by Catalyst
-        cmds.pip_install(
-            build_venv.python,
-            "--upgrade",
-            "flax==0.10.6",
-            use_uv=False,
-            quiet=False,
-        )
+    # if dev and execute:
```
Alan-eMartin, line 281: @doctorperceptron, do we want to keep this on hold for the time being, until the issues with the `dev` are resolved?

Alan-eMartin: _clarification_ the comment versions of these.

PR #1447, opened 2025-07-17
Fix shots use in `tutorial_quantum_circuit_cutting`

albi3ro:
**Title:**

**Summary:**

`tutorial_quantum_circuit_cutting.py` was providing shots as a pennylane numpy tensor.  A change in pennylane makes it so the demo could no longer get away with that.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**


andrijapau:
Tagging @doctorperceptron for visibility

doctorperceptron:
FYI @Alan-eMartin

PR #1465, opened 2025-07-24
Mark `tutorial_error_mitigation` as non-executable

SimoneGasperini:
The `tutorial_error_mitigation.py` has to be marked as non-executable to unblock the `dev` branch CI.

The issue is related to [this PR](https://github.com/PennyLaneAI/pennylane/pull/7909) in PennyLane which is introducing the deprecation of the `QuantumScript.to_openqasm` method in v0.43 (removal in v0.44). This method is still used in Mitiq's `from_pennylane` conversion function (see [here](https://github.com/unitaryfoundation/mitiq/blob/main/mitiq/interface/mitiq_pennylane/conversions.py#L28)) and it works fine for them because Mitiq currently has an explicit dependency on PennyLane 0.36.

We will be able to mark the tutorial executable again when they update their code to work with `pennylane>=0.43`.


SimoneGasperini:
@andrijapau Is this enough? I don't think we need to add any warning for the PennyLane version because it is already specified that the demo require `mitiq>=0.11` which had the correct PL version as a dependency.

andrijapau (approved):
Can you add the warning to the demo indicating it's only compatible with 0.36?

SimoneGasperini:
> Can you add the warning to the demo indicating it's only compatible with 0.36?

Suure! Done in [cc43015](https://github.com/PennyLaneAI/qml/pull/1465/commits/cc4301515818a1f965f62f055588c99d3e1863bd).

doctorperceptron:
@Alan-eMartin this will break our existing slugs for this demo ☹️

Alan-eMartin:
> @Alan-eMartin this will break our existing slugs for this demo ☹️

Thanks for the heads up!

andrijapau (approved):
Looks good to me, I think we can go forward with this --- can you add a story somewhere that reminds us to make it executable when mitiq updates?

comp-phys-marc (approved):

PR #1467, opened 2025-07-25
Optionally override demo execution via the metadata

doctorperceptron:


Alan-eMartin (approved):
Great, this will be a lot better, and prevent a lot of seo issues etc.

PR #1468, opened 2025-07-25
Fix v2 demo links to other demos

doctorperceptron:
The new V2 pipeline builds demos independently. One consequence of this is that Sphinx is unable to resolve references within a demo to other demos using the existing `:doc:` syntax. As an example, the `tutorial_adaptive_circuits` demo references another demo in its References:
Screenshot 2025-07-25 at 5 11 56 PM
In the above screen shot from production you can see the fourth reference renders the title correctly as a hyperlink to the referenced demo.

This doesn't work in the V2 pipeline, as can be seen in this screenshot for the same demo in staging (V2):
Screenshot 2025-07-25 at 5 13 45 PM

We can fix this by providing the intersphinx mapping for the demos, which references the objects.inv file that we host on pennylane.ai for all the demo files. The final step is to rename any reference to a demo to conform to the naming convention for referencing external demos.

This is a draft PR showing how to do this for a single demo. This has been tested and verified locally by building the demo.
 Once we are ready to go live with the V2 pipeline, we will need to merge the updated `conf.py` file (we could even do that in advance) and then go through all the demos and replace e.g. `:doc:demo_name` with `:doc:demos/demo_name` and similar where applicable. Note the lack of leading / before `demos`, as some demos include this it will need to be removed.

### DIFF
--- a/demonstrations_v2/tutorial_adaptive_circuits/demo.py
+++ b/demonstrations_v2/tutorial_adaptive_circuits/demo.py
@@ -411,7 +411,7 @@ for n in range(10):
 #
 # .. [#givenstutorial]
 #
-#     :doc:`tutorial_givens_rotations`
+#     :doc:`demos/tutorial_givens_rotations`
 #
 # .. [#grimsley2019]
 #


justinpickering:
[sc-96745]

PR #1482, opened 2025-08-01
`v2-sync` | Sync ongoing deprecation warnings into v2 demonstrations folders and  metadata changes

Alan-eMartin:
This PR syncs all changes from Deprecate-device-shot-settings to v2. ~It also updates the metadata within demonstrations_v2, adding an "executable" flag that is true if the directory name starts with tutorial_, and false otherwise.~

- This will easier conflict management easier as we are transitioning to V2


demonstrations_v2/error_mitigation/demo.py
doctorperceptron, line 1: Master has the `tutorial_` version of this demo, while dev has the non-tutorial version. Let's keep the tutorial version and then we will have to change it to non-executable when we merge into dev.

demonstrations_v2/tutorial_zne_catalyst/demo.py
doctorperceptron, line 1: To match master we should rename this one to `zne_catalyst`

doctorperceptron: Ah I see there is already a folder for zne, so we can move the demo over there and delete this one.

doctorperceptron (approved):

PR #1498, opened 2025-08-05
Bugfix |  Fix `new` command to add image pathn to metadata if valid

Alan-eMartin:
This pull request updates the `_setup_thumbnails` function in `lib/qml/app/app.py` to simplify how thumbnail files are handled. Instead of copying thumbnail files into the demo directory, the function now just verifies their existence and returns their absolute paths. This reduces complexity and avoids potential file operation errors.

**Links**
- https://app.shortcut.com/xanaduai/story/97075/fix-new-command-in-qml-tool

**Thumbnail handling changes:**

* Changed `_setup_thumbnails` to no longer copy thumbnail files; it now only checks if the files exist and returns their paths as strings. This removes error handling for file operations and simplifies the function logic.

### DIFF
--- a/.github/workflows/v2-deploy-pr.yml
+++ b/.github/workflows/v2-deploy-pr.yml
@@ -103,18 +103,8 @@ jobs:
         run: |
           demos="${{ needs.prepare-build-context.outputs.updated_demos }}"
           
-          comment="### Preview(s) are ready! :tada:\n"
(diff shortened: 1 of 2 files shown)


Alan-eMartin:
> Looks good, just to confirm we do upload thumbnails from their original directory itself correct? there is no other dependency for them

Correct, this is a path to the demo image, and it is uploaded from there

PR #1501, opened 2025-08-07
Remove TF usage in demos

isaacdevlugt:
**Title:** Remove TF usage

**Summary:** Remove TF usage in demos due to dep & removal. We will need to ensure that relevant demos are no longer executed in CI.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

### DIFF
--- a/demonstrations/learning2learn.py
+++ b/demonstrations/learning2learn.py
@@ -14,8 +14,11 @@ Learning to learn with quantum neural networks
 *Author: Stefano Mangini — Posted: 02 March 2021. Last updated: 15 September 2021.*
 
 .. warning::
-    This demo is only compatible with TensorFlow version ``2.9`` or below.
-    Otherwise, the output of some cells and plots may differ.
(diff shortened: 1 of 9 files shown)


andrijapau:
Also, @isaacdevlugt , you should re-base to `dev` so these changes don't go immediately live

doctorperceptron:
> Don't we need to update only the v2 workflows? Also could you update the metadata? 😊

Yes please. @isaacdevlugt we've switched over to the new (we call it V2) pipeline for demos. There's a [documentation PR](https://github.com/PennyLaneAI/qml/pull/1470) in progress. The points relevant to this PR are:
- No need to update the demos in `demonstrations/`, these are no longer being built and served to the website and are all going to be removed soon.
- The metadata can now control the execution of the demos. The fields `executable_stable` and `executable_latest` control whether or not the demo is executable in the master or dev branches, respectively. If you want to stop executing the `tutorial_` demos, you can change these to false. If these metadata fields are missing, then the pipeline falls back to executing demos that start with `tutorial_`.

PR #1526, opened 2025-08-20
rm demonstrations folder

Alan-eMartin:
**Summary**

This PR removes all legacy v1 workflows and the `demonstrations` v1 directory containing all v1 demos.

**Note:**
I have not completely removed everything from our `github` dir as I am unsure if some of these files and directories are still relevant.

I am wondering if @rashidnhm could and @mudit2812 could verify some of these before deleting.

- `generate_diffs.py` I believe this is the old, old version of diff checker?
- `html_parser.py` Also part of legacy diff checker?
- `qml_pipeline_utils/` directory.
- `artifact.yml`

Also — please let me know if I should restore anything, I did my best to investigate.


mudit2812:
> generate_diffs.py I believe this is the old, old version of diff checker?
html_parser.py Also part of legacy diff checker?

Yes, and yes. Those should actually have been removed years ago 😅

mudit2812 (commented):
I'm not sure about `qml_pipeline_utils` or `artifact.yml`, but it should be safe to delete the other two files you asked about. Otherwise, the removals seem fine to me. Will approve once those are removed.

mudit2812 (approved):
Nice 🚀

rashidnhm:
No issues from my end. `qml_pipeline_utils` is part of the previous v1 tooling.

PR #1527, opened 2025-08-20
Fix admontion on `tutorial_neutral_atoms`

Alan-eMartin:
## Summary

Fixes issue where "blue" styling doesn't appear. The issue is that the container class rst conflicts with tailwind classes

## Change:
- convert blocks to tip admonition

## Issue
- https://app.shortcut.com/xanaduai/story/93788/update-tutorial-neutral-atoms-to-use-admonition-note-instead-of-alert-class


alvaro-at-xanadu (approved):
Thank you for this @Alan-eMartin!

PR #1538, opened 2025-09-04
Update uv

doctorperceptron:
The Pytorch index has been flaky recently, causing [builds to fail](https://github.com/PennyLaneAI/qml/actions/runs/17419394443). This PR bumps the version of `uv`, which includes automatic retries in the event of a 503 response from the Pytorch index. 

This has been [tested](https://github.com/PennyLaneAI/qml/actions/runs/17444053054), but since we can't make the Pytorch index unavailable we won't know it works until we stop seeing failures due to 503s from Pytorch.


Alan-eMartin (approved):
Thanks, Paul! Hopefully, having this retry logic will help remedy the situation. 

LGTM.

andrijapau (approved):
Only seems like a good thing. Thanks! ✨

PR #1540, opened 2025-09-05
Fix device shots in `tutorial_xas`

JerryChen97:
Tutorial for XAS is a new demo that was published on 29 Aug 2025. It involved some deprecated usage of setting shots on device. Therefore we would like to fix it on `dev` branch.


andrijapau (approved):
Nice!

PR #1546, opened 2025-09-15
Custom demo rst processing

doctorperceptron:
This PR introduces a series of [Pandoc filters](https://pandoc.org/filters.html) to customize processing of various rst directives for the markdown output in the rendered Jupyter Notebook versions of the demos.

### DIFF
--- a/lib/filter_figures.py
+++ b/lib/filter_figures.py
@@ -8,9 +8,14 @@ from pandocfilters import toJSONFilter, Image
 
 def filter_images(key, value, format, _):
     if key == 'Image':
-        specs, url, alt = value
+        specs, [alt], src = value
+
+        try:
+            alt["c"] = ""
+        except KeyError:
+            alt = { "t": "Str", "c": "" }
         
-        return Image([specs, url, alt])
+        return Image(specs, [alt], src)
 
 if __name__ == '__main__':
     toJSONFilter(filter_images)
--- a/lib/filter_figures.py
+++ b/lib/filter_figures.py
@@ -16,15 +16,12 @@ def parse_img_source(src: str) -> str:
 
 def filter_images(key, value, format, _):
     if key == 'Image':
-        specs, [alt], [url, _] = value
+        _, _, [url, _] = value
         cleaned_url = parse_img_source(url)
 
-        try:
-            alt["c"] = cleaned_url
-        except KeyError:
-            alt = { "t": "Str", "c": cleaned_url }
+        alt = { "t": "Str", "c": cleaned_url }
         
-        return Image(specs, [alt], [cleaned_url, ""])
+        return Image(["", [], []], [alt], [cleaned_url, ""])
 
 if __name__ == '__main__':
     toJSONFilter(filter_images)
--- a/lib/filter_figures.py
+++ b/lib/filter_figures.py
@@ -3,6 +3,7 @@
 """
 Pandoc filter to convert figure directives to images.
 """
+
 import os
 from pandocfilters import toJSONFilter, Image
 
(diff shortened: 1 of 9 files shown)


lib/qml/lib/demo.py
```diff
@@ -311,6 +311,7 @@ def _build_demo(
         "GALLERY_OUTPUT_DIR": str(out_dir.resolve().relative_to(ctx.repo_root)),
         # Make sure demos can find scripts installed in the build venv
         "PATH": f"{os.environ['PATH']}:{build_venv.path / 'bin'}",
+        "CURRENT_DEMO": str(demo.name),
```
doctorperceptron, line 314: We need this to be able to add the cloud URL for figures referenced in the demo, since Pandoc doesn't know which demo it's processing.

lib/filter_links.py
```diff
@@ -0,0 +1,83 @@
...
+    # 1. Some text followed by <sphinx/link>
+    # 2. ~.pennylane.some.function.name
+    # 3. sphinx/link
+    if "<" in link:
+        text, key = link.split("<")
+        return text.strip(), key.strip(">").lstrip(".")
+    elif "~" == link[0]:
+        return link.split(".")[-1], link.lstrip("~.")
+    else:
+        return "", link
```
Alan-eMartin, line 51: **Nit:**
could we use `<string>.startswith(<char>)` here, instead of `link[0]`?

lib/filter_links.py
```diff
@@ -0,0 +1,83 @@
...
+
+def pandocify_string(string: str) -> list:
+    """Helper function to convert a string to a list of Pandoc Str nodes."""
+    pandoc_string = []
+    for substring in string.split(" "):
+        pandoc_string.append({"t": "Str", "c": substring},)
+        pandoc_string.append({"t": "Space"},)
+    pandoc_string.pop() # Remove the last space
```
Alan-eMartin, line 59: **Clarification:**
Does this produce a blank " " at the end of the list, hence the pop here?

**Suggestion/Nit:**
We could consider the following `for substring in string.rstrip().split(" "):` instead?

lib/filter_figures.py
```diff
@@ -0,0 +1,28 @@
...
+import os
+from pandocfilters import toJSONFilter, Image
+
+ASSETS_DIR = "https://blog-assets.cloud.pennylane.ai/demos/"
+
+def parse_img_source(src: str) -> str:
+    if "../_static/" in src:
+        return src.replace("../_", ASSETS_DIR + os.getenv("CURRENT_DEMO") + "/main/_assets/")
+    else:
+        return src
```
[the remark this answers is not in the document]
Alan-eMartin: There is a chance that these images haven't been uploaded yet @ashishks0522. If this is the first build of a demo, they have not yet reached the server to upload images to S3.

Also, this only accounts for prod, directories in S3, so when this runs in a PR preview, the links will be broken regardless.

Alan-eMartin (approved):
Hey @doctorperceptron, a few non-blocking nits/considerations.

But from my end, everything is being rendered as expected in the notebooks

lib/filter_figures.py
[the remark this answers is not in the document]
doctorperceptron: Yup what Alan said. I chose to return the original URL so we could in principle handle any broken cases further down the pipeline.

lib/filter_links.py
doctorperceptron: Not quite. In the Pandoc AST a sentence is represented by a list, where each word and space is a dict:
- word: {"t": "Str", "c": word}
- space: {"t": "Space"}

This loop breaks the sentence up into words, then adds a dict for each word, followed by a space. The pop then removes the last added space dict since an n-word sentence has n-1 spaces.

ashishks0522 (approved):
Thanks Paul!

Alan-eMartin (approved):
Thanks again Paul!

PR #1551, opened 2025-09-19
Patch filters

doctorperceptron:
This PR adds a patch to the `filter_links` Pandoc filter to process inline HTML. Two demos use this directive and are crashing the objects.inv build.

Verified that all demos now build [here](https://github.com/PennyLaneAI/qml/actions/runs/17861828175).


ashishks0522 (approved):
Thanks Paul!

Alan-eMartin (approved):

PR #1558, opened 2025-09-22
[PENNYPALOOZA] Modify kitchen sink demo

doctorperceptron:
Changes the description appearing in the kitchen sink demo that references quantum machine learning.


andrijapau (approved):
❤️

PR #1566, opened 2025-10-03
fix format option in tutorial_variational_classifier

jontoye:
## Summary

This PR adds a missing colon (:) to the `format` role option in the `html` role definition, fixing anchor tag rendering issues in the `tutorial_variational_classifier` demo.


jontoye:
> Thanks @jontoye can we also audit while we are here that this issue is not present in any other demo?

Thanks @ashishks0522, sorry should have mentioned in the description that I did search for and confirm that there are no more instances of the incorrect `format:` option in the repo. Anywhere else it is currently used has the correct `:format:` form. 👍

PR #1462, opened 2025-07-23
Add demo on resourcefulness analysis with generalised Fourier transforms

mariaschuld:
This is a demo we are creating with the first authors of [this paper](https://arxiv.org/abs/2506.19696).

PR #1577, opened 2025-10-16
Tutorial on resourcefulness (CI fix)

doctorperceptron:
This is a direct copy of PR #1462. That PR was open before the switch to the V2 pipeline which has caused CI to stall.

### DIFF
--- a/demonstrations_v2/tutorial_quantum_phase_transitions/metadata.json
+++ b/demonstrations_v2/tutorial_quantum_phase_transitions/metadata.json
@@ -8,8 +8,8 @@
             "username": "tirthshah2806"
         }
     ],
-    "dateOfPublication": "2025-10-27T00:00:00+00:00",
-    "dateOfLastModification": "2025-10-29T00:00:00+00:00",
+    "dateOfPublication": "2025-10-28T00:00:00+00:00",
+    "dateOfLastModification": "2025-10-30T00:00:00+00:00",
     "categories": ["Optimization","Quantum Machine Learning"],
     "tags": ["quantum phase transition", "Ising model"],
     "previewImages": [


demonstrations_v2/tutorial_resourcefulness/demo.py
```diff
@@ -0,0 +1,758 @@
...
+ax1.set_title("Fourier coefficients")
+ax2.bar(np.array(range(N))+0.05, np.imag(g_vals), color='lightpink', label="imaginary part")
+ax2.bar(range(N), np.real(g_vals), color='dimgray', label="real part")
+ax2.set_title(f"function g")
+plt.legend()
+plt.tight_layout()
+plt.show()
+
+######################################################################
+# The function with the least resource is constant! This makes sense: We know that the decay of the Fourier coefficients
```
peterse: I didn't follow what this means. What does 'the number of times a function is differentiable' mean?

demonstrations_v2/tutorial_resourcefulness/demo.py
```diff
@@ -0,0 +1,758 @@
...
+of quantum states as a resource, reproducing Figure 2 in [#Bermejo_Braccia]_.
+
+.. figure:: ../_static/demonstration_assets/resourcefulness/figure2_paper.png
+   :align: center
+   :width: 70%
+
+   Figure 1: GFD Purities of different states using 2-qubit entanglement as a resource.
+   A Bell state, which is maximally entangled, has high contributions in second-order GFD Purities, while
+   a tensor product state with no entanglement has contributions in the first-order Purities. The interpolation
+   between the two extremes, exemplified by a Haar random state, has a spectrum in between.
```
peterse: spectrum = GFD purity? or spectrum= power spectrum? 

Also, I liked the power spectrum analogy given in the main text, i think it is worth repeating here because this plot can only be understood on an intuitive level at this point in the demo,...

demonstrations_v2/tutorial_resourcefulness/demo.py
```diff
@@ -0,0 +1,758 @@
...
+++++++++++++++++++++++++++++++++++
+
+Let's start by recalling the `standard discrete Fourier transform <https://en.wikipedia.org/wiki/Discrete_Fourier_transform>`__.
+Given N real values :math:`x_0,...,x_{N-1}`, that we can interpret as the values
+of a function :math:`f(x)` over the integers :math:`x \in {0,...,N-1}`, the discrete Fourier transform
+computes the Fourier coefficients
+
+.. math:: \hat{f}(k) = \frac{1}{\sqrt{N}} \sum_{x=0}^{N-1} f(x) e^{\frac{2 \pi i}{N} k x}, \;\; k = 0,...,N-1
+
+In words, the Fourier coefficients are projections of :math:`f(x)` onto the *Fourier basis functions*
```
peterse: nit: projection is usually from a vector to a vector, this is more like the length or norm of a projection of f, right?

demonstrations_v2/tutorial_resourcefulness/demo.py
```diff
@@ -0,0 +1,758 @@
...
+    P[i, :] = np.roll(P[i-1, :], 1)
+
+print(P)
+
+
+######################################################################
+# 4. Find the basis that block-diagonalises the representation
+# ************************************************************
+# This was quite a bit of group jargon, but for a good reason: We are now guaranteed that there is
+# a basis change that block-diagonalises *all* circulant permutation matrices, or more generally,
```
peterse: is it obvious why this basis change exists, or does "guaranteed" mean that this a fact from representation theory that is not obvious here?

demonstrations_v2/tutorial_resourcefulness/demo.py
```diff
@@ -0,0 +1,758 @@
...
+# resource fingerprint we want to determine.
+#
+# For example, the GFD Purity for the third block (or "frequency") for the vector ``f_vec`` would be calculated as follows:
+#
+
+purity3 = np.abs(np.vdot(basis3, f_vec))**2
+
+######################################################################
+# What did we just do? We wrote a discrete function as a vector and computed the magnitude of its projection onto
+# a subspace spanned by rows of ``F``, the matrix that moves into the Fourier basis. This is exactly what we
```
peterse: > its projection onto a subspace spanned by rows of ``F``,

again i think its worth being explicit here that "rows of F" is for the non-abelian case, "a row of F" is for the abelian example. unless I have completely misunderstood this construction

demonstrations_v2/tutorial_resourcefulness/demo.py
```diff
@@ -0,0 +1,758 @@
...
+# did when computing the power spectrum. We have indeed just made the recipe for computing the power
+# spectrum more complicated. To confirm this, let's verify that the third entry of the power spectrum
+# is the third GFD Purity.
+#
+
+print("GFD Purity and power spectrum coincide:", np.isclose(power_spectrum[3], purity3))
+
+######################################################################
+# We now have a very different perspective on the power spectrum :math:`|\hat{f}(k)|^2`: It is a projection of the function
+# :math:`f` into irreducible subspaces revealed by moving into the basis that block-diagonalises circular matrices.
```
peterse: circular -> circulant (?)

peterse (commented):
Overall its quite clear! The introduction feels a bit rocky: There are lots of technical terms, but none of them are yet defined. I do not have a good idea on how to deal with that, compared to what you have written here. Otherwise the rest of the demo is conceptually very rewarding to a reader who pushes through, and I think that you build up an intuition for these objects so nicely.

Just two high-level comments
 - I don't fully agree with "complex" being interchangeable with "resource" (actually,  i don't like the "complex" is used here at all). Equating "complex"="many resource" implies that f(x)=sin(1000*x) is significantly more complex than a function of the form `f(x) = \sum_{\omega < 100} \hat{f}(\omega) sin(\omega*x)`. I don't see why that is. The former is much more compressible, much simpler to write down. There is not much followup on what 'complex' means (beyond the comparison "complex"="many resource" in the very first paragraph) so I do not think that the word 'complex' adds very much here, and could just be replaced with some language related to resource theory (expensive, not-smooth?)
 - It was jarring when the representation theory for states in H ended up using representations acting on density matrices in B(H) rather than states in H. It seemed like the introduction was working towards a generalization of the Z_n example, and then suddenly we use some representation that involves conjugating density matrices. What went wrong? _Why_ doesn't the naive generalization of what you already built up work for SU(2)\times SU(2)\times... ?? Everything you wrote was crystal clear, but the reasoning behind the choices is completely opaque...

PR #1586, opened 2025-10-28
Enable latest docs linking for dev builds

doctorperceptron:
This PR adds an environment variable for Sphinx to distinguish dev builds from stable builds. This enables Sphinx to use the latest or stable docs, respectively, for Intersphinx linking.

**Note**: We can't link to demos that only appear on `dev`, as this domain is not publicly reachable.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -54,6 +54,12 @@ extensions = [
     "extension",
 ]
 
+dev = os.getenv("DEV", "False")
+if dev == "True":
+    dev = True
+else:
+    dev = False
+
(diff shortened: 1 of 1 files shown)


conf.py
```diff
@@ -16,9 +16,6 @@
 import sys
 import warnings
 import numpy as np
-from jinja2 import FileSystemLoader, Environment
-import yaml
-from pathlib import Path
```
doctorperceptron: Cleanup. None of these imports are being used.

PR #1587, opened 2025-10-28
Daily qml builds with RC environment

runora95:
**Title:**
Daily QML builds with the RC environment when it exists. 

**Summary:**

- Adds a workflow to run daily QML builds with the RC environment. The workflow checks if an RC branch exists in Pennylane. If an RC branch exists it then gets the latest versions of pennylane, catalyst, and lightning from PyPI, updates the `demo.py` file with them, and builds the demos. 

**Relevant references:**
[Example demo.py configuration for an RC build](https://github.com/PennyLaneAI/qml/pull/1568/files)

**Possible Drawbacks:**
None

**Related GitHub Issues:**
[1579](https://github.com/PennyLaneAI/qml/pull/1579)


andrijapau (commented):
Have you tried this out?

.github/workflows/v2-build-rc-demos-daily.yml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Update demo versions
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          sed -i -E "s#git\+https://github.com/PennyLaneAI/pennylane.git\#egg=pennylane\",#pennylane<=${{ steps.setup-rc-versions.outputs.pennylane-version }}\",\\n\\t\\t\\t\"—extra-index-url\",\\n\\t\\t\\t\"https://test.pypi.org/simple/\",\\n\\t\\t\\tpre=True,#" lib/qml/lib/demo.py
+          sed -i "s/PennyLane-Lightning/pennylane-lightning<=${{ steps.setup-rc-versions.outputs.lightning-version }}/" lib/qml/lib/demo.py
+          sed -i "s/PennyLane-Catalyst/pennylane-catalyst<=${{ steps.setup-rc-versions.outputs.catalyst-version }}/" lib/qml/lib/demo.py
+          git commit -am "Set up RC versions for demo builds"
+          git push --set-upstream origin ${{ steps.setup-rc-versions.outputs.rc-build-branch }}
```
doctorperceptron: I can see the logic here, but I don't think manipulating the qml build tool files directly with sed in an Action is the right approach. I would prefer instead to add new command line argument(s) to the tool. We could even make this more general so you can build any version for PLC by passing in the arguments. We can then keep the logic that parses out the RC branches here in the Action.

doctorperceptron (changes requested):

PR #1603, opened 2025-11-20
Temporary copy of Low depth circuits Demo for preview

DSGuala:
PR https://github.com/PennyLaneAI/qml/pull/1554 has an error with the requirements and it is not rendering. Since that PR is coming from an external contributor, we created a fixed duplicate here to be able to see the preview while we wait for the fix to happend in the original PR.


DSGuala:
Closing because we already fixed this in the main PR

PR #1606, opened 2025-11-23
Add upper bound to summation of the QFT equation in QPE tutorial

antalszava:
**Summary:**

The expression for QFT in the QPE tutorial has a summation that could do with an upper bound and this PR adds that.


josh146:
@antalszava do you mind also updating the `dateOfLastModification` in the file https://github.com/antalszava/qml/blob/patch-1/demonstrations_v2/tutorial_qpe/metadata.json?

Edit: nevermind, looks like I have edit permissions :)

demonstrations_v2/tutorial_qpe/demo.py
```diff
@@ -81,7 +81,7 @@
 "When in doubt, take the Fourier transform"; or in our case, "When in doubt, take the quantum Fourier transform (QFT)".
 
 .. math::
-   \text{QFT}|\theta\rangle = \frac{1}{\sqrt{2^n}}\sum_{k=0} e^{2 \pi i\theta k} |k\rangle.
+   \text{QFT}|\theta\rangle = \frac{1}{\sqrt{2^n}}\sum^{n-1}_{k=0} e^{2 \pi i\theta k} |k\rangle.
```
KetpuntoG:
```suggestion
   \text{QFT}|\theta\rangle = \frac{1}{\sqrt{2^n}}\sum^{2^n-1}_{k=0} e^{2 \pi i\theta k} |k\rangle.
```

PR #1609, opened 2025-11-25
DQI mistakes fixes

daniela-angulo:
A series of changes in the code


josh146 (approved):
Thanks @daniela-angulo -- so it looks like this is predominantly an indexing by 1 issue?

Is it worth also asking @dwierichs for review as he was the original reviewer?

ashishks0522 (approved):

PR #1613, opened 2025-12-03
Update VQE demo to use dataset energy and geometry 

soranjh:
The demo titled "A brief overview of VQE" imports its example from the dataset. However, the reference energy and the coordinates in a Note section are still the original hard-coded ones. This gives slightly different results for the computed and expected energies.

This PR updates the demo to use dataset values in the following sections. This makes the results consistent everywhere in the demo. 

- `E_fci = -1.136189454088` is changed to `E_fci = dataset.fci_energy`.
- `coordinates = np.array([[0.0, 0.0, -0.6614], [0.0, 0.0, 0.6614]])` is changed to
  `coordinates = np.array([[-0.70108983, 0.0, 0.0], [0.70108983, 0.0, 0.0]])`


CatalinaAlbornoz (approved):
Looks good to me. Thanks for making this change Soran!

PR #1616, opened 2025-12-08
Actions cleanup

doctorperceptron:
The Actions that are run on a schedule are executed according to UTC, but the crons have been set assuming they will be run on ET. This PR adjusts the cron schedules so they will execute at 5am UTC (midnight EST, 1am ET). 

Additionally, the nightly merge of master into dev was only running on Sundays. This PR adjusts the schedule to run nightly. 

Finally, require metadata validation to pass in order to build and deploy a PR preview.


Lupeclan (approved):
LGTM. 📆

PR #1618, opened 2025-12-09
Demo: Active volume

dwierichs:
**Title: Active volume**

**Summary:**
Introduces the framework from [arXiv:2211.15465](https://arxiv.org/abs/2211.15465) for active volume-aware compilation.
Walks through the compilation of an exemplary Clifford circuit, showcasing the concepts of oriented ZX diagrams and logical networks. Complemented by explanations of reactive measurements/reaction time and state teleportation.

**Relevant references:**
https://arxiv.org/abs/2211.15465

**Possible Drawbacks:**

**Related GitHub Issues:**

----

  Strong fault-tolerant compilation framework of broad interest

* AUDIENCE — Who is this for?

  Researchers/students in quantum compilation and/or error correction.

- Active volume
- Quantum compilation
- Surface code/Quantum error correction 
- Fault-tolerant quantum computation
- (Lattice surgery)

- [ ] Tutorial
- [x] Demo
- [ ] How-to

### DIFF
--- a/demonstrations_v2/tutorial_active_volume/metadata.json
+++ b/demonstrations_v2/tutorial_active_volume/metadata.json
@@ -38,7 +38,34 @@
             "year": "2022",
             "journal": "arXiv",
             "url": "https://arxiv.org/abs/2211.15465"
-        }
+        },
+        {
+        "id": "Litinski2018",
+        "type": "article",
+        "title": "A Game of Surface Codes: Large-Scale Quantum Computing with Lattice Surgery",
+        "authors": "Daniel Litinski",
+        "year": "2018",
+        "journal": "arXiv",
+        "url": "https://arxiv.org/abs/1808.02892v3"
+    },
+    {
+        "id": "Gidney2018",
+        "type": "article",
+        "title": "Halving the cost of quantum addition",
+        "authors": "Craig Gidney",
+        "year": "2018",
+        "journal": "Quantum",
+        "url": "https://quantum-journal.org/papers/q-2018-06-18-74/"
+    },
+    {
+        "id": "PauliFrame",
+        "type": "article",
+        "title": "Pauli Frames for Quantum Computer Architectures",
+        "authors": "L. Riesebos, X. Fu, S. Varsamopoulos, C. G. Almudever, K. Bertels",
+        "year": "2017",
+        "journal": "Proceedings of the 54th Annual Design Automation Conference (DAC '17)",
+        "url": "https://doi.org/10.1145/3061639.3062300"
+    }
     ],
     "basedOnPapers": [
         "10.48550/arXiv.2211.15465"


demonstrations_v2/tutorial_active_volume/metadata.json
```diff
@@ -0,0 +1,64 @@
+{
+    "title": "Active volume",
```
Qottmann: is this the final title? 😅

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
+r"""Active volume
+=================
+
+In this demo, we will explore the concept of active volume of a quantum computation, and how compilation can exploit this concept to reduce the resources required to execute the computation on a suitable machine, an active volume quantum computer.
+We will look at circuits, ZX diagrams, and a new type of circuit representation called logical networks.
```
Qottmann:
```suggestion
We will look at circuits, :doc:`ZX diagrams <demos/tutorial_zx_calculus>`, and a new type of circuit representation called logical networks.
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+metric that expresses the combined cost of all resources used by a program. One such metric is
+the **spacetime volume** cost, which in fault-tolerant architectures can be understood as the
+total error-corrected qubits (space) taken up by the computation, times the total error
+correction cycles (time) required to perform it.
+
+To understand the impact of different quantum architectures and compilation strategies, we’ll
+often put an *algorithmic cost* metric (e.g. circuit volume) in relation to its
+*implementation cost* on hardware (i.e. the spacetime volume). For instance, the strategies
+described by Litinski’s :doc:`Game of Surface Codes <demos/tutorial_game_of_surface_codes>`
+incur a spacetime volume cost of roughly twice the circuit volume.
```
Qottmann: I would suggest to remove or move this paragraph, it's confusing at this point in the demo

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+(in terms of so-called “logical blocks”), and the systematic compilation framework introduced by
+Litinski and Nickerson [#Litinski2022]_, which then can be used to maximize the use of available
+computational resources. This framework combines the language of the ZX calculus with an
+abstraction of topologically error-corrected quantum computers, as well as the
+inherently-quantum technique of state teleportation.
+The result is a powerful approach to make the cost of quantum computations proportional to
+the active volume rather than the circuit volume, provided that we operate with a so-called
+**active volume computer**. We will not go into detail about the definition of such a computer,
+but emphasize that the compilation framework presented in this demo is specifically
+tailored to this type of quantum computer.
```
Qottmann: sounds very mysterious, is it that complicated to summarize (informally) what makes a qc an active volume qc?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+
+Without further a-do, let's get started!
+
+The input: a quantum circuit
+----------------------------
+
+The active volume framework can compile arbitrary quantum circuits, which we consider to
+be the input; Clifford gates are supported directly, as we will see for our example below.
+Discrete non-Clifford gates such as ``T`` or ``Toffoli`` are implemented in a surface
+code-corrected quantum computer by ingesting magic states via multi-qubit measurements.
```
Qottmann:
```suggestion
code corrected quantum computer by ingesting magic states via multi-qubit measurements.
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+And continuous non-Clifford gates such as arbitrary-angle Pauli product rotations (PPRs)
+are supported via standard discretization techniques like Gridsynth, repeat-until-success
+circuits, channel mixing, or phase gradient decompositions.
```
Qottmann: add links :)

Qottmann: gridsynth: https://docs.pennylane.ai/en/stable/code/api/pennylane.clifford_t_decomposition.html
RUS just the paper I guess
same of channel mixing,
phase gradient: https://pennylane.ai/compilation/phase-gradient

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+Logical network for the CNOT ladder
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Looking at the oriented ZX diagram of the CNOT ladder above, we see that the inserted spiders
+with two legs are connected to the original three-legged spiders with edges that do not satisfy
+the modified rule (4'.) yet. However, in each case, we may simply use the opposite port of those
+two-legged spiders. Changing the shape of the vertices, labeling them, and replacing drawn edges
+with vertex labels then leads us to a valid logical network:
+
+.. figure:: _static/demonstration_assets/active_volume/example_logical_network.png
```
Qottmann: this needs more explanation: what are the numbers? what is shown here in general? Are we avoiding connections for the sake of brevity/readability, or are there no connections anymore?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+   by N, U or E accordingly.
+#. Input (output) legs must connect to the down (up) port of a vertex.
+#. Edges (internal legs) must connect to ports at both vertices from the same of the three pairs.
+   That is, they must connect to the same port of the two vertices they connect, or to
+   opposite ports.
+#. Edges (internal legs) must connect vertices of the same type (X/Z) and same orientation, or
+   of different types and different orientations. (For Hadamarded edges, same (different) types
+   with different (same) orientations must be connected.)
+
+These rules may seem quite artificial and confusing. This is because they impose constraints
```
Qottmann: generaly more visualization aid would be helpful here

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+And this already concludes the compilation process of this simple Clifford operation, arriving at
+a fully parallelized computation. You may ask what we do with the logical network representation,
+and why we would prefer twelve custom blocks--with new complicated rules governing them--over a
+good old circuit diagram with three CNOTs. We turn to those questions next.
+
+Why logical networks?
+---------------------
+
+The properties of logical blocks and the set of rules we impose on logical networks are made
+such that any valid logical network can be realized with surface code-corrected logical qubits,
```
Qottmann: same as above, the hyphen looks off to me

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+While this may seem counterintuitive, it reproduces the footprint of a CNOT implemented via
+lattice surgery; there, we need to bring both the X and the Z boundaries of the two surface-code
+patches that encode the input qubits into contact, leading to a 2-by-2 square of patches overall.
+
+For the example of the CNOT ladder, we saw how a sequence of three Clifford gates can be
+transformed into a layer of simultaneously applied logical blocks, i.e., we **parallelized** the
+computation even though the individual operations **do not commute**.
+This is possible because the logical network compilation combines all parts of the input circuit
+into a single effect, creating a “CNOT ladder subroutine”, rather than keeping the three CNOTs
+as individual building blocks that must be executed in sequence.
```
Qottmann: is this really what is happening? If I treated the 4 qubit CNOT latter as a single unitary matrix and resynthesize it, I would also treat it as a CNOT ladder subroutine, but this doesnt offer me any time gains. I'd argue we still have individual building blocks, but those are the Z and X spiders and they have more freedom to move in time and space.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+While logical network synthesis allows us to parallelize and condense subroutines as much as
+possible, we will usually not want to compile a large-scale algorithm as one monolithic network,
+simply because this would not remain tractable. Instead, commonly used subroutines are compiled
+once and the full computation is encoded by composing their logical networks.
+To do this without re-introducing idle volume, we need to arrange the logical networks on the
+active volume computer such that they fill up the computational qubits at each time step.
+This can be done by procedurally parallelizing networks via state teleportation, and by splitting
+networks into multiple pieces that are executed sequentially to make them fit into the available
+computational space. We describe teleportation in more detail below but will not dwell on the
+splitting procedure here.
```
Qottmann: this is really hard to follow without any visuals 😿

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+simultaneously, without violating any physical laws or causal dependencies.
+
+Practical usefulness
+~~~~~~~~~~~~~~~~~~~~
+
+The previous discussion shows that it is *possible* to parallelize non-commuting gates at the cost
+of two additional qubits (we need two copies of the original qubit, plus a communication channel
+"back in time"), a Bell state preparation and Bell basis measurement (we need to entangle the
+communication channel), and classical compute (we need to transform the correction gates
+by :math:`\text{Ad}_B`). However, it is not so clear whether this is a useful tradeoff.
```
Qottmann: I dont think many readers will know what that means 😁 no need for it here

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+expensive to come by.
+
+For the surface code corrected quantum computer, and specifically for the active volume computer,
+the creation of qubit pairs in a Bell state and the Bell basis measurement are assumed to be
+much cheaper than arbitrary logical operations, because they can just be woven into the
+measurements, i.e., the code cycles, of the error correction code via lattice surgery.
+Similarly, Pauli correction gates are anyway tracked in software throughout, so that the only
+true price we are paying for the parallelization is the extra memory.
+As we are trying to condense a computation by parallelizing it and reducing idle volume, this
+tradeoff will often be beneficial.
```
Qottmann: does this imply we can use the idle volume for the parallelization tricks above? I.e., are the paralellization tricks used with _dirty_ qubits? (most idle regions are "dirty" in the diagram as far as I can tell)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+
+.. figure:: _static/demonstration_assets/active_volume/post-processed_measurement.png
+    :align: center
+    :width: 15%
+    :target: javascript:void(0)
+
+The Bell state measurements determine whether Pauli corrections need to be applied, which in
+turn affects the outcome (and only the outcome) of other Bell state measurements, which
+determines whether Pauli corrections need to be applied, and so on. But since we’re tracking
+Pauli corrections in software [cite pauli frame tracker], there is no immediate need to do
```
Qottmann: todo

Qottmann (changes requested):
Very cool demo! I really like the approach of explaining the compilation with a concrete example, makes it so much easier to follow!

I feel like the introduction could be significantly reduced, as it hints at a lot of things that at this stage make it more confusing than helpful (in particular when talking about new concepts are that are not yet introduced).

It's really good to have a lot of visual representations explaining the concepts rather intuitively. Sometimes these could be used more generously (i.e.: wouldnt hurt to add _even more_ more visualizations! 😎 )

I read the paper a while ago so I already forgot a lot of things, so saw this with relatively fresh eyes. I highlight the many points of confusion I encountered (which was a lot given the complexity of the topic)

I think logical blocks are not introduced anywhere? Its sometimes confusing what is being referred to (blocks, vertices, qubits,..?)
Its a bit weird that we're talking about an active volume quantum computer but dont say what that refers to. I'd either informally explain it, or not use the term. For the latter, from reading the demo it is not clear that I need any special properties of my QC to be able to perform all these operations _in principle_. Are there fundamental restrictions? Or is it just a matter of having the necessary software framework to create directed ZX diagrams, connect them in the right manner and track measuremtn dependent corrections?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+.. figure:: _static/demonstration_assets/active_volume/example_logical_network.png
+    :align: center
+    :width: 95%
+    :target: javascript:void(0)
+
+As we can see, we require only :math:`12` qubits to realize the network of :math:`12` blocks,
+removing the idle memory blocks. This is because the logical network compilation already made
+use of simplifications that could be made to the naively composed network.
+
+We see that logical network compilation achieves parallelization without the need for bridge
```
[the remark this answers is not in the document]
dime10: See my previous comment, but Path A is not a viable option, i.e. it does not represent an executable program. The options are:

1) compile each instruction to logical blocks (the native language of the computer), then schedule the block networks on the computer's resources (while maximizing utilization)
    - My read is that the first part, generating logical blocks from gates or other high-level instructions, is actually not done by an automated compiler (the one will run on each user program), but in advance for a fixed set of known instructions/subroutines.
    - The second part is though, i.e. for a given sequence of operations, insert the known logical networks for each operation, and then schedule them optimally.
2) compile an entire instruction sequence to a single logical network

And in practice you will most likely do both, using 1) as the default compilation mechanism or fallback, as it's unreasonable to compile an entire program end-to-end into a single logical network.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+We also assume that we can track the impact that :math:`B` has on Pauli operators, i.e., what
+happens to a Pauli operator when pulling it through :math:`B`. For our use case, this will be
+easily satisfied, because :math:`B` will be a Clifford operation that simply turns Paulis into
+new Pauli operators.
+
+For the parallelization, we first insert a state teleportation circuit between :math:`A` and
+:math:`B` using a pair of auxiliary qubits prepared in an entangled Bell state, see the
+:doc:`demo on state teleportation <tutorial_teleportation>` for details:
+
+.. figure:: _static/demonstration_assets/active_volume/teleportation_insert_teleport.png
```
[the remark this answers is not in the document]
dime10: A bell state, we can add that to the text :)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+    :target: javascript:void(0)
+
+We also assume that we can track the impact that :math:`B` has on Pauli operators, i.e., what
+happens to a Pauli operator when pulling it through :math:`B`. For our use case, this will be
+easily satisfied, because :math:`B` will be a Clifford operation that simply turns Paulis into
+new Pauli operators.
+
+For the parallelization, we first insert a state teleportation circuit between :math:`A` and
+:math:`B` using a pair of auxiliary qubits prepared in an entangled Bell state, see the
+:doc:`demo on state teleportation <tutorial_teleportation>` for details:
```
dime10: A bit unfortunate we already have B as an operation label 😅 
```suggestion
:math:`B` using a pair of auxiliary qubits prepared in an entangled Bell state
:math:`|\mathfrak{B}>`, see the :doc:`demo on state teleportation <tutorial_teleportation>`
for details:
```

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
Qottmann: I meant it more like this is how the current paragraph/section reads and it should be updated 😄

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+We will look at circuits, ZX diagrams, and a new type of circuit representation called logical networks.
+This demo is directly based on a seminal paper by Litinski and Nickerson [#Litinski2022]_. As the original work already does a great job at presenting the concepts and compilation techniques at multiple levels of detail and with plenty of visualizations, we will aim to take a complementary perspective, and walk slowly through a specific compilation example.
+
+For readers unfamiliar with ZX diagrams, it will be useful (but not necessary) to take a look at
+our :doc:`introduction to ZX calculus <demos/tutorial_zx_calculus>` first.
+Further, it can be instructive to get a :doc:`primer on lattice surgery <demos/tutorial_lattice_surgery>`
+and dive into the :doc:`Game of Surface Codes <demos/tutorial_game_of_surface_codes>`, but this
+is not a requirement either.
+
+.. figure:: _static/demo_thumbnails/large_demo_thumbnails/pennylane-demo-active-volume-large-thumbnail.png
```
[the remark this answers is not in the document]
dwierichs: mhmhm. Not sure what's best here. Maybe just some "meta" caption like 
"Only a portion of the spacetime volume occupied by a quantum circuit contributes to the actual computation, symbolized by active loudspeakers. The remaining volume sits idle without playing a crucial part."
What do you think? We could also skip the hero image.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+numerous key metrics of the quantum program during the compilation process. Typically, these
+metrics focus on specific properties of the program such as qubit count, and various gate counts,
+which often stand in contention against each other. To optimize effectively, we’d ideally like a
+metric that expresses the combined cost of all resources used by a program. One such metric is
+the **spacetime volume** cost, which in fault-tolerant architectures can be understood as the
+total error-corrected qubits (space) taken up by the computation, times the total error
+correction cycles (time) required to perform it.
+
+To understand the impact of different quantum architectures and compilation strategies, we’ll
+often put an *algorithmic cost* metric (e.g. circuit volume) in relation to its
```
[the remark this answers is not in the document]
dwierichs: I restructured this paragraph a bit so that the examples are given only in the sentence mentioning GoSC. This created a nice transition into the next paragraph where circuit volume is then explained :) Let me know what you think :)

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: Yeah, I think we need to cut some things off, and the idea for this demo was to be complementary. The paper does a great job of providing the abstract AV computer and putting the resulting cost model into context of the compilation. We could try and provide a more compact "primer on AV computer" that sets the stage and cost model. 🤔

Qottmann: Since it is meant to be complementary in the first place, we wouldnt require a full and detailed description of the AV computer. But some informal and visual primer on how computation is performed would be useful in order to understand the parallelization and utilization of idling circuit regions

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
Qottmann: Ah okay. I wouldnt skip it, and now that I think about it, also wouldnt put a caption. This may sound trivial but what about removing the blue background? I think this is what threw me off as the rectangular shape makes it look more like a graph/figure 😅 

Alternatively: put a "Introduction" header after the image to make the separation clear?

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
dwierichs: I'm confused by the confusion :D 
This section is basically putting the explicit teleportation into context with what the holistic synthesis of the full ladder does. As such, it aims at discussing the question "If I can teleport explicitly, why would I synthesize logical networks for larger chunks of a circuit beyond 1- and 2-qubit gates?". 

 I'll try to make more explicit that we do not have any use for a circuit diagram as output in this section, and that we want to discuss said question.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+If you are familiar with the Game of Surface Codes paper referenced earlier, you may wonder
+how the scheme presented there differs from what has been presented here. After all, Litinski
+already used teleportation back then to parallelize Pauli circuits and achieve reaction-limited
+computation (also referred to as “Fowler’s time-optimal scheme”).
+Fundamentally, teleportation always trades (execution) time for (memory) space. The crucial
+difference, intuitively, is that Litinski’s earlier techniques provide the additional space
+in the trade-off as brand new qubits, since each operation is in principle, and through
+techniques presented in the paper, assumed to take up the full width of a circuit. Meanwhile,
+Active Volume computation tries to reuse already available, but idle, qubit space at a
+fine-grained level, and to thus maximize the *efficiency* of the computer.
```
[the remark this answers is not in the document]
dwierichs: @dime10 wanna make a suggestion for this? :)

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
dwierichs: Replaced it with the opengraph version, which has white background :)

dwierichs:
Thanks for the thorough review @Qottmann 
I tried to address everything, but given that some parts received quite some restructuring/overhaul, I'd expect some things to maybe need some further polishing/iteration. However, it would be useful to get your feedback on the changes I made before polishing further :) 

> I feel like the introduction could be significantly reduced, as it hints at a lot of things that at this stage make it more confusing than helpful (in particular when talking about new concepts are that are not yet introduced).

This is not addressed yet. Maybe the other sections can settle a bit and then the introduction can be adjusted to that :)

> It's really good to have a lot of visual representations explaining the concepts rather intuitively. Sometimes these could be used more generously (i.e.: wouldnt hurt to add even more more visualizations! 😎 )

I added a bit more, but not crazy much. 

> I think logical blocks are not introduced anywhere? 

They are introduced, when formally introducing logical networks after oriented ZX diagrams. I tried to be more clear about it being something that will be introduced later whenever they are referenced before.

> Its sometimes confusing what is being referred to (blocks, vertices, qubits,..?)

I removed almost all uses of "vertex" and replaced them by "spider" instead. I hope that helps.

> Its a bit weird that we're talking about an active volume quantum computer but dont say what that refers to. I'd either informally explain it, or not use the term. 

I added an info box :) 

> For the latter, from reading the demo it is not clear that I need any special properties of my QC to be able to perform all these operations in principle. Are there fundamental restrictions? 

The AV computer abstraction is crucial for the cost model. We could just pass a circuit to a QC and have it execute that gate-by-gate, but AV compilation wants to compress the computation, lowering the cost compared to this baseline approach. In order to form a cost model that tells us whether we actually make something cheaper, we need to assume certain operations to be super cheap compared to others. Those assumptions then are the "AV computer". With the info box added and referenced throughout the demo, I hope that these points come across :)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+Further, it can be instructive to get a :doc:`primer on lattice surgery <demos/tutorial_lattice_surgery>`
+and dive into the :doc:`Game of Surface Codes <demos/tutorial_game_of_surface_codes>`, but this
+is not a requirement either.
+
+.. figure:: _static/demo_thumbnails/opengraph_demo_thumbnails/pennylane-demo-active-volume-open-graph.png
+    :align: center
+    :width: 65%
+    :target: javascript:void(0)
+
+:doc:`Quantum compilation <compilation/>` is fundamentally about bridging the
```
Qottmann: Add heading "Introduction" to separate it from the hero image
(optional)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+In addition to the bare translation between those representations, it also aims to optimize
+numerous key metrics of the quantum program during the compilation process. Typically, these
+metrics focus on specific properties of the program such as qubit count, and various gate counts,
+which often stand in contention against each other. To optimize effectively, we’d ideally like a
+metric that expresses the combined cost of all resources used by a program. One such metric is
+the **spacetime volume** cost, which in fault-tolerant architectures can be understood as the
+total error-corrected qubits (space) taken up by the computation, times the total error
+correction cycles (time) required to perform it.
+
+Intuitively, the **circuit volume** can be understood as the total “area” taken up by a
```
Qottmann: much better 👨‍🍳 👌

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+Now we can use the U and D ports of the central vertices to route the last edge. We choose
+both U ports, because this will be a convenient choice below, and arrive at the following
+oriented ZX diagram for a single CNOT:
+
+.. figure:: _static/demonstration_assets/active_volume/cnot_oriented_zx_complete.png
+    :align: center
+    :width: 50%
+    :target: javascript:void(0)
+
+Note that we could also connect the input state :math:`|b\rangle` to the D port of the left X spider.
```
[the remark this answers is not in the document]
Qottmann: see below

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+Oriented ZX diagram to logical network
+--------------------------------------
+
+The last step required to arrive at logical networks is comparatively small; the major effort
+is already done with the orientation of the ZX diagram.
+
+We introduce **logical blocks**, which are just the nodes from an oriented ZX diagram,
+drawn as hexagons to mark the six ports more clearly, with a modification of rule (4.) above:
+
+    4'. Edges (internal legs) must connect to **the same port** at both logical blocks.
```
Qottmann: It's still unclear to me what a logical block is 😬 Isnt rule 4' describing a logical block? What is meant that edges must connect to the same port at **both** blocks?

Like, is a logical block one of the hexagonal spiders? or a collection of them?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+and are considered expensive, from the block connectivity, which represents communication between
+blocks and is considered cheaper because it is implemented via fast SWAP connections (also see
+active volume computer info box).
+
+Logical network for the CNOT ladder
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Looking at the oriented ZX diagram of the CNOT ladder above, we see that the inserted spiders
+with two legs are connected to the original three-legged spiders with edges that do not satisfy
+the modified rule (4'.) yet. However, in each case, we may simply use the opposite port of those
```
Qottmann: would go a long to way to visualize this and be able to _see_ it, perhaps because I dont understand what logical blocks are, yet, but its unclear to me how rule 4' is violated above

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+two-legged spiders. We then arrive at a valid logical network by performing a few cosmetic
+changes: first, change the shape of the spiders to hexagons, with each corner corresponding
+to a port. Second, enumerate the hexagons for referencing. Any labels work, really. Third and last,
+cut the edges between hexagons and instead draw shortened legs, annotated by the label of the
+hexagon they connect to.
+For our example, we find the logical network
+
+.. figure:: _static/demonstration_assets/active_volume/example_logical_network.png
+    :align: center
```
Qottmann: optional: add connected version with messy new connections that then simplify to the shown logical network diagram

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
dime10: Yeah there is definitely a misunderstanding here. Both are doing AV compilation, the difference is that one is compiled as a single instruction (essentially I'm adding a "3-CNOT ladder" to the list of instructions I know how to directly execute on my computer), while the other looks at it as a sequence of 3 instructions, each of which I know (individually) how to execute on the computer. 
All programs consistent consist of instruction sequences, so I always need to know how to deal with those. The simplest way is to schedule them sequentially, but of course that is inefficient, so AV computation uses teleportation parallelization to schedule as many instructions simultaneously as possible.
It makes sense that there is an overhead to the "generic" approach, I think that is almost universally true. The difference is that one has an instruction set of say `{CNOT, H, S, T}`, while the other needs `{3-CNOT-ladder, CNOT, H, S, T}`. Bigger instruction sets will usually produce faster results, but of course it raises the complexity and you can't put all possible unitaries in there. 
Section 6 in the paper has a few notes on this concept as well.

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: > Like, is a logical block one of the hexagonal spiders? or a collection of them?

One logical block is really just one ZX spider from an oriented ZX diagram to which we apply the small modification from rule 4. to rule 4'. :) I'll rephrase to make the 1-to-1 correspondence explicit.

> What is meant that edges must connect to the same port at both blocks?

In 4., edges must connect to the same or the opposite ports of the two blocks they connect. In 4'., opposite ports are no longer good enough, the edge must connect the same port of both blocks. Would this be clearer ?

```suggestion
We introduce **logical blocks**, which are just the nodes from an oriented ZX diagram,
drawn as hexagons to mark the six ports more clearly, with a modification of rule (4.) above:

    4'. Edges (internal legs) must attach to **the same ports** at the two logical blocks they connect, e.g. N to N or W to W.
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+networks that resembles quantum circuits ensures that we do not venture too far into the space
+of abstract representations, as could happen with pure ZX diagrams.
+
+Under the hood of parallelization: state teleportation
+------------------------------------------------------
+
+In the CNOT ladder example circuit, which has a fundamentally sequential appearance in the
+circuit picture, logical network compilation achieved its parallelization by combining the
+effects of the three CNOTs into a single effect, creating a “CNOT ladder subroutine”.
+If we want to parallelize multiple blocks,
```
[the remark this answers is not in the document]
dwierichs: Oh yes, excellent catch!

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+
+:doc:`Quantum compilation <compilation/>` is fundamentally about bridging the
+gap between high-level descriptions of
+quantum algorithms, and low-level instructions that are actually executable on quantum hardware.
+In addition to the bare translation between those representations, it also aims to optimize
+numerous key metrics of the quantum program during the compilation process. Typically, these
+metrics focus on specific properties of the program such as qubit count, and various gate counts,
+which often stand in contention against each other. To optimize effectively, we’d ideally like a
+metric that expresses the combined cost of all resources used by a program. One such metric is
+the **spacetime volume** cost, which in fault-tolerant architectures can be understood as the
```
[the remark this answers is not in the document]
dwierichs: Yeah, I agree with @dime10 here. It may seem like some detail but because the AV compilation is tied so specifically to the AV computer abstraction, it really makes sense to differentiate between algorithmic volume and implementation volume, imo.

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: Added a figure redlining the problems, and a version just after the next sentence where it's fixed :)

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: See above.

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
dime10: Referenced the figure and attempted slight reformulation here: https://github.com/PennyLaneAI/qml/pull/1618/commits/e9ffe83238966948c70aee7b98a320139ff06e4c

demonstrations_v2/tutorial_active_volume/demo.py
dime10: https://github.com/PennyLaneAI/qml/pull/1618/commits/a7a25717a5ddf877bb5dfacebe2683d45f712c5a

This would be a nice addition to the compilation hub I think :)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+CNOT is a simple two-qubit Clifford gate. However, note that the active volume computer
+architecture operates with :doc:`lattice surgery <demos/tutorial_lattice_surgery>`, and
+executing a CNOT with lattice surgery also requires two additional logical qubits as intermediate
+bridge space. Thus, the oriented ZX diagram simply reproduces the true cost for a CNOT in this
+type of architecture.
+
+Note that we could also connect the input state :math:`|b\rangle` to the D port of the left X spider.
+This is because we could have inserted the additional blue spider on the output leg of the existing
+one, rather than between the inserted orange and the existing blue spiders.
+For this, the orientation of the spiders could be kept as-is.
```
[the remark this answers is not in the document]
Qottmann: perfect, thanks 👌

demonstrations_v2/tutorial_active_volume/demo.py
Qottmann: Ah! So a logical block is always just a single spider?
Then perhaps simply:
> 4'. Edges (internal legs) to other logical blocks must attach to **the same ports**.

I just found it confusing that we are talking about a single logical block and then say something about "both blocks"

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
dwierichs: I re-inserted the modified version of the paragraph that I had done before removing it. Let me know whether y'all are happy with that :D 
https://github.com/PennyLaneAI/qml/pull/1618/commits/c7c6d6a8580372afc28d3b932b2decd15ebe670b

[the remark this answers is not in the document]
Qottmann: I'm still a bit thrown by it at the point it is currently put. For me personally, it'd make more sense to put this _after_ defining circuit, idle, and active volume. This can be combined with bringing the following point home: That AV compilation is way to reduce the _implementation cost_ to the active volume (of the circuit?). IMO this is not stressed / put clearly enough right now in the intro.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+effects of the three CNOTs into a single effect, creating a “CNOT ladder subroutine”.
+If we want to parallelize multiple subroutines, however,
+without having to re-compile their joint logical network, we can do this procedurally through
+state teleportation, using a so-called bridge qubit. This technique parallelizes non-commuting
+operations without breaking physics, and it pays off because Bell state preparation and
+measurements are assumed to be fast on an active volume computer (see info box at the top).
+
+We will showcase this type of parallelization without resynthesizing the logical network(s) at the
+circuit level for convenience, and will then compare the resulting "procedurally parallelized"
+networks for the 3 CNOTs to the monolithic network that we obtained from synthesis earlier.
```
Qottmann:
```suggestion
We will showcase this type of parallelization by treating the three CNOT gates in our ladder example as three individual sub-routines. For convenience, we do this at the circuit level to showcase the state teleportation. We then compare the resulting "procedurally parallelized" circuit for the 3 CNOTs to the monolithic network that we obtained from synthesis earlier.
```

Qottmann (commented):
Looking really good now! Thanks for the additional visualizations, I think they are top notch 👌 

Thanks also for your patience and detailed explainers, they helped a lot to properly understand the concepts and how you were trying to convey them.
I think not all bits of the demo properly convey them the way you intended to. I dared make some explicit suggestions that I think could help make them. Feel free to ignore them if you disagree 🙃 

I wonder if there is a good way to make it clear that a logical **block** is really just a spider. In my mind I immediately assumed it could also be a collection of spiders (with the extra rules), which seems not to be the case? I made a suggestion in its intro that _might_ do so, but again feel free to ignore if you disagree.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+    :align: center
+    :width: 65%
+    :target: javascript:void(0)
+
+Introduction
+------------
+
+`Quantum compilation <https://pennylane.ai/compilation/>`__ is fundamentally about bridging the
+gap between high-level descriptions of
+quantum algorithms, and low-level instructions that are actually executable on quantum hardware.
```
daniela-angulo:
```suggestion
quantum algorithms and low-level instructions that are actually executable on quantum hardware.
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+        :align: center
+        :width: 65%
+        :target: javascript:void(0)
+
+    The following characteristics are crucial assumptions about the computer that shape the cost
+    model:
+
+    #. The information content of two qubit modules within range can be
+       exchanged quickly/cheaply. They are "quickswappable".
+    #. Single qubit modules can be prepared quickly/cheaply in the state :math:`|0\rangle` or :math:`|+\rangle`.
```
daniela-angulo:
```suggestion
    #. Single-qubit modules can be prepared quickly/cheaply in the state :math:`|0\rangle` or :math:`|+\rangle`.
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+       They can also be measured quickly/cheaply in the Pauli-:math:`X` or Pauli-:math:`Z` basis.
+    #. Two qubit modules within some specified range :math:`r` can be prepared quickly/cheaply in the Bell
```
daniela-angulo:
```suggestion
    #. Two-qubit modules within some specified range :math:`r` can be prepared quickly/cheaply in the Bell
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+       state :math:`|\phi\rangle = \tfrac{1}{\sqrt{2}} (|0\rangle +|1\rangle)`.
```
daniela-angulo: if they are two qubits, why is that "Bell state" written with a single one?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+
+    Given the characterization above, active volume compilation aims to maximize the utilization
+    of the computational qubit modules, because they execute the slow operations (logical blocks
+    and magic state distillation). Memory qubit modules are allowed to sit idle, providing
+    intermediate storage and space for routing information through quickswaps. Note that
+    long-range communication in the active volume computer is achieved through layers
+    of quickswaps, which remain cheap in the above cost model.
+
+In the following, we will demonstrate how to compile the circuit of a subroutine
+step-by-step into a so-called **logical network**, which is a representation that allows
```
daniela-angulo:
```suggestion
step by step into a so-called **logical network**, which is a representation that allows
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+Oriented ZX diagram to logical network
+--------------------------------------
+
+The last step required to arrive at logical networks is comparatively small; the major effort
+is already done with the orientation of the ZX diagram.
+
+We introduce **logical blocks**, which are just the nodes from an oriented ZX diagram,
+drawn as hexagons to mark the six ports more clearly, with a modification of rule (4.) above:
+
+    4'. Edges between two logical blocks must connect to **the same port** of both blocks.
```
daniela-angulo: this indented block renders a bit funny. It is ok if you want to leave it as is, it just doesn't look great in my opinion.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+    Image source: Daniel Litinski [#Litinski2018]_.
+
+This is where we quickly run into issues, as Clifford corrections are generally already considered
+too complex to be tracked classically. That means that we’ll have to physically implement them,
+and it is at this point that we require a concrete value of the measurement result (accounting
+for any prior Pauli corrections). In this sense, the Clifford corrections impose a limit on the
+ability to parallelize, as part of the computation will still need to happen sequentially.
+
+Doing corrections in-line as depicted in the above circuits has been shown to be inefficient
+actually, so a key optimization is to move the correction away from the data qubits and onto
```
daniela-angulo:
```suggestion
Doing corrections in-line as depicted in the above circuits has actually been shown to be 
inefficient, so a key optimization is to move the correction away from the data qubits and onto
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,772 @@
...
+
+If you are familiar with the Game of Surface Codes paper referenced earlier, you may wonder
+how the scheme described there differs from what has been presented here. After all, Litinski
+already used teleportation then (cf. fig. 25 in [#Litinski2018]_) to parallelize Pauli circuits
+and achieve reaction-limited computation (also referred to as “Fowler's time-optimal scheme”).
+Fundamentally, parallelization via teleportation always trades (execution) time for (memory) space.
+The crucial difference, intuitively, is that Litinski's earlier techniques have to provide the
+additional space as brand new qubits, since each operation is assumed to in principle span the full
+width of a circuit (a characteristic that is reinforced by the techniques presented in the paper).
+Parallelization thus happens on the *layer structure*, where whole circuit layers are executed
```
daniela-angulo:
```suggestion
Parallelization thus happens on the *layer structure*, where entire circuit layers are executed
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,771 @@
...
+In this demo, we will look at how to obtain the active volume of a quantum circuit
+(in terms of so-called “logical blocks”), and the systematic compilation framework introduced by
+Litinski and Nickerson [#Litinski2022]_, which then can be used to maximize the use of available
+computational resources. This framework combines the language of the ZX calculus with an
+abstraction of topologically error-corrected quantum computers, as well as the
+inherently-quantum technique of state teleportation.
+The result is a powerful approach to make the cost of quantum computations proportional to
+the active volume rather than the circuit volume, provided that we operate with a so-called
+**active volume computer**. We will not go into detail about the definition of such a computer,
+but only outline its rough characteristics.
```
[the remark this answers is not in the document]
dwierichs: Nice, I like it!

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,771 @@
...
+    :target: javascript:void(0)
+
+This is a valid oriented ZX diagram for a single CNOT gate. Note that the spiders we had
+to insert above directly correspond to additional information processing steps, and will take up
+additional qubits in the execution of the CNOT. This may come as a surprise, given that
+CNOT is a simple two-qubit Clifford gate. However, note that the active volume computer
+architecture operates with :doc:`lattice surgery <demos/tutorial_lattice_surgery>`, and
+executing a CNOT with lattice surgery also requires two additional logical qubits as intermediate
+bridge space. Thus, the oriented ZX diagram simply reproduces the true cost for a CNOT in this
+type of architecture.
```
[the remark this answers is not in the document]
dwierichs: I think I'd just ignore this fact, because one qubit for two time steps and two qubits for one time step costs the same volume, and the latter is more in line with the "full parallelization into a single time step" that logical network synthesis pursues.

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: Very nice, thanks! Used an adapted version of this 👍

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,774 @@
...
+As we can see, the state teleportation together with the ability to track correction gates through
+:math:`B` makes it possible to implement the two non-commuting operations :math:`A` and :math:`B`
+simultaneously, without violating any physical laws or causal dependencies.
+
+Practical usefulness
+~~~~~~~~~~~~~~~~~~~~
+
+The previous discussion shows that it is *possible* to parallelize non-commuting gates at the cost
+of two additional qubits (we need two copies of the original qubit, plus a communication channel
+"back in time"), a Bell state preparation and Bell basis measurement (we need to entangle the
```
Qottmann: What is this communication channel "back in time" in the diagram? Is it the 2 Bell state qubits? Or the classical control (or both?) and why "back in time"?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,712 @@
...
+deformations where "rigid" quantum circuits do not, but the additional structure of logical
+networks that resembles quantum circuits ensures that we do not venture too far into the space
+of abstract representations, as could happen with pure ZX diagrams.
+
+Under the hood of parallelization: state teleportation
+------------------------------------------------------
+
+In the CNOT ladder example circuit, which has a fundamentally sequential appearance in the
+circuit picture, logical network compilation achieved its parallelization by combining the
+effects of the three CNOTs into a single effect, creating a “CNOT ladder subroutine”.
```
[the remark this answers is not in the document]
dwierichs: I fully agree that ZX diagrams are the underlying tool that unlock the parallelization without introducing manual teleportation routines. But as the demo is focusing on the active volume compilation, which uses the ZX-diagrams, I'd rather stress that the logical network synthesis achieves this. If a reader thinks of ZX-calculus as the sota compilation technique, this may sound a little off, but coming from GoSC or standard circuit compilation, I think this is the nicer comparison.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,774 @@
...
+The Bell state measurements determine whether Pauli corrections need to be applied, which in
+turn affects the outcome (and only the outcome) of other Bell state measurements, which
+determines whether Pauli corrections need to be applied, and so on. But since we’re tracking
+Pauli corrections in software [#PauliFrame], there is no immediate need to do anything with the
+Bell measurement results, and we can happily continue to apply operations in parallel or any
+other order without worrying about data dependencies between them. So do we *ever* need to affect
+the quantum state using the information from measurements and Pauli corrections?
+
+To answer this question, we need to consider non-Clifford operations. As mentioned further up,
+these are implemented via magic state injection. As a general rule, operations implemented via
```
Qottmann:
```suggestion
these are implemented via `magic state injection <https://pennylane.ai/qml/glossary/what-are-magic-states>`__. As a general rule, operations implemented via
```

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: This was left over from a previous styling :) Thanks!

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+    :target: javascript:void(0)
+
+We also assume that we can track the impact that :math:`B` has on Pauli operators, i.e., what
+happens to a Pauli operator when pulling it through :math:`B`. For our use case, this will be
+easily satisfied, because :math:`B` will be a Clifford operation that simply turns Paulis into
+new Pauli operators.
+
+For the parallelization, we first insert a state teleportation circuit between :math:`A` and
+:math:`B` using a pair of auxiliary qubits prepared in an entangled Bell state
+:math:`|\phi\rangle=\tfrac{1}{\sqrt{2}}(|0\rangle +|1\rangle)`, see the
```
daniela-angulo:
```suggestion
:math:`|\phi\rangle = \tfrac{1}{\sqrt{2}} (|00\rangle +|11\rangle)`, see the
```

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
Qottmann: Good point! Now that I understand it better, I'm a bit confused (lol) by the first method (logical network compilation): Is this completely circumventing the need for classically controlled corrections (that we get from teleportation), or are those just swept under the rug and _implicit_ in this technique?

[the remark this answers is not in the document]
Qottmann: As for the statemet itself, up until this point in the demo we have not seen teleportation, and just used the translation to ZX diagrams to enable the parallelism. Perhaps reformulating the sentence to focus on something different than parallelism already resolves the issue of the sentence :)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,774 @@
...
+    :width: 40%
+    :target: javascript:void(0)
+
+We fully parallelized the CNOT ladder, using :math:`8` instead of the original :math:`4` qubits.
+
+Following the approach mentioned above, we now could
+naively plug in the logical network for a single CNOT from earlier. This would lead to yet
+another :math:`3\cdot 2` additional qubits, arriving at :math:`12` logical blocks and two idling
+blocks (blocks with only U and D ports occupied that can be executed by memory modules),
+acting on :math:`14` qubits:
```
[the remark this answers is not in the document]
Qottmann: is that because those things only apply to clifford circuits, and explicitly exclude T gates?

_static/demonstration_assets/active_volume/example-parallelized-explicitly.png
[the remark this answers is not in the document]
dwierichs: Ohhh, you are absolutely right! I think I'd just give those states some labels and state that they are measured in the Bell basis?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+    Active and idle volumes are represented by areas occupied by gates (green)
+    and areas without gates (red), respectively, in a standard circuit diagram.
+    Adapted from [1].
+
+In this demo, we will discuss the active volume compilation framework introduced by
+Litinski and Nickerson [#Litinski2022]_ that allows one to reduce the **implementation cost**
+of an algorithm to being proportional to the active volume of the circuit. This is in contrast to
+the framework of [#Litinski2018]_, where the implementation cost is proportional to the circuit volume.
+The active volume is always less than or equal to the circuit volume, so that this typically
+implies a cost reduction.
```
dime10: Just want to be careful with strong statements, do we know for sure this is true?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+       a slow/expensive operation as well.
+
+    A more detailed characterization is provided in Sec. 1 of [#Litinski2022]_.
+
+    Given the characterization above, active volume compilation aims to maximize the utilization
+    of the computational qubit modules, because they execute the slow operations (logical blocks
+    and magic state distillation). Memory qubit modules are allowed to sit idle, providing
+    intermediate storage and space for routing information through quickswaps. Note that
+    long-range communication in the active volume computer is achieved through layers
+    of quickswaps, which remain cheap in the above cost model.
```
dime10: I think this is slightly clearer given the previous comment?
```suggestion
    arbitrary long-range communication in the active volume computer is achieved through layers
    of quickswaps, which in typically scenarios remain low-depth and thus cheap in the above cost
    model.
```
Technically I'm not sure that the long-range comms are cheap (in the sense that cheap means O(1) instead of O(d)), but I suppose cheap in the sense that they don't slow the computer down (i.e. swap depth < d).

_static/demonstration_assets/active_volume/oriented-zx-rules.png
dime10: Just leaving a note here so that we don't forget to fix the illegal connections in the bottom right of the diagram

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+    :align: center
+    :width: 90%
+    :target: javascript:void(0)
+
+In order to connect the “individual” CNOT networks side-by-side, we have moved some of the input
+qubit legs to what was previously an “internal” node. This is allowed as there is no preference
+for where to place logical inputs and outputs in a network. Note how some of the qubits appear
+to be routed laterally through the network, hinting at the parallelization of what were previously
+sequential gates. As for the single CNOT, this is not a unique solution.
+We are almost done with the parallelization of the CNOT ladder; one last step will be needed to
```
dime10:
```suggestion
We are almost done with the compilation of the CNOT ladder; one last step will be needed to
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+We introduce **logical networks**, which are just oriented ZX-diagrams with spiders
+drawn as hexagons to mark the six ports more clearly, with a modification of rule (4.) above:
+
+4'. Edges between two oriented spiders must connect to **the same port** of both spiders.
+
+From now on, we refer to these hexagonal spiders inside a logical network as **logical blocks**.
+
+Depending on the concrete circuit we are compiling, this change might just lead to minor
+modifications such as relabeling the ports of two connected blocks, or it might lead to cascades of
+re-routing that eventually force us to insert a new block, even.
```
dime10: Oh wow, I know I raised doubts that the process was trivial, but did you end encountering an example that required new blocks even? :o

demonstrations_v2/tutorial_active_volume/demo.py
dwierichs: I thought this is trivially true because "circuit volume = active volume + idle volume" and "idle volume >= 0" 🤔

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+    .. figure:: _static/demonstration_assets/active_volume/active-volume-computer.png
+        :align: center
+        :width: 65%
+        :target: javascript:void(0)
+
+    The following characteristics are crucial assumptions about the computer that
+    determine our cost model:
+
+    #. The information content of two qubit modules within range can be
+       exchanged quickly/cheaply. They are "quickswappable".
```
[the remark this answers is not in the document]
dwierichs: Ah, I actually misunderstood this, I think :/ 
I like your minimalistic edit to fix this! Thanks :)

_static/demonstration_assets/active_volume/oriented-zx-rules.png
dwierichs: Ah, had fixed it and not uploaded the fix :D Thanks 👍

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,656 @@
...
+rotations:
+
+.. figure:: _static/demonstration_assets/active_volume/auto-magic.png
+    :align: center
+    :width: 55%
+    :target: javascript:void(0)
+
+    Auto-corrected non-Clifford Pauli rotation. Image source: Daniel Litinski [#Litinski2018]_.
+
+In essence, we have swapped the Clifford correction for a multiplexed measurement, where the
```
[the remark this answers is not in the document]
dime10: I edited the original GoSC figure to show the newer AV construction to be in-line with the latest hand-drawn figure added by David, hopefully that is allowed 😄 
https://github.com/PennyLaneAI/qml/pull/1618/commits/6af90eea90d4baff4f6657029d3d5c83dde4dea7

demonstrations_v2/tutorial_active_volume/demo.py
dime10: Well, I think that's just the intuitive picture 😅 For hard, numeric comparisons the paper defines AV as the total sum of logical blocks needed, while CV is the product of algo qubits and T gates.
From those definitions it's not immediately obvious to me that one is always less than the other. Actually, I just thought of a counter example which would be a Clifford circuit, which has no T gates but does have an active volume?

_static/demonstration_assets/active_volume/example-parallelized-explicitly.png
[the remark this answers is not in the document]
dwierichs: Did the above :) Let me know whether this works.

demonstrations_v2/tutorial_active_volume/demo.py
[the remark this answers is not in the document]
dime10: > What are both of your opinions on the latest version of this sentence?

Quoting the current version:
> In the CNOT ladder example circuit, which has a fundamentally sequential appearance in the circuit picture, logical network compilation parallelized it by exploiting the lack of time ordering in ZX-diagrams, combining the effects of the three CNOTs into a single effect.

I am happy with this sentence 👍 

However, I am concerned about the wider framing of the comparison, I'm going through it end-to-end again leaving some comments that hopefully could help with that?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,784 @@
...
+and why we would prefer twelve custom blocks--with new complicated rules governing them--over a
+good old circuit diagram with three CNOTs.
+A key advantage of the logical network representation lies in its homogeneous expression of
+computational steps within the active volume computer, c.f. the `info box at the top <AV info box_>`_;
+all computation is made up of logical blocks that can be scheduled densely in the computational
+qubit modules of the computer in order to remove idle volume. If logical networks turn out too
+large for the space available at a given time step, they can be split into two networks that
+are executed sequentially.
+
+For our example circuit, this idea of gradual parallelization generalizes the idea that we are
```
dime10: I can see how these paragraphs below give the impression that logical network compilation exists primarily to parallelize computation and do space-time trade-offs, seemingly rendering teleportation obsolete, so I can see where @Qottmann is coming from. 
I don't know what to with this yet but leaving this comment so that we can revisit this passage.

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,784 @@
...
+We see that logical network compilation achieves parallelization without the need for bridge
+qubits, making the resulting network more memory-efficient than a procedurally parallelized
+network (the additional memory is the bridge qubit that idles throughout the execution of the
+network).
+Thus, even though the number of computational blocks is the same, resynthesizing the composed
+network of three CNOTs leads to a slightly cheaper network in terms of routing. It is noteworthy
+that this kind of optimization is of second-order, because the active volume computer is assumed
+to be very fast at routing (see `info box at the top <AV info box_>`_). However, for larger computations, e.g.,
+already for a Toffoli gate realized via state injection, the number of computational blocks
+itself will be reduced, too.
```
dime10: I think this is a very important point, that unfortunately our example doesn't show. @Qottmann does this help contextualize why both techniques are important/relevant?

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,787 @@
...
+From now on, we refer to these hexagonal spiders inside a logical network as **logical blocks**.
+
+Depending on the concrete circuit we are compiling, this change might just lead to minor
+modifications such as relabeling the ports of two connected blocks, or it might lead to cascades of
+re-routing that eventually force us to insert a new block, even.
+**Logical networks** are diagrams, or networks, of logical blocks. We label the logical blocks
+uniquely and replace drawing edges between them by annotating the ports with the labels of the
+blocks they connect to. This allows us to separate logical blocks, which represent computation
+and are considered expensive, from the block connectivity, which represents communication between
+blocks and is considered cheaper because it is implemented via fast SWAP connections (also see
```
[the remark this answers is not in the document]
dime10: My understanding is that you cannot have block connections that are out of range at all, and I'm not sure how you would resolve this with swap operations. All network connections need to exist simultaneously and remain in place for d code rounds, if you couldn't swap the qubits in range before the network runs, how can you do it during? 🤔 

Litinski says that a range of 12 is sufficient for all subroutines that he compiles in the paper, so I think in practice this is probably not an issue. I don't know what it would look like, but I imagine some sort of large, highly-connected network that you can't lay out neatly in a line with blocks talking to their neighbours? 😅

demonstrations_v2/tutorial_active_volume/demo.py
dime10: I mean, I would love to know the answer to this, and I tried to find statements in the paper, but I don't see anything definitive. Only things like:

> The active volume can be orders of magnitude lower than the circuit volume.

> The active volume is typically much lower than the circuit volume.

> but, typically, the active volume will not scale with the total number of qubits in the computation.

All those statement leave room for uncertainty. Going by the definitions alone (which are admittedly presented very off-handedly), I can't see a guarantee of this fact. Maybe we just follow suit and add a "typically" into the sentence? 😆

demonstrations_v2/tutorial_active_volume/demo.py
Qottmann: yes! though it is a bit of a side sentence (so might easily be missed)

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,784 @@
...
+quantum computer, allowing them to be adjusted to available hardware resources and promoting
+space-time tradeoffs to first-class program transformations.
+
+From a mathematical perspective, logical networks distill the best out of the two worlds of
+quantum circuits and ZX-diagrams; ZX-diagrams allow for continuous
+deformations where "rigid" quantum circuits do not, but the additional structure of logical
+networks that resembles quantum circuits ensures that we obtain instructions that can be executed
+on the computer.
+
+Under the hood of parallelization: state teleportation
```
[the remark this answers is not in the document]
Qottmann: yeah perhaps 
```suggestion
Additional parallelization via state teleportation
```

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,693 @@
...
+Second, logical networks can be parallelized via state teleportation, as we just discussed. While
+this is true for components of a quantum circuit as well, there is a third feature that makes
+parallelization more useful:
+third, logical networks can be divided into multiple networks, which then can be executed in
+sequence. We will not go into detail about this, but refer to Sec. 2 and Fig. 11 of
+[#Litinski2022]_. This allows us to split up a network that otherwise would be too large to be
+executed in parallel with a previous network.
+We may sketch the idea behind the scheduling via splitting and parallelization:
+
+.. figure:: _static/demonstration_assets/active_volume/scheduling.png
```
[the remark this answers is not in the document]
Qottmann: other wise 👨‍🍳 👌 figure

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,693 @@
...
+In the CNOT ladder example circuit, which has a fundamentally sequential appearance in the
+circuit picture, logical network compilation parallelized it by exploiting the lack of time
+ordering in ZX-diagrams, combining the effects of the three CNOTs into a single effect.
+If we want to parallelize multiple subroutines, however,
+without having to re-compile their joint logical network, we can do this procedurally through
+state teleportation, using a so-called bridge qubit. This technique parallelizes non-commuting
+operations without breaking physics, and it pays off because Bell state preparation and
+measurements are assumed to be fast on an active volume computer (see `info box at the top <AV info box_>`_).
+
+We will showcase this type of parallelization by treating the three CNOT gates in our ladder
```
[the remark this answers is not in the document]
dwierichs: Oh, good catch!

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,689 @@
...
+       operations is a property of the computer.
+    #. Qubit modules can execute so-called logical blocks, which is a slow/expensive operation.
+       Computations are represented by logical networks (see below) made up of connected logical
+       blocks. The qubit modules executing connected blocks must be within range.
+    #. Qubit modules can be grouped and occupied for some time to distill magic states. This is
+       a slow/expensive operation as well.
+
+    A more detailed characterization is provided in Sec. 1 of [#Litinski2022]_.
+
+    |
```
dime10: Attempt to preserve a new line as the box looks a little cramped

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,687 @@
...
+the active volume computer to use its computing resources with high efficiency.
+For this, we will transform the circuit into a ZX-diagram, impose a set of rules that
+lead us to an **oriented** ZX-diagram, and then rewrite it into the aforementioned
+logical network. This representation will allow us to capture the essence of quantum information
+processing without the restrictive representation as quantum circuits, while enforcing enough
+structure to arrive at a program that can be executed by an active volume computer.
+
+To put this rather formal process into context, we will analyze the same
+circuit under the lens of parallelization via quantum state teleportation, and relate the
+result to the derived logical network.
```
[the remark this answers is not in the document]
dwierichs: Oh, good catch yes

demonstrations_v2/tutorial_active_volume/demo.py
```diff
@@ -0,0 +1,687 @@
...
+Depending on the concrete circuit we are compiling, this change might just lead to minor
+modifications such as relabeling the ports of two connected blocks, or it might lead to cascades of
+re-routing that eventually force us to insert a new block, even.
+**Logical networks** are diagrams, or networks, of logical blocks. We label the logical blocks
+uniquely and replace drawing edges between them by annotating the ports with the labels of the
+blocks they connect to. This allows us to separate logical blocks, which represent computation
+and are considered expensive, from the block connectivity, which represents communication between
+blocks and is considered cheaper because it is natively supported by modules within range of
+each other, or can be implemented via fast SWAP connections (also see
+the `active volume computer info box <AV info box_>`_).
```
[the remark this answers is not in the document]
dwierichs: I did this:

>The one thing I could imagine that makes sense to say is that not drawing the edges allows us a bit more "visual" flexibility to rearrange blocks when scheduling them into qubit modules 

in a short sentence, just to give a bit of a motivation for leaving out crucial information from the visual representation, beyond just "for convenience" :D

PR #1626, opened 2025-12-19
Demo notebook fixes

doctorperceptron:
This PR introduces a few fixes for processing the demo scripts into Jupyter Notebooks:

- Replace direct references to thumbnail images with cloud URL
- Process doc URLs
  - A lot of these end with a dollar sign, which can confuse the renderer into thinking the following text is math. The dollar sign isn't needed in the URL so we strip it.
- Fix headings
  - Some demos have custom classes at the very start of the document. These aren't used and conflict with the heading, so they are removed.
- Use html for figures
  - This is the only way to apply scaling to the figures. 
- Misc. formatting issues and broken links.

**Note**: I used 50% scaling for inline images and 25% for tables. _Most_ images are huge, so this seems reasonable, but happy to entertain other suggestions.

### DIFF
--- a/lib/filter_helpers.py
+++ b/lib/filter_helpers.py
@@ -3,12 +3,12 @@ from pandocfilters import stringify, walk
 
 ASSETS_DIR = "https://blog-assets.cloud.pennylane.ai/demos/"
 
-def make_html_figure(src: str, width: int = 800, height: int = 500) -> str:
+def make_html_figure(src: str, width: int = 50) -> str:
     """Make an HTML figure from a source string.
     Returns an HTML string for the figure.
     """
     cleaned_url = parse_img_source(src)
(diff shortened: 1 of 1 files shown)


Lupeclan (approved):
LGTM, I was always curious about the
```
# This cell is added by sphinx-gallery
# It can be customized to whatever you like
```
block in the Notebooks.

doctorperceptron:
> LGTM, I was always curious about the
> 
> ```
> # This cell is added by sphinx-gallery
> # It can be customized to whatever you like
> ```
> 
> block in the Notebooks.

Haha that's been driving me nuts :)

anthayes92 (approved):
Thanks for the tidy up @doctorperceptron 🧹 ✨

PR #1630, opened 2025-12-24
Demo Magic states

daniela-angulo:


demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
+r"""What are magic states?
+==========================
+
+Quantum computers rely on very fragile physical systems. They are easily disturbed, leading to 
+the rapid generation and propagation of errors. In order to scale and harness the full potential 
+of quantum computing, we must develop fault-tolerant quantum computers (FTQC). For this purpose, besides 
+simply detecting and correcting errors, we must be able to realize gates that correctly perform their 
+intended operations without introducing or spreading noise across the encoded information.
```
drdren: I wonder if it is possible to breeze through the QEC part because I think it is reasonable to assume that the community understands that current quantum hardware is prone to errors and therefore it is necessary to use codes/protocols to do error correction.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+To achieve universal quantum computing (UQC), we need to have access to a universal gate set, such as
+the :math:`\textrm{Clifford + T}` set, ``{H, S, CNOT, T}``. As previously stated, all these gates must be 
+executed in a fault-tolerant manner. 
+For most quantum error correction architectures, gates of the Clifford group are much 
+simpler to implement, often using `transversal operations <https://arthurpesah.me/blog/2023-12-25-transversal-gates/>`__,
```
drdren: Is it more accurate to state "Clifford gates introduce less error in an error correction scheme than non-Clifford gates do"?

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+that when injected into a circuit implement a specific non-Clifford gate. 
+
+There are different types of magic states and their nomenclature often varies across the literature. 
+Let’s examine a specific state, which we will denote as :math:`|H\rangle`, to see it in action:
+
+.. math:: |H\rangle=\frac{1}{\sqrt{2}}(|0\rangle+e^{i\pi/4}|1\rangle)=T|+\rangle.
+
+Notice that this state is obtained by applying a T gate to the :math:`|+\rangle` state 
+(the +1 eigenstate of the Pauli X operator).  
+Using **magic state injection** (see the circuit illustration below), we can apply a T operation to an
```
drdren: "Injection" and "teleportation" appear to be used interchangeably. Indeed, they are strongly related concepts. However, consistently using only one of the two may lead to better understanding.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+print(qml.draw(t_gate_teleportation_circuit)(np.pi / 3))
+print(t_gate_teleportation_circuit(np.pi / 3))
+
+######################################################################
+# The output displays the density matrix of the target state after the T gate has been applied via
+# teleportation:
+# 
+# .. math:: T|\psi\rangle=T (R_y(\pi/3)|0\rangle)=\frac{\sqrt{3}}{2}|0\rangle+\frac{1}{2}e^{i\pi/4}|1\rangle.
```
drdren: It may be nice to explain that |psi\rangle is meant to represent any arbitrary quantum state into which a T gate can be injected.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# UQC via magic states but also proposed a method to prepare them starting from imperfect copies
+# of magic states. This purification process relies solely on Clifford operations and Pauli basis measurements. 
+# The synergy of these two properties---enabling universality and being distillable through restricted 
+# (Clifford) operations---is precisely why they were named "magic states".
+#
+# Preparing magic states 
+# ----------------------
+#
+# While magic states offer an elegant workaround for complex non-Clifford gate implementations, 
+# they remain computationally expensive to prepare. Let's examine two methods for their preparation.
```
drdren: It would be nice to have a quick explanation for why they are expensive to create. After all, the demo has presented that a magic state "only" consists of a H gate and a T gate. Why is that so challenging to create?

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# logical, being executed across multiple error-correcting blocks, resulting in a significant resource overhead. 
+#
+# See this `demo <https://pennylane.ai/qml/demos/tutorial_magic_state_distillation>`__ 
+# for an implementation of a distillation protocol using Catalyst. 
+#
+# Magic state cultivation
+# ~~~~~~~~~~~~~~~~~~~~~~~
+#
+# Currently considered the state-of-the-art method for generating high-fidelity magic states, 
+# formulated by Gidney et al [#Gidney2024]_ as a practical optimization.
```
drdren: It is unclear what "practical optimization" means.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# 4. Decoding: determine whether to accept the final state using standard error correction. Since the circuit 
+#    is now too large for efficient post-selection, a decoder computes a complementary gap. This metric acts as 
+#    a confidence score for the final state, allowing the system to accept or discard the state accordingly.
+#     
+#    
+# Current research
+# ----------------
+#
+# Despite the theoretical foundations of magic states being well-established, translating them into 
+# scalable hardware remains a challenge. Current research in the field focuses on developing more
```
drdren: Could you elaborate on why this is a challenge?

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# Current research
+# ----------------
+#
+# Despite the theoretical foundations of magic states being well-established, translating them into 
+# scalable hardware remains a challenge. Current research in the field focuses on developing more 
+# efficient protocols for generating high-fidelity magic states and achieving robust experimental 
+# implementations across various qubit architectures. 
+#
+# A distillation protocol was recently implemented at the logical level on a neutral-atom quantum 
+# computer [#QuEra2025]_, leveraging the parallel control and high connectivity offered by atom platforms.
```
drdren: As a proof of concept!

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+#    is now too large for efficient post-selection, a decoder computes a complementary gap. This metric acts as 
+#    a confidence score for the final state, allowing the system to accept or discard the state accordingly.
+#     
+#    
+# Current research
+# ----------------
+#
+# Despite the theoretical foundations of magic states being well-established, translating them into 
+# scalable hardware remains a challenge. Current research in the field focuses on developing more 
+# efficient protocols for generating high-fidelity magic states and achieving robust experimental
```
drdren: The current phrasing implies that MSC is a 'breakthrough' method but now this section suggests that its efficiency is still inadequate. Is that correct?

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# providing alternatives to the "grafting" method, used in the original MSC proposal, such as 
+# lattice surgery [#Hirano2025]_.  
+#
+# Conclusion
+# ----------
+#
+# In this tutorial, we reviewed the concept of magic states, an essential and creative tool for achieving FTQC and UQC.
+# While traditional distillation provided the theoretical foundation for universality, the shift 
+# toward cultivation reflects a new era of practical fault-tolerance focused on resource efficiency.
+# As we move toward large-scale implementations, the challenge lies in optimizing these "magic state factories."
```
drdren: This is the first mention of "magic state factories," so it is not quite clear what they are. You previously used the terms "developing more efficient protocols" in the "Current research" section, but now you use the term "optimizing." Are you referring to the same process?

demonstrations_v2/tutorial_magic_states/demo.py
daniela-angulo: No, that is magic state injection. When you have the good and nice magic state, then you perform injection to implement the T gate in an arbitrary state. But before injection, many things happened to actually obtain the good magic state and that's what it is described in this section. I added a sentence saying that we will see how they are difficult in what follows.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# The synergy of these two properties---enabling universality and being distillable through restricted 
+# (Clifford) operations---is precisely why they were named "magic states".
+#
+# Preparing magic states 
+# ----------------------
+#
+# While magic states offer an elegant workaround for complex non-Clifford gate implementations, 
+# they remain computationally expensive to prepare. Let's examine two methods for their preparation.
+#
+# Magic state distillation
```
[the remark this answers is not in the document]
daniela-angulo: I think that since there is not a simple illustration for Cultivation, we won't take this approach. I don't think the diagram will do much better than saying it in words. However, I could also mention a particular example of this implementation like a 5 to 1 (where you need five noisy to distill one good), if that sounds better. Maybe even a more particular example where I mention the amount of physical qubits given the inner and outer code structures.

demonstrations_v2/tutorial_magic_states/demo.py
daniela-angulo: I think I improved it. Let me know.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+# The primary objective is to synthesize magic states with the specific fidelities required 
+# for large-scale quantum computations as efficiently as possible.
+#
+# Unlike distillation, which consumes numerous noisy states to filter out a clean one, cultivation 
+# starts with a single *seed* state and improves it "in-place." The entire process occurs 
+# within a single code patch using physical-level operations.
+#
+# The protocol consists of four primary stages:
+#
+# 1. Injection: prepare an initial, noisy magic state encoded in a small code of distance 3 or better.
```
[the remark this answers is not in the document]
daniela-angulo: Gave titles for both sections and used bold font

demonstrations_v2/tutorial_magic_states/demo.py
drdren: Ah I see! That is very interesting. I did not know this nuance. If that is the case, then keep the phrasing!

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,256 @@
...
+#    of the small code would limit the state's purity. This stage then involves cycles of growing
+#    the code, stabilizing it, and checking the magic state. 
+# 3. Escape: rapidly expand the code hosting the state. Once the cultivation stage is complete, the magic state 
+#    reaches its target fidelity, and becomes "too good for the code". 
+#    To preserve this high fidelity, the state needs to *escape*
+#    into a much larger code as quickly as possible, typically via `grafting <https://arxiv.org/abs/2409.17595>`__
+#    `code morphing <https://arxiv.org/abs/2112.01446>`__ or 
+#    `lattice surgery <https://pennylane.ai/qml/demos/tutorial_lattice_surgery>`__.
+# 4. Decoding: determine whether to accept the final state using standard error correction. Since the circuit 
+#    is now too large for efficient post-selection, a decoder computes a complementary gap. This metric acts as
```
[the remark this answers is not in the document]
drdren: Oh wow! That is an improvement for MSC over distillation! Yes, I think a quick little bit about how cultivation reduced the spacetime cost by an order of magnitude is great. I think you don't need to go into tremendous detail about what 'spacetime costs' are. Perhaps you could link to the Active Volume demo, if appropriate.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,279 @@
...
+
+Where is the magic?
+-------------------
+
+Let's begin with a little bit of history. Bravyi and Kitaev formalized the concept and coined the
+term "magic states" in [#Bravyi2005]_. 
+In this work, they proved that the ability to prepare magic states, when combined with a set of ideal Clifford 
+gates, the preparation of :math:`|0\rangle` ancillas, and Z-basis measurement capabilities 
+on all qubits, is sufficient to enable UQC. Essentially, magic states are a class of states 
+that when injected into a circuit implement a specific non-Clifford gate.
```
drdren: , such as a T-gate.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,279 @@
...
+# 
+# An important detail is that, as one might suspect, there is a strict threshold fidelity for these preparation
+# protocols: the initial noisy state(s) :math:`\rho` must have a fidelity above a certain limit for the process 
+# to converge toward a pure state [#Bravyi2005]_. If the initial states are too noisy, the protocol will 
+# fail to improve them.
+#      
+# Outlook and recent developments
+# -------------------------------
+#
+# Current research in the field focuses of magic states on developing more
```
drdren: "Current research in this field focuses on developing more efficient protocols for generating ..." would be better.

demonstrations_v2/tutorial_magic_states/demo.py
```diff
@@ -0,0 +1,279 @@
...
+# Outlook and recent developments
+# -------------------------------
+#
+# Current research in the field focuses of magic states on developing more 
+# efficient protocols for generating high-fidelity magic states and achieving robust experimental 
+# implementations across various qubit architectures.
+#
+# A 5-to-1 distillation protocol was recently implemented as a proof-of-principle at the logical level 
+# on a neutral-atom quantum computer [#QuEra2025]_. This achievement leverage the parallel control and 
+# reconfigurability offered by atom platforms. However, in order to run deep computations,
```
drdren: "neutral atom platforms" would be less ambiguous because there are trapped ion platforms.

demonstrations_v2/tutorial_magic_states/demo.py
daniela-angulo: I don't know why I wrote it like that. Dyslexic moment.

PR #1639, opened 2026-01-07
Add seed to MCM demo

comp-phys-marc:
**Title:** Add seed to MCM demo

**Summary:** We keep seeing this demo getting flagged by the diff checker. Add a seed so that the diff does not always show differences that are inconsequential.

**Possible Drawbacks:** N/A


andrijapau (approved):

JerryChen97 (changes requested):
I think it should target `dev`? @andrijapau @doctorperceptron

PR #1646, opened 2026-01-08
0.44.0 QA changes for Braket demos

mudit2812:
Updating Braket demos based on information found during the QA process for the v0.44.0 release.
* Add warnings about old Python versions, HW execution costs, and bugs.
* Render a deprecation warning in the `ahs_aquila` demo.
* Add the Braket plugin as a demo dependency to the `requirements.in` file of all the Braket demos.

Change unrelated to Braket demos:
* ~Updated the versions of `jax` and `jaxlib` to 0.7.1, which is the only supported version for Catalyst and PennyLane program capture in v0.44.0.~ This change is removed because it broke dependency installation.


dependencies/constraints-dev.txt
```diff
@@ -19,8 +19,8 @@ h5py==3.11.0
 ipython==8.16.1
 
 chex<1.0.0
-jax==0.6.2
-jaxlib==0.6.2
+jax==0.7.1
+jaxlib==0.7.1
```
mudit2812: The JAX dependency was updated for the 0.44.0 release.

andrijapau (approved):
Renders LGTM, minor suggestion about including the bug fix version.

demonstrations_v2/braket-parallel-gradients/demo.py
```diff
@@ -451,7 +457,9 @@ def circuit(params, **kwargs):
 # :doc:`getting started with hybrid jobs <demos/getting_started_with_hybrid_jobs>`.
 #
 # .. warning::
-#     The following demo is only compatible with Python version 3.10.
+#
+#     The remaining demo is only compatible with Python version 3.10. Consequently, the latest supported
+#     PennyLane version is ``pennylane==0.42``.
```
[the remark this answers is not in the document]
mudit2812: The reason why I didn't include the patch version in the warning is because the warning is about the highest _minor_ supported version, but it's a pretty small detail so I'm fine either way. Tagging @gabrielasd for input; I'll update the message accordingly 🙂

demonstrations_v2/ahs_aquila/demo.py
```diff
@@ -744,8 +744,14 @@ def circuit(params):
...
-# To run this without connecting to the hardware, switch the aquila device out with the ``rydberg_simulator`` below.
-# Note that running on hardware is a paid service and will incur a fee.
+#
+#  .. note:: To run this without connecting to the hardware, switch the aquila device out with the ``rydberg_simulator`` below.
+#
+#  .. warning::
+#
+#      Note that running on hardware is a paid service and will incur a fee. Ensure that you are aware of
+#      these costs and comfortable with them before proceeding.
+#
```
gabrielasd: This is optional but would it be worth specifying whose hardware are we referring to here?

I notice one of the other demos had this warning message that refers to the AWS hardware:
"The following lines are computationally intensive. Remember that running it will result in simulation fees charged to your AWS account. We recommend monitoring your usage on the AWS dashboard."

demonstrations_v2/braket-parallel-gradients/demo.py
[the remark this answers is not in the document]
gabrielasd: I think leaving v0.42 is fine, but if in other demos the mini version has been updated, I'd say that we be consistent

demonstrations_v2/ahs_aquila/demo.py
mudit2812: I'll add it to the warning :)

PR #1648, opened 2026-01-08
Test shors demo

JerryChen97:


JerryChen97:
There's really nothing wrong with the demo running on 3rc. Will be closed after the entire jax dynamic shape concern has been gone.

PR #1663, opened 2026-01-19
Update the constraints of master to align with the dev

JerryChen97:
This PR is a remedy to patch the necessary step that has been missed during the release week of PLv0.44.

By applying this patch, we expect to see the external dependencies are correctly handled on the `master` branch.

Addtionally:
`tutorial_QUBO` has been marked as non-executable due to multiple outdated dependencies. It will be better for us to do so sooner than later, since this demo's core dependency `openqaoa` has been out of maintenance for years, so it's just a matter of time for it to generate more and more conflicts before we finally decide to drop its building from CI.

### DIFF
--- a/demonstrations_v2/tutorial_QUBO/metadata.json
+++ b/demonstrations_v2/tutorial_QUBO/metadata.json
@@ -5,8 +5,8 @@
             "username": "alejomonbar"
         }
     ],
-    "executable_stable": true,
-    "executable_latest": true,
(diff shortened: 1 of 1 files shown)


JerryChen97:
Build triggered here: https://github.com/PennyLaneAI/qml/actions/runs/21149908877

Lupeclan (approved):
LGTM, I appreciate the fast turn around on this issue! 🥇

PR #1687, opened 2026-02-09
Pin qiskit, install demo requirements last

doctorperceptron:
This PR adds some changes to get the dev branch building again:
- Pin `ibm-platform-services>=0.73`. For some reason a very old version is getting pulled in and failing to build for the ibm demos. Could be ibm or qiskit plugin related, but can't confirm root cause.
- Move `jaxopt` to the global `requirements-dev.txt` file, since most demos that were importing jax also use it.
- For dev builds, re-install the demo-specific requirements _after_ installing the dev versions of PLC. This is due to several demos requiring numpy<2.4, but the PLC installs upgrade numpy to 2.4
- Cleaned up all requirements.txt files that install packages that are already part of core PLC.

This has been tested successfully in CI and all the dev and master demos build successfully using this branch.

### DIFF
--- a/demonstrations_v2/tutorial_iqp_circuit_optimization_jax/demo.py
+++ b/demonstrations_v2/tutorial_iqp_circuit_optimization_jax/demo.py
@@ -15,9 +15,9 @@ estimating expectation values of certain observables is easy.
     :alt: IQP circuit optimization
 
 .. note::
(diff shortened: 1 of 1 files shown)


Lupeclan (approved):
LGTM :shipit:

PR #1694, opened 2026-02-12
Cache objects.inv when building notebooks

doctorperceptron:
Sphinx has a pretty sophisticated in-memory caching mechanism for handling the intersphinx objects.inv inventory files it downloads to create dynamic links to our documentation. We also need these files in the Pandoc filters for creating links when building the notebooks, but we do not have access to Sphinx in these scripts so have been downloading the inventory files each time we process a notebook. This can add up to quite a few network calls, and they do occasionally time out or get blocked, which kills the run.

This PR adds a caching class for downloading these files on first-run of the Pandoc filters, and then persisting them to disc. This results in one download per batch of demos processed, rather than one download per demo processed. 

A command line argument is also added to the `qml` tool so this cache can be cleared when working locally.

### DIFF
--- a/lib/filter_links.py
+++ b/lib/filter_links.py
@@ -7,6 +7,7 @@ from inventories import IntersphinxInventories
 from pandocfilters import toJSONFilter, Link, RawInline
 
 DEMOS_URL = "https://pennylane.ai/qml/demos/"
+inventories = IntersphinxInventories()
 
 def process_link(text: str, key: str) -> tuple[str, str]:
     """Process a link to a PennyLane or Catalyst object."""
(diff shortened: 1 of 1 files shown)


lib/filter_links.py
```diff
@@ -3,59 +3,18 @@
...
 
 def process_link(text: str, key: str) -> tuple[str, str]:
     """Process a link to a PennyLane or Catalyst object."""
     # Check if it's a PennyLane object. These are more popular, so check first.
-    if key in pl_obj_inv.keys():
-        return text if text else pl_obj_inv[key].dispname, PL_OBJ_INV_URL + pl_obj_inv[key].uri
-    if key in cat_obj_inv.keys():
-        return text if text else cat_obj_inv[key].dispname, CAT_OBJ_INV_URL + cat_obj_inv[key].uri
+    if key in inventories.pennylane.keys():
+        return text if text else inventories.pennylane[key].dispname, inventories.pennylane_base_url + inventories.pennylane[key].uri
```
ashishks0522: @doctorperceptron do need to account for `latest` and `stable` here as well in the base of URL? 🤔

PR #1704, opened 2026-02-17
NA

AbdullahKazi500:


AbdullahKazi500:
Hi @DSGuala deleted my puzzles submission

CatalinaAlbornoz:
Hi @AbdullahKazi500 , thanks for updating your submission. Unfortunately this again looks like it's fully made by AI. I noticed some mistakes where you're using Qiskit functions or arguments for example. While it's ok to use AI for refinements, we want to see human-made contributions. With that in mind, you're free to contribute a new GIF if you'd like by Friday.

AbdullahKazi500:
> Hi @AbdullahKazi500 , It's good to see that you did work on the GIF yourself.
> 
> When I mentioned Qiskit I didn't mean that you were using the Qiskit package, I was referring to the fact that some of the code shown in some of the doodles is not functional PennyLane code, meaning that it won't actually run if you tried to code it up. It's just a detail but the main message here is that one way you can improve your GIFs is by looking into the quality of the execution. If you verify the code shown in your doodles, and if you can ensure that the different items don't overlap for example, this can make the GIFs clearer. Remember that one of the evaluation criteria is "Clarity in concept and execution" so if you can find ways to improve these that would be good.

Thank you yes even I was confused about the Ai part as my submission never had any Ai assistance 
Yes I agree with you on the clarity of the concept part I believe it should be definitely looked into , I will try coming up with a more good gif for the next submission

PR #1715, opened 2026-03-05
[Demo] FT threshold theorem

obliviateandsurrender:
**Title:** Understanding Fault-tolerant Threshold Theorem in Practice

**Summary:** A hands-on guide to understanding the fault-tolerant threshold theorem and simulating surface code pseudo-thresholds.

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

----

* AUDIENCE — Who is this for?

- [ ] Tutorial
- [ ] Demo
- [ ] How-to


Qottmann (commented):
Had a brief first read, looks very nice overall! 

My main feedback is that while the content itself is great, the start after the abstract is a bit steep, and could potentially could be better motivated by providing some big picture context and anticipate what is going to happen in the demo

Qottmann:
> to better help readers visualize the function.

which function are you referring to?

I think it could be nice to anticipate these diagrams and what they mean already in the abstract/intro
image

i.e. creating a hand-drawn mock of them and explaining the different regions and what these diagrams say

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,455 @@
...
+In this demo, we will learn the theoretical framework of the fault-tolerant threshold
+theorem and observe this phenomenon in action. We will construct a QEC scheme called the
+*rotated surface code*, and subject it to simulated quantum noise across a range of physical
+error rates, computing two key metrics: (i) the pseudo-threshold, which is the break-even
+point where a specific error-correcting code becomes better than doing nothing at all, and
+(ii) the fault-tolerant threshold, the fundamental crossing point below which increasing our
+code size provides an exponential suppression of errors.
+
+.. figure::
+    ../_static/demo_thumbnails/opengraph_demo_thumbnails/pennylane-demo-stabilizer-codes-open-graph.png
```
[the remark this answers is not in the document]
Qottmann: not sure how complicated it would be, but perhaps a second or third tick on the x axis could help see that the x axis is logarithmic as well (minor thing)

Qottmann (approved):
Looks goot to me! Great demo @obliviateandsurrender 

just two math rendering issues and of course updating the thumbnail!

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+        for q in range(num_wires):
+            if x_basis:
+                qp.H(wires=q)
+            qp.DepolarizingChannel(noise_param, wires=q)
+        return [qp.sample(op) for op in meas_ops]
+
+    # Circuit 1: |+...+⟩ -> noise -> measure X-stabs + X-logical  (Z-error syndrome)
+    # Circuit 2: |0...0⟩ -> noise -> measure Z-stabs + Z-logical  (X-error syndrome)
+    z_stab_res = np.column_stack(syndrome_circuit(x_meas_ops, x_basis=True))
+    x_stab_res = np.column_stack(syndrome_circuit(z_meas_ops, x_basis=False))
```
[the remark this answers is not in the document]
obliviateandsurrender: Yes! The inversion exists because we measure X-stabilizers to capture phase flips and Z-stabilizers to capture bit flips.

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+point where a specific error-correcting code becomes better than doing nothing at all, and
+(ii) the fault-tolerant threshold, the fundamental crossing point below which increasing our
+code size provides an exponential suppression of errors.
+
+.. figure::
+    ../_static/demonstration_assets/ft_threshold/cartoon_thresholds.png
+    :align: center
+    :width: 85%
+    :target: javascript:void(0)
+
```
drdren, line 42: You refer to this figure a few times throughout the demo because it is very important. I think it would be best to label this Figure 1 and then refer back to Figure 1 when you need to in the text. 

It would be good, too, to define the axes. p = number of errors / number of operations. How do you define p_L?

Also, the figure is beautiful. Did you make it? If not, then you'll need to cite the source!

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+i.e., below a critical physical error rate, adding more qubits exponentially suppresses the
+logical error rate.
+
+In this demo, we will learn the theoretical framework of the fault-tolerant threshold
+theorem and observe this phenomenon in action. We will construct a QEC scheme called the
+*rotated surface code*, and subject it to simulated quantum noise across a range of physical
+error rates, computing two key metrics: (i) the pseudo-threshold, which is the break-even
+point where a specific error-correcting code becomes better than doing nothing at all, and
+(ii) the fault-tolerant threshold, the fundamental crossing point below which increasing our
+code size provides an exponential suppression of errors.
```
drdren:
```suggestion
error rates, computing two key metrics: (i) the pseudo-threshold :math:`p_{pseudo}`, which is the break-even
point where a specific error-correcting code becomes better than doing nothing at all, and
(ii) the fault-tolerant threshold :math:`p_{th}`, the fundamental crossing point below which increasing our
code size provides an exponential suppression of errors.
```

drdren: I would say that the two thresholds don't seem very different to each other. Surely, if increasing the code size means that errors go down, then that means that QEC is better than doing nothing at all? The other way around is true too; if applying QEC is better than doing nothing, then that implies that increasing distance d should improve errors?

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+with computational errors.
+
+Building a fault-tolerant quantum computer requires Quantum Error Correction (QEC),
+which involves redundantly encoding a single logical qubit across many physical qubits
+so that errors can be detected and corrected without ever directly reading the encoded
+information. But because the physical operations performing this encoding are themselves
+noisy, QEC introduces new opportunities for errors, raising a fundamental question:
+*Can we ever get ahead of the noise?*
+
+Think of QEC like bailing water from a leaky boat. If the hole is too large, water rushes
```
drdren, line 17: The analogy is great! Very engaging.

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+
+Fault-tolerant Threshold Theorem
+---------------------------------
+
+The threshold theorem is the mathematical bedrock of scalable quantum computing.
+Intuitively, it states that a fault-tolerant quantum computation of any size can be 
+accurately executed on imperfect hardware, provided that the base error rate of the 
+physical operations, :math:`p`, remains strictly below a specific, non-zero constant 
+known as the *threshold*, :math:`p_{th}`, the rightmost crossing point in the
+schematic above.
```
drdren:
```suggestion
known as the *threshold*, :math:`p_{th}`, the rightmost crossing point in Figure 1.
```

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+
+More rigorously, assuming a local stochastic error model where :math:`p < p_{th}`,
+we can take any ideal circuit :math:`\mathcal{C}` of size :math:`N` and construct a 
+corresponding fault-tolerant circuit :math:`\mathcal{C}^{\prime}`. Even when subjected
+to continuous noise, this corrected circuit is guaranteed to yield an output that deviates
+from the ideal outcome by no more than an arbitrarily small tolerance, :math:`\epsilon > 0`.
```
drdren: I was surprised about this paragraph based on the previous paragraph. Is this phrasing correct, based on my understanding of the theorem? Is p only applicable to the ideal circuit or all of the qubits added to comprise the bigger fault-tolerant circuit? 


```suggestion
More rigorously, the theorem proves that if physical qubits continuously experiencing local stochastic error with :math:`p < p_{th}` 
as part of an ideal circuit :math:`\mathcal{C}` of size :math:`N`, the corresponding error-corrected circuit :math:`\mathcal{C}^{\prime}` is guaranteed to yield an output that deviates
from the ideal outcome by no more than an arbitrarily small tolerance, :math:`\epsilon > 0`. 
When :math:`p > p_{th}`, however, the error-corrected circuit yields a more erroneous output than the ideal circuit. 
```

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+Crucially, the theorem ensures that this correction is practically achievable and the
+required hardware overhead, i.e., the total number of physical qubits and time steps
+needed for :math:`\mathcal{C}^{\prime}`, grows at most by a polylogarithmic factor,
+:math:`\mathcal{O}(\log^{c}(N/\epsilon))` for some positive constant :math:`c` [#threshold]_.
+
+In other words, operating in the green region (where :math:`p < p_{th}`) of the schematic
```
drdren:
```suggestion
In other words, operating in the green region (where :math:`p < p_{th}`) of Figure 1 
```

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+physical operations, :math:`p`, remains strictly below a specific, non-zero constant 
+known as the *threshold*, :math:`p_{th}`, the rightmost crossing point in the
+schematic above.
+
+More rigorously, assuming a local stochastic error model where :math:`p < p_{th}`,
+we can take any ideal circuit :math:`\mathcal{C}` of size :math:`N` and construct a 
+corresponding fault-tolerant circuit :math:`\mathcal{C}^{\prime}`. Even when subjected
+to continuous noise, this corrected circuit is guaranteed to yield an output that deviates
+from the ideal outcome by no more than an arbitrarily small tolerance, :math:`\epsilon > 0`.
+Crucially, the theorem ensures that this correction is practically achievable and the
```
drdren: If I am understanding correctly: 

```suggestion
Crucially, the theorem ensures that this correction is practically achievable because the
```

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+required hardware overhead, i.e., the total number of physical qubits and time steps
+needed for :math:`\mathcal{C}^{\prime}`, grows at most by a polylogarithmic factor,
+:math:`\mathcal{O}(\log^{c}(N/\epsilon))` for some positive constant :math:`c` [#threshold]_.
+
+In other words, operating in the green region (where :math:`p < p_{th}`) of the schematic
+is not merely better, it is the condition under which arbitrarily large and reliable quantum
```
drdren: What does better mean? Less erroneous when encoding the circuit with a QEC code?

demonstrations_v2/tutorial_ft_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+circuits become physically constructable. Although the original theoretical framework relied
+on specific assumptions like independent stochastic noise, the threshold theorem is robust
+enough to apply to highly realistic, correlated noise environments as well. Moreover, it
+assures that the required number of physical qubits grows non-exponentially with the size
```
drdren: My overall comment for this paragraph is that some information is repeated, namely, the fact that the FT threshold theorem guarantees a reasonable resource scaling. I do like the part where you go "which means there is no fundamental physical barrier standing in the
way of large-scale quantum computers. At least, theoretically!" 

I think a part of this issue comes from the fact that you have the sentence "Although the original theoretical framework relied
on specific assumptions like independent stochastic noise, the threshold theorem is robust
enough to apply to highly realistic, correlated noise environments as well."

I wonder if you placed that sentence at the end of the previous paragraph, then maybe you could make this current paragraph more concise? :)

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+assures that the required number of physical qubits grows non-exponentially with the size
+of the computation, which means there is no fundamental physical barrier standing in the
+way of large-scale quantum computers. At least, theoretically!
+
+To test the threshold theorem in practice, we use one of the initial candidates for
+quantum error correction, the *surface code*, which is a topological code where qubits
+are arranged on a 2D grid, with stabilizer measurements locally checking for
+parity errors among nearest neighbors. If you would like to deepen your understanding
+of surface codes before proceeding, our `Introducing Lattice Surgery <tutorial_lattice_surgery>`_
+and `A Game of Surface Codes <tutorial_game_of_surface_codes>`_ demos are a great starting point.
```
drdren, line 84: Dang, we ought to have our own surface code demo! I will come back to link to the surface code demo when it is written. Nothing to do on your end!

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+# blue curve sits *above* the red line, meaning QEC is making things worse because
+# the extra circuit operations introduce more noise than they correct. Moving
+# leftward to lower physical error rates, the blue curve eventually dips
+# *below* the red line. That crossing point is the *pseudo-threshold* for
+# our distance :math:`d=3` code.
+#
+# Simulating The Threshold
+# -------------------------
+#
+# While the pseudo-threshold we computed previously tells us when a
```
drdren, line 363: So the threshold is not the same as calculating the pseudo threshold for a range of distances? The threshold is defined as when the errors reduce as code distance increases? 

Why is there guaranteed to be a single point when all the logical error rates cross over? I could expect that there are multiple points when the lines cross over. 

So is the goal to engineer the hardware for the threshold or the pseudo threshold?

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+Here, we specifically look at its *rotated* variant, which requires only :math:`d^2`
+data qubits to achieve the same distance :math:`d`. This gives a 50% reduction in
+qubit overhead when compared to the standard surface codes.
+
+.. figure::
+    ../_static/demonstration_assets/ft_threshold/rotated_surface_code.jpg
+    :align: center
+    :width: 85%
+    :target: javascript:void(0)
+
```
drdren, line 94: If the first schematic is labelled at Figure 1, then this should be labelled as Figure 2.

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+plt.ylabel(r"Logical Error Rate (p$_{L}$)", fontsize=10)
+plt.yscale("log")
+plt.xscale("log")
+plt.grid(True, which="both", ls="--", alpha=0.6)
+plt.legend(fontsize=10)
+plt.tight_layout()
+plt.show()
+
+######################################################################
+# The red dashed line is the baseline, i.e., the error rate you would see with
```
drdren, line 352: Same idea: Figure 3

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+plt.ylabel("Logical Error Rate (p$_{L}$)", fontsize=10)
+plt.yscale("log")
+plt.xscale("log")
+plt.grid(True, which="both", ls="--", alpha=0.6)
+plt.legend(fontsize=10)
+plt.tight_layout()
+plt.show()
+
+######################################################################
+# The curves for different distances cross at a single point, the *threshold*.
```
drdren, line 422: Same idea: Figure 4

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
```diff
@@ -0,0 +1,465 @@
...
+error rates, computing two key metrics: (i) the pseudo-threshold, which is the break-even
+point where a specific error-correcting code becomes better than doing nothing at all, and
+(ii) the fault-tolerant threshold, the fundamental crossing point below which increasing our
+code size provides an exponential suppression of errors.
+
+.. figure::
+    ../_static/demonstration_assets/ft_threshold/cartoon_thresholds.png
+    :align: center
+    :width: 85%
+    :target: javascript:void(0)
```
drdren, line 41: Actually, based on reading the whole demo and coming back to this figure, I am a little confused. The threshold is a single point where increasing code distance always improves error. But the figure shows that d=5 and 7 are noisier than the unencoded case. 

I think the pseudo threshold part is OK; every code has lower error rates than the unencoded case.

drdren (changes requested):
Hi @obliviateandsurrender , I really enjoyed your demo. I think the readers will love reading it. I had few issues with the flow or the wording. Rather, I think that the definitions of the two thresholds could be made clearer. 

The concept is tangible at a high level: If the physical qubits are very erroneous, then encoding qubits with a QEC code using many more qubits will only create noisier results. However, introducing the two different thresholds makes that story more complicated. 

Can you clarify the two thresholds? Which should a person eager to implement an algorithm on hardware care about?

demonstrations_v2/tutorial_ft_threshold/demo.py
obliviateandsurrender: Updated it slightly differently, you can take a second look and let me know how it reads now.

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
obliviateandsurrender: Added the said label, the definitions of the axes in the figures and the caption.

Also, I made it myself!

demonstrations_v2/tutorial_ft_threshold/demo.py
obliviateandsurrender: Yes, they are not very different (at least in purpose), but rather in being a property of a code family or an individual code. Since generally, you would start with a single code implementation for your hardware, the pseudo-threshold is what you would like to establish, which is also known as the break-even testing. You would proceed to observe the latter only after the former is confirmed (as per my intuitive understanding).

demonstrations_v2/tutorial_ft_threshold/demo.py
obliviateandsurrender: I have rewritten this, let me know if it makes more sense now.

demonstrations_v2/tutorial_ft_threshold/demo.py
obliviateandsurrender: Yes, also the fact that since you can achieve larger suppression with a controlled distance you overall physical qubit requirements remain non-exponential.

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
obliviateandsurrender: Yes and yes! You could consider it as a classical phase transition, which is to happen at the same point. I added a line about it in the text.

You start with a pseudo-threshold and then go for a threshold. Adding this in the intro section!

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
obliviateandsurrender: You were absolutely right to be confused by the old figure! I have replaced it with a new one, let me know if the confusion now persists.

drdren: Very impressive! Great work!

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
drdren: Thank you very much!

demonstrations_v2/tutorial_fault_tolerant_threshold/demo.py
drdren: Thank you very much!

demonstrations_v2/tutorial_fault_tolerant_threshold/metadata.json
```diff
@@ -0,0 +1,64 @@
+{
+    "title": "Understanding Fault-tolerant Threshold Theorem in Practice",
+    "authors": [{
+        "username": "whatsis"
+    }],
+    "executable_stable": true,
+    "executable_latest": true,
+    "dateOfPublication": "2026-03-17T00:00:00+00:00",
+    "dateOfLastModification": "2026-03-17T00:00:00+00:00",
+    "categories": [
```
drdren, line 10: We will change this in time for a marketing release next week!

drdren (approved):
I really enjoyed reading this demo, @obliviateandsurrender! I think it is clear and well written. Amazing graphics. How? I am happy to approve this demo.

drdren (commented):
Let's be sure to add the thumbnail when it is ready!

drdren (changes requested):
Thumbnail!

drdren (approved):
With the thumbnails and dates changed, I am happy to approve this!

demonstrations_v2/tutorial_fault_tolerant_threshold/metadata.json
```diff
@@ -0,0 +1,64 @@
+{
+    "title": "Understanding Fault-tolerant Threshold Theorem in Practice",
+    "authors": [{
+        "username": "whatsis"
+    }],
+    "executable_stable": true,
+    "executable_latest": true,
+    "dateOfPublication": "2026-04-17T00:00:00+00:00",
+    "dateOfLastModification": "2026-04-17T00:00:00+00:00",
```
drdren:
```suggestion
    "dateOfPublication": "2026-04-24T15:00:00+00:00",
    "dateOfLastModification": "2026-04-24T15:00:00+00:00",
```

drdren (approved):
Update dates

PR #1725, opened 2026-03-23
Update date of modification for active volume demo to reflect image change

ashishks0522:
## Changes
- Update date to reflect changes in https://github.com/PennyLaneAI/demos/pull/1724


_static/demonstration_assets/active_volume/reaction-limited-computation.png
dwierichs: The reaction-limited path in the lower left was edited. Note how it started with straight horizontal path originally, but it actually should flow through the first X/Y measurement box.

PR #1737, opened 2026-04-07
update BlueQbit

JerryChen97:
Trying to resolve https://github.com/PennyLaneAI/demos/actions/runs/23935399156/job/69810732510

### DIFF
--- a/demonstrations_v2/tutorial_bluequbit/metadata.json
+++ b/demonstrations_v2/tutorial_bluequbit/metadata.json
@@ -8,7 +8,7 @@
     "executable_stable": true,
     "executable_latest": true,
     "dateOfPublication": "2024-09-24T00:00:00+00:00",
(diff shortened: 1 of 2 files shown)


doctorperceptron:
LGTM, but we should merge to master rather than dev.

JerryChen97:
> LGTM, but we should merge to master rather than dev.

Thanks for quick review! Just I found actually [rerun the failed CI of master and it's gone](https://github.com/PennyLaneAI/demos/actions/runs/24020478434/job/70297754852), so I'm thinking very likely it is also not a version issue; and if there is indeed something beyond fluke, more probably tmrw API will still hang, unless there is some very peculiar nuance with BQ's API and SDK ver.

doctorperceptron:
Master and dev builds are passing, so this looks like it was a one-off issue. Closing this PR. cc- @JerryChen97

PR #1714, opened 2026-03-03
[Demo] qLDPC codes

obliviateandsurrender:
**Title:** Quantum low-density parity-check (qLDPC) codes for quantum error correction

**Summary:** Explain the motivation, construction and decoding of various types of qLDPC codes.

**Relevant references:** [Will add]

**Possible Drawbacks:** N/A

**Related GitHub Issues:**

----
If you are writing a demonstration, please answer these questions to facilitate the marketing process.

* GOALS — Why are we working on this now?
  *Eg. Promote a new PL feature or show a PL implementation of a recent paper.*
	* To have resources regarding quantum error correction.

* AUDIENCE — Who is this for?
    *Eg. Chemistry researchers, PL educators, and beginners in quantum computing.*
    * Error correction researchers and enthusiasts.

* KEYWORDS — What words should be included in the marketing post?
	* qLDPC codes, quantum error correction 

* Which of the following types of documentation is most similar to your file? 
(more details [here](https://www.notion.so/xanaduai/Different-kinds-of-documentation-69200645fe59442991c71f9e7d8a77f8))
    
- [ ] Tutorial
- [X] Demo
- [ ] How-to

PR #1745, opened 2026-04-13
Port qLDPC codes demo

obliviateandsurrender:
**Summary:** Ports qLDPC demo #1714 from `dev` to `master` with manual implementation of `binary_matrix_rank` function, which is currently available only in PL's latest branch.  

```py
def binary_matrix_rank(binary_matrix: np.ndarray) -> int:
    r"""Returns the rank of a binary matrix over :math:`\mathbb{Z}_2`."""
    rank, matrix = 0, np.asarray(binary_matrix, dtype=bool).copy()
    while len(matrix):
        matrix, pivot = matrix[:-1], matrix[-1]
        if not pivot.any():
            continue
        rank += 1 # New pivot found
        rows_with_bit = matrix[:, np.flatnonzero(pivot)[-1]]
        matrix[rows_with_bit] ^= pivot
    return rank
```


drdren (approved):
Looks good to me

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,826 @@
...
+plt.figure(figsize=(6, 3))
+pos = nx.bipartite_layout(G, var_nodes)
+colors = ["#70CEFF"] * num_vars + ["#C756B2"] * num_checks
+nx.draw(G, pos, with_labels=True, node_color=colors, node_size=500)
+text_options = {"verticalalignment": "center", "fontsize": 12}
+plt.text(-0.75, 0.0, "Variable Nodes", rotation=90, color=colors[0], **text_options)
+plt.text(1.1, 0.0, "Check Nodes", rotation=270, color=colors[-1], **text_options)
+plt.show()
+
+######################################################################
```
drdren, line 102: Hey @obliviateandsurrender , I think there is a plot missing here!

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,826 @@
...
+# error by computing the syndrome :math:`s = H\vec{e} \pmod{2}`, which flags the
+# parity checks that are violated. We can see this with an error on bit ``0``:
+#
+
+error_vec = np.array([1, 0, 0, 0, 0]) # bit flip on bit 0
+syndrome = (H @ error_vec) % 2
+
+print(f"Syndrome: {syndrome}")
+
+######################################################################
```
drdren, line 114: The syndrome is missing here.

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,826 @@
...
+(m1, n1), (m2, n2) = h1.shape, h2.shape
+
+css_code = np.hstack((
+        np.vstack([np.zeros((m1, n1), dtype=np.uint8), h1]),
+        np.vstack([h2, np.zeros((m2, n2), dtype=np.uint8)])
+))
+
+print(f"Shape of the CSS code: {css_code.shape}")
+
+######################################################################
```
drdren, line 169: The output is missing here.

drdren (changes requested):
Actually, upon closer inspection, the plots are absent! @obliviateandsurrender

drdren (approved):
Now it all looks great!

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,836 @@
...
+# For these codes, all stabilizers must commute, which is ensured by having each of the
+# :math:`X - Z` stabilizer pairs overlap on an even number of qubits. Mathematically, this is
+# equivalent to the symplectic orthogonality condition :math:`H_X(H_Z)^T = 0 \pmod{2}`,
+# which can easily be verified like so:
+#
+
+hx, hz = css_code[:m1, :n1], css_code[m1:, n1:] # Extract individual components.
```
JerryChen97: Is this correct? I thought according to the standard CSS picture we adopted, `css_code` is
0 Hz
Hx 0
and if so shouldn't `css_code[:m1, :n1], css_code[m1:, n1:]` be taking the zero matrices at bottom-right and top-left? If there's some tricks applied somewhere that I missed, I guess we better add a note there to indicate that our construction differs from the texts

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,836 @@
...
+#
+# Here, the algebraic properties of the tensor product ensure that :math:`H_X` and :math:`H_Z`
+# satisfy the symplectic orthogonality condition. Furthermore, the transposed matrix :math:`H_i^T`
+# defines the transpose code, which has its own parameters :math:`[m_i, k_i^t, d_i^t]`, where
+# the superscript :math:`t` simply labels the dimension and distance of this new code.
+# For example, look at the following HGP code constructed from two :math:`d_1=3` and :math:`d_2=3`
+# repetition codes, which is equivalent to a toric code :math:`[[13, 1, 3]]`:
+#
+
+from pennylane.qchem.tapering import _kernel as binary_matrix_kernel
```
JerryChen97, line 233: I'm worried this might become a fragile dependency on our internal code. Since we are already implementing `binary_matrix_rank` locally, I would suggest just rewrite the local GF2 kernel as well

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,836 @@
...
+#
+# To see why these modern constructions are so powerful, let us look at a simplified construction
+# of the QT codes. The function ``tanner_code`` below takes the base parity-check matrices of
+# an HGP code and iteratively applies a row-and-column elimination procedure to enforce the
+# local-code structure of the Tanner graph, which removes the inter-block redundancies in the
+# HGP parity-checks.
+#
+
+def tanner_code(h1: np.ndarray, h2: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+    """Construct Tanner code parity check matrices [arXiv:2309.11719]."""
```
JerryChen97, line 365: This [paper ](https://arxiv.org/pdf/2309.11719) talks about LRESC (Long-Range-Enhanced Surface Codes). Though quite in the same field, I wonder if we have other more direct source, e.g. 2202.13641; also, 2309.11719 only shows up here, not in the ref list.

demonstrations_v2/tutorial_qldpc_codes/demo.py
```diff
@@ -0,0 +1,836 @@
...
+        for label, evolved in [("Lx", evolved_lx), ("Lz", evolved_lz)]:
+            commutes = all(evolved.commutes(g) for g in generators)
+            ntrivial = str(evolved) not in stabilizers
+            valid = commutes and ntrivial
+            print(f"{label}[{itr}] --> {evolved} | "\
+                    f"commutes={commutes}, nontrivial={ntrivial}")
+            if not valid:
+                result = False
+                break
+    return result
```
JerryChen97, line 743: It looks like that this helper only checks "each evolved operator is some nontrivial normalizer element." Does it prove that the evolved $(L_x, L_z)$ remain a valid logical basis?

JerryChen97 (changes requested):

demonstrations_v2/tutorial_qldpc_codes/demo.py
drdren: Hi @JerryChen97 , I will comment for context. We have to implement binary_matrix_rank locally because it is not available in the current version. We're happy to remove this local implementation when v0.45 releases.

obliviateandsurrender: Or we should shift `_kernel` to `math` module as `binary_matrix_kernel`. The only reason I am implementing `binary_matrix_rank` as well is that I want to have this work on stable, otherwise the `dev` version was already using the new function.

JerryChen97: Oh I actually didn't think implementation of `binary_matrix_rank` is bad here; quite in the contrary I feel that a demo better to be fully self-contained, and we could just also implment `binary_matrix_kernel` as well so that the demo is not depending on that internal func `pennylane.qchem.tapering._kernel`

JerryChen97: Let's ignore this one since it is rather unimportant lol

demonstrations_v2/tutorial_qldpc_codes/demo.py
obliviateandsurrender: Yeah, but the `Appendix B` from this paper discusses the said transformation. I can add it to the `ref` list if you'd like, but I didn't think it was required as I was quoting the identifier here directly.

demonstrations_v2/tutorial_qldpc_codes/demo.py
obliviateandsurrender: Since the Clifford gates preserve the symplectic inner product by definition, I was skipping this check as it would be redundant and make the function a little bit more complex. I am adding a loop to check this explicitly (I did test this while writing this function lol). Let me know if you like that.

demonstrations_v2/tutorial_qldpc_codes/demo.py
obliviateandsurrender: Nice catch, updating this now!

JerryChen97 (approved):
Thanks for the prompt fix!

PR #1790, opened 2026-05-19
[SC-118904] Remove `/qml/` from Sphinx config & `objects.inv` S3 upload path

ant0nmsv:
### Changes

This PR implements the following changes:

- Removes `/qml` from all URL paths in the Sphinx config (`conf.py` + supplementary files).
- Updates the S3 path that the `objects.inv` file gets uploaded to.
  - The corresponding infra changes for the `objects.inv` file are in the following PR: https://github.com/XanaduAI/xanadu-terraform-provisioning/pull/1844
- Some minor auto-linting that was triggered on save.


doctorperceptron (approved):

.github/workflows/v2-sync-objects-dot-inv.yml
```diff
@@ -155,4 +155,4 @@ jobs:
           ${DEV_FLAG}
 
       - name: Sync objects.inv to HTML Bucket
-        run: aws s3 cp ./_build/objects.inv s3://$AWS_S3_HTML_BUCKET_ID/qml/objects.inv
+        run: aws s3 cp ./_build/objects.inv s3://$AWS_S3_HTML_BUCKET_ID/objects.inv
```
doctorperceptron: We might need to update this environment variable. I believe the existing AWS_S3_HTML_BUCKET_ID points to an S3 bucket that hosted old HTML site code, and suspect this is not where we want to put the objects.inv file long-term.

ashishks0522: I think you are right Paul, @ant0nmsv we should update the bucket to be `PL_SITE_PROD_S3_BUCKET_NAME` and test it. This way it will automatically be in the right bucket for static website code at the base URL

ant0nmsv: Hey guys, I think `AWS_S3_HTML_BUCKET_ID` already points to `PL_SITE_DEV_S3_BUCKET_NAME` in dev and `PL_SITE_PROD_S3_BUCKET_NAME` in prod. Just to clarify, what do we need to update exactly?
Also, I've noticed that this path now returns a 404:
https://dev.pennylane.ai/objects.inv

Are there any other deployments we need to do for the changes to take effect?

ashishks0522: Not doing this anymore

ashishks0522 (approved):

PR #1803, opened 2026-05-29
Use poetry 2.1

doctorperceptron:
The pyproject.toml file requires `poetry>=2.1`, but the `poetry-plugin-export` package version was pulling a lower value, which masked the fact that we had some old syntax sticking around for poetry<2.1. The recent Dependabot PR to bump `dulwich` (#1801 ) exposed this mess. This PR bumps the appropriate packages and cleans up the old poetry code that is no longer needed in CI. It also upgrades PennyLane to 0.45.0 in the project build itself, which was missed post-release.

Build tested [here](https://github.com/PennyLaneAI/demos/actions/runs/26641980390) ✅


poetry.lock
```diff
@@ -1665,79 +1912,67 @@ type = ["mypy (>=1.18.2)"]
 
 [[package]]
 name = "poetry"
-version = "1.8.5"
+version = "2.4.1"
```
doctorperceptron, line 1915: Relevant lockfile changes for poetry.

Lupeclan (approved):

JerryChen97 (approved):

PR #1804, opened 2026-05-29
Trotterization Demo

nobese:
### Before submitting

      guidelines specified in the [README](https://github.com/PennyLaneAI/demos/blob/master/README.md).

**Title:**

**Summary:**

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

----

* AUDIENCE — Who is this for?

- [ ] Tutorial
- [ ] Demo
- [ ] How-to


demonstrations_v2/trotterization_with_phase_gradient_demos/metadata.json
```diff
@@ -0,0 +1,25 @@
...
+  "executable_latest": true,
+  "dateOfPublication": "2026-05-26T00:00:00+00:00",
+  "dateOfLastModification": "2026-05-26T00:00:00+00:00",
+  "categories": ["Algorithms"],
+  "tags": ["Trotterization","Phase Gradient", "Simulation"],
+  "previewImages": [
+    {
+      "type": "thumbnail",
+      "uri": "/_static/demo_thumbnails/regular_demo_thumbnails/thumbnail_placeholder.png"
+  ],
```
doctorperceptron, line 24:
```suggestion
  "previewImages": [
    {
      "type": "thumbnail",
      "uri": "/_static/demo_thumbnails/regular_demo_thumbnails/thumbnail_placeholder.png"
    }
  ],
```

nobese: Thank you!

PabloAMC (approved):

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+Even more frustrating, things tend to change with time, meaning we cannot neglect time evolution as we set out to model realistic systems. To execute this successfully, though, tends to be incredibly computationally intense. In the case of particle systems, which constitute a lot of interesting, cutting-edge simulation work, the exponential nature of the system's scaling  very quickly renders classical, sequential simulation impossible, even for short time scales. Take, for example, a system of :math:`n` particles, in which there are :math:`2^n` possible configurations, meaning the Hamiltonian energy matrix of the system used in the time evolution operator :math:`U(t)=e^{-iHt}` would be :math:`2^n \times 2^n`. That's a lot to deal with!
+
+Luckily, quantum computers have shown promise in addressing this issue since, instead of having to represent each possible state contained in the problem's Hilbert space, qubits can themselves act as analogues for the particles that make up the system, enabling polynomial scaling with particle count rather than exponential scaling. `Maybe Feynman was onto something <https://s2.smu.edu/~mitch/class/5395/papers/feynman-quantum-1981.pdf>`_!
```
josh146: This is a bit of a hard sentence to parse simply based on how long it is (there are a lot of clauses and commas)

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+Luckily, quantum computers have shown promise in addressing this issue since, instead of having to represent each possible state contained in the problem's Hilbert space, qubits can themselves act as analogues for the particles that make up the system, enabling polynomial scaling with particle count rather than exponential scaling. `Maybe Feynman was onto something <https://s2.smu.edu/~mitch/class/5395/papers/feynman-quantum-1981.pdf>`_! 
+
+To make this feasible, though, we need to be strategic with how we simulate time evolution for our systems. **Trotterization** is a method used in Hamiltonian simulation that enables step-by-step time evolution of systems composed of non-commuting components. This demo will demonstrate that the strategic use of mathematical tricks that allow these complicated Hamiltonians to be exponentated opens the door to achieving realistic, accurate time evolution simulations.
+
```
josh146: Overall I like this intro @nobese, but my one piece of feedback is that it is perhaps focusing too hard on the engaging hook, and the actual details of what the demo is for is lost.

Would you be able to rework this to find a bit of a better balance? Have an engaging hook, but avoid burying the lede (that is, make sure the reason to keep reading the demo itself is clear. What will it show/do? Is trotterization the tool I need in my toolbox for hamiltonian simulation?)

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+The Commutation Problem
+-----------------------
+For a completely isolated free particle experiencing no external potential, the system's Hamiltonian can be simply defined in terms of kinetic energy and applied to the system via a unitary gate representing the time evolution operator :math:`U_{free}(t)=e^{-iHt}=e^{-iTt}`, where :math:`T` is the kinetic energy term. Sticking to this idealized case would, of course, be useless.
```
josh146: why would it be useless?

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+For a completely isolated free particle experiencing no external potential, the system's Hamiltonian can be simply defined in terms of kinetic energy and applied to the system via a unitary gate representing the time evolution operator :math:`U_{free}(t)=e^{-iHt}=e^{-iTt}`, where :math:`T` is the kinetic energy term. Sticking to this idealized case would, of course, be useless. 
+
+With each term added to a Hamiltonian, the feasibility of the time evolution operator :math:`e^{-iHt}` being executable within reasonable computational resource bounds diminishes. This brings to mind a supposedly simple solution, why not just split the Hamiltonian into smaller pieces that are computationally executable? If we take the representation :math:`H=H_1+H_2+...+H_n`, we could naïvely say that our time evolution operator becomes :math:`e^{-i(H_1+H_2+...+H_n)t}=e^{-iH_1t}e^{-iH_2t}...e^{-iH_nt}`. In our naïveté, however, we would neglect to consider the commutation relations of the split Hamiltonian components. If certain fragments of the Hamiltonian do not commute (:math:`[H_i,H_j] \neq 0`), we cannot exponentiate it in its entirety without deviating from our expected outcome.
```
josh146: Maybe just a reader clarification: so it is the number of terms in the Hamiltonian that determine the complexity of execution?

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+   |
+
+   :math:`(A+B)^2 = A^2+AB+BA+B^2`
+
+   |
+
+   So, the Taylor expansion :math:`e^{A+B}=I+(A+B)+\frac{1}{2}(A+B)^2+...` differs in the two cases!
+
+As we know, most physical systems require a combination of non-commuting operators to be fully described, but splitting the Hamiltonian this way in a non-commuting scenario would be inaccurate. To imagine this more physically, consider what it would mean to exponentiate two dependent properties in this way. If we took :math:`A` and :math:`B` to be our non-commuting operators, our naïve approach would lead us to blindly exponentiate as :math:`e^{-iHt}=e^{-iAt}e^{-iBt}`. However, applying these operators sequentially assumes that the system is accurately described by iterating the position *then* iterating the momentum, one after the other. This completely ignores how they influence each other, thus rendering the simulation inaccurate. Rats!
```
josh146: @nobese I wonder if this aside is necessary --- can we assume that the reader has prior knowledge of the Baker–Campbell–Hausdorff formula?

This feels like something that we can have as assumed knowledge, or link out to other demos/resources/codebook

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+Luckily for us, the problem of exponentiating non-commuting equations is not new. The `Lie-Trotter product formula <https://en.wikipedia.org/wiki/Lie_product_formula>`_ is a main result of this effort. The theorem states that for arbitrary :math:`m \times m` matrices :math:`A` and :math:`B`
+
+.. math::
+   e^{A+B}=\lim_{r \to \infty}(e^{A/r}e^{B/r})^r.
+
+Turning back to the Hamiltonian picture, if we take a large, finite :math:`r`, letting :math:`r` represent the number of time steps taken in the desired time evolution, we can approximate the Lie-Trotter formula to the first order Trotterization expression
+
+.. math::
+   e^{-iHt}=(\prod_j e^{-iH_j t/r})^r.
```
josh146:
```suggestion
   e^{-iHt}=\left(\prod_j e^{-iH_j t/r}\right)^r.
```

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+The form of the Lie-Trotter approximation implies a clear order of operations that must be executed to simulate the time evolution of a non-commuting system. 
+
+The first step of carrying out Hamiltonian simulation using Trotterization is splitting the Hamiltonian. As a rule of thumb, feasible algorithms should strive for a minimal gate count and minimized opportunity for error, meaning we should try to minimize the number of operators used. To achieve this, the Hamiltonian should be split only where mathematically necessary, meaning commuting terms should always be grouped together to the greatest extent possible. Take, for example, a Hamiltonian containing the terms :math:`Z_1Z_3`, :math:`Z_2Z_4`, and :math:`X_1X_2`, where :math:`Z_i` and :math:`X_j` are Pauli operators. The first and second terms will always commute since they consist completely of Z operators, but neither of the first two operators commute with the third operator. Therefore, the most optimal splitting would yield :math:`H_1=Z_1Z_3+Z_2Z_4` and :math:`H_2=X_1X_2`. This is not always so obvious in complex physical systems, so thorough analysis should be carried out to ensure optimal operator pairing.
```
josh146: You could link out to our Hamiltonian grouping functionality, or even show this in code!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+Once the Hamiltonian is appropriately split, we need to ensure that the fragments we are working with can actually be implemented using realistic quantum devices. This tends to require some kind of `transformation <https://pennylane.ai/demos/tutorial_mapping>`_ between the native representation of the system and a computationally compatible representation. Recalling the relevance of this method in simulating particle systems, one relevant example is the mapping of Fermionic states to qubit states via the `Jordan-Wigner transformation <https://docs.pennylane.ai/en/stable/code/api/pennylane.jordan_wigner.html>`_. For the purposes of the following simple demonstration, we will assume our Hamiltonian was originally defined in the Pauli basis and forgo the transformation step. Let our Hamiltonian be given by
```
josh146: Recommend using cross references here, e.g., `:doc:` (for the demo), and `:func:` (for the JW transform)

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+@qp.qnode(dev)
+def TrotterStepper(t,r,coeffs):
+    del_t = t/r
+
+    #Apply the rotation r times
+    for i in range(r):
+        qp.RX(2*coeffs[0]*del_t, wires=0)
+        qp.RZ(2*coeffs[1]*del_t, wires=0)
+
+    return [qp.expval(qp.PauliX(0)), qp.expval(qp.PauliY(0)), qp.expval(qp.PauliZ(0))]
```
josh146, line 132: Always good to execute code to show it being used, alongside relevant output!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+import pennylane as qp
+import numpy as np
+from scipy.linalg import expm
+import pennylane.estimator as qre
+from pennylane.transforms.rz_phase_gradient import rz_phase_gradient
+
+#Define System Parameters
+coeffs = [0.2, 1.3] #Define Hamiltonian coefficients
+observables = [qp.PauliX(0), qp.PauliZ(0)] #Define observables
```
josh146: It was confusing to me that this does not follow on from the same example as used in the grouping discussion. I think keeping consistency in the example through related sections would be helpful here

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+def TrotterStepper(t,r,coeffs):
+    del_t = t/r
+
+    #Apply the rotation r times
+    for i in range(r):
+        qp.RX(2*coeffs[0]*del_t, wires=0)
+        qp.RZ(2*coeffs[1]*del_t, wires=0)
+
+    return [qp.expval(qp.PauliX(0)), qp.expval(qp.PauliY(0)), qp.expval(qp.PauliZ(0))]
+###############################################################################
```
josh146, line 138: Nice first section! One comment although - as a reader, I am wondering where PL's trotter functionality fits into it. Should I be aware of them? Should there be code examples showing them after the manual trotter implementation?

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+#
+# In the simple example implemented in this demo, an observed Trotter error can be achieved by calculating time evolution of the system analytically and carrying out a comparison. This would, of course, be unfeasible for large, complicated (in other words, useful) models or long time scales, but it is sufficient for this example. `Methods for investigating Trotter error in non-analytical cases <https://simons.berkeley.edu/sites/default/files/docs/15639/trottererrortheorysimons.pdf>`_ are a continually developing field of study.
```
josh146: @nobese timings-wise, this feels like a good opportunity to mention or (minimally) showcase results from the Xanadu team:

- https://arxiv.org/abs/2606.30738
- https://arxiv.org/abs/2606.30741

josh146: (we should probably also put this in the intro)

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+#
+# When we discuss selecting a gate synthesis method, our initial instinct might be to minimize the T count in favour of implementability and efficiency. From this perspective, it seems obvious to always select the method with a low cost, such as the phase gradient approach. Using PennyLane's `rz_phase_gradient() <https://docs.pennylane.ai/en/stable/code/api/pennylane.transforms.rz_phase_gradient.html>`_ transformation, the expensive :math:`R_z(\phi)` rotations implemented in the above Trotterization can be translated into phase gradient additions and compared to the naïve approach. Since this method only transforms :math:`R_z` operations, we can perform our desired :math:`R_x` rotations using an :math:`R_z` rotation sandwiched between two Hadamard gates. For this example, a step count of :math:`r=200` will be taken in the resource estimation step.
```
josh146: This is a great point to link your phase gradient demo

josh146: Since the phase gradient method is not described here, I think it will be important to link out to it for readers that are not familiar with it

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+
+#Resource estimation
+test_r = 200
+Trotter_resources = qre.estimate(TrotterStepper)(t,test_r,coeffs) #Repeat Until Success!
+Trotter_resources_PG = qre.estimate(TrotterStepperPG)(t,test_r,coeffs)
+print("Repeat Until Success")
+print(Trotter_resources)
+print("Phase Gradient")
+print(Trotter_resources_PG)
+###############################################################################
```
josh146: My recommendation here with codeblocks is to break this up by text; avoid print statements, but instead print statements are replaced by rendered text explanation.

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+        qp.Hadamard(wires=0)
+        
+        U_B = qp.RZ(2*coeffs[1]*del_t, wires=0)
+
+    return [qp.expval(qp.PauliX(0)), qp.expval(qp.PauliY(0)), qp.expval(qp.PauliZ(0))]
+
+
+#Resource estimation
+test_r = 200
+Trotter_resources = qre.estimate(TrotterStepper)(t,test_r,coeffs) #Repeat Until Success!
```
josh146: Why do you say repeat until success here?

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+print(f"{'r':>5} | {'X error':>8} | {'Y error':>8} | {'Z error':>8} | {'Total error':>11}")
+print("-" * 51)
+
+for r in R:
+    resultPG = TrotterStepperPG(t,r,coeffs)
+    X_error_PG = abs(resultPG[0]-exact_exp_X)
+    Y_error_PG = abs(resultPG[1]-exact_exp_Y)
+    Z_error_PG = abs(resultPG[2]-exact_exp_Z)
+    total_error_PG = np.sqrt(X_error_PG**2+Y_error_PG**2+Z_error_PG**2)
+    print(f"{r:>5} | {X_error_PG:>8.5f} | {Y_error_PG:>8.5f} | {Z_error_PG:>8.5f} | {total_error_PG:>11.5f}")
```
josh146: Similar comment here as above, but rather than one large codeblock with multiple print statements, I suggest breaking this up with rendered text to make it easier to parse and read

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,308 @@
...
+#
+# Higher-Order Trotterizations
+# ----------------------------
+# To this point we have taken the first-order Trotterization approach, in which the root equation is truncated to the first order term. While this approach is sufficient for simple, short time scale problems (ex. systems with weak interaction), ignoring higher order terms and, therefore, reducing precision can result in ignorance of important system characteristics and unnecessarily high error that requires increased resources to account for [#Childs2021]_. Thus, the use of higher-order Trotterizations is often necessary to achieve realistic simulation.
```
josh146:
```suggestion
# To this point we have taken the first-order Trotterization approach, in which the root equation is truncated to the first order term. While this approach is sufficient for simple, short time scale problems (e.g., systems with weak interaction), ignoring higher order terms and, therefore, reducing precision can result in ignorance of important system characteristics and unnecessarily high error that requires increased resources to account for [#Childs2021]_. Thus, the use of higher-order Trotterizations is often necessary to achieve realistic simulation.
```

josh146 (changes requested):
Nice work @nobese! Left a review, but have a few high-level comments:

- I recommend avoiding having large codeblocks with lots of print statements, and instead naturally breaking up the codeblocks with rendered text. It is okay (and preferable!) to have text before 1-2 lines of code that print out information, as the text can help explain what is happening.

- I found the middle section (about the discretization) a bit out of left field --- it felt a bit like a tangent rather than part of a cohesive narrative. I understand the reasoning for including it (as I previously read the phase gradient demo), but I'm wondering if there is a way to make it a 'tighter' story for those who have come to this demo without reading the phase gradient demo?

- It feels like the demo should introduce or reference PL's own trotter functions/operations, even if it doesn't use them for pedagogical purposes

demonstrations_v2/exploring_trotterization/demo.py
nobese: Agreed, I split it up!

demonstrations_v2/exploring_trotterization/demo.py
nobese: This is one factor that occurs in tandem with non-commutivity! I changed the wording slightly to make this more clear

demonstrations_v2/exploring_trotterization/demo.py
nobese: I personally don't think BCH can be assumed as prior knowledge, especially because Pablo mentioned to incorporate it in the technical review. I cut it down a bit to hopefully reduce distraction, but let me know if you think it should just go entirely!

demonstrations_v2/exploring_trotterization/demo.py
nobese: I added qp.trotterize! Let me know if there is anything else that you think would be relevant to highlight

demonstrations_v2/exploring_trotterization/demo.py
nobese: It is linked and briefly described in the paragraph above! If that is too unclear, let me know.

demonstrations_v2/exploring_trotterization/demo.py
nobese: removed, this was from a previous misconception!

demonstrations_v2/exploring_trotterization/demo.py
nobese: I added the next section on qp.Trotterize that includes a comparison to the output of this function!

demonstrations_v2/exploring_trotterization/demo.py
nobese: Let me know if this works!

demonstrations_v2/exploring_trotterization/demo.py
nobese: I added one into the intro and one into the Trotter error section!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+
+Luckily, quantum computers have shown promise in addressing this issue. Instead of having to represent each possible state contained in the problem's Hilbert space, qubits can themselves act as analogues for the particles that make up the system. This has the potential to enable polynomial scaling with particle count rather than exponential scaling. `Maybe Feynman was onto something <https://s2.smu.edu/~mitch/class/5395/papers/feynman-quantum-1981.pdf>`_! 
+
+To make this feasible, tools have been put forward to aid in the task of `Hamiltonian simulation <https://pennylane.ai/topics/hamiltonian-simulation>`_. This demo will focus on **Trotterization**, a simulation methods that implements time evolution by segmenting and iteratively walking a Hamiltonian forward in time. Together, we will explore why Trotterization is an essential tool for quantum algorithms, work through a simple implementation to various orders, and determine what trade-offs may be necessary in the fault tolerant picture.
```
josh146: this is great!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+
+If complexity is the issue, why not just split the Hamiltonian into smaller pieces that *are* computationally executable? If we take the representation :math:`H=H_1+H_2+...+H_n`, we could naïvely say that our time evolution operator becomes :math:`e^{-i(H_1+H_2+...+H_n)t}=e^{-iH_1t}e^{-iH_2t}...e^{-iH_nt}`. In our naïveté, however, we would neglect to consider the commutation relations of the split Hamiltonian components. If certain fragments of the Hamiltonian do not commute (:math:`[H_i,H_j] \neq 0`), we cannot exponentiate it in its entirety without deviating from our expected outcome.
```
josh146: One thing I'm not sure I fully understand @nobese, is the argument that U(t)=e^{-iHt} can be implemented via a unitary gate, but this becomes infeasible as the number of terms increases.

- My first question is, what is the naive/straightforward implementation of U(t)? I am unaware of any off the top of my head. Or do you mean, essentially, for a given $H$, compute the exponential directly and decompose this down into a gate set?

- My second question, is _why_ does the feasibility diminish as the number of terms increases? And by feasibility, do you mean resources, error, or something else?

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+
+   For commuting matrices, where :math:`AB=BA` and it is implied that A and B are completely independent,
+
+   |
+
+   :math:`(A+B)^2 = A^2+2AB+B^2`.
+
+   |
+
+   For non-commuting matrices, where :math:`AB \neq BA` and it is implied that A and B have some level of dependency,
```
josh146:
```suggestion
   For non-commuting matrices, where :math:`AB \neq BA` and it is implied that $A$ and $B$ have some level of dependency,
```

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+If complexity is the issue, why not just split the Hamiltonian into smaller pieces that *are* computationally executable? If we take the representation :math:`H=H_1+H_2+...+H_n`, we could naïvely say that our time evolution operator becomes :math:`e^{-i(H_1+H_2+...+H_n)t}=e^{-iH_1t}e^{-iH_2t}...e^{-iH_nt}`. In our naïveté, however, we would neglect to consider the commutation relations of the split Hamiltonian components. If certain fragments of the Hamiltonian do not commute (:math:`[H_i,H_j] \neq 0`), we cannot exponentiate it in its entirety without deviating from our expected outcome.
+
+.. admonition:: Recall
+   :class: note
+
+   For commuting matrices, where :math:`AB=BA` and it is implied that A and B are completely independent,
+
+   |
+
+   :math:`(A+B)^2 = A^2+2AB+B^2`.
```
josh146: This would be best as display math

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+
+   |
+
+   For non-commuting matrices, where :math:`AB \neq BA` and it is implied that A and B have some level of dependency,
+
+   |
+
+   :math:`(A+B)^2 = A^2+AB+BA+B^2`
```
josh146: Same here, this would be best in display math

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+
+   |
+
+   So, the Taylor expansion :math:`e^{A+B}=I+(A+B)+\frac{1}{2}(A+B)^2+...` differs in the two cases!
```
josh146, line 50: Just to play devils advocate here, is this aside needed? (no strong feelings here, it might make sense to keep -- my main thought is due to disturbing the flow).

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+    del_t = t/r
+
+    #Apply the rotation r times
+    for i in range(r):
+        qp.RX(2*coeffs[0]*del_t, wires=0)
+        qp.RZ(2*coeffs[1]*del_t, wires=0)
+
+    return [qp.expval(qp.PauliX(0)), qp.expval(qp.PauliY(0)), qp.expval(qp.PauliZ(0))]
+###############################################################################
+# Luckily for us, PennyLane has the tools to make this much simpler. Using :func:`~qp.Trotterize`, the exact same procedure can be carried out on the target Hamiltonian.
```
josh146: This is great!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+    qp.trotterize(
+        first_order_expansion, 
+        n = num_trotter_steps,
+        order = 1
+    )(time, theta, phi, wires=['a','b'])
+    return [qp.expval(qp.PauliX(0)), qp.expval(qp.PauliY(0)), qp.expval(qp.PauliZ(0))]
+
+#Compare results for 50 Trotter steps
+print(TrotterStepper(t, R[4], coeffs))
+print(BuiltInTrotter(t, coeffs[0], coeffs[1], R[4]))
```
josh146: Only minor modification recommendation here: to move the first print statement after the first trotter stepper qfunc definition (line 116). This allows readers to immediately see the output, and it is close enough to this print statement that they can still compare by eye.

demonstrations_v2/exploring_trotterization/demo.py
josh146: The link however is not working -- it should be
```suggestion
# Luckily for us, PennyLane has the tools to make this much simpler. Using :func:`~pennylane.Trotterize`, the exact same procedure can be carried out on the target Hamiltonian.
```

josh146: Worth going through and double checking that all links are correctly resolving

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+
+###############################################################################
+# That's a good match! This procedure is advantageous for dealing with larger Hamiltonians or working Trotterization into a larger PennyLane workflow. For the purposes of this demonstration, we will continue using our DIY solution, but the choice is yours!
+#
+# For this simple Hamiltonian, fragmentation was handled by definition. In more complex scenarios, fragmentation must be carried out in its own step. Take, for example, a Hamiltonian containing the terms :math:`Z_1Z_3`, :math:`Z_2Z_4`, and :math:`X_1X_2`, where :math:`Z_i` and :math:`X_j` are Pauli operators. The first and second terms will always commute since they consist completely of Z operators, but neither of the first two operators commute with the third operator. Therefore, the most optimal splitting would yield :math:`H_1=Z_1Z_3+Z_2Z_4` and :math:`H_2=X_1X_2`. This is often not obvious in complex physical systems, so thorough analysis should be carried out to ensure optimal operator pairing. Tools such as :func:`~qp.pauli.group_observables` are very helpful in this case!
```
josh146:
```suggestion
# For this simple Hamiltonian, fragmentation was handled by definition. In more complex scenarios, fragmentation must be carried out in its own step. Take, for example, a Hamiltonian containing the terms :math:`Z_1Z_3`, :math:`Z_2Z_4`, and :math:`X_1X_2`, where :math:`Z_i` and :math:`X_j` are Pauli operators. The first and second terms will always commute since they consist completely of Z operators, but neither of the first two operators commute with the third operator. Therefore, the most optimal splitting would yield :math:`H_1=Z_1Z_3+Z_2Z_4` and :math:`H_2=X_1X_2`. This is often not obvious in complex physical systems, so thorough analysis should be carried out to ensure optimal operator pairing. Tools such as :func:`~pennylane.pauli.group_observables` are very helpful in this case!
```

josh146: Flows much better now!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+#    e^{-i\alpha Xt}e^{-i\beta Zt}=(e^{-i\alpha Xt/2r}e^{-i\beta Zt/r}e^{-i\alpha Xt/2r})
+#
+# which can be implemented as
+#
+# .. math::
+#    S_2(t)=R_X(\alpha t)R_Z(2\beta t)R_X(\alpha t).
+#
+# A benefit of imposing a symmetric formula such as this is that it cancels out the dominant error term discussed previously. The symmetry, essentially, does not allow for even powers to survive the operator application process, eliminating the dominant :math:`t^2` term discussed previously.
+#
+# Altering the TrotterStepper() function to a second-order Trotterization shows the impact this symmetry has on the Trotter error.
```
josh146:
```suggestion
# Altering the ``TrotterStepper()`` function to a second-order Trotterization shows the impact this symmetry has on the Trotter error.
```

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+#
+# Gate Synthesis Considerations
+# -----------------------------
+# Before we can officially add Trotterization to our list of capabilities, we must consider how our simulations will interface with eventual fault tolerant quantum hardware. Whether we like it or not, the resources provided by quantum hardware will be, in some way, limited. With this in mind, it is important that the resource requirements of the systems we work with are quantified and evaluated. One metric of interest is the number of `T-gates <https://pennylane.ai/blog/2025/01/optimizing-with-op-t-mize-dataset>`_ required to synthesize our operations.
+#
+# The PennyLane :func:`~qp.estimate` tool can be implemented to approximate the number of gates used to carry out a given process, taking a naïve approach. There are ample optimized methods that can be used to reduce resource requirements. Alternatively, the :doc:`multiplexed phase gradient method <demos/efficient_rotations_with_phase_gradient_states`, which takes advantage of a static register that holds spatially dependent phase values which can be *added* to a target state as needed, can be used for synthesis.
```
josh146:
```suggestion
# The PennyLane :func:`~pennylane.estimate` tool can be implemented to approximate the number of gates used to carry out a given process, taking a naïve approach. There are ample optimized methods that can be used to reduce resource requirements. Alternatively, the :doc:`multiplexed phase gradient method <demos/efficient_rotations_with_phase_gradient_states>`, which takes advantage of a static register that holds spatially dependent phase values which can be *added* to a target state as needed, can be used for synthesis.
```

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,341 @@
...
+#
+# Conclusion
+# ----------
+# There is a phenomenon that is renewed over and over again in which the field of mathematics is continually years ahead of physics (especially applied physics). The use of product formulas as tools for time evolution in Hamiltonian simulation is a beautiful example of this, in which a purely mathematical description of how to exponentiate non-commuting operators has become a defining method for time-evolution simulation in today's quantum pursuits. Understanding the basics of how Trotter products can be used and optimized for various applications of quantum simulation opens the door to not only increased utility but a heightened awareness of how the input of many fields is required to achieve viable outcomes. Keep calm and Trotter on!
```
josh146: @nobese are there any CTAs or links you would include in the conclusion?

josh146 (changes requested):
Looking good @nobese, very nearly there! Left a final round of comments

demonstrations_v2/exploring_trotterization/demo.py
nobese: I reworked these two paragraphs to hopefully make this more clear! To answer your questions here for context (to make sure the description aligns), the "feasibility" is that as the Hamiltonian grows, it is less and less likely that we can implement it in its entirety with available gates and hardware. So, we are inclined to fragment it into components that can be implemented using available gates and resources, but this means we must consider how these fragments commute since that dictates the error in our system. To answer your question directly, the naive approach would be exactly that, compute the Hamiltonian exponential completely and generate a gate set, which becomes a bigger/impossible task as the Hamiltonian becomes more complex.

demonstrations_v2/exploring_trotterization/demo.py
nobese: I think so! I was inspired by the binary number review in the QPE demo. I think it's something that will likely be accepted by most people without the elaboration but that having the elaboration will make things much clearer as they go forward.

demonstrations_v2/exploring_trotterization/demo.py
nobese: I added the codebook chapter! I could also link to some papers here, but I feel like the codebook is very actionable.

nobese:
Thank you for the comments @josh146 ! Sorry for the delay in addressing them, I missed the notification. I believe that I hit all of your points, particularly reworking the explanation of why increased complexity makes it harder to implement the time evolution unitary. Let me know if you have any more comments!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,344 @@
...
+
+Let's get started, time is of the essence!
+
+The Commutation Problem
+-----------------------
+For a completely isolated free particle experiencing no external potential, the system's Hamiltonian can be simply defined in terms of kinetic energy and applied via a unitary gate representing the time evolution operator :math:`U_{free}(t)=e^{-iHt}=e^{-iTt}`, where :math:`T` is the kinetic energy term. If this is the only scenario we can deal with, however, our capabilities would be incredibly limited. As a Hamiltonian becomes more complex (e.g., additional terms are added), the chance that the time evolution operator :math:`e^{-iHt}` can be implemented with available gates (determined by a process of calculation and decomposiiton into a relevant gate set) and computational resources diminishes.
```
josh146:
```suggestion
For a completely isolated free particle experiencing no external potential, the system's Hamiltonian can be simply defined in terms of kinetic energy and applied via a unitary gate representing the time evolution operator :math:`U_{free}(t)=e^{-iHt}=e^{-iTt}`, where :math:`T` is the kinetic energy term. If this is the only scenario we can deal with, however, our capabilities would be incredibly limited. As a Hamiltonian becomes more complex (e.g., additional terms are added), the chance that the time evolution operator :math:`e^{-iHt}` can be implemented with available gates (determined by a process of calculation and decomposition into a relevant gate set) and computational resources diminishes. 
```

demonstrations_v2/exploring_trotterization/demo.py
josh146: thanks!

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,344 @@
...
+
+The first step of carrying out Hamiltonian simulation using Trotterization is splitting the Hamiltonian. As a rule of thumb, feasible algorithms should strive for a minimal gate count and reduced opportunity for error, meaning we should try to minimize the number of operators required. To achieve this, the Hamiltonian should be split only where mathematically necessary, meaning commuting terms should always be grouped together to the greatest extent possible. 
+
+Once the Hamiltonian is appropriately split, we need to ensure that the fragments we are working with can actually be implemented using realistic quantum devices. This tends to require some kind of :doc:`transformation <demos/tutorial_mapping>` between the native representation of the system and a computationally compatible representation. Recalling the relevance of this method in simulating particle systems, one relevant example is the mapping of Fermionic states to qubit states via the :func:`~qp.jordan_wigner` transformation. For the purposes of the following simple demonstration, we will assume our Hamiltonian was originally defined in the Pauli basis and forgo the transformation step.
```
josh146:
```suggestion
Once the Hamiltonian is appropriately split, we need to ensure that the fragments we are working with can actually be implemented using realistic quantum devices. This tends to require some kind of :doc:`transformation <demos/tutorial_mapping>` between the native representation of the system and a computationally compatible representation. Recalling the relevance of this method in simulating particle systems, one relevant example is the mapping of Fermionic states to qubit states via the :func:`~pennylane.jordan_wigner` transformation. For the purposes of the following simple demonstration, we will assume our Hamiltonian was originally defined in the Pauli basis and forgo the transformation step. 
```

(worth verifying after merging this suggestion that the link resolves correctly)

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,344 @@
...
+    #Apply the rotation r times
+    for i in range(r):
+        qp.RX(2*coeffs[0]*del_t, wires=0)
+        qp.RZ(2*coeffs[1]*del_t, wires=0)
+
+    return [qp.expval(qp.PauliX(0)), qp.expval(qp.PauliY(0)), qp.expval(qp.PauliZ(0))]
+
+print(TrotterStepper(t, coeffs, R[4]))
+###############################################################################
+# Luckily for us, PennyLane has the tools to make this much simpler. Using :func:`~pennylane.Trotterize`, the exact same procedure can be carried out on the target Hamiltonian.
```
josh146: Link is not resolving :(

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,344 @@
...
+print(BuiltInTrotter(t, coeffs[0], coeffs[1], R[4]))
+
+###############################################################################
+# That's a good match! This procedure is advantageous for dealing with larger Hamiltonians or working Trotterization into a larger PennyLane workflow. For the purposes of this demonstration, we will continue using our DIY solution, but the choice is yours!
+#
+# For this simple Hamiltonian, fragmentation was handled by definition. In more complex scenarios, fragmentation must be carried out in its own step. Take, for example, a Hamiltonian containing the terms :math:`Z_1Z_3`, :math:`Z_2Z_4`, and :math:`X_1X_2`, where :math:`Z_i` and :math:`X_j` are Pauli operators. The first and second terms will always commute since they consist completely of Z operators, but neither of the first two operators commute with the third operator. Therefore, the most optimal splitting would yield :math:`H_1=Z_1Z_3+Z_2Z_4` and :math:`H_2=X_1X_2`. This is often not obvious in complex physical systems, so thorough analysis should be carried out to ensure optimal operator pairing. Tools such as :func:`~pennylane.pauli.group_observables` are very helpful in this case!
```
josh146: Minor but there are a lot of explanation marks at the end of these sentence, might make sense to reduce this a bit

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,344 @@
...
+#
+# Gate Synthesis Considerations
+# -----------------------------
+# Before we can officially add Trotterization to our list of capabilities, we must consider how our simulations will interface with eventual fault tolerant quantum hardware. Whether we like it or not, the resources provided by quantum hardware will be, in some way, limited. With this in mind, it is important that the resource requirements of the systems we work with are quantified and evaluated. One metric of interest is the number of `T-gates <https://pennylane.ai/blog/2025/01/optimizing-with-op-t-mize-dataset>`_ required to synthesize our operations.
+#
+# The PennyLane :func:`~pennylane.estimate` tool can be implemented to approximate the number of gates used to carry out a given process, taking a naïve approach. There are ample optimized methods that can be used to reduce resource requirements. Alternatively, the :doc:`multiplexed phase gradient method <demos/efficient_rotations_with_phase_gradient_states`, which takes advantage of a static register that holds spatially dependent phase values which can be *added* to a target state as needed, can be used for synthesis.
```
josh146: Both links here are also not resolving correctly.

The second one (the `:doc:` can be fixed by adding the closing `>`. For the :func:` one, not sure 🤔

demonstrations_v2/exploring_trotterization/demo.py
```diff
@@ -0,0 +1,344 @@
...
+#
+# When we select a gate synthesis method, our initial instinct might be to minimize the T count in favour of implementability and efficiency above all. From this perspective, it seems obvious to always select a low-cost method, such as the aforementioned phase gradient approach. Using PennyLane's :func:`~qp.rz_phase_gradient` transformation, the expensive :math:`R_z(\phi)` rotations implemented in the above Trotterization can be translated into phase gradient additions and compared to the naïve approach. Since this method only transforms :math:`R_z` operations, we can perform our desired :math:`R_x` rotations using an :math:`R_z` rotation sandwiched between two Hadamard gates. For this example, a step count of :math:`r=200` will be taken in the resource estimation step.
```
josh146:
```suggestion
# When we select a gate synthesis method, our initial instinct might be to minimize the T count in favour of implementability and efficiency above all. From this perspective, it seems obvious to always select a low-cost method, such as the aforementioned phase gradient approach. Using PennyLane's :func:`~pennylane.rz_phase_gradient` transformation, the expensive :math:`R_z(\phi)` rotations implemented in the above Trotterization can be translated into phase gradient additions and compared to the naïve approach. Since this method only transforms :math:`R_z` operations, we can perform our desired :math:`R_x` rotations using an :math:`R_z` rotation sandwiched between two Hadamard gates. For this example, a step count of :math:`r=200` will be taken in the resource estimation step.
```

josh146 (approved):
Great work @nobese! Happy to approve, just a few typos and links to correct before merging.

CatalinaAlbornoz (changes requested):
Hi @nobese, I used AI to review the demo and it noticed a couple major issues (and a few minor ones) that need to be resolved before we can merge. 

---
🔴 Blocking / high-priority

1. Stray file named git committed at the repo root (empty, 0 bytes). This looks like an accidental redirect (e.g. ... > git). Please delete it.
2. .gitignore change (+1 −1 in demonstrations_v2/.gitignore) — please confirm this is intentional and not a stray edit.
3. PEP8 / black non-conformance (the PR checklist claims conformance). black -l 100 --check fails. Concretely:
  - Missing spaces around operators: 2*coeffs[0]*del_t, t/r, etc.
  - Spaces around = in keyword args: wires = 0, n = num_trotter_steps, order = 1 should be wires=0, etc.
  - Comments without a space after # (E265): 11 instances (#Define System Parameters, #Apply the rotation r times, …).
  - ~20 lines with trailing whitespace in the docstring.
  - Please run black -l 100 demo.py.
4. CamelCase function names violate PEP8 (functions should be snake_case): TrotterStepper, BuiltInTrotter, TrotterStepperSO, TrotterStepperPG → e.g. trotter_stepper, built_in_trotter, trotter_stepper_so, trotter_stepper_pg.

---
📚 References & metadata

5. Su2020 is missing from metadata.json. It's cited inline ([#Su2020]_) and appears in the RST reference list, but not in the metadata references array.
6. Two orphan references — Childs2010 ("Limitations on the simulation of non-sparse Hamiltonians") and Paetznick2014 ("Repeat-Until-Success…") are defined in the RST reference list and in metadata but are never cited inline, and their topics don't match this demo. These look like leftover template references — uncited targets also produce Sphinx warnings. Please remove them or cite them.
7. Duplicated author in metadata.json for Cirstoiu2020: "Cirstoiu, Holmes, Holmes, Iosue, …" — "Holmes" appears twice.
8. Title mismatch: metadata.json title is "Exploring Trotterization", but the demo's H1 is just "Trotterization". Align them.
9. [#Verlet1967] title has unescaped nested quotes in demo.py: "Computer "Experiments" on Classical Fluids…" — this will render awkwardly. The metadata version uses single quotes ('Experiments'); mirror that or escape the inner quotes.
10. [#Suzuki1990]: doi:\...`is missing the space afterdoi: that every other reference entry has (doi: `...``).
11. Minor: [#Su2020] is a talk ("presented at the Simons Institute…") but is typed as "article" in metadata.

---
✍️ Spelling

12. Line 44: "eachother" → "each other".
13. Line 44: "non-commutivity" → "non-commutativity".

---
〰️ Hyphenation & consistency

14. Inconsistent ordinal-modifier hyphenation. Compound modifiers before a noun should be hyphenated. Currently mixed: "first**-**order" is used in most places but not in "first order Trotterization expression" (l. 51), "first order term" (l. 192), or "higher order terms" (l. 192).
15. "well known" → "well-known" (ll. 46, 238).
16. "tradeoff" (l. 318) vs "trade-offs" (l. 11) — pick one (PennyLane tends to use "trade-off").
17. "sublinear" vs "sub-linear" — both appear in the same sentence (l. 320).
18. "fault tolerant" → "fault-tolerant" when used as a modifier ("fault-tolerant quantum hardware", "the fault-tolerant picture").

---
🔤 Grammar

19. Line 243: "allowing for resources to be allocated to only the instances they are justified" — missing a word; → "…to only the instances where they are justified" (or "in which they are justified").
20. Line 320: "requires the Hamiltonian have a known analytical solution" — reads better as "requires that the Hamiltonian have a known analytical solution."

---
➗ Math rigor / presentation

21. Second-order formula (≈ l. 197):
e^{-iαXt}e^{-iβZt} = (e^{-iαXt/2r} e^{-iβZt/r} e^{-iαXt/2r})
The left side is the full first-order product over time t, while the right side is a single per-step term (note the /r and /2r) — these aren't equal, and there's no outer ^r. Consider writing it as the per-step symmetric operator S_2(Δt) = e^{-iαXΔt/2} e^{-iβZΔt} e^{-iαXΔt/2} with ≈ and the (...)^r for the full evolution.
22. Several exact-looking = signs are really approximations, e.g. l. 51 e^{-iHt}=(∏_j e^{-iH_j t/r})^r and l. 88 U(t)=e^{-iβZt}e^{-iαXt}=…. Since the whole point of the demo is that Trotterization is approximate, consider ≈ where appropriate.

---
🎨 Style vs other PennyLane demos

23. Section headings use Title Case ("The Commutation Problem", "Implementing the Trotter Method", "Gate Synthesis Considerations", "Higher-Order Trotterizations"). PennyLane's style guide generally prefers sentence case for headings (e.g. "The commutation problem"). Worth aligning.
24. Tone/exclamation density. The voice is charming, but the exclamations and asides pile up ("Rats!", "Not too shabby!", "Ah ha!", "Keep calm and Trotter on!", "time is of the essence!", "This becomes unaffordable very quickly!"). PennyLane demos are friendly but a bit more even — consider trimming a few so the strongest ones land.

---
🧹 Code cleanups (non-blocking)

25. first_order_expansion(time, theta, phi, wires) ignores its wires argument and hardcodes wires=0. Since trotterize passes wires in, use the parameter (wires=wires[0]) for correctness and clarity.
26. observables (l. 103) is defined but never used.
27. TrotterStepperSO has dead assignments: U_A_half = qp.RX(...) (reassigned) and U_B = qp.RZ(...) are never used; the gate calls have side effects, so drop the assignments. Same pattern in TrotterStepperPG (U_B = qp.RZ(...)).
28. Loop counters for i in range(r) don't use i — use for _ in range(r).

---
Summary

The demo runs correctly on v0.45.1 and is scientifically sound — no functional blockers. Before merge I'd prioritize: (1) removing the stray git file, (2) black/PEP8 + snake_case cleanup, and (3) the reference/metadata reconciliation (missing Su2020, orphan Childs2010/Paetznick2014, duplicated "Holmes", title mismatch). The spelling, hyphenation, and math-presentation items are quick polish passes on top.

nobese:
Hi @CatalinaAlbornoz ! Thanks for the feedback, I have implemented the applicable changes and re-requested review : )

CatalinaAlbornoz (approved):

PR #1824, opened 2026-06-30
Install Pytorch from PyPI

doctorperceptron:
The pipeline has been installing Pytorch from the Pytorch index, but this often times out resulting in failed builds. This PR switches to installing torch from PyPI. Tested by [building master successfully](https://github.com/PennyLaneAI/demos/actions/runs/28451085504).

### DIFF
--- a/dependencies/constraints-dev.txt
+++ b/dependencies/constraints-dev.txt
@@ -3,7 +3,6 @@
 # are handled separately in the demo cli.
 
 --extra-index-url https://test.pypi.org/simple/
---extra-index-url https://download.pytorch.org/whl/cpu
 
 # Install plug-ins required for demos
 pennylane-cirq @ git+https://github.com/PennyLaneAI/pennylane-cirq.git#egg=pennylane-cirq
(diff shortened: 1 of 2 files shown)


JerryChen97 (approved):

JerryChen97:
Just a message that recently in Pennylane we also transited from PyPI source to Pytorch source, exactly because we noticed that there're some huge delay sometimes which oculd be lasting for one hour or so. However, with PyTorch source we also started to often see the same issues as demos.
In a word, seems that there's no real solution here, we could only wish one is better than the other :pray:

doctorperceptron:
> Just a message that recently in Pennylane we also transited from PyPI source to Pytorch source, exactly because we noticed that there're some huge delay sometimes which oculd be lasting for one hour or so. However, with PyTorch source we also started to often see the same issues as demos. In a word, seems that there's no real solution here, we could only wish one is better than the other 🙏

For the demos pipeline, we often see a build failure due to a random package trying to be installed from the torch index but timing out. If as you say, using PyPI works but could add a delay to the successful build, that's currently acceptable.

In the future will look into using uv to install _just_ torch and torchvision from the pytorch index, which should be possible 🤞

ashishks0522 (approved):

PR #1833, opened 2026-07-10
fix: mark error prop demo as non-executable with PL<0.46

andrijapau:
This demo is no longer executable on latest due to https://github.com/PennyLaneAI/pennylane/pull/9786. 

Screenshot 2026-07-10 at 9 22 50 AM


demonstrations_v2/tutorial_error_prop/metadata.json
```diff
@@ -6,9 +6,10 @@
         }
     ],
     "executable_stable": true,
```
doctorperceptron: We should change this to false, otherwise this will fail when we merge dev into master following the next PennyLane release.

doctorperceptron (dismissed):

demonstrations_v2/tutorial_error_prop/metadata.json
andrijapau: Done

drdren (approved):
Hi @andrijapau ! From a content perspective, this looks great to me!

doctorperceptron (approved):

josh146 (changes requested):
Since this demo is non-executable, will need to copy in the output

drdren:
I am on it!

andrijapau:
> I am on it!

Thank you!!

josh146 (approved):

PR #1818, opened 2026-06-19
Upgrade python version to 3.12 in github actions for dev builds in demos [sc-122880]

derekokelly-rgb:
### Before submitting

Please complete the following checklist when submitting a PR:

- [ ] Ensure that your tutorial executes correctly, and conforms to the
      guidelines specified in the [README](https://github.com/PennyLaneAI/demos/blob/master/README.md).

- [ ] Remember to do a grammar check of the content you include.
- [ ] All tutorials conform to
      [PEP8 standards](https://www.python.org/dev/peps/pep-0008/).
      To auto format files, simply `pip install black`, and then
      run `black -l 100 path/to/file.py`.

When all the above are checked, delete everything above the dashed
line and fill in the pull request template.

------------------------------------------------------------------------------------------------------------

**Title:**

**Summary:**

**Relevant references:**

**Possible Drawbacks:**

**Related GitHub Issues:**

----
If you are writing a demonstration, please answer these questions to facilitate the marketing process.

* GOALS — Why are we working on this now?

  *Eg. Promote a new PL feature or show a PL implementation of a recent paper.*

* AUDIENCE — Who is this for?

  *Eg. Chemistry researchers, PL educators, beginners in quantum computing.*

* KEYWORDS — What words should be included in the marketing post?

* Which of the following types of documentation is most similar to your file? 
(more details [here](https://www.notion.so/xanaduai/Different-kinds-of-documentation-69200645fe59442991c71f9e7d8a77f8))
    
- [ ] Tutorial
- [ ] Demo
- [ ] How-to

PR #1835, opened 2026-07-14
Upgrade V2 Sync objects.inv only to use 3.12

derekokelly-rgb:
#1818 caused V2 Sync objects.inv workflow to fail in certain scenarios (e.g. [here](https://github.com/PennyLaneAI/demos/actions/runs/27852517986/job/82434041514)) due to a Python version mismatch between `master` and `dev` branches.

This PR bumps that workflow's Python version in `master` to match `dev` (v3.12)


doctorperceptron (approved):
Thanks @derekokelly-rgb! Let's give it a go and see what happens 🤞

JerryChen97 (approved):

